PR #17, opened 2017-08-09
Finishing checkbox

sarahkf:
Finishing the checkbox component

### DIFF
--- a/src/CheckBox/README.md
+++ b/src/CheckBox/README.md
@@ -1,0 +1,17 @@
+# Checkbox Component
+
+Checkbox based off of the [WAI-ARIA authoring guidelines for the checkbox component](https://www.w3.org/TR/wai-aria-1.1/#checkbox). The checkbox is an HTMl input element with added attributes to ensure proper functionality and accessibility. The checkbox is wrapped in an `asInput` component. The `asInput` wrapper is passed the input element as well as a second parameter set to `false` to ensure that the label is placed to the right of the checkbox.
+
+The following inputs should be passed into every checkbox component:
+* `name`: `name` attribute - type `String`
+* `label`: label to be placed next to the checkbox - type `String`
+* `checked`: `"true"` if the default state should be checked, `"false"` otherwise - type `String`
+
+The following input can optionally be passed into a checkbox component:
+* `disabled`: `true` if the checkbox should be disabled, `false` otherwise - type `Boolean`
+* `onChangeState`: function to be called when the checkbox changes state
+
+The implementation of the checkbox contains the following functions:
+* `constructor()`: The constructor sets the `id` for the checkbox and sets whether the initial state should be checked or unchecked
+* `handleClick()`: Switches the state of the checkbox; also calls the `onChangeState()` function if one has been passed in
+* '`render()`: Returns the checkbox as an input element with seven attributes: `id`, `type`, `name`, `defaultChecked`, `aria-checked`, `onClick`, and `disabled`.
--- a/src/CheckBox/README.md
+++ b/src/CheckBox/README.md
@@ -2,13 +2,13 @@
 
 Checkbox based off of the [WAI-ARIA authoring guidelines for the checkbox component](https://www.w3.org/TR/wai-aria-1.1/#checkbox). The checkbox is an HTMl input element with added attributes to ensure proper functionality and accessibility. The checkbox is wrapped in an `asInput` component. The `asInput` wrapper is passed the input element as well as a second parameter set to `false` to ensure that the label is placed to the right of the checkbox.
 
-The following inputs should be passed into every checkbox component:
-* `name`: `name` attribute - type `String`
-* `label`: label to be placed next to the checkbox - type `String`
-* `checked`: `"true"` if the default state should be checked, `"false"` otherwise - type `String`
+The following parameters should be passed into every checkbox component:
+* `name` (`String`): `name` attribute
+* `label` (`String`): label to be placed next to the checkbox
 
-The following input can optionally be passed into a checkbox component:
-* `disabled`: `true` if the checkbox should be disabled, `false` otherwise - type `Boolean`
+The following parameters can optionally be passed into a checkbox component:
+* `checked` (`Boolean`): `true` if the default state should be checked, `false` otherwise
+* `disabled` (`Boolean`): `true` if the checkbox should be disabled, `false` otherwise
 * `onChangeState`: function to be called when the checkbox changes state
 
 The implementation of the checkbox contains the following functions:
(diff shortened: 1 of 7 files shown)


src/asInput/index.jsx
```diff
@@ -85,10 +85,28 @@ const asInput = (WrappedComponent) => {
...
+              ]}
+              describedBy={describedBy}
+              onChange={this.handleChange}
+              onBlur={this.handleBlur}
+            />
+            {error}
+            {description}
+          </div>
+        );
+      }
```
arizzitano: Copy-pasting this block of code is the fastest solution, but that doesn't mean it's the best solution. When you write software, you should strive to keep it as DRY (http://programmer.97things.oreilly.com/wiki/index.php/Don%27t_Repeat_Yourself) as possible. Both this `WrappedComponent` and its label now exist in two places within a few lines of each other. Duplicating data and logic makes future maintenance unnecessarily complicated (a future maintainer would need to update `WrappedComponent` or the <label> in both places). Please store both of these in variables.

Multiple return statements can also be problematic, although it depends how they're used. Typically you use multiple returns if you need to exit out of a loop or a particularly long control chain, but not in a case like this. [This article](https://www.tomdalling.com/blog/coding-tips/coding-tip-have-a-single-exit-point/) gives a pretty good summary of why it's an antipattern. Instead, your if statement here should modify the variable you store `WrappedComponent` in (either appending or prepending the label based on the value of `labelFirst`) so you can return the whole thing in a single return statement at the bottom of this render method.

src/CheckBox/index.jsx
```diff
@@ -46,24 +46,26 @@ class CheckBox extends React.Component {
...
     );
   }
 }
 
-CheckBox.propTypes = inputProps;
+Check.propTypes = inputProps;
+
+const CheckBox = asInput(Check, false);
+
+CheckBox.propTupes = {
```
arizzitano: Careful here. What do `Check.propTypes` and `CheckBox.propTypes` contain?

src/CheckBox/README.md
```diff
@@ -0,0 +1,17 @@
+# Checkbox Component
+
+Checkbox based off of the [WAI-ARIA authoring guidelines for the checkbox component](https://www.w3.org/TR/wai-aria-1.1/#checkbox). The checkbox is an HTMl input element with added attributes to ensure proper functionality and accessibility. The checkbox is wrapped in an `asInput` component. The `asInput` wrapper is passed the input element as well as a second parameter set to `false` to ensure that the label is placed to the right of the checkbox.
+
+The following inputs should be passed into every checkbox component:
+* `name`: `name` attribute - type `String`
+* `label`: label to be placed next to the checkbox - type `String`
+* `checked`: `"true"` if the default state should be checked, `"false"` otherwise - type `String`
```
arizzitano: Can you remind me why `checked` is a string and not a boolean?

src/CheckBox/README.md
```diff
@@ -0,0 +1,17 @@
+# Checkbox Component
+
+Checkbox based off of the [WAI-ARIA authoring guidelines for the checkbox component](https://www.w3.org/TR/wai-aria-1.1/#checkbox). The checkbox is an HTMl input element with added attributes to ensure proper functionality and accessibility. The checkbox is wrapped in an `asInput` component. The `asInput` wrapper is passed the input element as well as a second parameter set to `false` to ensure that the label is placed to the right of the checkbox.
+
+The following inputs should be passed into every checkbox component:
+* `name`: `name` attribute - type `String`
```
arizzitano: :+1: for declaring the type beside each of these props, although I'd place it at the beginning of the line, e.g.:
```
`name` (string): `name` attribute
```

src/CheckBox/index.jsx
```diff
@@ -36,8 +36,8 @@ class CheckBox extends React.Component {
         checked: true,
       });
     }
-    if (this.onChange) {
-      this.onChange();
+    if (this.onChangeState) {
+      this.onChangeState();
```
arizzitano: Take a look at `asInput`'s `handleChange` function. It fires whatever `onChange` function has been passed to its descendant component, so it should handle this for you automatically. What happens if you change your parameter name back to `onChange` and remove this `onChangeState` call from the checkbox component?

arizzitano: I'm also curious what would happen if you try just commenting out this `handleClick` method entirely. I built `asInput` with the intention that it would handle all change events for its descendant components, so the descendants wouldn't have to mess around with their own logic at all. Does `asInput`'s `handleChange` fire when this checkbox's value changes?

src/CheckBox/README.md
sarahkf: I'm don't think there is an actual reason. I changed it to a boolean which makes more sense and also allows me to write more concise code.

src/CheckBox/index.jsx
sarahkf: So, this seems to be a little complicated. In the `asInput` wrapper, the `onChange` attribute is used instead of `onClick` like in the checkbox. It seems like `onChange` is not called when the checkbox is clicked. So, I’m not sure if we can get rid of the `handleClick` function. 

Further, `asInput`’s `onChange` is hit when the checkbox calls the`console.log(“the checkbox changed state”)` function. Hence why I had to change the name from `onChange` to `onChangeState` because when the `console.log(“the checkbox changed state”)` function is called, asInput’s `onChange` handler is called which then calls the passed in `onChange` function a second time and I get errors.

src/CheckBox/index.jsx
```diff
@@ -1,69 +1,62 @@
...
+};
+
+Check.defaultProps = {
+  checked: false,
+  onChange: () => {},
+};
+
+const CheckBox = asInput(Check, false);
+
+CheckBox.propTypes = {
```
arizzitano: You should be able to remove this propTypes entirely. The propTypes and defaultProps above aren't ideal, but that's not your fault and it's something I can solve in asInput later on.

src/CheckBox/index.jsx
```diff
@@ -1,69 +1,62 @@
...
-        checked: true,
-      };
-    } else {
-      this.state = {
-        id,
-        checked: false,
-      };
-    }
+    this.state = {
+      id,
```
arizzitano: is this `id` being used anywhere? if not, you can remove it (and the `const id` above, and the `newId` import)

sarahkf: When I get rid of the id, I lose the functionality where when you click on the label the checkbox changes state.

src/CheckBox/index.jsx
sarahkf: So, should I add ignore rules to the linter?

PR #20, opened 2017-08-23
prepare to publish to npm

arizzitano:


arizzitano:
@efischer19 good call! I added it but Github was a little slow showing the change... https://github.com/edx/paragon/pull/20/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R20

PR #21, opened 2017-08-25
[DONTMERGE] convert to independently versioned monorepo

arizzitano:
Update: we are back burnering this for now as I'm a little worried about the surface area. Not ruling out Lerna/monorepo for future work, but it's just not a priority for the work we're doing right now. I'm going to leave the PR up for reference, though.

(god, that's a scary sounding PR title, isn't it?)

Currently exploring [lerna](https://github.com/lerna/lerna) to facilitate the release of components as individual packages. I'll explain my reasoning later.


arizzitano:
WONTMERGE, so closing this. 
image

PR #32, opened 2017-09-12
table component

arizzitano:
Adds a very basic table with tests and stories to demo functionality.

Looks like this:
image

For details on the table's API, check out [the readme](https://github.com/edx/paragon/blob/37a01bc388ca252233d4c24236260b65b7210b66/src/Table/README.md).

### DIFF
--- a/.storybook/__snapshots__/Storyshots.test.js.snap
+++ b/.storybook/__snapshots__/Storyshots.test.js.snap
@@ -500,7 +500,7 @@ exports[`Storyshots Paragon Welcome 1`] = `
 
 exports[`Storyshots Table default heading 1`] = `
 <table
-  className="table table-responsive"
+  className="table"
 >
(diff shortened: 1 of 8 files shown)


.storybook/__snapshots__/Storyshots.test.js.snap
```diff
@@ -498,6 +498,358 @@ exports[`Storyshots Paragon Welcome 1`] = `
 </div>
 `;
 
+exports[`Storyshots Table default heading 1`] = `
+<table
+  className="table"
+>
+  <caption>
+    Famous Internet Cats
```
thallada: Best example ever :cat:

arizzitano:
Yeah, good point about using a vendor library @thallada. No use reinventing the wheel if we don't need to. Let's reevaluate once we get to that point.

PR #33, opened 2017-09-19
patch dropdown for bootstrap v4

arizzitano:
No functional change here, just some changes to classnames and markup that were affected by the release of Bootstrap v4 beta.


src/Dropdown/index.jsx
```diff
@@ -138,14 +137,17 @@ class Dropdown extends React.Component {
           type="button"
           inputRef={(toggleElem) => { this.toggleElem = toggleElem; }}
         />
-        <ul
+        <div
```
arizzitano: Although the convention is to use a `ul` for dropdown lists, @cptvitamin says it's just fine not to.

PR #34, opened 2017-09-21
add a modal component

MichaelRoytman:


src/Modal/index.jsx
```diff
@@ -0,0 +1,140 @@
...
+            <div className={styles['modal-body']}>
+              <p>{this.props.body}</p>
+            </div>
+            <div className={styles['modal-footer']}>
+              {this.renderButtons()}
+              <Button
+                display="Close"
+                buttonType="secondary"
+                onClick={this.close}
+                inputRef={(input) => { this.buttons[this.props.buttons.length + 1] = input; }}
```
MichaelRoytman: Is this okay to do? What if the user/developer wants buttons to appear dynamically? Will props change accordingly?

src/Modal/index.jsx
```diff
@@ -0,0 +1,167 @@
...
+      }
+    }
+  }
+
+  renderButtons() {
+    return this.props.buttons.map((button, i) => {
+      let buttonElement = button;
+      let buttonProps = button.props;
+
+      if (button.type !== Button) {
```
MichaelRoytman: This is a little odd looking, but essentially I found that if a button is passed in as a prop as a Button object, the Button should be constructed with the Button's props. Otherwise, you can just create a Button object using the object that is passed in and has the shape of the Button propTypes.

src/Modal/index.jsx
```diff
@@ -0,0 +1,167 @@
...
+        buttonProps = button;
+      }
+
+      buttonElement = (<Button
+        {...buttonProps}
+        key={i}
+        inputRef={(input) => { this.buttons[i + 1] = input; }}
+        onKeyDown={this.handleKeyDown}
+      />);
```
MichaelRoytman: I had to end up creating each button passed in as a prop, even Button objects, because I had to add props for the proper functionality of the modal (i.e. key, inputRef, and onKeyDown).

src/Modal/index.jsx
```diff
@@ -0,0 +1,167 @@
...
+        <div className={styles['modal-dialog']} role="dialog" >
+          <div className={styles['modal-content']}>
+            <div className={styles['modal-header']}>
+              <h5 className={styles['modal-title']}> {this.props.title} </h5>
+              <Button
+                display="&times;"
+                aria-label="Close"
+                buttonType="light"
+                onClick={this.close}
+                inputRef={(input) => { this.buttons[0] = input; }}
```
MichaelRoytman: I assume the X button will always be the first button to be in focus, so I hard-coded the index here.

src/Modal/index.jsx
MichaelRoytman: As a follow on, originally I had attempted to do something like inputRef={(input) => { this.buttons[this.buttons.length + 1] = input; }}. This was having un-intended side-effects, where the buttons array would grow and have null buttons. I think it's because the renderButtons function is executed asynchronously, and therefore, the close button reference may be inserted in to the array before the other buttons have been rendered, causing weird array structure.

src/Modal/index.jsx
```diff
@@ -0,0 +1,167 @@
...
+  // open() {
+  //   this.setState({ open: true });
+  //   this.closeButton.focus();
+  // }
+
+  handleKeyDown(e) {
+    if (e.key === 'Escape') {
+      this.close();
+    } else if (e.key === 'Tab') {
+      e.preventDefault();
```
MichaelRoytman: I'm using preventDefault here instead of at the top of the function so that the default enter/space functionality is maintained for button selection.

src/Modal/index.jsx
```diff
@@ -0,0 +1,167 @@
...
+  // what does true mean here?
+  addEvents() {
+    document.addEventListener('click', this.handleDocumentClick, true);
+  }
+
+  removeEvents() {
+    document.removeEventListener('click', this.handleDocumentClick, true);
+  }
+
+  handleDocumentClick(e) {
```
MichaelRoytman: Right now, this just ensures that the focusIndex is updated when someone clicks on a button and the focus moves to that button.

src/Modal/Modal.stories.jsx
```diff
@@ -0,0 +1,117 @@
...
+          buttonType: 'danger',
+        },
+        <Button
+          display="Green button!"
+          buttonType="success"
+        />,
+      ]}
+    />
+  ))
+  .add('configurable title and body (creppy edition)', () => (
```
MichaelRoytman: I'll fix this to be something more professional.

src/Modal/index.jsx
```diff
@@ -0,0 +1,131 @@
...
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.open !== this.props.open) {
+      this.setState({ open: nextProps.open });
+    }
+  }
+
+  componentDidUpdate(prevState) {
+    if (this.state.open && !prevState.open) {
+      this.xButton.focus();
```
MichaelRoytman: I'm unsure about this. My intent was to put focus on the X button when the modal is opened via a button and not on the initial render.

MichaelRoytman:
@arizzitano I think this is good for a look when you have some time.

TODO: Figure out the ```Warning: Failed prop type: Invalid prop `buttons[0]` supplied to `Modal`. in Modal``` warning in Storybook.
TODO: Write tests.
TODO: Accessibility review.
TODO: More?

src/Button/index.jsx
```diff
@@ -37,7 +37,7 @@ function Button(props) {
   );
 }
 
-Button.propTypes = {
+export const buttonPropTypes = {
```
MichaelRoytman: This is so we can reference the buttonPropTypes as part of the Modal PropTypes.

src/Modal/Modal.stories.jsx
```diff
@@ -0,0 +1,117 @@
...
+  .add('configurable buttons that do things', () => (
+    <Modal
+      open
+      title="Modal title"
+      body="Modal body"
+      buttons={[
+        <Button
+          display="I am a button!"
+          buttonType="light"
+          onClick={() => console.log('I made a log!')}
```
MichaelRoytman: ESLint complains about this log statement :information_desk_person:.

src/Modal/index.jsx
```diff
@@ -0,0 +1,133 @@
...
+    );
+  }
+}
+
+Modal.propTypes = {
+  open: PropTypes.bool,
+  title: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
+  body: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
+  buttons: PropTypes.arrayOf(PropTypes.oneOfType([
+    PropTypes.element,
```
MichaelRoytman: TODO: This could use a custom validator.

MichaelRoytman:
I'm not sure what's going on with the Button coverage?

src/Modal/Modal.test.jsx
```diff
@@ -0,0 +1,196 @@
...
+  describe('invalid keystrokes do nothing', () => {
+    beforeEach(() => {
+      wrapper = mount(
+        <Modal
+          {...defaultProps}
+        />,
+      );
+    });
+
+    it('does nothing on invalid keystroke q', () => {
```
arizzitano: Lol I love this. I kind of feel like "invalid keystroke q" is going to become the go-to for coverage fixes.

src/Modal/Modal.test.jsx
```diff
@@ -0,0 +1,196 @@
...
+      buttons.last().simulate('keyDown', { key: 'Tab' });
+      expect(buttons.at(0).matchesElement(document.activeElement)).toEqual(true);
+    });
+    it('traps focus backwards on shift + tab keystroke', () => {
+      expect(buttons.at(0).matchesElement(document.activeElement)).toEqual(true);
+      buttons.at(0).simulate('keyDown', { key: 'Tab', shiftKey: true });
+      expect(buttons.last().matchesElement(document.activeElement)).toEqual(true);
+    });
+  });
+});
```
arizzitano: Great coverage here 🙏

src/Modal/index.jsx
arizzitano: I'm kinda split on this. On the one hand, there's out-of-the-box validators we could use to make this easy (https://www.npmjs.com/package/react-element-proptypes), but on the other hand HRH Abramov [discourages typechecking that's this granular](https://github.com/facebook/react/pull/7119#issuecomment-228615386). I'm inclined to say just leave it as-is -- it'll make things easier if we end up adopting something like [Flow](https://flow.org/) instead of propTypes for typechecking.

src/Modal/index.jsx
```diff
@@ -0,0 +1,133 @@
...
+    PropTypes.shape(buttonPropTypes),
+  ])),
+  onClose: PropTypes.func,
```
arizzitano: I just realized something: we've hardcoded the string `Close` in a couple places within this modal (lines 88 and 101), which isn't super flexible with regard to i18n. Could you add another optional prop `closeText` with a default of `'Close'`, so consumers can override it if necessary? Probably would be good to add a story to demo this as well.

src/Modal/Modal.stories.jsx
```diff
@@ -0,0 +1,117 @@
...
+        />,
+      ]}
+    />
+  ))
+  .add('modal invoked via a button', () => (
+    <ModalWrapper
+      title="I am the modal!"
+      body="I was invoked by a button!"
+    />
+  ));
```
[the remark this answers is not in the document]
MichaelRoytman: That's a good idea! Do you think adding, say, a button to the above example that would, theoretically, send you an e-mail with a book (theoretically, as I have no book to send), would be useful? I'd, of course, create some kind of story in which a button would actually be functional and wouldn't required sending e-mail. It would require another wrapper class, though, akin to ModalWrapper, and I fear that may be too complicated for the purposes of the storybook. I was thinking something along the lines of... select your favorite month, click a button, and I'll tell you what season it's in (maybe with a coordinating emoji), or something silly like that. Either way is fine with me!

src/Modal/Modal.scss
```diff
@@ -0,0 +1,5 @@
+@import "~bootstrap/scss/_modal";
+
+.modal-open {
+    display: block;
+}
```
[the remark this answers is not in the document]
MichaelRoytman: Done; thanks for the helpful hint!

src/Modal/index.jsx
MichaelRoytman: Sounds good to me!

MichaelRoytman:
@arizzitano This should be good for a re-review when you have a chance. I believe I've addressed all of your comments.

src/Modal/index.jsx
```diff
@@ -0,0 +1,146 @@
...
+  render() {
+    return (
+      <div className={classNames(styles.modal, styles.show, { [styles['modal-open']]: this.state.open })} role="dialog" aria-modal aria-labelledby={this.headerId}>
+        <div className={styles['modal-dialog']}>
+          <div className={styles['modal-content']}>
+            <div className={styles['modal-header']}>
+              <h5 className={styles['modal-title']} id={this.headerId}>{this.props.title}</h5>
+              <Button
+                display="&times;"
+                aria-label="Close"
```
arizzitano: I'd change this aria-label to `{this.props.closeText}` as well.

PR #42, opened 2017-10-03
Create a Status Alert

fysheets:
Create a new component for StatusAlert that is accessible and reusable.


src/StatusAlert/StatusAlert.stories.jsx
```diff
@@ -0,0 +1,124 @@
...
+  .add('Non-dismissible alert', () => (
+    <StatusAlert
+      alertType="danger"
+      dismissible={false}
+      dialog="You can't get rid of me!"
+      onClose={() => {}}
+      open
+    />
+  ))
+  .add('modal invoked via a button', () => (
```
MichaelRoytman: I think you mean alert here.

src/StatusAlert/StatusAlert.stories.jsx
```diff
@@ -0,0 +1,124 @@
...
+    <StatusAlertWrapper
+      alertType="success"
+      dialog="Success! You triggered the alert!"
+    />
+  ))
+  .add('modal with a link', () => (
```
MichaelRoytman: Alert here, too!

src/StatusAlert/index.jsx
```diff
@@ -0,0 +1,125 @@
...
+    return (
+      <div className="alert-dialog">
+        { dialog }
+      </div>
+    );
+  }
+
+  renderDismissible() {
+    const { dismissible } = this.props;
+    let dismissButton = (<div />);
```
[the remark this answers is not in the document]
fysheets: I can definitely give this a go -- it didn't cross my mind that null was a valid thing that could be returned as part of a render.

src/StatusAlert/index.jsx
```diff
@@ -0,0 +1,125 @@
...
+    if (dismissible) {
+      dismissButton = (
+        <button
```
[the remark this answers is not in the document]
fysheets: Noting down our discussion -- the dismissible button formatting is generated from both the alert & close SASS files in order to render properly -- Will try to include the close partial into Button to see if it will continue to render correctly.

src/StatusAlert/StatusAlert.stories.jsx
fysheets: Thank you Michael :)

src/StatusAlert/index.jsx
```diff
@@ -0,0 +1,125 @@
...
+    );
+  }
+}
+
+StatusAlert.propTypes = {
+  alertType: PropTypes.string,
+  className: PropTypes.arrayOf(PropTypes.string),
+  dialog: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
+  dismissible: PropTypes.bool,
+  onClose: PropTypes.func.isRequired,
```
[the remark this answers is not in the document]
fysheets: That totally makes sense and that package looks interesting! I will read up and see if we can leverage it.

src/StatusAlert/index.jsx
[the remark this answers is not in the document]
fysheets: @arizzitano I updated this to be null instead of the empty div.  When I made my first attempt to turn it into a ternary operator it didn't render happily with the <button> spanning over multiple lines.  I will revisit this though when I test out the options to switch to the paragon Button from the above comment.

.storybook/__snapshots__/Storyshots.test.js.snap
```diff
@@ -982,6 +982,179 @@ exports[`Storyshots Paragon Welcome 1`] = `
...
+    className="alert-dialog"
+  >
+    You can't get rid of me!
+  </div>
+</div>
+`;
+
+exports[`Storyshots StatusAlert alert invoked via a button 1`] = `
+<div>
+  <div
```
fysheets: @arizzitano while playing with the screenreader yesterday, I think that I need to add an aria-hidden on the status alert that is tied with the "open" state field, but when I did my first attempt it didn't appear to work as I had expected.  Advice on the logic that should go here?

fysheets: Spoke with Mark -- he noted that we could leverage `display: 'none'` and `display: 'block'` for this instead of `aria-hidden` so I am looking into how to correctly integrate that logic into classNames

src/StatusAlert/README.md
```diff
@@ -0,0 +1,23 @@
...
+### `className` (string array; optional)
+`className` is a string array that defines the classes to be used within the status alert.
+
+### `dialog` (string or element; required)
+`dialog` is a string or an element that is rendered inside of the status alert as the main data.
+
+### `dismissible` (boolean; optional)
+`dismissible` specifies if the status alert will include the dismissible X button to close the alert. It defaults to true.
+
+### `onClose` (function; required)
```
arizzitano: Can you add a note about `onClose` not being required if `dismissible` is true?

src/StatusAlert/index.jsx
```diff
@@ -0,0 +1,120 @@
...
+          styles.fade,
+        ], {
+          [styles['alert-dismissible']]: dismissible,
+        }, {
+          [styles[`alert-${alertType}`]]: alertType !== undefined,
+        }, {
+          [styles.show]: this.state.open,
+        })}
+        role="alert"
+        hidden={!this.state.open}
```
arizzitano: Great find re: `hidden`! 👏👏👏

src/StatusAlert/README.md
fysheets: For sure -- how does this look:

### `onClose` (function; conditionally required)
`onClose` is a function that is called on close. It can be used to perform actions upon closing of the status alert, such as restoring focus to the previous logical focusable element.  It is only required if `dismissible` is set to `true` and not required if the alert if not `dismissible`.

arizzitano: Perfect!

PR #50, opened 2017-10-18
feat(build): static and themeable production builds

arizzitano:
### WHAT

Two separate production builds for Paragon, both exported from the same `@edx/paragon` package. The main build, `themeable`, is intended for use within Bootstrap environments. Components come with stock Bootstrap classnames and will inherit any existing Bootstrap styles and themes on consumers' pages. The alternative build, `static`, is intended for use within legacy code. It is less customizable, but contains protection against CSS conflicts.

### WHY

Paragon components need to work consistently across consumers' various different runtime environments. Currently, consumers skirt this by importing source files directly (from `/src`), but this requires special webpack adjustments in the consuming applications, which is messy and brittle.

Requirements for production builds are as follows:
- component styles should not conflict with the styles on the rest of the page, and vice versa
- components should be themeable
- components should have consistent top-level classnames for integration testing purposes
- consumers should not need to customize their builds for Paragon

### HOW

On build, we provide multiple export targets:
```
paragon
├─┬ static
│ ├── index.js -- scoped classnames for all styles
│ └── paragon.min.css -- css classes scoped to the components, including bootstrap styles
└─┬ themeable
  ├── index.js -- plaintext bootstrap classnames, scoped top-level classnames
  └── paragon.min.css -- contains ONLY components' custom, non-bootstrap styles
```

To import static components:
```
import { Button } from @edx/paragon/static;
```
To import themeable (Bootstrap-ready plaintext classname) components:
```
import { Button } from @edx/paragon;
```

### OPEN QUESTIONS

1. What should we name the different builds?
- `static` vs `themeable` (my personal favorite)
- `scoped` vs `unscoped` (original idea, maybe too generic)
- `scoped` vs `bootstrap` (a red herring, since they both use bootstrap)

2. What should be the default export (that is, the one you get when you `import { Component } from @edx/paragon`)? I think themeable should be the default. Static components are intended to be a stop-gap measure whereas themeable ones are our bread and butter.

3. Should the static components inject their own styles into the page (via style-loader, which can cause a FOUC, but makes asset management MUCH easier for consumers) or build separate assets (a stylesheet and font awesome assets) for consumers to work into their own asset bundling processes?


arizzitano:
I've been publishing this as `@arizzitano/paragon@0.0.1-development-{rev}` to test in other envs.

arizzitano:
@edx/educator-dahlia / @edx/fedx-team, I'd love to get your thoughts on this PR, primarily around the open questions. Fortunately, this change is non-breaking and won't affect any of the existing Paragon embeds, but I'd like to start moving consuming IDAs (currently just SFE and edx-platform) to start using the Paragon production builds as soon as we can.

cc @tuchfarber

MichaelRoytman:
**Open Questions:**

1. I like the first suggestion as well. I agree that the last suggestion contains a red herring, and the second suggestion, although it pleases me more because it feels more semantically symmetric, does leave a lot to the imagination (i.e. what does "unscoped" mean?). I think the first is clear and specific.
2. I would personally expect themeable to be the default, since it is the one that we should be actively moving toward, while static is to support embedding Paragon components into legacy code, if I'm understanding correctly.
3. Hm. I'm not sure that I am familiar enough with the considerations for this question. I guess I'd be interested to know how others feel about FOUC and whether that's enough of a reason to make asset management more difficult to avoid it. 

I'm very excited for this. If I'm understanding this correctly, having a separate themeable build will allow for studio-frontend to have a much easier time styling components. This would allow a consumer to rely on/leverage the stock Bootstrap classnames in their SCSS files (for adding additional styling or overrides), right? I don't think we should have any problems adopting the themeable build into the existing pages in studio-frontend, if that is the case.

config/webpack.static.config.js
```diff
@@ -0,0 +1,50 @@
...
+      {
+        test: /\.scss|\.css$/,
+        use: ExtractTextPlugin.extract({
+          use: [
+            {
+              loader: 'css-loader',
+              options: {
+                modules: true,
+                localIdentName: 'paragon__[local]',
+                sourceMap: true,
```
thallada: Shouldn't there also be [`minimize: true`](https://github.com/webpack-contrib/css-loader#minimize) here so the CSS is actually minimized?

arizzitano: Good catch!

config/webpack.config.js
```diff
@@ -0,0 +1,68 @@
...
+const MinifyPlugin = require('babel-minify-webpack-plugin');
+
+const staticConfig = require('./webpack.static.config');
+const themeableConfig = require('./webpack.themeable.config');
+
+const commonConfig = {
+  // 'source-map' raises errors (something about the combination)
+  // of MinifyPlugin and ExtractTextPlugin) -- see:
+  // https://github.com/webpack-contrib/babel-minify-webpack-plugin/issues/68
+  devtool: 'cheap-source-map',
```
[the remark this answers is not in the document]
arizzitano: I originally picked babel-minify (nee babili) over uglify because uglify (at the time -- last summer) came with some nasty compatibility issues. I'll give uglify a shot and see if it does any better.

arizzitano:
Static size:
image

Themeable size:
image

Both clock in <100k for the JS, which isn't so bad -- obviously not as small as letting consumers import each individual component as necessary, but IMO the reduced cognitive overhead is a worthwhile payoff. We can finetune filesize and build details as we go -- for now my main concern is getting this out so we can start reworking consumers' webpack configs.

thallada:
Yeah definitely :+1: it looks like the difference between the builds is pretty minimal.

webpack.config.js
```diff
@@ -1,114 +1,111 @@
 const path = require('path');
-const MinifyPlugin = require('babel-minify-webpack-plugin');
+const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
 
-const env = process.env.NODE_ENV || 'dev';
+// we build the library for two different build targets:
+// static (with scoped styles) and themeable (with stock,
+// overrideable classnames)
+const targetProperties = [{
```
arizzitano: I know I could have written this as a one-liner, but I split it out into its own variable for clarity's sake.

arizzitano:
Thanks for the comments @thallada and @MichaelRoytman. I addressed them and also reorganized the separate configs into one single file (on @cpennington's recommendation) if you'd like to take another look.

PR #51, opened 2017-10-19
sortable table

MichaelRoytman:


src/Table/Table.test.jsx
```diff
@@ -17,6 +17,35 @@ const props = {
...
+  },
+  sq: {
+    columnSortable: false,
+  },
+};
+
+const sortableColumns = props.columns.map(column => ({
+  ...column,
+  // columnSortable: column.key !== 'sq',
+  // onSort: () => {},
```
MichaelRoytman: Oops. I nee to remove this.

src/Table/index.jsx
```diff
@@ -73,14 +170,22 @@ Table.propTypes = {
       PropTypes.string,
       PropTypes.element,
     ]).isRequired,
+    columnSortable: isRequiredIf(PropTypes.bool, props => props.tableSortable),
```
MichaelRoytman: I'm not sure that this conditional require and the one below it actually work correctly.

src/Table/index.jsx
```diff
@@ -1,16 +1,110 @@
...
 import PropTypes from 'prop-types';
 
 import styles from './Table.scss';
+import Button from '../Button';
 
 class Table extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = props.tableSortable ?
```
MichaelRoytman: If the table is sortable, set the appropriate state variable to the defaultSortedColumn and defaultSortDirection; otherwise, leave both empty.

MichaelRoytman:
TODO: Update README!

src/Table/index.jsx
MichaelRoytman: I don't believe that this conditional requirement works correctly, since it refers outside of column. I've been able to exclude this prop when initializing a sortable table without propTypes complaining.

src/Table/Table.stories.jsx
```diff
@@ -46,6 +46,17 @@ const catColumns = [
   },
 ];
 
+const sort = function sort(firstElement, secondElement, key, direction) {
```
MichaelRoytman: This is just a utility function to sort catData by key and in a particular direction, since Javascript's Arrays.sort() function doesn't support sorting objects, obviously.

src/Table/Table.stories.jsx
```diff
@@ -77,4 +88,24 @@ storiesOf('Table', module)
       caption="Famous Internet Cats"
       className={['table-responsive']}
     />
-  ));
+  ))
+  .add('sortable', () => {
+    const catDataSortable = catData.slice();
```
MichaelRoytman: I'm making a copy to keep the other stories unaffected by the sorting in this story.

src/Table/Table.stories.jsx
```diff
@@ -77,4 +88,24 @@ storiesOf('Table', module)
...
+
+    return (<Table
+      data={catDataSortable.sort((firstElement, secondElement) => sort(firstElement, secondElement, catColumns[0].key, 'desc'))}
```
MichaelRoytman: This just performs an initial sort so that the table is initially displayed in some kind of default sorted order. Something we should consider is whether or not we expect the user to pass in pre-sorted data. I assume yes, since we don't want to do client-side sorting, but I just wanted to double-check.

src/Table/Table.stories.jsx
```diff
@@ -77,4 +88,24 @@ storiesOf('Table', module)
       caption="Famous Internet Cats"
       className={['table-responsive']}
     />
-  ));
+  ))
+  .add('sortable', () => {
```
MichaelRoytman: I wanted to give an example of an actual sortable table here, so I had to do some hocus-pocus to actually sort the data here.

src/Table/Table.test.jsx
```diff
@@ -17,6 +17,33 @@ const props = {
   ],
 };
 
+const sortableColumnProps = {
```
MichaelRoytman: All these variables are used for creating props for a sortable table.

src/Table/index.jsx
```diff
@@ -1,16 +1,110 @@
...
+    this.onSortClick = this.onSortClick.bind(this);
+  }
+
+  onSortClick(columnKey) {
+    let newDirection = 'desc';
+
+    if (this.state.sortedColumn === columnKey) {
+      newDirection = (this.state.sortDirection === 'desc' ? 'asc' : 'desc');
+
+      this.setState({
```
MichaelRoytman: There are two setStates here that have some redundnacy in the `sortDirection: newDirection` line. I could either leave them this way, so that there is only one setState call per sort click, or I can have a setState at line 38 that only sets `sortedColumn: columnKey`, and then another setState at line 43 that sets `sortDirection: newDirection`. There would be the potential for two sortState calls on one sortClick in that case. I believe setStates are batched together, so I'm not sure it would be a performance issue with regards to re-rendering, but just wanted to double-check here.

src/Table/index.jsx
```diff
@@ -1,16 +1,110 @@
...
+    currentlySortedColumn.onSort(newDirection);
+  }
+
   getCaption() {
     return this.props.caption && (
       <caption>{this.props.caption}</caption>
     );
   }
 
+  getHeadingText(columnKey) {
```
MichaelRoytman: This gets the sr-only text for the heading. I'm thinking this function name could be more clear.

src/Table/index.jsx
```diff
@@ -1,16 +1,110 @@
...
+    if (this.state.sortedColumn === columnKey) {
+      text = this.state.sortDirection === 'desc' ? 'sort descending' : 'sort ascending';
+    } else {
+      text = 'click to sort';
+    }
+
+    return text;
+  }
+
+  /* eslint-disable class-methods-use-this */
```
MichaelRoytman: Do we want this in the eslintrc, like it is in studio-frontend?

src/Table/index.jsx
fysheets: I may be missing some blunt logic, but is there a downside to resetting the `sortColumn` to `columnKey` when it's the same column being toggled? So only have one `setState` that sets both values regardless of which column is being toggled?

src/Table/index.jsx
fysheets: Maybe something like `getA11yHeadingText`?

src/Table/Table.test.jsx
```diff
@@ -36,5 +63,198 @@ describe('<Table />', () => {
...
+      expect(numSpy).toBeCalledWith('desc');
+
+      buttons.at(1).simulate('click');
+
+      expect(numSpy).toHaveBeenCalledTimes(2);
+      expect(x2Spy).toHaveBeenCalledTimes(1);
+
+      expect(x2Spy).toBeCalledWith('desc');
+    });
   });
```
fysheets: Nice tests! 👍

src/Table/index.jsx
MichaelRoytman: Oh, that's a very good point. I guess I was trying to focus too much on what's the absolute minimum change, but that change would simplify the code at a low cost. I guess if the page is going to re-render as a result of the set state anyway, it doesn't matter if I change only some or all of the state. Thanks!

src/Table/index.jsx
MichaelRoytman: That works for me. I think I'll wait and see whether the table needs any additional accessibility work first, and rename things after, once I know they're actually covering accessibility cases.

src/Table/index.jsx
```diff
@@ -1,16 +1,110 @@
...
+    this.state = props.tableSortable ?
+      {
+        sortedColumn: this.props.defaultSortedColumn,
+        sortDirection: this.props.defaultSortDirection,
+      }
+      :
+      {
+        sortedColumn: '',
+        sortDirection: '',
+      };
```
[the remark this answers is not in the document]
MichaelRoytman: This looks a lot nice. I think I'll do the ternary, just because that's what I've used before, but if it's stylistically more consistent for me to use the short-circuit, I can do that as well. 

I think I just didn't realize I could do it for each individual attribute of the state and had to do on the actual object level, which is why I had the ugly way of doing it.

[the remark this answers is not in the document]
arizzitano: Short-circuits are janky and some purists (not me tho) frown on them, so the ternary is a good instinct 😛

src/Table/index.jsx
```diff
@@ -1,16 +1,110 @@
...
+      case 'asc':
+        sortIconClassName = 'fa-sort-asc';
+        break;
+      case 'desc':
+        sortIconClassName = 'fa-sort-desc';
+        break;
+      default:
+        sortIconClassName = 'fa-sort';
+        break;
+    }
```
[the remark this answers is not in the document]
MichaelRoytman: That sounds good. Does ```['fa-sort', sortDirection].join('-');``` work as well? I tried it in the console it worked the same way. Does the filter have a purpose I'm just not seeing?

src/Table/index.jsx
MichaelRoytman: I think so. I'm not positive, because the [documentation](https://github.com/evcohen/react-proptype-conditional-require) isn't super clear. The conditional (props => props.tableSortable in this case) has access to props, propName, and componentName. If I console.log those from the conditional function, props is

image

which are the props for the column and not the actual Table :(. Should I remove the isRequiredIf if it's not really doing anything? Should I replace it with something? A note in the docs?

src/Table/index.jsx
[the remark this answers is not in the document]
MichaelRoytman: Just kidding, I figured it out. Never mind!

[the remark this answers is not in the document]
arizzitano: Yeah, `n => n` removes nulls/undefineds/falsies. So if `sortDirection` is undefined it won't add a trailing hyphen!

src/Table/index.jsx
arizzitano: I just cut `eslint-config-edx@4.0.1` so you can add that whenever and then these `/* eslint-disable class-methods-use-this */` statements can be removed.

MichaelRoytman:
@arizzitano I have addressed your comments, so if you could take a look and let me know if it's good to go, that would be appreciated. Thanks!

src/Table/index.jsx
arizzitano: Just to follow up: this looks like an issue with the actual library and I'm going to see if I can help patch it upstream.

src/Table/index.jsx
```diff
@@ -1,16 +1,88 @@
...
     );
   }
 
+  getSortButtonScreenReaderText(columnKey) {
+    let text = '';
+
+    if (this.state.sortedColumn === columnKey) {
+      text = this.props.sortButtonsScreenReaderText[this.state.sortDirection];
+    } else {
+      text = this.props.sortButtonsScreenReaderText[''];
```
arizzitano: This is clever code! But it's not especially readable. Can you use a more explicit keyname for the default state in `sortButtonsScreenReaderText`, like `defaultText` or something?

src/Table/index.jsx
```diff
@@ -73,14 +146,28 @@ Table.propTypes = {
...
+    columnSortable: isRequiredIf(PropTypes.bool, props => props.tableSortable),
+    onSort: isRequiredIf(PropTypes.func, props => props.columnSortable),
   })).isRequired,
   headingClassName: PropTypes.arrayOf(PropTypes.string),
+  tableSortable: PropTypes.bool,
+  /* eslint-disable react/require-default-props */
+  defaultSortedColumn: isRequiredIf(PropTypes.string, props => props.tableSortable),
+  /* eslint-disable react/require-default-props */
+  defaultSortDirection: isRequiredIf(PropTypes.string, props => props.tableSortable),
+  sortButtonsScreenReaderText: isRequiredIf(PropTypes.object, props => props.columnSortable),
```
arizzitano: Does `isRequiredIf` not allow you to specify a shape? It'd be good to be explicit about the keys required in this object.

src/Table/README.md
```diff
@@ -5,16 +5,32 @@ Provides a very basic table component with col-scoped headings displayed in the
...
 
 ### `className` (string array; optional)
 Specifies Bootstrap class names to apply to the table. See [Bootstrap's table documentation](https://getbootstrap.com/docs/4.0/content/tables/) for a list of applicable class names.
 
 ### `headingClassName` (string array; optional)
-Specifies Bootstrap class names to apply to the table heading. Options are detailed in [Bootstrap's docs](https://getbootstrap.com/docs/4.0/content/tables/#table-head-options)
+Specifies Bootstrap class names to apply to the table heading. Options are detailed in [Bootstrap's docs](https://getbootstrap.com/docs/4.0/content/tables/#table-head-options).
+
+### `tableSortable` (boolean; optional)
+Specifies whether the table is sortable. This setting supercedes column-level sortability, so if it is `false`, no sortable components will be rendered.
```
arizzitano: Can you add `sortButtonsScreenReaderText` as well here?

src/Table/index.jsx
```diff
@@ -20,10 +92,11 @@ class Table extends React.Component {
         <tr>
           {this.props.columns.map(col => (
             <th
+              className={this.props.tableSortable ? classNames({ sortable: col.columnSortable }) : ''}
```
efischer19: Flagging this line as a potential naive solution to [EDU-1535](https://openedx.atlassian.net/browse/EDUCATOR-1535) - if we're okay generalizing "all non-sortable column headings are hidden", we can change the else clause here to `'sr-only'`.

Is that a generalization we want to have? I'm worried we may eventually want to display a non-sortable heading.

MichaelRoytman: I think I could certainly see a case where someone would want to display a non-sortable heading, and I'd worry about inserting logic that benefits our page here at the expense of others interested in using the table, but I could be wrong.

efischer19: Cool, thanks for the confirmation. I was leaning that way as well, so I think I'll investigate getting a more specifically targeted approach working.

PR #54, opened 2017-10-24
change Button display prop to label

MichaelRoytman:


MichaelRoytman:
I've changed the display prop in studio-frontend to label (https://github.com/edx/paragon/pull/54). I'm not sure where else the Button has been used. I figured it would be less painful to do this now rather than later when the Button is more widely used.

PR #59, opened 2017-11-01
Hide table headers when marked hidden

efischer19:
# [EDUCATOR-1535](https://openedx.atlassian.net/browse/EDUCATOR-1535)

Adds `sr-only` class to table headers that are marked hidden.

This PR needs to merge first so that I can release a new version of paragon.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 node_modules
 npm-debug.log
 coverage
+jest*
--- a/src/Table/index.jsx
+++ b/src/Table/index.jsx
@@ -63,24 +63,28 @@ class Table extends React.Component {
   }
 
   getTableHeading(column) {
-    return (
-      this.props.tableSortable && column.columnSortable ?
-        <Button
-          label={
-            <span>
-              {column.label}
-              <span className={classNames(styles['sr-only'])}>
-                {' '}
-                {this.getSortButtonScreenReaderText(column.key)}
-              </span>
+    let heading;
+    if (this.props.tableSortable && column.columnSortable) {
+      heading = (<Button
+        label={
+          <span>
+            {column.label}
+            <span className={classNames(styles['sr-only'])}>
               {' '}
-              {this.getSortIcon(column.key === this.state.sortedColumn ? this.state.sortDirection : '')}
-            </span>}
-          onClick={() => this.onSortClick(column.key)}
-        />
-        :
-        column.label
-    );
+              {this.getSortButtonScreenReaderText(column.key)}
+            </span>
+            {' '}
+            {this.getSortIcon(column.key === this.state.sortedColumn ? this.state.sortDirection : '')}
+          </span>}
+        onClick={() => this.onSortClick(column.key)}
+      />);
+    } else if (column.hideHeader) {
+      heading = (<span className={classNames(styles['sr-only'])} />);
+    } else {
+      heading = column.label;
+    }
+
+    return heading;
   }
 
   getHeadings() {
--- a/.storybook/__snapshots__/Storyshots.test.js.snap
+++ b/.storybook/__snapshots__/Storyshots.test.js.snap
@@ -1241,7 +1241,9 @@ exports[`Storyshots Table default heading 1`] = `
         className=""
         scope="col"
       >
-        Coat Color
+        <span
+          className="sr-only"
+        />
       </th>
     </tr>
   </thead>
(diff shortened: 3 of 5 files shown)


efischer19:
I'll need to figure something out test-wise to fix the coverage drop

src/Table/index.jsx
```diff
@@ -63,24 +63,28 @@ class Table extends React.Component {
...
+          </span>}
+        onClick={() => this.onSortClick(column.key)}
+      />);
+    } else if (column.hide) {
+      heading = (<span className={classNames(styles['sr-only'])} />);
+    } else {
+      heading = column.label;
+    }
+
+    return heading;
```
efischer19: This diff looks a lot scarier than it actually is due to whitespace shifting, check the super handy [?w=1 "reviewing whitespace changes is for chumps" url shortcut](https://github.com/edx/paragon/pull/59/files?w=1)

.gitignore
```diff
@@ -3,3 +3,4 @@
 node_modules
 npm-debug.log
 coverage
+jest*
```
efischer19: After running tests locally, I had a folder titled `jest_0` in the repo directory. Cool to gitignore those?

src/Table/index.jsx
```diff
@@ -63,24 +63,28 @@ class Table extends React.Component {
...
-          label={
-            <span>
-              {column.label}
-              <span className={classNames(styles['sr-only'])}>
-                {' '}
-                {this.getSortButtonScreenReaderText(column.key)}
-              </span>
+    let heading;
+    if (this.props.tableSortable && column.columnSortable) {
+      heading = (<Button
```
jaebradley: why not just `return` in each branch instead of assigning to a variable and then returning that variable?

.gitignore
jaebradley: I don't see any `jest` directory in the repo, so probably 👍 to add this to the gitignore. I also don't see any `jest*` files in the repo but do you think, just to be safe, `jest*/` is better? I don't think it's a big deal either way, though it might be clearer that `jest*/` takes care of any directories and then `**/jest*` takes care of any files.

src/Table/index.jsx
jaebradley: ...errr and move the `if (column.hide)` branch first

efischer19:
storybook screenshot:

screen shot 2017-11-02 at 5 05 26 pm

src/Table/index.jsx
jaebradley: Regardless, is the intention that if there's a `column` with `columnSortable` = `true`, and `hide` = `true`, and `this.props.tableSortable` = `true` to render the column with a heading?

src/Table/index.jsx
```diff
@@ -63,24 +63,28 @@ class Table extends React.Component {
...
-        column.label
-    );
+              {this.getSortButtonScreenReaderText(column.key)}
+            </span>
+            {' '}
+            {this.getSortIcon(column.key === this.state.sortedColumn ? this.state.sortDirection : '')}
+          </span>}
+        onClick={() => this.onSortClick(column.key)}
+      />);
+    } else if (column.hide) {
```
jaebradley: Nitpick: would a better name be `column.hideHeader`? `column.hide` makes it sound like you want to hide the column.

And if we ever want to implement a feature in the future that does actually hide a column, it would be pretty confusing to implement it and still maintain this particular feature.

src/Table/index.jsx
efischer19: > why not just return in each branch instead of assigning to a variable and then returning that variable?

I've always been a fan of the single entry, single exit philosophy of writing methods. It's not nearly as important of a principle here in js as it is in old-school C code, but I find it's still a benefit for debugging flow-of-execution without having any huge drawbacks.

> ...errr and move the if (column.hide) branch up

Any particular reason why? The way this table is currently being used, column headings are usually sortable and unhidden; hidden headings are somewhat unusual.

> Regardless, is the intention that if there's a column with columnSortable = true, and hide = true, and this.props.tableSortable = true to render the column with a heading? One could argue this is the fault of the API user for misusing the API, was just curious if this behavior was intentional.

Yeah, we talked about it and decided that it's non-sensical to hide a sortable column, as having a heading to click is a necessary condition of sorting a column. So `this.props.tableSortable && column.columnSortable` takes precedence, if that's true we can ignore the value of `hide` entirely.

src/Table/index.jsx
efischer19: good idea!

src/Table/index.jsx
jaebradley: Cool, all makes sense.

My comment about moving the `column.hide` branch up was my thinking that the `hide` property would take precedence, but clearly that was intentionally not the case.

PR #62, opened 2017-11-06
Prevent Unnecessary Radio Button and Radio Button Group Rerendering

jaebradley:
# Discussion
I stumbled across [this great repository for React patterns / anti-patterns](https://github.com/vasanthk/react-bits/) and I realized that the current implementation of `RadioButton` and `RadioButtonGroup` is woefully guilty of
[this React bad practice](https://github.com/vasanthk/react-bits/blob/master/gotchas/01.pure-render-checks.md#bad-1)

```javascript
class App extends PureComponent {
  render() {
    return <MyInput
      onChange={e => this.props.update(e.target.value)}/>;
  }
}
```

Why is this bad? 

Because `e => this.props.update(e.target.value)` creates a new function, which has a different reference, which means that React's shallow equality check is always `false`, which means that this will __always__ re-render 😭 .

Just to make sure, I ran the debugger in `Storybook` and on any `RadioButtonGroup` change every single `RadioButton` re-rendered.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@edx/paragon",
-  "version": "1.1.4",
+  "version": "1.1.5",
   "description": "Accessible, responsive UI component library based on Bootstrap.",
   "main": "src/index.js",
   "author": "arizzitano",
@@ -23,6 +23,7 @@
     "prop-types": "^15.5.8",
     "react": "^15.5.4",
     "react-dom": "^15.5.4",
+    "react-element-proptypes": "^1.0.0",
     "react-proptype-conditional-require": "^1.0.4"
   },
   "devDependencies": {
--- a/src/RadioButtonGroup/index.jsx
+++ b/src/RadioButtonGroup/index.jsx
@@ -1,37 +1,54 @@
+/* eslint-disable react/no-multi-comp */
+
 import React from 'react';
 import PropTypes from 'prop-types';
+import ElementPropTypes from 'react-element-proptypes';
 
-function RadioButton(props) {
-  const {
-    children,
-    index,
-    isChecked,
-    name,
-    onBlur,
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
     "prop-types": "^15.5.8",
     "react": "^15.5.4",
     "react-dom": "^15.5.4",
+    "react-element-proptypes": "^1.0.0",
```
jaebradley: There was [an issue in the `React` repo that asked for element-specific `PropType` checking](https://github.com/facebook/react/pull/7119). 

Dan Abramov had [this to say](https://github.com/facebook/react/pull/7119#issuecomment-228615386)
> My personal opinion is that requiring elements to be of a specific type is an antipattern more often than not. I’m not saying it never happens, but it encourages tightly coupling components and makes it harder to introduce “transparent” wrappers around children.

> So while there may be use cases for this, I would say we’d rather not encourage this pattern, and it would better live as a third party library. Additionally we are very hesitant to expand the surface API area of PropTypes which we plan to eventually extract from React.

So why do I think it would be useful in [this use-case](https://github.com/edx/paragon/pull/62/files#diff-3a3b7078694a1e5b19d02a0bd3dcc9c2R172)? 

Because we actually want to encourage tight coupling - a `RadioButtonGroup` should not have children that aren't `RadioButton`s.

src/RadioButtonGroup/index.jsx
```diff
@@ -1,37 +1,54 @@
+/* eslint-disable react/no-multi-comp */
```
jaebradley: Since `RadioButton` should `bind` it's event handler, it can no longer be a functional component (right?). However, I don't think it should live in another file, as these two components are (right now at least) always going to be coupled.

PR #63, opened 2017-11-06
Fix missing repository warning and add unmet peer dev dependency

jaebradley:
I was seeing npm warnings for a missing repository field and for an unmet peer dev dependency (`@storybook/react^3.2.12`).

~~There are still two unmet peer dependency warnings for `jquery` and `popper.js` due to `@edx/edx-bootstrap`, which [I created a PR for](https://github.com/edx/edx-bootstrap/pull/22).~~ This should be merged in!

### TODO
- [x] Increase the version before merge


jaebradley:
Err I guess this is related to #52

PR #65, opened 2017-11-06
update docs and add proptype validation for hideHeader

MichaelRoytman:


src/Table/index.jsx
```diff
@@ -152,6 +152,7 @@ Table.propTypes = {
     ]).isRequired,
     columnSortable: isRequiredIf(PropTypes.bool, props => props.tableSortable),
     onSort: isRequiredIf(PropTypes.func, props => props.columnSortable),
+    hideHeader: PropTypes.bool,
```
jaebradley: should this have an `isRequiredIf` as well?

jaebradley: hmmmm no, no it shouldn't

MichaelRoytman: What would the condition be for requiring hideHeader?

jaebradley: I guess in the same way we would require for `columnSortable`?

MichaelRoytman: Ah, okay. I thought that the ability to hide the column would be independent of whether the column could be sorted, because a sortable table could have hidden columns and a non-sortable table could have hidden columns. Maybe I'm misunderstanding?

jaebradley: I could totally be misunderstanding PropTypes but wouldn't adding `isRequiredIf(PropTypes.bool, props => props.hideHeader)` be saying

> In the required `shape` there is a prop named `hideHeader`. Require this prop if it is part of our collection of props?

PR #67, opened 2017-11-07
Hyperlink Component

### LINKED ISSUE #23: Link component
Suggested functionality:
* a required `href` prop
* an `external` prop. This will add a `target="_blank"` to the output `<a>`, plus an icon and screen reader text indicating it's an external link

jaebradley:
# Discussion

Resolves #23 

## Accessibility

It looks like the only necessary [accessibility specification](https://www.w3.org/TR/wai-aria-practices-1.1/#link) is

> Enter: Executes the link and moves focus to the link target.

This behavior is already supported on Chrome, Safari, and Firefox.

## Design Decisions

* Why `Hyperlink` vs. `Link`?
  * This is more of a nitty specification on my part. Since you can create an `<a>` tag without an `href` attribute, when we require the `href` attribute, all our components will be hyperlinks, so why not use the more exact name? 
  * I don't feel *that* strongly about this though...
* Why a `destination` prop instead of an `href` prop?
  * In a similarly nitty vein, `href` seems like an implementation detail of a specific attribute that an `<a>` tag might need. 
  * `destination` seems more descriptive.
  * I also don't feel that strongly about this.
* I made the `content` prop a required `string`. I'm happy to open this up to `string` + `element` though in terms of PropTypes.
* The current interface looks something like
```javascript
<Hyperlink destination={'http://www.edx.org'} content={'Click me!'} />
```
I could also see an argument being made for an interface that looks like
```javascript
<Hyperlink destination={'http://www.edx.org'}>{'Click me!'}</Hyperlink>
```
* Why not make the `target` prop a `oneOf` PropType?
  * I initially went down this path, but in addition to `_blank`, `_self`, `_parent`, and `_top` you can pass in a custom frame name as the target.

### External Link With Icon
external link with icon

image

### External Link Without Icon
non external link

### TODO:
- [x] Check version before merging

### DIFF
--- a/src/Hyperlink/index.jsx
+++ b/src/Hyperlink/index.jsx
@@ -1,0 +1,61 @@
+import React from 'react';
+import classNames from 'classnames';
+import FontAwesomeStyles from 'font-awesome/css/font-awesome.min.css';
+import PropTypes from 'prop-types';
+import isRequiredIf from 'react-proptype-conditional-require';
+
+function Hyperlink(props) {
+  const {
+    destination,
+    content,
+    target,
+    onClick,
+    externalLinkAlternativeText,
+    externalLinkTitle,
+    ...other
+  } = props;
+
+  let externalLinkIcon;
+
+  if (target === '_blank') {
+    externalLinkIcon = (
+      // Space between content and icon
+      <span>{' '}
+        <span
+          className={classNames(FontAwesomeStyles.fa, FontAwesomeStyles['fa-external-link'])}
+          aria-hidden={false}
+          aria-label={externalLinkAlternativeText}
+          title={externalLinkTitle}
+        />
+      </span>
+    );
+  }
+
+  return (
+    <a
+      href={destination}
+      target={target}
+      onClick={onClick}
+      {...other}
+    >{content}{externalLinkIcon}
+    </a>
+  );
+}
+
+Hyperlink.defaultProps = {
+  target: '_self',
+  onClick: () => {},
+  externalLinkAlternativeText: 'Opens in a new window',
+  externalLinkTitle: 'Opens in a new window',
+};
+
+Hyperlink.propTypes = {
+  destination: PropTypes.string.isRequired,
+  content: PropTypes.string.isRequired,
+  target: PropTypes.string,
+  onClick: PropTypes.func,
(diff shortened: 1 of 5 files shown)


src/Hyperlink/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+  let externalLinkIcon;
+
+  if (target === '_blank') {
+    externalLinkIcon = (
+      // Space between content and icon
+      <span>{' '}
+        <span
+          className={classNames(FontAwesomeStyles.fa, FontAwesomeStyles['fa-external-link'])}
+          aria-hidden={false}
+          aria-label={externalLinkAlternativeText}
```
jaebradley: This `aria-label` field is meaningful because for screen-readers it will display the text (e.g. "open in a new tab") instead of a less useful `[?]` value that's supposed to represent the icon.

src/Hyperlink/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+      onClick={onClick}
+      {...other}
+    >{content}{externalLinkIcon}
+    </a>
+  );
+}
+
+Hyperlink.defaultProps = {
+  target: '_self',
+  onClick: () => {},
```
arizzitano: If this is a proper hyperlink, what's the use case for an `onClick`? I'm not sure we need it.

arizzitano: (...maybe for event tracking?)

jaebradley: That's an interesting point - I guess I was kind've future-proofing here...if you think it's unnecessary I can remove it.

jaebradley:
@arizzitano I believe Storybook uses iframes for component previewing and many websites prevent being displayed in an iframe

>You cannot display a lot of websites inside an iFrame. Reason being that they send an "X-Frame-Options: SAMEORIGIN" response header. This option prevents the browser from displaying iFrames that are not hosted on the same domain as the parent page. This is a security feature to prevent click-jacking.

[Relevant StackOverflow](https://stackoverflow.com/a/41522672/5225575)

I'm going to change the destination of that example Hyperlink!

src/Hyperlink/index.jsx
jaebradley: Actually, we definitely use `onClick` event handlers for analytics in Marketing so I think we should keep this.

PR #68, opened 2017-11-08
convert functional Button component to class to handle focus on click

MichaelRoytman:


src/Button/index.jsx
```diff
@@ -4,40 +4,65 @@ import PropTypes from 'prop-types';
...
+        ], {
+          [styles[`btn-${this.props.buttonType}`]]: this.props.buttonType !== undefined,
+        }, {
+          [styles.close]: this.props.isClose,
+        })}
+        onBlur={this.props.onBlur}
+        onClick={this.onClick}
+        onKeyDown={this.props.onKeyDown}
+        type={this.props.type}
+        ref={(input) => { this.getRefs(input); }}
```
jaebradley: Will this create a new function each render? If so, then I think this will force a re-render for every button since React's shallow equality check will fail.

[Here are some more examples](https://github.com/vasanthk/react-bits/blob/master/gotchas/01.pure-render-checks.md#bad-1).

src/Button/index.jsx
```diff
@@ -4,40 +4,65 @@ import PropTypes from 'prop-types';
...
+      <button
+        className={classNames([
+          ...this.props.className,
+          styles.btn,
+        ], {
+          [styles[`btn-${this.props.buttonType}`]]: this.props.buttonType !== undefined,
+        }, {
+          [styles.close]: this.props.isClose,
+        })}
+        onBlur={this.props.onBlur}
```
jaebradley: `this.onBlur`?

src/Button/index.jsx
```diff
@@ -4,40 +4,65 @@ import PropTypes from 'prop-types';
...
+        onClick={this.onClick}
+        onKeyDown={this.props.onKeyDown}
```
jaebradley: `this.onKeyDown`?

src/Button/index.jsx
```diff
@@ -4,40 +4,65 @@ import PropTypes from 'prop-types';
...
+      inputRef,
+      isClose,
+      onBlur,
+      onClick,
+      onKeyDown,
+      type,
+      ...other
+    } = this.props;
+
+    return (
```
jaebradley: I don't think you need a lot of `this.props` in the following logic since you destructured many of the variables above.

jaebradley:
@MichaelRoytman sorry I didn't see the WIP!

MichaelRoytman:
@jaebradley There wasn't one before you posted. I forgot to change the title when I opened the PR. No worries! Those React bits you linked to look really helpful.

src/Button/index.jsx
```diff
@@ -4,40 +4,74 @@ import PropTypes from 'prop-types';
...
+          [styles.close]: isClose,
+        })}
+        onBlur={onBlur}
+        onClick={onClick}
+        onKeyDown={onKeyDown}
+        type={type}
+        ref={this.setRefs}
+        {...other}
+      >
+        {this.props.label}
```
jaebradley: just `label`?

src/Button/index.jsx
```diff
@@ -4,40 +4,74 @@ import PropTypes from 'prop-types';
...
+        className={classNames([
+          ...className,
+          styles.btn,
+        ], {
+          [styles[`btn-${buttonType}`]]: buttonType !== undefined,
+        }, {
+          [styles.close]: isClose,
+        })}
+        onBlur={onBlur}
+        onClick={onClick}
```
jaebradley: should these be `this.onBlur`, `this.onClick`, and `this.onKeyDown`?

MichaelRoytman: Yeah, they should be! I forgot to push up my changes last week. Oops. I'll do that now.

jaebradley: oh no worries, sorry for the notification spam if this is still a WIP

src/Button/Button.stories.jsx
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable import/no-extraneous-dependencies */
+/* eslint-disable no-console */
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import Button from './index';
+
+storiesOf('Button', module)
```
MichaelRoytman: There were no stories for the Button, so I have added a very basic one here.

src/Button/index.jsx
```diff
@@ -4,40 +4,68 @@ import PropTypes from 'prop-types';
 
 import styles from './Button.scss';
 
-function Button(props) {
```
MichaelRoytman: I have removed this destructuring because props that are not used immediately in the render but by other auxiliary functions get passed to the HTML button component as part of other. For example, inputRef is now used in setRefs and not in the render, so to prevent it from being passed explicitly to the HTML button, it would need to be explicitly named so that is not part of other, which is conceptually kind of confusing, since it's not being used in the render.

src/Button/index.jsx
```diff
@@ -4,40 +4,83 @@ import PropTypes from 'prop-types';
...
+      buttonType,
+      className,
+      label,
+      isClose,
+      type,
+      /* inputRef is not used directly in the render, but it needs to be assigned
+        here to prevent it from being passed to the HTML button component as part
+        of other.
+      */
+      inputRef,
```
MichaelRoytman: I don't love that I have to do this here... I wanted to remove this destructuring because props that are not used immediately in the render but by other auxiliary functions get passed to the HTML button component as part of other. For example, inputRef is now used in setRefs and not in the render, so to prevent it from being passed explicitly to the HTML button, I need to be explicitly assign it here so that is not part of other, which is conceptually kind of confusing, since it's not being used in the render. If I remove the destructuring, however, I remove the ability to pass other attributes to the HTML button that are not explicit props of the Button. Any insight/recommendation about how to proceed?

MichaelRoytman:
@jaebradley This is ready for a re-review, if you're interested! I believe I've addressed all your comments from before, but let me know if you find something else/that isn't true.

src/Button/index.jsx
jaebradley: This is the same thing I did for [`RadioButtons`](https://github.com/edx/paragon/blob/master/src/RadioButtonGroup/index.jsx#L103-L114)...I'm not sure there's a better solution.

alt-text

MichaelRoytman: Oh, well! Thanks for taking a look!

PR #76, opened 2017-11-17
fix(Table): add aria-hidden to sort icons in Table

MichaelRoytman:
add aria-hidden tag to the table headings of sortable columns in a sortable Table so icons are not
read aloud to screen-reader users


MichaelRoytman:
Please review this PR at your convenience.

PR #77, opened 2017-11-17
fix(semantic-commit-linting): add semantic commit linting

jaebradley:
The `commitmsg` hook fails when you try and add a commit message via `git commit` that does not follow conventional commit convention

image

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  extends: ['@commitlint/config-angular'],
+};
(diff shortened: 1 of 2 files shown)


arizzitano:
This is a valuable bit of automation, but IMO too heavyweight to run for every commit. Some developers (myself included) push up a lot of WIP stuff as they're working on it, just so others can see it or CI can run against it, and those commit messages are generally meaningless. Besides, people tend to get antsy when you prevent them from committing -- even the pre-commit linter annoys some folks. Why enforce conventional commit syntax when it doesn't have any deeper implications?

The only commit message that really matters is the squash commit, right? What do you think about using the [`prepare-commit-msg`](https://git-scm.com/docs/githooks#_prepare_commit_msg) hook instead, and only running it if the hook receives the `squash` param?

jaebradley:
@arizzitano that all makes sense - I guess you could still commit using the `--no-verify` option? Either way, I don't have that big of a preference!

PR #83, opened 2017-11-20
fix(modal): Add Backdrop style

MichaelRoytman:


MichaelRoytman:
TODO: 
- add tests around some of the dynamic className functionality
- add alert "theme" to the modal

fysheets:
This PR will add the bootstrap modal-backdrop class to Modals so that the backdrop when a Modal is open is grayed out.

src/Modal/index.jsx
```diff
@@ -100,7 +100,19 @@ class Modal extends React.Component {
...
     return (
-      <div className={classNames(styles.modal, styles.show, { [styles['modal-open']]: this.state.open })} role="dialog" aria-modal aria-labelledby={this.headerId}>
+      <div
+        className={classNames(
+          styles.modal,
+          {
+            [styles['modal-open']]: this.state.open,
+            [styles['modal-backdrop']]: this.state.open,
+          },
+          this.state.open ? [styles.show] : [styles.fade],
```
jaebradley: This totally works - you could consider grouping all the classnames that are gated on `this.state.open` together
```javascript
{
  [styles['modal-open']]: this.state.open,
  [styles['modal-backdrop']]: this.state.open,
  [styles.show]: this.state.open,
  [styles.fade]: !this.state.open,
},
```
I find this slightly more readable since everything is grouped together but it doesn't really matter!

src/Modal/Modal.test.jsx
```diff
@@ -7,7 +7,8 @@ import Button from '../Button';
 
 
 const modalOpen = (isOpen, wrapper) => {
-  expect(wrapper.childAt(0).hasClass('modal-open')).toEqual(isOpen);
+  expect(wrapper.find('.modal').hasClass('modal-open')).toEqual(isOpen);
+  expect(wrapper.find('.modal').hasClass('modal-backdrop')).toEqual(isOpen);
```
jaebradley: should there be a test for `show` and `fade` or am I missing something?

src/Modal/index.jsx
```diff
@@ -100,7 +100,19 @@ class Modal extends React.Component {
 
   render() {
     return (
-      <div className={classNames(styles.modal, styles.show, { [styles['modal-open']]: this.state.open })} role="dialog" aria-modal aria-labelledby={this.headerId}>
+      <div
```
jaebradley: Nit: you could also consider pulling out `open` from `this.state` like
```
const { open } = this.state;
```
so you don't have to rewrite `this.state.open` everywhere, but again, it doesn't really matter.

src/Modal/Modal.test.jsx
fysheets: Added :)

PR #88, opened 2017-11-28
feat(TextArea): New TextArea Component

fysheets:
New bootstrap form-control TextArea component - leverages Paragon asInput

Note* based on: https://github.com/edx/paragon/tree/master/src/InputText

Ticket: [EDUCATOR-1824](https://openedx.atlassian.net/browse/EDUCATOR-1824)

### DIFF
--- a/src/TextArea/TextArea.stories.jsx
+++ b/src/TextArea/TextArea.stories.jsx
@@ -1,0 +1,48 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import TextArea from './index';
+
+storiesOf('Textarea', module)
+  .add('minimal usage', () => (
+    <TextArea
+      name="name"
+      label="First Name"
+      value="Foo Bar"
+    />
+  ))
+  .add('scrollable', () => (
+    <TextArea
+      name="name"
+      label="Information"
+      value="Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
+       incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud
+       exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure
+       dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
+       Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt
+       mollit anim id est laborum. Lorem ipsum dolor sit amet, consectetur adipiscing elit,
+       sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
+       quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute
+       irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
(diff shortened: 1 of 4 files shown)


fysheets:
FYI @edx/educator-dahlia @dsjen @arizzitano - based on the ticket title uncertainty, I can rename this to InputArea if we want.  Also this autmatically sets width to 100% with the bootstrap styling (same as InputText) so for reference if we want the width to be less, we should use bootstrap's grid/col styles. Also this means that resizing will be specifically for height, not height & width unless we overwrite the styles explicitly. See bootstrap example here, https://getbootstrap.com/docs/4.0/components/forms/#form-controls

src/TextArea/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+      onChange={props.onChange}
+      onBlur={props.onBlur}
+      aria-invalid={!props.isValid}
+      disabled={props.disabled}
+      required={props.required}
+    />
+  );
+}
+
+Text.propTypes = inputProps;
```
jaebradley: Were you thinking about adding `propTypes` for props here that aren't covered by `inputProps` (for example, `isValid`)?

I think you could do something like

```javascript
Text.propTypes = Object.assign({}, inputProps, {
  id: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.bool]),
  placeholder: PropTypes.string,
  describedBy: PropTypes.string,
  isValid: PropTypes.bool,
  ...
});
```
but there could be a better way

src/TextArea/index.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import asInput, { inputProps } from '../asInput';
+
+function Text(props) {
```
jaebradley: If you want to go down the syntactic sugar rabbit hole, you could write this as 

```javascript
const Text = props => (<textarea  ...props />)
```

and save a couple lines - obviously not a big deal either way

src/TextArea/TextArea.stories.jsx
```diff
@@ -0,0 +1,48 @@
...
+       mollit anim id est laborum."
+    />
+  ))
+  .add('validation', () => (
+    <TextArea
+      name="username"
+      label="Username"
+      description="The unique name that identifies you throughout the site."
+      validator={(value) => {
+        let feedback = { isValid: true };
```
jaebradley: Not a big deal but why not
```javascript
if (value.length < 3) {
  return { }
}

return { }
```
I get the single exit point argument but it does seem a lot simpler than dealing with an (IMHO) unnecessary variable assignment.

jaebradley:
@fysheets don't forget to use [the Angular commit convention](https://gist.github.com/stephenparish/9941e89d80e2bc58a153#format-of-the-commit-message) particularly for the __merged commit message__ as the [`semantic-release` package uses the commit message format to inform package version upgrades](https://github.com/semantic-release/semantic-release#default-commit-message-format).

src/TextArea/index.jsx
fysheets: I kind of have the same answer for all of your feedback haha -- I used the InputText component as the base for the creation of this component, so the syntax choices essentially match up to that. If we think these changes should be made I'm happy to make them, but if not I'm also happy to have the two Components share similarities for uniformity sake.

fysheets:
@jaebradley thanks for the links! I was going to find you prior to any merging to figure out if we had documentation of like a step by step guide of "the proper way to commit/merge/push in paragon" or anything.  I'll likely still snag you to ensure process is done correctly.

PR #89, opened 2017-12-02
fix(StatusAlert): Fix focus error for dismissible alert.

MichaelRoytman:
StatusAlert requires a check for existence of xButton in componentDidUpdate to prevent errors when
displaying dismissible alerts.

### DIFF
--- a/src/StatusAlert/index.jsx
+++ b/src/StatusAlert/index.jsx
@@ -32,7 +32,7 @@ class StatusAlert extends React.Component {
   }
 
   componentDidUpdate(prevState) {
-    if (this.state.open && !prevState.open) {
+    if (this.state.open && !prevState.open && this.xButton) {
       this.xButton.focus();
     }
   }


jaebradley:
Angular convention specifies that the first letter of the commit message `subject` should not be capitalized.

> <subject> text
> * use imperative, present tense: “change” not “changed” nor “changes”
> * don't capitalize first letter
> * no dot (.) at the end

- [Source](https://gist.github.com/stephenparish/9941e89d80e2bc58a153#subject-text)

Also, I didn't see anything regarding the `scope` of the commit message (`StatusAlert`, in this case) but I _think_ it should be lowercase? Just to be safe, can you lowercase it when you change the `subject`?

I have #77 that's pretty close for review / merge that hopefully (🤞 ) will eliminate commit message hiccups in the future. Apologies that #77 hasn't been merged in earlier and that there isn't go-to documentation.

jaebradley:
Are there any additional tests to add so that we know if we change the component's behavior going forward?

MichaelRoytman:
@jaebradley I think that, if I were to add testing for validation of focus-related behavior, I'd want to add test for focus logic in the StatusAlert as a whole, and I think it might be out-of-scope for me at the moment. I'm inclined to leave that until a later date, since the coverage is essentially unchanged. I'll make a note of it, though. Thank you for the review!

PR #91, opened 2017-12-05
fix(asinput): Add danger theme to asInput component

MichaelRoytman:
Add danger theme to asInput component to add alert icon, red text for validation message, and red border to input element

### DIFF
--- a/src/asInput/index.jsx
+++ b/src/asInput/index.jsx
@@ -1,6 +1,8 @@
 /* eslint-disable react/no-unused-prop-types */
 import React from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import FontAwesomeStyles from 'font-awesome/css/font-awesome.min.css';
 
 import newId from '../utils/newId';
 import styles from './asInput.scss';
@@ -23,6 +25,7 @@ export const inputProps = {
   onBlur: PropTypes.func,
   validator: PropTypes.func,
   className: PropTypes.arrayOf(PropTypes.string),
+  themes: PropTypes.arrayOf(PropTypes.string),
 };
 
 const asInput = (WrappedComponent, labelFirst = true) => {
@@ -50,9 +53,19 @@ const asInput = (WrappedComponent, labelFirst = true) => {
       const desc = {};
 
       if (!this.state.isValid) {
+        const hasDangerTheme = this.hasDangerTheme();
+
         desc.error = (
-          <div className={styles['form-control-feedback']} id={errorId} key="0">
-            {this.state.validationMessage}
+          <div className={classNames(styles['form-control-feedback'], { [styles['invalid-feedback']]: hasDangerTheme })} id={errorId} key="0">
+            { hasDangerTheme &&
+            <span
+              className={classNames(FontAwesomeStyles.fa, FontAwesomeStyles['fa-exclamation-circle'], styles['fa-icon-spacing'])}
+              aria-hidden
+            />
+            }
+            <span>
+              {this.state.validationMessage}
+            </span>
           </div>
         );
         desc.describedBy = errorId;
@@ -68,6 +81,10 @@ const asInput = (WrappedComponent, labelFirst = true) => {
       }
 
       return desc;
+    }
+
+    hasDangerTheme() {
+      return this.props.themes.includes('danger');
     }
 
     handleBlur(event) {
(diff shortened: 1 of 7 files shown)


src/asInput/index.jsx
```diff
@@ -89,15 +101,17 @@ const asInput = (WrappedComponent, labelFirst = true) => {
...
           <WrappedComponent
             {...this.props}
             {...this.state}
-            className={[
+            className={[classNames(
               styles['form-control'],
-              ...this.props.className,
-            ]}
+              styles['form-inline'],
+              { [styles['is-invalid']]: !this.state.isValid && this.props.theme.includes('danger') },
```
MichaelRoytman: I have no strong feelings about this theme's name. Any opinions?

src/asInput/index.jsx
```diff
@@ -50,8 +53,17 @@ const asInput = (WrappedComponent, labelFirst = true) => {
 
       if (!this.state.isValid) {
         desc.error = (
-          <div className={styles['form-control-feedback']} id={errorId} key="0">
-            {this.state.validationMessage}
+
+          <div className={classNames(styles['form-control-feedback'], { [styles['invalid-feedback']]: this.props.theme.includes('danger') })} id={errorId} key="0">
```
MichaelRoytman: I can pull out the `this.props.theme.includes('danger')` condition into a variable, as the condition is being used twice here, but I was interested to see whether it would be an issue as we add more themes. Is the idea that, as we add more themes, we conditionally apply classnames this way? I'll add a 'hasDangerTheme' variable if this looks good to everyone, otherwise.

src/asInput/index.jsx
```diff
@@ -50,8 +53,17 @@ const asInput = (WrappedComponent, labelFirst = true) => {
...
+            { this.props.theme.includes('danger') &&
+            <span
```
MichaelRoytman: This is an alert icon, which follows UI/UX guidelines laid out [here](https://openedx.atlassian.net/wiki/spaces/UX/pages/156737282/Forms).

src/asInput/index.jsx
```diff
@@ -89,15 +101,17 @@ const asInput = (WrappedComponent, labelFirst = true) => {
     render() {
       const { description, error, describedBy } = this.getDescriptions();
       return (
-        <div className={styles['form-group']}>
+        <div className={classNames(styles['form-group'])}>
```
MichaelRoytman: Any harm in adding classNames?

MichaelRoytman:
I'd be interested in thoughts about this "theme". It's a solution we discussed when considering the modal and styling that studio-frontend needed for the modal, but I haven't gotten around to it because of the accessibility form. Does this look like what was envisioned?

src/asInput/index.jsx
```diff
@@ -22,6 +24,7 @@ export const inputProps = {
   onBlur: PropTypes.func,
   validator: PropTypes.func,
   className: PropTypes.arrayOf(PropTypes.string),
+  theme: PropTypes.arrayOf(PropTypes.string),
```
jaebradley: Naming nit: if it is an array of `string`s should this be called `themes` instead?

src/asInput/index.jsx
```diff
@@ -49,9 +52,20 @@ const asInput = (WrappedComponent, labelFirst = true) => {
       const desc = {};
 
       if (!this.state.isValid) {
+        const hasDangerTheme = this.props.theme.includes('danger');
```
jaebradley: Since the `this.props.theme.includes('danger')` check is implemented in `getDescriptions` and `render` you could consider moving this to it's own helper method and that way if you ever have to adjust it / add a theme value, you can do it in one place instead of having to remember to do it in two places.

src/asInput/index.jsx
jaebradley: I don't think there is any harm, but why do you think it is useful here if there are no class names to join?

src/asInput/index.jsx
jaebradley: My only thought is that if we ever want to add additional styling in the future whenever `!this.state.isValid` (like always making the background color `red` or something) we'd probably want to name _that_ style `is-invalid`, right?  

But even if we have to rename this class it won't be a breaking change (right?), so I don't think it's a big deal either way.

src/asInput/index.jsx
MichaelRoytman: Ah, my bad. I had added class names and then removed them. I'll switch it back.

src/asInput/index.jsx
MichaelRoytman: Good catch!

dsjen:
@arizzitano -- do you want to take a look? We are under time pressure, and if you intend to look at this, we'd appreciate if you could do so at your earliest convenience so that we can wire everything up.

jaebradley:
Hold on, I'm investigating if `style` will bump the version.

PR #92, opened 2017-12-05
 fix(input): Use prop.id/prop.type when available and allow for ref on inputs

fysheets:
Leverage the ID that is passed through as prop when it's available, allow for InputText to accept a type prop, and allow for input components to be assigned a ref

### DIFF
--- a/src/InputText/index.jsx
+++ b/src/InputText/index.jsx
@@ -18,6 +18,7 @@ function Text(props) {
       aria-invalid={!props.isValid}
       disabled={props.disabled}
       required={props.required}
+      ref={props.inputRef}
     />
   );
 }
--- a/src/InputText/index.jsx
+++ b/src/InputText/index.jsx
@@ -8,7 +8,7 @@ function Text(props) {
     <input
       id={props.id}
       className={classNames(props.className)}
-      type="text"
+      type={props.type || 'text'}
       name={props.name}
       value={props.value}
       placeholder={props.placeholder}
--- a/src/asInput/index.jsx
+++ b/src/asInput/index.jsx
@@ -32,7 +32,7 @@ const asInput = (WrappedComponent, labelFirst = true) => {
       this.handleChange = this.handleChange.bind(this);
       this.handleBlur = this.handleBlur.bind(this);
 
-      const id = this.props.id ? this.props.id : newId('asInput');
+      const id = this.props.id || newId('asInput');
       this.state = {
         id,
         value: this.props.value,
@@ -118,7 +118,7 @@ const asInput = (WrappedComponent, labelFirst = true) => {
   NewComponent.defaultProps = {
     onChange: () => {},
     onBlur: () => {},
-    id: undefined,
+    id: newId('asInput'),
     value: '',
     description: undefined,
     disabled: false,
--- a/src/asInput/index.jsx
+++ b/src/asInput/index.jsx
@@ -92,8 +92,8 @@ const asInput = (WrappedComponent, labelFirst = true) => {
         <div className={styles['form-group']}>
           {labelFirst && <label htmlFor={this.state.id}>{this.props.label}</label>}
           <WrappedComponent
-            {...this.props}
             {...this.state}
+            {...this.props}
             className={[
               styles['form-control'],
               ...this.props.className,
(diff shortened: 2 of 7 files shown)


fysheets:
@arizzitano / @jaebradley not sure if this should be a feat or a fix when I squash/merge

src/asInput/index.jsx
```diff
@@ -31,7 +32,7 @@ const asInput = (WrappedComponent, labelFirst = true) => {
       this.handleChange = this.handleChange.bind(this);
       this.handleBlur = this.handleBlur.bind(this);
 
-      const id = newId('asInput');
+      const id = this.props.id ? this.props.id : newId('asInput');
```
jaebradley: would `this.props.id || newId('asInput');` be slightly more readable?

src/asInput/index.jsx
```diff
@@ -117,6 +118,7 @@ const asInput = (WrappedComponent, labelFirst = true) => {
   NewComponent.defaultProps = {
     onChange: () => {},
     onBlur: () => {},
+    id: undefined,
```
jaebradley: if we want to default to `newId('asInput')` should this also be the default prop value?

src/asInput/index.jsx
fysheets: Sure I can make that swap - I tend to lean ternary as ternaries are what I assume to be more familiar across languages, but I know we use the `||` syntax elsewhere for react so it won't be a stand out.

Note -- just saw your second comment, see reply there as that might make this unnecessary.

src/asInput/index.jsx
```diff
@@ -11,6 +11,7 @@ export const getDisplayName = WrappedComponent =>
 export const inputProps = {
   label: PropTypes.string.isRequired,
   name: PropTypes.string.isRequired,
+  id: PropTypes.string,
```
jaebradley: Does the `id` attribute have to be a `string`?  Should this also support `number`?

src/asInput/index.jsx
fysheets: Reading this made me realize -- If I set the default value here, would I even need to do the check above?

src/asInput/index.jsx
fysheets: `newId` returns a string and if were to come from the input prop I would expect something along the lines of `<InputText id="theID" ... />` and even if the ID there is `id={someID}` I think it would be cleaner to enforce that value to a `string`.

src/asInput/index.jsx
jaebradley: Yes, because this only fills in the `id` value if somebody doesn't pass in the `id` prop, but if somebody passes in an `undefined` value for the `id` prop we probably want to default to the `newId` id, right?

fysheets: Ah yes, that is indeed an edge case. I'll update this default and change to the || above.

jaebradley: Actually, I'm not sure about this after all - I'm currently investigating.

jaebradley: Sorry @fysheets I might have led you astray - check for yourself, but I think the `defaultProps` might also coerce `undefined` values.

The only edge case is that [React treats `undefined` and `null` prop values differently](https://github.com/facebook/react/issues/2166) - it *won't* coerce `null` - do you think it makes sense to create a default id value if the prop value is `null`? If so, then we should keep the previous check.

src/InputText/index.jsx
```diff
@@ -8,7 +8,7 @@ function Text(props) {
     <input
       id={props.id}
       className={classNames(props.className)}
-      type="text"
+      type={props.type || 'text'}
```
jaebradley: While this does the job of `defaultProps`, I think it might be useful to add a `defaultProps` for the `type` prop.

I think it's useful because we only would keep the `props.type || 'text'` logic to deal with case where `type` *is* defined (but `null`) not where `type` is defined or not.

Basically, while there would be no difference in the behavior of the component, there is a subtle difference in the intention of `defaultProps` vs. `props.type || 'text'` (in my opinion at least)

fysheets: I see where you're coming from on this one, but it's not within the scope at the moment for this task. Willing to revisit once we get this a11y page out the door.

fysheets:
This should be ready for a review - it is needed in part for the Studio Accessibility Feedback page that needs to be released by 12/12 but our goal is to have it ready merged in by EOD Thursday 12/7 so that it can be validated with enough time for updates. Thanks!
Input on if this should be squashed/merged as a fix vs feat would be appreciated.

jaebradley:
@fysheets thanks for doing this! This feels like more of a fix than a feat since it's a relatively minor change, but I don't think it matters much one way or the other.

PR #93, opened 2017-12-07
fix(asinput): Remove control of state.value from asInput

MichaelRoytman:
asInput component was overriding state.value that was set by the parent component. This allows parent to call setState on input components.

### DIFF
--- a/src/asInput/asInput.test.jsx
+++ b/src/asInput/asInput.test.jsx
@@ -41,7 +41,7 @@ describe('asInput()', () => {
     const wrapper = mount(<InputTestComponent {...props} />);
     expect(wrapper.find('label').text()).toEqual(props.label);
     expect(wrapper.find('#description-asInput1').text()).toEqual(props.description);
-    expect(wrapper.state('value')).toEqual(props.value);
+    expect(wrapper.prop('value')).toEqual(props.value);
   });
 
   it('creates generic prop id', () => {
@@ -79,6 +79,21 @@ describe('asInput()', () => {
     expect(wrapper.find('label').prop('id')).toEqual(`label-${testId}`);
     expect(wrapper.find('input').prop('id')).toEqual(testId);
     expect(wrapper.find('small').prop('id')).toEqual(`description-${testId}`);
+  });
+
+  it('overrides state value when props value changes', () => {
+    const initValue = 'foofoo';
+    const newValue = 'barbar';
+    const props = {
+      ...baseProps,
+      value: initValue,
+    };
+    const wrapper = mount(<InputTestComponent {...props} />);
+    expect(wrapper.state('value')).toEqual(initValue);
+    wrapper.setProps({
+      value: newValue,
+    });
+    expect(wrapper.state('value')).toEqual(newValue);
   });
 
   describe('fires', () => {
@@ -149,7 +164,7 @@ describe('asInput()', () => {
           const err = wrapper.find('.form-control-feedback');
           expect(err.exists()).toEqual(true);
           expect(err.text()).toEqual(validationResult.validationMessage);
-          // expect(err.hasClass('invalid-feedback')).toEqual(true);
+          expect(err.hasClass('invalid-feedback')).toEqual(true);
 
           const dangerIcon = wrapper.find('.fa-exclamation-circle');
           expect(dangerIcon.exists()).toEqual(true);
--- a/src/asInput/index.jsx
+++ b/src/asInput/index.jsx
@@ -46,6 +46,14 @@ const asInput = (WrappedComponent, labelFirst = true) => {
       };
     }
 
+    componentWillReceiveProps(nextProps) {
+      if (nextProps.value !== this.props.value) {
+        this.setState({
+          value: nextProps.value,
+        });
+      }
+    }
+
     getDescriptions() {
       // possible future work: multiple feedback msgs?
       const errorId = `error-${this.state.id}`;


src/asInput/asInput.test.jsx
```diff
@@ -149,7 +149,7 @@ describe('asInput()', () => {
           const err = wrapper.find('.form-control-feedback');
           expect(err.exists()).toEqual(true);
           expect(err.text()).toEqual(validationResult.validationMessage);
-          // expect(err.hasClass('invalid-feedback')).toEqual(true);
+          expect(err.hasClass('invalid-feedback')).toEqual(true);
```
MichaelRoytman: I don't know why I left this commented out before.

jaebradley:
I wanted to write this down before I forgot.

[SFE#61](https://github.com/edx/studio-frontend/pull/61) is using `InputText` as part of a form. After a form is submitted, the form fields should be cleared for future use. Unfortunately, passing an empty string to the `InputText` `value` prop is not clearing the form field. 

I think this is because of
1. How the `asInput` component handles change events for the `InputText` component
2. How the `asInput` component applies `props` and `state` to `InputText` during render

Here's [the crucial piece of code from `asInput`](https://github.com/edx/paragon/blob/master/src/asInput/index.jsx#L99-L130)

```javascript
    handleChange(event) {
      this.setState({ value: event.target.value });
      this.props.onChange(
        event.target.type === 'checkbox' ? event.target.checked : event.target.value,
        this.props.name,
      );
    }

    render() {
      return (
        <div className={styles['form-group']}>
          {labelFirst && <label id={`label-${this.state.id}`} htmlFor={this.state.id}>{this.props.label}</label>}
          <WrappedComponent
            {...this.props}
            {...this.state}
            className={[classNames(
              styles['form-control'],
              { [styles['is-invalid']]: !this.state.isValid && this.hasDangerTheme() },
              { ...this.props.className },
            )]}
            describedBy={describedBy}
            onChange={this.handleChange}
            onBlur={this.handleBlur}
          />
          {!labelFirst && <label htmlFor={this.state.id}>{this.props.label}</label>}
          {error}
          {description}
        </div>
      );
    }
  }
```

The first thing to remember is that `asInput` takes some component and then returns a new component (in this case, literally called `NewComponent`) that implements styling as well as some event handling, essentially wrapping the passed-in component (`InputText`, for example).

Part of this event handling is [the `onChange` handler on the `WrappedComponent` in the `render` method of `asInput`](https://github.com/edx/paragon/blob/master/src/asInput/index.jsx#L121).  The `onChange` handler calls the `handleChange` method, which changes the state of `NewComponent` and then calls the passed in (from props) `onChange` event handler.

So as somebody is filling out the email address portion of the form the `state` of `NewComponent` is updating (first they type `j` => `state.value` = `j`, then they type `b` =>`state.value` = `jb`, etc.).

Now let's say that after the email address portion of the form is filled out (i.e. `state.value` is `jbradley@edx.org`) we want to clear this portion of the form by setting the `value` prop of `InputText` to an empty string - what happens?

Here's how the `WrappedComponent` gets rendered in `asInput`.
```javascript
<WrappedComponent
    {...this.props}
    {...this.state}
    className={[classNames(
      styles['form-control'],
      { [styles['is-invalid']]: !this.state.isValid && this.hasDangerTheme() },
      { ...this.props.className },
    )]}
    describedBy={describedBy}
    onChange={this.handleChange}
    onBlur={this.handleBlur}
/>
```
Note how `{...this.props}` comes __before__ `{...this.state}`. This means that if they share a key, the `state` key value will override the `props` key value.

Here's [an example in RunKit](https://runkit.com/jaebradley/5a29f8259f13ba0011c12506).

So thus, the empty string we pass to `InputText` gets overridden by the `jbradley@edx.org` stored in `state` and thus we can't "clear" the `InputText` form field.

src/asInput/index.jsx
```diff
@@ -46,6 +46,14 @@ const asInput = (WrappedComponent, labelFirst = true) => {
       };
     }
 
+    componentWillReceiveProps(nextProps) {
+      if (nextProps.value !== this.props.value) {
```
jaebradley: do we want to compare `nextProps` against `this.props.value` or `this.state.value`?

arizzitano: I think we want to compare it with props, per https://reactjs.org/docs/react-component.html#componentwillreceiveprops:
```
If you need to update the state in response to prop changes (for example, to reset it), you may compare this.props and nextProps and perform state transitions using this.setState() in this method.
```

jaebradley: So imagine that I have a `render` like this

```javascript
render() {
  return <InputText value={this.state.emailAddress}>;
}
```
As I type into this component, `state.value` of `myInput` will change via the `handleChange` method. 

But if `this.state.emailAddress` does not change (for example, there's no `onChange` handler) then `nextProps.value` will equal `this.props.value` which means that the `state.value` of `NewComponent` will not reflect the value of the `value` prop.

fysheets: @jaebradley I believe this should be props. Do you have a story that tests that hypothetical and shows that's it doesn't work? I think it'll be easier to discuss this more if it's not a conceptual example and something tangible to poke at.

arizzitano:
@edx/educator-dahlia -- would one of you guys mind verifying that this branch fixes the form-not-resetting issue in SFE?

src/asInput/index.jsx
jaebradley: yeah this looks good after all - I was confusing what `value` means; now it kind've means initial value or "reset" value.

fysheets:
Just validated with SFE locally -- triggering the state change to clear the fields properly clears the input forms on the page, so 👍 here

MichaelRoytman:
Great! I'm going to squash and merge. Thank you to everyone for getting this resolved :).

PR #94, opened 2017-12-07
fix(statusalert): Expose focus function on StatusAlert component

MichaelRoytman:
In order to make focus management for accessible pages easier, expose a focus function on the component to allow parent to call focus on the component ref.


MichaelRoytman:
For reference, we're using the solution in  Draft.js's [documentation](https://draftjs.org/docs/advanced-topics-managing-focus.html) in order to make focus management of React components easier. It avoid needing access to an HTMLElement to focus on from parent components.

PR #96, opened 2017-12-11
docs(jest): Document how to run jest tests in Chrome DevTools

thallada:
Also adds an npm run test-debug script.

I thought this was useful enough to make a command so everyone else can easily do it.

Also, I think jest-cli was missing from the devDependencies? I couldn't run the tests until I installed that.


README.md
```diff
@@ -71,6 +71,14 @@ To run the unit tests, run:
 npm run test
 ```
 
+To run the unit tests in the Chrome DevTools inspector, run:
+
+```
+npm run test-debug
+```
+
+Then, open `chrome://inspect` in your Chrome browser and select the "node_modules/.bin/jest" target to open the Chrome DevTools. You can set breakpoints in Chrome DevTools or insert a `debugger;` statement into the code to pause execution at that point.
```
jaebradley: @thallada this looks great - thanks for doing this!

A couple (very nitty) style things (just my opinion so feel free to not implement these suggestions):

1. Maybe move this to the end of the `Unit Testing` section and then add a `Debug Tests Via Chrome` (sub) header or something? Might help to highlight how to debug tests.
2. Maybe add a screenshot (something like this perhaps?)
image just to add a little bit more context?

jaebradley:
@thallada I noticed that this deletes the `package-lock.json` file which [should be committed to source control as it's automatically created by NPM 5](https://stackoverflow.com/a/44210813/5225575). Can you restore this file?

package.json
```diff
@@ -21,7 +21,8 @@
     "semantic-release": "semantic-release pre && npm publish && semantic-release post",
     "snapshot": "jest --updateSnapshot",
     "start": "start-storybook -p 6006",
-    "test": "jest --coverage"
+    "test": "jest --coverage",
+    "test-debug": "node --inspect-brk node_modules/.bin/jest --runInBand --coverage"
```
jaebradley: Super nitty and I don't feel that strongly either way (nor do I think it matters that much): `test-debug` feels a little backwards - I get that it's trying to keep with the `test` paradigm, but `debug-test` feels a lot more natural. Just my ✌💰.

**tl;dr: it doesn't really matter**

jaebradley:
@thallada @arizzitano one additional thought - since a commit message with the `docs` type doesn't create a new release, the updated `README` will only be reflected in GitHub (and not NPM). If we want the docs to be consistent between NPM and GitHub, the commit message should change.

package.json
thallada: Hah, I actually had it like that at first, but changed it to test-debug because that's how it was in [insights](https://github.com/edx/edx-analytics-dashboard/blob/master/gulpfile.js#L40). Doesn't really matter though, I can make it debug-test.

thallada:
@jaebradley I didn't delete the `package-lock.json` file, mine just removes `fsevents` for some reason. I don't quite get the package lock thing yet, I can just not commit any changes to it and leave it alone if that's better.

jaebradley:
@thallada thanks for the clarification!

Looks like the last comment on that issue says
> this is fixed in npm 5.6.0 - [comment](https://github.com/npm/npm/issues/18135#issuecomment-351085000)

What npm version are you using?

thallada:
Upgraded to 5.6.0 and now it doesn't remove fsevents!

thallada:
@jaebradley Can we make a docs commit make a new release?

jaebradley:
@thallada if you want to create a GH issue to make the `docs` commit message `type` create a new release, we can discuss the implementation steps there. I think changing the default commit analyzer behavior is non-trivial so if you'd like to get this in sooner rather than later I'd suggest changing the commit message `type` to `fix`.

PR #98, opened 2017-12-12
docs(readme): add semantic-release documentation

jaebradley:
The additional README documentation hopefully answers questions around `semantic-release` specifically on the structure of commit messages, and why that structure is important from a release perspective.

If we decide that documentation updates need to be consistent between NPM and GitHub, I will change the commit `type` to be a `fix` instead of a `docs`.

I considering putting this in it's own markdown file - let me know if this separation is better. If we decide to put it in it's own markdown then we don't need to change the commit type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,3 +88,75 @@ If the snapshot tests fail, it's generally pretty easy to tell whether it's happ
 ### Coverage
 
 Paragon measures code coverage using Jest's built-in `--coverage` flag (which I believe uses istanbul under the hood) and report it via [Coveralls](https://coveralls.io/github/edx/paragon). Shoot for 100% test coverage on your PRs, but use your best judgment if you're really struggling to cover those last few lines. At the very least, don't *reduce* total coverage. Coveralls will fail your build if your PR reduces coverage.
+
+## Semantic Release
+
+Paragon uses the [`semantic-release` package](https://github.com/semantic-release/semantic-release) to automate its release process (creating Git tags, creating GitHub releases, and publishing to NPM).
+
+### Commit Messages
+
+[`semantic-release` analyzes commit messages to determine whether to create a `major`, `minor`, or `patch` release](https://github.com/semantic-release/semantic-release#default-commit-message-format) (or to skip a release).
+Paragon currently uses [the default commit analyzer release rules](https://github.com/semantic-release/commit-analyzer/blob/master/lib/default-release-rules.js#L8-L11) which means that there are **4** commit types that will trigger a release:
+1. `feat` (`minor` release)
+2. `fix` (`patch` release)
+3. `perf` (`patch` release)
+4. `breaking` (`major` release)
+
+#### `Angular` Commit Message Convention
+
+Paragon currently uses [the `Angular` commit message convention](https://gist.github.com/stephenparish/9941e89d80e2bc58a153). As documented in the previously linked gist, a commit that follows the `Angular` commit message convention has four parts:
+1. A `type` - is this commit a `feat`, `fix`, `chore`, `docs`, etc.? There is a set of `type` values to choose from, but again, **only the `feat`, `fix`, `perf`, and `breaking`** `type` values will trigger a release.
+2. A `scope` - what is this commit impacting? Did you fix a bug in the `Hyperlink` component? Did you add a new feature to the `RadioButtonGroup` component? Currently, the `scope` must be lower-case and `-` separated though switching this to being `camelCase` is currently being investigated.
+3. A `subject` - provide a short description of _what_ your change is.
+  * use imperative, present tense language (so `change` not `changed` or `changes`)
+  * don't capitalize the first letter
+  * don't add a period (`.`) at the end
+  * there is a 50 character limit
+4. A `body` (optional) - add more detail about your change
+5. A `footer` (optional)
+  * > All breaking changes have to be mentioned in footer with the description of the change, justification and migration notes - [`Angular` commit message specification](https://gist.github.com/stephenparish/9941e89d80e2bc58a153#breaking-changes)
+
+##### Examples
+This will lead to a patch version bump
+> fix(someComponent): fix tab accessibility issue in someComponent
+
+This will lead to a minor version bump
+> feat(newComponent): add newComponent`
+
+This will lead to a patch version bump - note the body
+>fix(anotherComponent): fix escape key accessibility issue
+>
+> Unable to clear anotherComponent's popup using escape key.
+> By updating the onKeyPress event handler to close popup window on escape key press, accessibility issue was resolved.
+
+
+#### Using `commitlint` in Paragon
+
+Paragon uses the [`commitlint`](https://github.com/marionebl/commitlint) package to lint commit messages. Paragon currently has [a `commit message` hook](https://github.com/edx/paragon/blob/master/package.json#L20) that runs `commitlint`'s, well, commit linting, on the given commit message using the configuration specified in [`commitlint.config.js`](https://github.com/edx/paragon/blob/master/commitlint.config.js). If a commit message fails linting, the commit associated with that message does not end up getting committed.
+
+`commitlint` also comes with [a helpful CLI](https://github.com/marionebl/commitlint/tree/master/@commitlint/cli) that walks one through the entire semantic commit process. This CLI can be triggered by running the `npm run gc` command.
+
+![example-commitlint](https://media.giphy.com/media/3ohs7H8y9Qi7HOHWko/giphy.gif)
+
+As noted above, the only required fields are **`type`, `scope`, and `subject`**. The `body` and `footer` can be skipped through the CLI by typing `:skip`.
+
+### Pull Requests
+
+Since all pull requests should be `squashed` / `rebased` down to a single semantically-formatted commit, it is recommended that **all pull requests have a title that matches the commit message**. This way, whether you're merging a pull request, squashing and merging a pull request, or rebasing and merging a pull request, the correct commit message will be analyzed by `semantic-release`.
+
+### Merging, Building, and Releasing
+
+> When `semantic-release` is set up it will do that after every successful continuous integration build of your master branch (or any other branch you specify) and publish the new version for you
+> - [`semantic-release` README](https://github.com/semantic-release/semantic-release)
+
+Release-related activity only happens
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -88,3 +88,61 @@ If the snapshot tests fail, it's generally pretty easy to tell whether it's happ
...
 Paragon measures code coverage using Jest's built-in `--coverage` flag (which I believe uses istanbul under the hood) and report it via [Coveralls](https://coveralls.io/github/edx/paragon). Shoot for 100% test coverage on your PRs, but use your best judgment if you're really struggling to cover those last few lines. At the very least, don't *reduce* total coverage. Coveralls will fail your build if your PR reduces coverage.
+
+## Semantic Release
+
+Paragon uses the [`semantic-release` package](https://github.com/semantic-release/semantic-release) to automate it's release process (creating Git tags, creating GitHub releases, and publishing to NPM).
+
+### Commit Messages
+
+[`semantic-release` analyzes commit messages to determine the code changes will be associated with a `major`, `minor`, or `patch` release](https://github.com/semantic-release/semantic-release#default-commit-message-format).
+Paragon currently uses [the default commit analyzer release rules](https://github.com/semantic-release/commit-analyzer/blob/master/lib/default-release-rules.js#L8-L11) which means that there are **3** commit types that will trigger a release:
```
jaebradley: `4` not `3` - apparently I can't count

README.md
```diff
@@ -88,3 +88,61 @@ If the snapshot tests fail, it's generally pretty easy to tell whether it's happ
...
+As noted above, the only required fields are **`type`, `scope`, and `subject`**. The `body` and `footer` can be skipped through the CLI by typing `:skip`.
+
+### Pull Requests
+
+Since all pull requests should be `squashed` / `rebased` down to a single semantically-formatted commit, it is recommended that **all pull requests have a title that matches the commit message**. This way, whether you're merging a pull request, squashing and merging a pull request, or rebasing and merging a pull request, the correct commit message will be analyzed by `semantic-release`.
+
+### Merging, Building, and Releasing
+
+> When `semantic-release` is set up it will do that after every successful continuous integration build of your master branch (or any other branch you specify) and publish the new version for you
+> [`semantic-release` README](https://github.com/semantic-release/semantic-release)
```
jaebradley: Formatting error here - probably need to add a `-` to indicate it's from a source

README.md
```diff
@@ -88,3 +88,61 @@ If the snapshot tests fail, it's generally pretty easy to tell whether it's happ
...
+
+Release-related activity only happens
+1. After a successful Travis build on `master` ([the `semantic-release` NPM command is only executed as part of the `after_success` Travis hook](https://github.com/edx/paragon/blob/master/.travis.yml#L21-L22).))
```
jaebradley: Extra `)`

README.md
```diff
@@ -88,3 +88,61 @@ If the snapshot tests fail, it's generally pretty easy to tell whether it's happ
...
 
 Paragon measures code coverage using Jest's built-in `--coverage` flag (which I believe uses istanbul under the hood) and report it via [Coveralls](https://coveralls.io/github/edx/paragon). Shoot for 100% test coverage on your PRs, but use your best judgment if you're really struggling to cover those last few lines. At the very least, don't *reduce* total coverage. Coveralls will fail your build if your PR reduces coverage.
+
+## Semantic Release
+
+Paragon uses the [`semantic-release` package](https://github.com/semantic-release/semantic-release) to automate it's release process (creating Git tags, creating GitHub releases, and publishing to NPM).
+
+### Commit Messages
+
+[`semantic-release` analyzes commit messages to determine the code changes will be associated with a `major`, `minor`, or `patch` release](https://github.com/semantic-release/semantic-release#default-commit-message-format).
```
jaebradley: missing `whether`, also add possibility that code changes are not associated with a release

README.md
```diff
@@ -88,3 +88,61 @@ If the snapshot tests fail, it's generally pretty easy to tell whether it's happ
...
+2. A `scope` - what is this commit impacting? Did you fix a bug in the `Hyperlink` component? Did you add a new feature to the `RadioButtonGroup` component? Currently, the `scope` must be lower-case and `-` separated though switching this to being `camelCase` is currently being investigated.
+3. A `subject` - provide a short description of _what_ your change is.
+  * use imperative, present tense language (so `change` not `changed` or `changes`)
+  * don't capitalize the first letter
+  * don't add a period (`.`) at the end
+  * there is a 50 character limit
+4. A `body` (optional) - add more detail about your change
+5. A `footer` (optional)
+  * > All breaking changes have to be mentioned in footer with the description of the change, justification and migration notes - [`Angular` commit message specification](https://gist.github.com/stephenparish/9941e89d80e2bc58a153#breaking-changes)
+
```
fysheets: Can an example be added to this? This gives all the rules, but it's probably also good to have a tangible example to sanity check against when looking between your commit message and the documentation.

fysheets: Ah - noticed the example at the bottom of the doc, can that just be copied to be up here as well since it's also relevant to this part of the document when committing regularly and not only on merge/build/release process.

jaebradley: Good point! I can add an example commit message.

README.md
```diff
@@ -88,3 +88,61 @@ If the snapshot tests fail, it's generally pretty easy to tell whether it's happ
 ### Coverage
 
 Paragon measures code coverage using Jest's built-in `--coverage` flag (which I believe uses istanbul under the hood) and report it via [Coveralls](https://coveralls.io/github/edx/paragon). Shoot for 100% test coverage on your PRs, but use your best judgment if you're really struggling to cover those last few lines. At the very least, don't *reduce* total coverage. Coveralls will fail your build if your PR reduces coverage.
+
+## Semantic Release
+
+Paragon uses the [`semantic-release` package](https://github.com/semantic-release/semantic-release) to automate it's release process (creating Git tags, creating GitHub releases, and publishing to NPM).
```
MichaelRoytman: Nit: it's -> its

README.md
```diff
@@ -88,3 +88,61 @@ If the snapshot tests fail, it's generally pretty easy to tell whether it's happ
...
+3. `perf` (`patch` release)
+4. `breaking` (`major` release)
+
+#### `Angular` Commit Message Convention
+
+Paragon currently uses [the `Angular` commit message convention](https://gist.github.com/stephenparish/9941e89d80e2bc58a153). As documented in the previously linked gist, a commit that follows the `Angular` commit message convention has four parts:
+1. A `type` - is this commit a `feat`, `fix`, `chore`, `doc`, etc.? There is a set of `type` values to choose from, but again, **only the `feat`, `fix`, `perf`, and `breaking`** `type` values will trigger a release.
+2. A `scope` - what is this commit impacting? Did you fix a bug in the `Hyperlink` component? Did you add a new feature to the `RadioButtonGroup` component? Currently, the `scope` must be lower-case and `-` separated though switching this to being `camelCase` is currently being investigated.
+3. A `subject` - provide a short description of _what_ your change is.
+  * use imperative, present tense language (so `change` not `changed` or `changes`)
```
MichaelRoytman: This is helpful. I, shamefully, have to Google what the imperative tense is each time I write a commit message 😛 .

README.md
```diff
@@ -88,3 +88,61 @@ If the snapshot tests fail, it's generally pretty easy to tell whether it's happ
...
+  * there is a 50 character limit
+4. A `body` (optional) - add more detail about your change
+5. A `footer` (optional)
+  * > All breaking changes have to be mentioned in footer with the description of the change, justification and migration notes - [`Angular` commit message specification](https://gist.github.com/stephenparish/9941e89d80e2bc58a153#breaking-changes)
+
+#### Using `commitlint` in Paragon
+
+Paragon uses the [`commitlint`](https://github.com/marionebl/commitlint) package to lint commit messages. Paragon currently has [a `commit message` hook](https://github.com/edx/paragon/blob/master/package.json#L20) that runs `commitlint`'s, well, commit linting, on the given commit message using the configuration specified in [`commitlint.config.js`](https://github.com/edx/paragon/blob/master/commitlint.config.js). If a commit message fails linting, the commit associated with that message does not end up getting committed.
+
+`commitlint` also comes with [a helpful CLI](https://github.com/marionebl/commitlint/tree/master/@commitlint/cli) that walks one through the entire semantic commit process.
```
MichaelRoytman: Quick question. This CLI is invoked via `npm run gc`, right? Should we document that anywhere? Actually, now that I look at package.json, it looks like there's a `commitmsg` command. Is that the CLI command?

README.md
```diff
@@ -88,3 +88,61 @@ If the snapshot tests fail, it's generally pretty easy to tell whether it's happ
...
+1. `feat` (`minor` release)
+2. `fix` (`patch` release)
+3. `perf` (`patch` release)
+4. `breaking` (`major` release)
+
+#### `Angular` Commit Message Convention
+
+Paragon currently uses [the `Angular` commit message convention](https://gist.github.com/stephenparish/9941e89d80e2bc58a153). As documented in the previously linked gist, a commit that follows the `Angular` commit message convention has four parts:
+1. A `type` - is this commit a `feat`, `fix`, `chore`, `doc`, etc.? There is a set of `type` values to choose from, but again, **only the `feat`, `fix`, `perf`, and `breaking`** `type` values will trigger a release.
+2. A `scope` - what is this commit impacting? Did you fix a bug in the `Hyperlink` component? Did you add a new feature to the `RadioButtonGroup` component? Currently, the `scope` must be lower-case and `-` separated though switching this to being `camelCase` is currently being investigated.
```
MichaelRoytman: Ah, I forgot to make the camelCase thing a GitHub issue. Sorry about that. Would you still like me to do that?

jaebradley: You can though I actually think this is [a `commitlint`](https://github.com/marionebl/commitlint) implementation detail and not one with Paragon. So if you can, see if this is a configurable setting in `commitlint` and then if so, create an issue giving some insight into how to configure this setting.

README.md
jaebradley: lol I can add that note about `npm run gc`

jaebradley: `commitmsg` actually refers to [the `husky` `commitmsg` hook](https://github.com/typicode/husky/blob/master/HOOKS.md).

jaebradley:
* @arizzitano @thallada any additional thoughts?
* @fysheets @MichaelRoytman have I addressed your comments?

PR #101, opened 2017-12-13
fix(input-text): add prop types for the inner Text component

jaebradley:
Follow up to [this comment](https://github.com/edx/paragon/pull/88#discussion_r153579701) from #88.

We should still do prop type checking for wrapped components.


jaebradley:
Looking into the missing test coverage

image

jaebradley:
I know this is a pretty nitty PR @arizzitano but are there any issues here regarding default prop types in `InputText` vs. `asInput`?

PR #110, opened 2017-12-20
feat(modal-icon): Add variant capability to Modal & create Icon

fysheets:
Add ability to apply a variant to Modal - based on the requirement for a Warning Modal in Studio-Frontend that needs an Icon as well.

### DIFF
--- a/src/Icon/index.jsx
+++ b/src/Icon/index.jsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+import styles from './Icon.scss';
+import newId from '../utils/newId';
+
+function Icon(props) {
+  return (
+    <div>
+      <span
+        id={props.id ? props.id : newId('Icon')}
+        className={classNames(props.className)}
+        aria-hidden={props.hidden}
+      />
+      { props.screenReaderText &&
+        <span className={classNames(styles['sr-only'])}>
+          {props.screenReaderText}
+        </span>
+      }
+    </div>
+  );
+}
+
+Icon.propTypes = {
+  id: PropTypes.string,
+  className: PropTypes.arrayOf(PropTypes.string).isRequired,
+  hidden: PropTypes.bool,
+  screenReaderText: PropTypes.string,
+};
+
+Icon.defaultProps = {
+  id: newId('Icon'),
+  hidden: true,
+  screenReaderText: undefined,
+};
+
+export default Icon;
--- a/src/Modal/index.jsx
+++ b/src/Modal/index.jsx
@@ -2,9 +2,12 @@ import React from 'react';
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
 
+import FontAwesomeStyles from 'font-awesome/css/font-awesome.min.css';
 import styles from './Modal.scss';
 import Button, { buttonPropTypes } from '../Button';
+import Icon from '../Icon';
 import newId from '../utils/newId';
+import Variant from '../utils/constants';
 
 class Modal extends React.Component {
   constructor(props) {
@@ -48,6 +51,50 @@ class Modal extends React.Component {
     this.closeButton = input;
   }
 
+  getVariantIconClassName() {
+    const { variant } = this.props;
+    let variantIconClassName;
+
+    switch (variant.status) {
+      case Variant.status.WARNING:
+        variantIconClassName = classNames(
+          FontAwesomeStyles.fa,
+          FontAwesomeStyles['fa-exclamation-triangle'],
+          FontAwesomeStyles['fa-3x'],
+          styles[`text-${variant.status.toLowerCase()}`],
+        );
+        break;
+      default:
+        break;
+    }
+
+    return variantIconClassName;
+  }
+
+  getVariantGridBody(body) {
+    const { variant } = this.props;
+
+    return (
+      <div className={styles['container-fluid']}>
+        <div className={styles.row}>
+          <div className={styles['col-md-10']}>
+            <div>
+              {body}
+            </div>
+          </div>
+          <div className={styles.col}>
+            <Icon
+              id={newId(`Modal-${variant.status}`)}
+              className={[
+                this.getVariantIconClassName(),
+              ]}
+            />
+          </div>
+        </div>
+      </div>
+    );
+  }
+
   close() {
     this.setState({ open: false });
     this.props.onClose();
@@ -89,10 +136,15 @@ class Modal extends React.Component {
   }
 
   renderBody() {
+    const { variant } = this.props;
     let { body } = this.props;
 
     if (typeof body === 'string') {
       body = <p>{body}</p>;
+    }
+
+    if (variant.status) {
+      body = this.getVariantGridBody(body);
     }
 
     return body;
@@ -119,7 +171,7 @@ class Modal extends React.Component {
         <div className={styles['modal-dialog']}>
           <div className={styles['modal-content']}>
             <div className={styles['modal-header']}>
-              <h5 className={styles['modal-title']} id={this.headerId}>{this.props.title}</h5>
+              <h2 className={styles['modal-title']} id={this.headerId}>{this.props.title}</h2>
               <Button
                 label={<span aria-hidden="true">&times;</span>}
                 aria-label={this.props.closeText}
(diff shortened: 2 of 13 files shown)


src/Icon/index.jsx
```diff
@@ -0,0 +1,39 @@
...
+        </span>
+      }
+    </div>
+  );
+}
+
+Icon.propTypes = {
+  id: PropTypes.string.isRequired,
+  className: PropTypes.arrayOf(PropTypes.string).isRequired,
+  describedBy: PropTypes.string,
```
fysheets: @cptvitamin does it ever make sense for an Icon to have an aria-describedBy attribute? I was trying to create a story for this to demo it, but was having a hard time thinking of when this would be used.

cptvitamin: @fysheets nothing [explicitly prevents](https://www.w3.org/TR/html-aria/#allowed-aria-roles-states-and-properties) aria-describedby from being used on any element.  That said, I personally can't think of a valid use case for using it on an icon...  Icons are usually used to convey something pretty concise, like STOP, WARNING, ERROR, EMAIL, UNREAD etc.  If the icon requires more nuance to describe it, i would think that nuance would be in accessible text that is adjacent to the icon.

fysheets:
Removed aria-describedBy from Icon for now, can be added back if a use case arises - also rebased on top of differentiated builds.

src/utils/constants.js
```diff
@@ -0,0 +1,10 @@
+const Settings = {
```
fysheets: @arizzitano I know we've gone back and forth on this some more, but I was thinking, Settings still seems too broad for me (but am fine keeping it if we still think it makes the most sense) but what if we called these Patterns? I remember hearing UI more recently use "Pattern Library" as terminology even if it's not the same meaning as the deprecated Pattern Library we had?

fysheets:
@edx/educator-dahlia @edx/fedx-team ready for review :)

src/Icon/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+          {props.srText}
+        </span>
+      }
+    </div>
+  );
+}
+
+Icon.propTypes = {
+  id: PropTypes.string.isRequired,
+  className: PropTypes.arrayOf(PropTypes.string).isRequired,
```
jaebradley: If this is an array of strings, should the name indicate multiple values? Obviously, there's a conflict with `classNames` but maybe there's some other name that could make sense?

src/Icon/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+function Icon(props) {
+  return (
+    <div>
+      <span
+        id={props.id}
+        className={classNames(props.className)}
+        aria-hidden={props.hidden}
+      />
+      { props.srText &&
+        <span className={classNames(styles['sr-only'])}>
```
jaebradley: is this `classNames` necessary?

src/Icon/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+Icon.propTypes = {
+  id: PropTypes.string.isRequired,
+  className: PropTypes.arrayOf(PropTypes.string).isRequired,
+  hidden: PropTypes.bool,
+  srText: PropTypes.string,
+};
+
+Icon.defaultProps = {
+  hidden: true,
+  srText: undefined,
```
jaebradley: The documentation makes it really clear what `srText` is supposed to represent (which is great). However, I'm wondering if the added clarity of the prop being called `screenReaderText` is worth the cost of a slightly more verbose name.

src/Icon/index.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+import styles from './Icon.scss';
+
+function Icon(props) {
```
jaebradley: you could also consider doing 
```javascript
function Icon({ id, className, hidden, srText })
```
and thus wouldn't have to write `props.X` everywhere.

Not a big deal either way!

src/Modal/index.jsx
```diff
@@ -48,6 +51,50 @@ class Modal extends React.Component {
     this.closeButton = input;
   }
 
+  getVariantIcon() {
```
jaebradley: Nit: it looks like this method is not technically getting a variant icon but rather getting the variant icon class name?

src/Modal/index.jsx
```diff
@@ -119,7 +171,7 @@ class Modal extends React.Component {
         <div className={styles['modal-dialog']}>
           <div className={styles['modal-content']}>
             <div className={styles['modal-header']}>
-              <h5 className={styles['modal-title']} id={this.headerId}>{this.props.title}</h5>
+              <h2 className={styles['modal-title']} id={this.headerId}>{this.props.title}</h2>
```
jaebradley: Will changing this negatively impact any of our current modals?

src/utils/constants.js
```diff
@@ -0,0 +1,10 @@
+const Variant = {
```
jaebradley: Should this have [`Object.freeze`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) applied to it?

src/Icon/index.jsx
fysheets: `className` is the name for the string array format on all other Paragon components that take that as a prop.

src/Icon/index.jsx
fysheets: I find that the `props.X` is more clear especially with the format of other components but I get where you're coming from too.

src/Icon/index.jsx
fysheets: It was more for uniformity and removing the question of "wait why was this used in the above element but not this span?"

src/Icon/index.jsx
fysheets: This is a fair point, I think I defaulted this way because of the class name of `sr-only`, but the more verbose name here will definitely help with understanding for folks not immediately familiar with a11y rules.

src/Modal/index.jsx
fysheets: Good point - I will clarify this one as well for clarity.

src/Modal/index.jsx
fysheets: It should not - the change here to `h2` comes from a quick discussion with Ari & Mark S. noting that it would be expected to be that type.

src/utils/constants.js
fysheets: Possibly, this is one of those things that I'm not sure of the expectations / necessity of that but would be willing to discuss more about it :)

fysheets: Actually the more I read about `Object.freeze` I'm leaning more towards just a yes on this one

src/Icon/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+        <span className={classNames(styles['sr-only'])}>
+          {props.screenReaderText}
+        </span>
+      }
+    </div>
+  );
+}
+
+Icon.propTypes = {
+  id: PropTypes.string.isRequired,
```
arizzitano: Why is `id` required? Could we leverage `utils/newId` instead?

fysheets: Good spot -- It doesn't need to be required, I think I just did this out of habit and didn't realize -- will update to use the newId as we do elsewhere.

src/Icon/index.jsx
```diff
@@ -0,0 +1,38 @@
...
+import PropTypes from 'prop-types';
+
+import styles from './Icon.scss';
+import newId from '../utils/newId';
+
+function Icon(props) {
+  return (
+    <div>
+      <span
+        id={props.id ? props.id : newId('Icon')}
```
jaebradley: Consider using `||` here instead - [`airbnb/javascript#15.7`](https://github.com/airbnb/javascript#comparison--unneeded-ternary)

src/Icon/index.jsx
jaebradley: Consistency is important but we could consider in the future

1. Renaming the `classNames` import within the file and thus, resolving the naming conflict
2. Using something like `classes` instead of `className`

src/Modal/index.jsx
jaebradley: 👍  sweet, just wanted to check

src/utils/constants.js
```diff
@@ -0,0 +1,10 @@
+const Variant = Object.freeze({
```
jaebradley: should this file be called `variant.js` ([`airbnb/javascript#23.6`](https://github.com/airbnb/javascript#naming--filename-matches-export)) or are there plans to add other constants to this file?

fysheets: There are no active plans from my part, so this could probably be renamed

src/Icon/index.jsx
fysheets: I went with the ternary based on the previous findings we had with `null/undefined` to ensure that cases were properly caught

PR #115, opened 2018-01-05
feat(checkboxgroup): New CheckBoxGroup, update CheckBox & asInput styles

fysheets:
# [EDUCATOR-2048](https://openedx.atlassian.net/browse/EDUCATOR-2048)

### DIFF
--- a/src/CheckBoxGroup/index.jsx
+++ b/src/CheckBoxGroup/index.jsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ElementPropTypes from 'react-element-proptypes';
+
+import CheckBox from '../CheckBox';
+import styles from './CheckBoxGroup.scss';
+
+function CheckBoxGroup(props) {
+  return (
+    <div className={styles['form-group']}>
+      {props.children}
+    </div>
+  );
+}
+
+CheckBoxGroup.propTypes = {
+  children: PropTypes.arrayOf(ElementPropTypes.elementOfType(CheckBox)).isRequired,
+};
+
+export default CheckBoxGroup;
--- a/src/asInput/index.jsx
+++ b/src/asInput/index.jsx
@@ -41,7 +41,7 @@ export const defaultProps = {
   themes: [],
 };
 
-const asInput = (WrappedComponent, labelFirst = true) => {
+const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) => {
   class NewComponent extends React.Component {
     constructor(props) {
       super(props);
@@ -125,8 +125,31 @@ const asInput = (WrappedComponent, labelFirst = true) => {
       return desc;
     }
 
+    getLabel() {
+      return (
+        <label
+          id={`label-${this.state.id}`}
+          htmlFor={this.state.id}
+          className={[classNames({
+            [styles['form-check-label']]: this.isGroupedInput(),
+          })]}
+        >
+          {this.props.label}
+        </label>
+      );
+    }
+
     hasDangerTheme() {
       return this.props.themes.includes('danger');
+    }
+
+    isGroupedInput() {
+      switch (inputType) {
+        case 'checkbox':
+          return true;
+        default:
+          return false;
+      }
     }
 
     handleBlur(event) {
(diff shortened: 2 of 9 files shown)


src/asInput/index.jsx
```diff
@@ -41,7 +41,7 @@ export const defaultProps = {
   themes: [],
 };
 
-const asInput = (WrappedComponent, labelFirst = true) => {
+const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) => {
```
fysheets: Adding in this `inputType` to help with styling logic. I could see this being leverage for the upcoming Search input box that Dahlia is going to be working on (label layout inline with input field instead of above). Also if any other input types need specific classes/formatting this will help with that logic. If there is a better way to determine this that I was unaware of, let me know! :)

src/asInput/index.jsx
```diff
@@ -125,6 +125,20 @@ const asInput = (WrappedComponent, labelFirst = true) => {
...
+          htmlFor={this.state.id}
+          className={[classNames({
+            [styles['form-check-label']]: inputType === 'checkbox',
+          })]}
+        >
+          {this.props.label}
+        </label>
+      );
+    }
+
```
fysheets: Moved logic for creating the label element out to help with readability and styling. As noted in the above comment, this will likely need a different class name / formatting for the upcoming Search input type that Dahlia will be working on shortly.

src/asInput/index.jsx
```diff
@@ -149,21 +163,28 @@ const asInput = (WrappedComponent, labelFirst = true) => {
     render() {
       const { description, error, describedBy } = this.getDescriptions();
       return (
-        <div className={styles['form-group']}>
-          {labelFirst && <label id={`label-${this.state.id}`} htmlFor={this.state.id}>{this.props.label}</label>}
+        <div className={[classNames({
+            [styles['form-group']]: inputType !== 'checkbox',
+            [styles['form-check']]: inputType === 'checkbox',
+          })]}
```
fysheets: Bootstrap documentation shows that checkboxes should be wrapped with the form-check style, the CheckBoxGroup will wrap multiple checkboxes within the same group (instead of each in their own group) as expected. https://getbootstrap.com/docs/4.0/components/forms/#checkboxes-and-radios

src/asInput/index.jsx
```diff
@@ -149,21 +163,28 @@ const asInput = (WrappedComponent, labelFirst = true) => {
...
             {...this.props}
             {...this.state}
             className={[classNames(
-              styles['form-control'],
-              { [styles['is-invalid']]: !this.state.isValid && this.hasDangerTheme() },
+              {
+                [styles['form-control']]: inputType !== 'checkbox',
+                [styles['form-check-input']]: inputType === 'checkbox',
+                [styles['is-invalid']]: !this.state.isValid && this.hasDangerTheme(),
+              },
```
fysheets: Bootstrap has a different class name to be applied to the input field when that input is a checkbox. https://getbootstrap.com/docs/4.0/components/forms/#checkboxes-and-radios

fysheets:
This should be ready for initial review - Looking forward to discussion that will arise around architecture.

src/asInput/index.jsx
```diff
@@ -149,21 +163,28 @@ const asInput = (WrappedComponent, labelFirst = true) => {
     render() {
       const { description, error, describedBy } = this.getDescriptions();
       return (
-        <div className={styles['form-group']}>
-          {labelFirst && <label id={`label-${this.state.id}`} htmlFor={this.state.id}>{this.props.label}</label>}
+        <div className={[classNames({
+            [styles['form-group']]: inputType !== 'checkbox',
```
arizzitano: All these checks for `inputType === 'checkbox'` feel a little brittle to me. Could you maybe encapsulate this check as a method like, say, `isGroupedInput`? That way we could easily update it when/if we pass `RadioButton` through `asInput` as well.

src/asInput/index.jsx
arizzitano: This is probably a more flexible approach than the existing `labelFirst` param. I wonder if we should kill `labelFirst` and just lean primarily on `inputType` for label styling logic?

src/CheckBoxGroup/index.jsx
```diff
@@ -0,0 +1,20 @@
...
+import CheckBox from '../CheckBox';
+import styles from './CheckBoxGroup.scss';
+
+function CheckBoxGroup(props) {
+  return (
+    <div className={styles['form-group']}>
+      {props.children}
+    </div>
+  );
+}
```
arizzitano: So to be clear here, the primary purpose of the `CheckBoxGroup` is to wrap the checkboxes in a div with the proper styles? Not opposed to this, but curious whether we might want to eventually build additional functionality into this component. Like, would this be the place for something like a `clearAll` method?

src/asInput/index.jsx
fysheets: I could see an argument for a single input type wanting to be able to support both a label before/after the input field, which is why I didn't simply leverage the labelFirst option for the checks I was doing below. I'm inclined to leave it for now, until we're certain that we won't want it later.

src/CheckBoxGroup/index.jsx
fysheets: I could definitely see this component growing in scope, with some sort of `clearAll` method or anything else that I likely haven't thought of yet. But yeah, for the purposes of this PR it was mostly to get the correct stylings in place.

src/asInput/index.jsx
fysheets: Good call -- I'll move that through and get the PR updated shortly

PR #116, opened 2018-01-05
fix(table): Buttons in table headings look like headings

dsjen:
UX would like the table headings to look like the standard bootstrap headings. When the column is sortable, the heading is a button. This PR makes those buttons look like headings.

screen shot 2018-01-05 at 3 01 44 pm

@edx/educator-dahlia @arizzitano


dsjen:
Thanks for the quick review!

PR #130, opened 2018-01-25
feat(checkbox): change CheckBox to be controllable via props

MichaelRoytman:
BREAKING CHANGE: add a componentWillReceiveProps function to CheckBox to allow for CheckBox to be updated by changes to prop, prop defaultChecked has been replaced by prop checked.

### DIFF
--- a/src/CheckBox/CheckBox.stories.jsx
+++ b/src/CheckBox/CheckBox.stories.jsx
@@ -4,6 +4,42 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 
 import CheckBox from './index';
+import Button from '../Button';
+
+class CheckBoxWrapper extends React.Component {
+  constructor(props) {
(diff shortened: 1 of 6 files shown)


MichaelRoytman:
This is ready for a review! @edx/educator-dahlia @arizzitano @jaebradley

src/CheckBox/CheckBox.stories.jsx
```diff
@@ -4,6 +4,42 @@ import React from 'react';
...
+    return (
+      <div>
+        <Button
+          onClick={this.toggleCheckBox}
+          label="Click me to toggle the check box!"
+          buttonType="light"
+        />
+        <CheckBox
+          name="checkbox"
+          label=""
```
MichaelRoytman: I can't think of a useful label, but it's required. Is an empty string okay?

PR #133, opened 2018-01-29
fix(input-label): extend asInput to allow label prop as element

fysheets:
Extend input labels to accept elements as well as strings. The use case we are trying to handle is wrapping labels within spans specifying their language to manage i18n fallbacks correctly.

### DIFF
--- a/.storybook/__snapshots__/Storyshots.test.js.snap
+++ b/.storybook/__snapshots__/Storyshots.test.js.snap
@@ -881,6 +881,54 @@ exports[`Storyshots InputText focus test 1`] = `
 </div>
 `;
 
+exports[`Storyshots InputText label as element 1`] = `
+<div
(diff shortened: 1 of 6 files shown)


jaebradley:
@fysheets thanks for doing this - mind updating the [`asInput#label`](https://github.com/edx/paragon/blob/master/src/asInput/README.md#label-string-required) part of the documentation as well?

fysheets:
@jaebradley bah! I knew I was forgetting something, what would be the best way to go about doing this now that it has been merged?

jaebradley:
@fysheets probably a straight-forward `docs` commit would do the trick -- I did something like this in #131. A `docs` commit will not trigger a new release.

fysheets:
@jaebradley cool - thanks!

PR #137, opened 2018-01-30
feat(mailtolink): add MailtoLink component

jaebradley:
### Description

Creating a `mailto` hyperlink should be simple:

* There shouldn't be any string interpolation
```javascript
<a href=`mailto:${someEmailAddress}`>ew</a>
```
* There shouldn't be any URI encoding
```javascript
<a href=`mailto:${someEmailAddress}?subject=${encodeURI('foo bar baz')}`>gross</a>
```

### Example `MailtoLink` usage
```javascript
<MailtoLink to={someEmailAddress} content={'ew'} />
```

```javascript
<MailtoLink to={someEmailAddress} subject={'foo bar baz'} content={'gross'}/>
```

```javascript
<MailtoLink
  to=['foo@example.com', 'bar@example.com', 'baz@example.com']
  cc=['foo@example.com', 'bar@example.com', 'baz@example.com']
  bcc=['foo@example.com', 'bar@example.com', 'baz@example.com']
  subject="Check out this mailto component!"
  body="This mailto component is awesome!"
  content="email with subject and body"
/>
```

### Discussion

`MailtoLink` is a thin wrapper on top of [the `Hyperlink` component](https://github.com/edx/paragon/blob/master/src/Hyperlink/index.jsx).

It generates the `mailto` URL using [the `mailto-link` library](https://www.npmjs.com/package/mailto-link). The `mailto-link` library does not have a ton of downloads but [the source](https://github.com/bendrucker/mailto-link/blob/master/index.js) looks pretty good (it's basically how I started implementing it). If we feel uncomfortable relying on this package, we could roll our own.

For the email address `props`, I used [the `email-prop-type` library](https://github.com/jaebradley/email-prop-type) for `prop` type validation.

**Full disclosure: I am the author and maintainer of the `email-prop-type` package**

The [`email-prop-type` library's source](https://github.com/jaebradley/email-prop-type/blob/master/src/index.js) is very simple (it relies on [the popular `isemail` package](https://www.npmjs.com/package/isemail) for it's email validation) - we could replicate the behavior as a custom prop validator if we feel uncomfortable relying on the package.


src/MailtoLink/index.jsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react'; // eslint-disable-line no-unused-vars
```
jaebradley: I was getting an `eslint` error here for some weird reason that I couldn't get rid of

alt-text

src/Hyperlink/index.jsx
```diff
@@ -51,7 +51,7 @@ Hyperlink.defaultProps = {
 
 Hyperlink.propTypes = {
   destination: PropTypes.string.isRequired,
-  content: PropTypes.string.isRequired,
+  content: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
```
jaebradley: I want to explicitly call out the addition of the `element` prop type in `Hyperlink`.

src/MailtoLink/index.jsx
arizzitano: This is bizarre. I'd guess it's because it's a functional component and doesn't explicitly use `React`, but we already have a number of existing components that do that. I'm fine leaving in the ignore rule for now but if this happens much more we should think about a large-scale ignore.

storybook-static/index.html
```diff
@@ -0,0 +1,44 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
arizzitano: Actually, we probably don't want to gitignore it since it's used on `gh-pages`, but I don't think we need to check it into the master branch.

PR #144, opened 2018-02-10
build(storybook): add README plugin

jaebradley:
[`storybook-readme`](https://github.com/tuchk4/storybook-readme) adds a `README` panel. This is useful for end users as it provides context about the component / story they are viewing.

image

I only added this add-on for the Hyperlink component.

We can / should clean up our READMEs so that they look nicer, but I'm going to leave that task to a follow-on PR.


jaebradley:
Ugh, nevermind...([`storybook-readme#41`](https://github.com/tuchk4/storybook-readme/issues/41))

image


alt-text

PR #145, opened 2018-02-10
build(storybook): use addon-info

jaebradley:
Ok, let's try this again (after #144 crashed and burned).

[`storybook/addon-info`](https://github.com/storybooks/storybook/tree/master/addons/info) allows you to build storybooks that look like

image

I would argue this is a significantly better user experience when looking at various stories.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "@commitlint/prompt-cli": "^6.0.2",
     "@storybook/addon-actions": "^3.3.10",
     "@storybook/addon-console": "^1.0.0",
+    "@storybook/addon-info": "^3.3.12",
     "@storybook/addon-options": "^3.3.10",
     "@storybook/addon-storyshots": "^3.3.10",
     "@storybook/addons": "^3.3.10",
@@ -76,6 +77,7 @@
     "identity-obj-proxy": "^3.0.0",
     "jest": "^22.0.4",
     "jest-cli": "^22.0.4",
+    "markdown-loader-jest": "^0.1.1",
     "node-sass": "^4.5.3",
     "postcss-scss": "^1.0.1",
     "react-router-dom": "^4.1.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -76,6 +77,7 @@
     "identity-obj-proxy": "^3.0.0",
     "jest": "^22.0.4",
     "jest-cli": "^22.0.4",
+    "markdown-loader-jest": "^0.1.1",
```
jaebradley: Need [`markdown-loader-jest`](https://github.com/peerigon/markdown-loader) in order to import `README.md` files for stories.

package.json
```diff
@@ -91,6 +93,10 @@
     "webpack-dev-server": "^2.4.4"
   },
   "jest": {
+    "transform": {
+      "^.+\\.md?$": "markdown-loader-jest",
+      "^.+\\.jsx?$": "<rootDir>/node_modules/babel-jest"
```
jaebradley: When `markdown-loader-jest` is added here, [the `babel-jest` transformer is no longer loaded automatically](https://facebook.github.io/jest/docs/en/configuration.html#transform-object-string-string), so we have to specify it manually.

image

src/Hyperlink/Hyperlink.stories.jsx
```diff
@@ -16,31 +19,27 @@ const onClick = (event) => {
...
-      onClick={onClick}
-    />
-  ))
-  .add('with icon as content', () => (
-    <Hyperlink
-      destination="https://www.edx.org"
-      content={(<span className="fa fa-book" />)}
-    />
-  ));
+  .add('minimal usage', wrapComponent(<Hyperlink
```
jaebradley: I'm not a big fan of this format, but there are `eslint` violations (`unexpected new line`) when `<Hyperlink />`  is on it's own line. 😞

PR #146, opened 2018-02-15
feat(asinput): add "inline" prop

thallada:
Allows users of `asInput` to display [forms inline](https://v4-alpha.getbootstrap.com/components/forms/#inline-forms).

### DIFF
--- a/.storybook/__snapshots__/Storyshots.test.js.snap
+++ b/.storybook/__snapshots__/Storyshots.test.js.snap
@@ -3242,6 +3242,50 @@ exports[`Storyshots InputText different textual input types 1`] = `
 </form>
 `;
 
+exports[`Storyshots InputText displayed inline 1`] = `
+<div
+  className={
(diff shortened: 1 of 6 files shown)


jaebradley:
I think this makes sense - a question I have is where do we want to draw the line (**if** we want to draw a line at all) between a component's CSS vs. an API user's "custom" CSS for their implementation of that component?

Is it something like "if this is a "common" use case" (whatever "common" might mean)?

thallada:
@jaebradley I agree that the line is pretty murky here. We usually solve this problem with the `className` prop which allows users to apply any styles via tacking on an extra class name to the element. However, in this case there are a bunch of nested elements and it's not clear which element that `classNames` should affect.

I would argue that this `inline` case is a common one and not "custom" because it is a feature that Bootstrap provides. I see this as on the same level as the `disabled` prop that already exists.

PR #147, opened 2018-02-15
docs(inputtext): demo "type" prop w/ new story

thallada:
I thought this would be useful for users of Paragon to know. There are many different [textual inputs](https://v4-alpha.getbootstrap.com/components/forms/#textual-inputs) that Bootstrap provides which will change the way text is entered.


jaebradley:
The `fix` commit message will publish a new `patch`ed version of Paragon. None of these changes seem to deal with the underlying API, so it doesn't seem necessary.

I view `Stories` as documentation, so I'm not opposed to the `docs` commit type (which would not publish a new version of Paragon), but I'm open to other suggestions.

thallada:
Forgot about the `docs` type. That makes more sense.

I also decided to just demo all the different textual input types in that story.

PR #153, opened 2018-02-23
feat(modal): add parentSelector prop, use portals

Mjloturco:
Modal currently does not render on bootstrap pages, because paragon uses class names different from the default bootstrap class names. This PR brings the modal more in line with bootstrap's default modal and renders the modal via [```createPortal()```](https://reactjs.org/docs/portals.html)

### DIFF
--- a/src/Modal/Modal.stories.jsx
+++ b/src/Modal/Modal.stories.jsx
@@ -34,6 +34,7 @@ class ModalWrapper extends React.Component {
           open={this.state.open}
           title={this.props.title}
           body={this.props.body}
+          parentSelector={this.props.parentSelector}
           onClose={this.resetModalWrapperState}
         />
         <Button
@@ -50,6 +51,11 @@ class ModalWrapper extends React.Component {
 ModalWrapper.propTypes = {
   title: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
   body: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
+  parentSelector: PropTypes.string,
+};
+
+ModalWrapper.defaultProps = {
+  parentSelector: 'body',
 };
 
 storiesOf('Modal', module)
@@ -172,4 +178,51 @@ storiesOf('Modal', module)
       onClose={() => {}}
       variant={{ status: Variant.status.WARNING }}
     />
+  ))
+  .add('modal inside special div', () => (
+    <div>
+      <div>
+        <div className="special-div" />
+      </div>
+      <ModalWrapper
+        title="I am the modal!"
+        body="I was invoked by a button!"
+        parentSelector=".special-div"
+      />
+    </div>
+  ))
+  .add('two modals with the same target', () => (
+    <div>
+      <div>
+        <div className="target-div" />
+      </div>
+      <ModalWrapper
+        title="I am the first modal!"
+        body="I target one"
+        parentSelector=".target-div"
+      />
+      <ModalWrapper
+        title="I am the second modal!"
+        body="I target one"
+        parentSelector=".target-div"
+      />
+    </div>
+  ))
+  .add('two modals with seperate targets', () => (
+    <div>
+      <div>
+        <div className="target-div-one" />
+        <div className="target-div-two" />
+      </div>
+      <ModalWrapper
+        title="I am the first modal!"
+        body="I target one"
+        parentSelector=".target-div-one"
+      />
+      <ModalWrapper
+        title="I am the second modal!"
+        body="I target two"
+        parentSelector=".target-div-two"
+      />
+    </div>
   ));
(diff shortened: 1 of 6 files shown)


Mjloturco:
@reviewers, just noticed that it looks like a bunch of build changes were pulled into the diff. I'll look to remove those before merging anything.

src/Modal/index.jsx
```diff
@@ -43,6 +58,10 @@ class Modal extends React.Component {
     }
   }
 
+  componentWillUnmount() {
+    modalRoot.removeChild(this.el);
```
jaebradley: Just out of curiosity are we missing anything that we'd get from [`ReactDOM.unmountComponentAtNode`](https://reactjs.org/docs/react-dom.html#unmountcomponentatnode)?

src/Modal/index.jsx
```diff
@@ -20,12 +26,21 @@ class Modal extends React.Component {
 
     this.headerId = newId();
 
+    this.el = document.createElement('div');
+
+    if (!modalRoot) {
+      modalRoot = document.createElement('div');
+      modalRoot.id = modalRootId;
+      document.body.appendChild(modalRoot);
```
jaebradley: What happens if I want to append this modal to a specific parent in the DOM? Should we add a `parentSelector` prop that defaults to the document body?

src/Modal/index.jsx
```diff
@@ -20,12 +26,21 @@ class Modal extends React.Component {
...
+    }
+
     this.state = {
       open: props.open,
     };
   }
 
   componentDidMount() {
+    modalRoot.appendChild(this.el);
```
jaebradley: So I'm probably totally missing something here...

It looks like we're creating a `modalRoot` `div` that we're appending to document body. Then for this `modalRoot` `div` we're appending another `div` (`this.el`), which is where we're creating the portal.

Is there any reason we're not appending `this.el` to document body? In other words, what is `modalRoot` doing that `this.el` couldn't do?

src/Modal/index.jsx
```diff
@@ -9,6 +10,11 @@ import Icon from '../Icon';
 import newId from '../utils/newId';
 import Variant from '../utils/constants';
 
+
+const modalRootId = 'modal-root';
+let modalRoot = document.getElementById(modalRootId);
```
jaebradley: Having this outside the scope of Modal makes me nervous - what happens if somebody deletes this element in the DOM? `modalRoot` will still probably be defined right?

src/Modal/index.jsx
Mjloturco: We could add a parentSelector prop, though I believe we do expect to want to always render the modal at the top level component, that was at a suggestion from @arizzitano who may have more thoughts on that.

src/Modal/index.jsx
Mjloturco: I'm not sure, the docs for unmountComponentAtNode specifically mentions cleaning up eventListeners and such, which removeChild does not. I'll check to see if unmountComponentAtNode accomplishes what we want.

Mjloturco:
@fysheets hi

jaebradley:
Just because I have a terrible memory and have already forgotten the reason for these changes here's a relevant conversation from the FedX channel

image

Just want to add this for historical purposes.

src/Modal/Modal.stories.jsx
```diff
@@ -50,6 +51,11 @@ class ModalWrapper extends React.Component {
 ModalWrapper.propTypes = {
   title: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
   body: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
+  parentSelector: PropTypes.string,
+};
+
+ModalWrapper.defaultProps = {
+  parentSelector: '',
```
jaebradley: I'm being uber-paranoid, but can we make the default value 'body' or something similar?

fysheets: screen shot 2018-03-07 at 4 27 39 pm
Seconded :)

fysheets:
Unblocking from my end - as we've discussed and I've tested to get a plan of attack from our team's end. Will let others continue with additional technical reviews etc. Also, obviously pending tests :)

src/Modal/index.jsx
```diff
@@ -43,6 +63,10 @@ class Modal extends React.Component {
     }
   }
 
+  componentWillUnmount() {
+    ReactDOM.unmountComponentAtNode(this.el);
```
jaebradley: Does `modal-root` need to get unmounted as well?

src/Modal/index.jsx
```diff
@@ -20,12 +22,30 @@ class Modal extends React.Component {
 
     this.headerId = newId();
 
+    const modalRootId = 'modal-root';
+    this.modalRoot = document.getElementById(modalRootId);
```
jaebradley: What happens if there are two modals on the same page if `id`s are supposed to be unique across the DOM?

PR #156, opened 2018-02-26
Pass Disabled prop down in InputText

Mjloturco:
The InputText component currently cannot be disabled, this PR ensures that the disabled prop gets passed through our component down to the html select element.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "eslint-config-airbnb": "^16.0.0",
     "eslint-config-edx": "^4.0.1",
     "eslint-import-resolver-webpack": "^0.8.1",
-    "eslint-plugin-compat": "^2.1.0",
+    "eslint-plugin-compat": "^2.2.0",
     "eslint-plugin-jsx-a11y": "^5.1.0",
     "eslint-plugin-react": "^7.1.0",
     "extract-text-webpack-plugin": "^3.0.1",
--- a/src/InputSelect/index.jsx
+++ b/src/InputSelect/index.jsx
@@ -50,6 +50,7 @@ class Select extends React.Component {
         onChange={props.onChange}
         onBlur={props.onBlur}
         ref={props.inputRef}
+        disabled={props.disabled}
       >
         {options}
       </select>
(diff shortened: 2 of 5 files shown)


jaebradley:
@Mjloturco check out [our commit message syntax](https://github.com/edx/paragon#commit-messages)

jaebradley:
@Mjloturco can you add a story that illustrates the `disabled` prop?

jaebradley:
@Mjloturco do you have commit linting setup correctly? Upper-case scopes should not pass commit message linting

image

Mjloturco:
@jaebradley I will add a story to illustrate and i'll fix the snapshots (though I haven't looked at what those are yet). I may have had a workaround to get that commit through... will rename

src/InputSelect/index.jsx
```diff
@@ -50,6 +50,7 @@ class Select extends React.Component {
         onChange={props.onChange}
         onBlur={props.onBlur}
         ref={props.inputRef}
+        disabled={props.disabled}
```
jaebradley: Can you add this to [the `Input Select README.md`](https://github.com/edx/paragon/blob/master/src/InputSelect/README.md)

jaebradley: No you don't need to do this

package.json
```diff
@@ -71,7 +71,7 @@
     "eslint-config-airbnb": "^16.0.0",
     "eslint-config-edx": "^4.0.1",
     "eslint-import-resolver-webpack": "^0.8.1",
-    "eslint-plugin-compat": "^2.1.0",
+    "eslint-plugin-compat": "^2.2.0",
```
jaebradley: Why was this updated?

jaebradley: So it looks like [`eslint-plugin-compat` is pinned to `2.2.0` in `package-lock.json`](https://github.com/edx/paragon/blob/master/package-lock.json#L6213)

image

So this shouldn't change anything.

jaebradley:
alt-text

PR #157, opened 2018-02-26
build(webpack): keep babel config only in .babelrc

thallada:
I think that specifying the preset option in the webpack config will override whatever we have in the `.babelrc`, which was causing ES6 code to end up in the themeable target (the "env" preset compiles down to ES6 stuff by default).

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -41,9 +41,6 @@ module.exports = targetProperties.map(config => ({
         use: [
           {
             loader: 'babel-loader',
-            options: {
-              presets: ['env'],
-            },
           },
           { loader: 'source-map-loader' },
         ],


jaebradley:
I think `env` compiles down to `ES5`
> A Babel preset that compiles ES2015+ down to ES5 by automatically determining the Babel plugins and polyfills you need based on your targeted browser or runtime environments.
- [`babel-preset-env`](https://www.npmjs.com/package/babel-preset-env)

That being said, it could be useful to move all our `babel`ing to one place.

jaebradley:
@thallada if you can confirm that everything was compiling down to `ES5` in the first place can you change this to a non-release triggering commit message (`build`, `refactor`, maybe even `style`) since there's nothing here that should theoretically generate a different release (right?).

Additionally, (and I'm sorry I'm a nit about this), do you mind changing the PR name and commit message to reflect that this change?

thallada:
@jaebradley I changed to a `build` commit. I'm going to merge this now.

PR #158, opened 2018-02-27
Greenkeeper/webpack 4.0.1

jaebradley:

PR #170, opened 2018-03-08
chore(package): update webpack to version 4.1.0

jaebradley:
Closes #154 and #158 

@thallada I tried to follow as much of [edx/front-end-cookie-cutter-application#11](https://github.com/edx/front-end-cookie-cutter-application/pull/11/) as possible.

Let's work off of this PR since it's the latest webpack 4 version (~~I take that back, `4.1.1` got published yesterday `#lesigh`~~ should now be updated to `4.1.1`).

What's the right way to make sure our builds haven't regressed in some way? Install Paragon from this commit hash in SFE & edX Platform?

[Here's an article about webpack 4 that was published by the webpack core team.](https://medium.com/webpack/webpack-4-released-today-6cdb994702d4)

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "eslint-plugin-compat": "^2.1.0",
     "eslint-plugin-jsx-a11y": "^5.1.0",
     "eslint-plugin-react": "^7.1.0",
-    "extract-text-webpack-plugin": "^3.0.1",
+    "extract-text-webpack-plugin": "^4.0.0-beta.0",
     "file-loader": "^1.1.4",
     "greenkeeper-lockfile": "^1.7.1",
     "husky": "^0.14.1",
@@ -92,8 +92,9 @@
     "style-loader": "^0.20.1",
     "travis-deploy-once": "^4.3.2",
     "uglifyjs-webpack-plugin": "^1.1.8",
-    "webpack": "^3.0.0",
+    "webpack": "^4.1.1",
     "webpack-build-notifier": "^0.1.22",
+    "webpack-cli": "^2.0.10",
     "webpack-dev-server": "^3.0.0"
   },
   "jest": {
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,5 +1,4 @@
 const path = require('path');
-const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 
 // we build the library for two different build targets:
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -72,7 +72,7 @@
     "eslint-plugin-compat": "^2.1.0",
     "eslint-plugin-jsx-a11y": "^5.1.0",
     "eslint-plugin-react": "^7.1.0",
-    "extract-text-webpack-plugin": "^3.0.1",
+    "extract-text-webpack-plugin": "^4.0.0-beta.0",
```
jaebradley: I prefer specifying a semver value instead of `next` if we can help it.

package.json
```diff
@@ -90,8 +90,9 @@
     "style-loader": "^0.20.1",
     "travis-deploy-once": "^4.3.2",
     "uglifyjs-webpack-plugin": "^1.1.8",
-    "webpack": "^3.0.0",
+    "webpack": "^4.1.1",
     "webpack-build-notifier": "^0.1.22",
+    "webpack-cli": "^2.0.10",
```
jaebradley: `webpack-cli` is necessary for webpack 4

webpack.config.js
```diff
@@ -1,5 +1,4 @@
 const path = require('path');
-const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
```
arizzitano: Does Webpack 4 handle minification/uglification out of the box?

arizzitano:
Just to sanity check that this works, I'd recommend installing a copy of Paragon built from this branch into edx-platform and/or SFE. @thallada suggested adding a `prepare` script to run `npm run build` -- if we do that, we can just install directly from this branch without having to check in any dist files (i.e. `"paragon": "edx/paragon/greenkeeper/webpack-4.1.0"`)

thallada:
I'm having some trouble running `npm start` on this branch. I think we should hold off on this change until [storybook supports Webpack 4](https://github.com/storybooks/storybook/pull/3148).

Once that happens. How do you all feel about using webpack-merge like we do in other projects so that we could have a separate `webpack.dev.config.js`, `webpack.dev.config.js`, and `webpack.prod.config.js`? I don't think we need a separate themeable and static build for development so it might speed things up a little.

thallada:
Re: https://github.com/edx/paragon/pull/187, we should go straight to webpack 4.2.0.

It looks like storybook's webpack 4 support might be released as an alpha soon.

thallada:
Storybook has released an alpha with Webpack 4 support to NPM https://www.npmjs.com/package/@storybook/react We can now install it by specifying `"alpha"` in the `package.json`.

PR #175, opened 2018-03-14
fix(webpack): transpile isemail to es5

thallada:
Paragon's themeable and static builds were including some es6 syntax (arrow functions) because of the isemail package. We have to transpile it to es5 ourselves.

Also, I'm using `include` instead of `exclude` to make it more explicit about what babel is touching.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -37,7 +37,11 @@ module.exports = targetProperties.map(config => ({
     rules: [
       {
         test: /\.(js|jsx)$/,
-        exclude: /(node_modules)/,
+        include: [
+          path.resolve(__dirname, 'src'),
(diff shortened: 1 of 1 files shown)


jaebradley:
It's in the comments, but just for full visibility here's the issue: https://github.com/hapijs/isemail/issues/158

jaebradley:
It's not just us: https://github.com/lemurheavy/coveralls-public/issues/1086

jaebradley:
This should be solved by [`email-prop-type@1.1.4`](https://github.com/jaebradley/email-prop-type/releases/tag/v1.1.4).

thallada:
Yeah, I'll close this in favor of #178

PR #177, opened 2018-03-15
refactor(es-check): use es-check to validate es5

jaebradley:
[`es-check`](https://github.com/dollarshaveclub/es-check) is a nifty little package that lets you validate JavaScript files against a certain ECMA version.

Paragon exports `ES5` files, so I added two additional Travis scripts
1. Build the `static` and `themable` builds
1. Execute `npm run is-es5` which checks the JavaScript files in the `static` and `themable` builds to see if they are `ES5`.


jaebradley:
Failing because built files are non-`ES5` (see `const` usage in `static/index.js`, for example)

image

This is due to `isemail` not being `ES5`.

PR #178, opened 2018-03-15
fix(build): builds are now es5 compliant

jaebradley:
As documented in #175, [`hapijs/isemail`](https://github.com/hapijs/isemail) is not `ES5` compliant.

I forked `isemail` to [`jaebradley/isemail`](https://github.com/jaebradley/isemail) and published an `ES5` compliant package at [`isemail-es5`](https://www.npmjs.com/package/isemail-es5).

I then updated [`email-prop-type` to `1.1.4`](https://github.com/jaebradley/email-prop-type/releases/tag/v1.1.4) which uses `isemail-es5` instead of `isemail`.

`static/index.js` and `themable/index.js` now seem to be `ES5` compliant.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5373,11 +5373,11 @@
       }
     },
     "email-prop-type": {
-      "version": "1.1.3",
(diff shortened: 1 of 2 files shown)


jaebradley:
This blocks #177.

PR #179, opened 2018-03-15
feat(table): Add tableFixed option to set col widths

thallada:
This is a feature that studio-frontend needs so that long filenames will not make the columns jump around jarringly when filtering or paging. I pulled the classnames from this [stackoverflow answer](https://stackoverflow.com/a/37928581/6620612).

FYI @edx/educator-dahlia

### DIFF
--- a/src/Table/README.md
+++ b/src/Table/README.md
@@ -12,6 +12,7 @@ Provides a very basic table component with col-scoped headings displayed in the 
 3. `columnSortable` (boolean; optional) specifies at the column-level whether the column is sortable. If `columnSortable` is `true`, a sort button will be rendered in the column table heading. It is only required if `tableSortable` is set to `true`.
 4. `onSort` (function; conditionally required) specifies what function is called when a sortable column is clicked. It is only required if the column's `columnSortable` is set to `true`.
 5. `hideHeader` (boolean; optional) specifies at the column-level whether the column label is visible. A column that is sortable cannot have its label be hidden.
+6. `width` (string; conditionally required) only if `hasFixedColumnWidths` is set to `true`, the `<td>` elements' `class` attributes will be set to this value. This allows restricting columns to specific widths. See [Bootstrap's grid documentation](https://getbootstrap.com/docs/4.0/layout/grid/) for `col` class names that can be used.
 
 The order of objects in `columns` specifies the order of the columns in the table.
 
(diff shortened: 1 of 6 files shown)


src/Table/README.md
```diff
@@ -44,3 +45,6 @@ Specifies the screen reader only text that accompanies the sort buttons for sort
 3. `defaultText`: (string) specifies the screen reader only text for sort buttons that are not engaged.
 
 It is only required if `tableSortable` is set to `true`.
+
+### `tableFixed` (boolean; optional)
```
jaebradley: I know this isn't the case for the other `props` in this README but I think it would be valuable to add the default prop value?

src/Table/index.jsx
```diff
@@ -93,10 +93,13 @@ class Table extends React.Component {
       <thead
         className={classNames(...this.props.headingClassName.map(className => styles[className]))}
       >
-        <tr>
+        <tr className={classNames({ 'd-flex': this.props.tableFixed })}>
           {this.props.columns.map(col => (
```
jaebradley: If you want to get even more terse you could consider doing

```javascript
{ this.props.columns.map(({ key, columnSortable, tableFixed, width }) => ... }
```

so it doesn't have to be `col.____` inside the mapping function.

src/Table/index.jsx
```diff
@@ -156,6 +161,7 @@ Table.propTypes = {
   })).isRequired,
   headingClassName: PropTypes.arrayOf(PropTypes.string),
   tableSortable: PropTypes.bool,
+  tableFixed: PropTypes.bool,
```
jaebradley: Should `width` be added to the `shape` prop under `columns`?

jaebradley: Also, I understand that `tableFixed` is following the `boolean` `prop` naming convention but `hasFixedColumnWidths` seems like a more accurate description of what effect this `prop` has on the `Table` component.

src/Table/Table.test.jsx
```diff
@@ -260,4 +278,25 @@ describe('<Table />', () => {
...
+    it('with col width classnames on headings', () => {
+      const tableHeadings = wrapper.find('th');
+
+      expect(tableHeadings).toHaveLength(fixedProps.columns.length);
+      expect(tableHeadings.at(0).hasClass('col-4')).toEqual(true);
+      expect(tableHeadings.at(1).hasClass('col-2')).toEqual(true);
+      expect(tableHeadings.at(2).hasClass('col-3')).toEqual(true);
+    });
+
+    it('with col width classnames on cells', () => {
```
jaebradley: Would it be useful to add a test with column widths that doesn't have `tableFixed` equal to `true` to show that we shouldn't see the fixed behavior even if column widths are defined but `tableFixed` is `false`?

src/Table/index.jsx
thallada: I'd do that, but I need a reference to `col` a few lines down inside the map function to pass it to `this.getTableHeading`.

jaebradley: Oh I didn't see that - good call!

I don't think it's useful here then.

Maybe more relevant for [`line 119`](https://github.com/edx/paragon/pull/179/files#diff-f387f1e6539c577f164d015dc0c5f453R119)?

Either way, it's a nit and doesn't really matter.

thallada: I can do that in another map below this though. :+1:

src/Table/Table.test.jsx
thallada: :+1: added those tests [here](https://github.com/edx/paragon/pull/179/files#diff-57ba469286d69d3fbdda0a86dda12540R307).

PR #184, opened 2018-03-20
build(eslint): remove superfluous eslint packages

jaebradley:
If you run `npm ls` on `master` you'll see warnings like

image

The peer dep warnings around `eslint-plugin-jsx-a11y@^6.0.2` are because

1. `eslint-airbnb-config@16.0.0` is imported directly into `package.json` as is `eslint-plugin-jsx-a11y@6.0.2` instead of relying directly on the `eslint-config-edx` configuration.
1. `eslint-config-edx` is at `4.0.1` and in `4.0.3` these peer dependencies were introduced in [`eslint-config-edx#34`](https://github.com/edx/eslint-config-edx/pull/34).

```json
"dependencies": {
  "eslint": "^4.5.0",
  "eslint-config-airbnb": "^16.1.0",
  "eslint-plugin-dollar-sign": "^1.0.1",
  "eslint-plugin-import": "^2.8.0",
  "eslint-plugin-jsx-a11y": "^6.0.2",
  "eslint-plugin-react": "^7.5.1"
}
```

The solution is to remove all `eslint` configurations that are already included in `eslint-config-edx`.

There was one additional accessibility-related `eslint` violation that occurred due to this change around [the `jsx-a11y/label-has-for` rule](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md). I've ignored this violation for now, but we should investigate this further.


src/asInput/index.jsx
```diff
@@ -129,6 +129,7 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) =>
 
     getLabel() {
       return (
+        // eslint-disable-next-line jsx-a11y/label-has-for
```
jaebradley: I'm seeing a [`jsx-a11y/label-has-for`](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md) violation here

image

@fysheets you were the last person to touch this - do you have any idea why this accessibility violation is occurring and what the solution might be?

fysheets: Looking at the documentation that you linked, for components example is using a nested `Label` component where the ID value is passed explicitly through. Our code is using `{this.state.id}` which I'm going to assume the linter isn't smart enough to match up to the ID being set/passed through for the input.

PR #190, opened 2018-03-22
fix(modal): allow reopening after close

thallada:
Modal should re-open if a parent updates its props to `{open: true}`.

### DIFF
--- a/src/Modal/index.jsx
+++ b/src/Modal/index.jsx
@@ -31,9 +31,9 @@ class Modal extends React.Component {
     }
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.open !== this.props.open) {
-      this.setState({ open: nextProps.open });
+  componentWillReceiveProps({ open }) {
+    if (open !== this.state.open) {
+      this.setState({ open });
     }
   }
 
--- a/src/Modal/Modal.test.jsx
+++ b/src/Modal/Modal.test.jsx
@@ -134,6 +134,13 @@ describe('<Modal />', () => {
       wrapper.find('button').at(0).simulate('click');
       expect(spy).toHaveBeenCalledTimes(1);
     });
+    it('reopens after closed', () => {
+      modalOpen(true, wrapper);
+      wrapper.find('button').at(0).simulate('click');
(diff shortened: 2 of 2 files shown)


src/Modal/index.jsx
```diff
@@ -32,7 +32,7 @@ class Modal extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (nextProps.open !== this.props.open) {
+    if (nextProps.open !== this.state.open) {
```
jaebradley: Apologies in advance for the rubber 🦆 -ing.

Will this cause a fundamentally different behavior in `Modal`?

Consider this newly-initialized, simplified `Modal`

```javascript
<Modal open />
```

The `open` `state` value is `true` and the `open` `prop` value is `true`.

Let's say I close this `Modal`.

For reference, here's the `close` method

```javascript
close() {
  this.setState({ open: false });
  this.props.onClose();
}
```

So now, my `Modal`'s `open` `state` value is `false` and the `open` `prop` value is `true`.

`componentWillReceiveProps` has not been called yet because, well, the `Modal` hasn't received any new `props`.

Let's say we update the `Modal` dynamically so that it's `prop` value is now

```javascript
<Modal open={false} />
```

So in the old implementation, `nextProps.open` (which is `false`) is not equal to `this.props.open` (which is `true`), so then `setState` would get called, and a new `render` would occur, but the `open` `state` value would remain `false`.

In the new implementation, `nextProps.open` is equal to `this.state.open` (which is `false`) so the `setState` will not get called.

In both cases, the `Modal` will remain closed.

**tl;dr: It doesn't look like this will change the behavior post-close, but it might eliminate a `render`**

**bonus tl;dr: If you got here, thanks for being patient**

jaebradley: Style nit - consider destructuring for added tersnessosity (not a word, but should be for irony)

```javascript
componentWillReceiveProps({ open }) {
  if (open !== this.state.open) {
    this.setState({ open });
  }
}
```

thallada: Thanks for thoroughly checking the post-close behavior!

PR #192, opened 2018-03-27
fix(license): switch from MIT to Apache 2.0

thallada:
I noticed that we don't have a LICENSE file and that the [MIT](https://tldrlegal.com/license/mit-license) license is specified in the `package.json`.

I think that [Apache 2.0](https://tldrlegal.com/license/apache-license-2.0-(apache-2.0)) is preferred at edX for libraries that have a broad interest which we want other developers to use, see: https://openedx.atlassian.net/wiki/spaces/OPEN/pages/13959332/Licensing

I made this a `fix` so that semantic version will update the patch version for this change in case. If someone is already using this library and for whatever reason cannot comply with the Apache license then they can pin to the version of Paragon before this change.

FYI: @nedbat

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,177 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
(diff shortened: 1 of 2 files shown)


jaebradley:
Nit: I think the SPDX identifier for Apache 2.0 is Apache-2.0

PR #193, opened 2018-03-28
docs(inputtext): include type property

arizzitano:


src/InputText/README.md
```diff
@@ -6,3 +6,6 @@ Provides an input component called InputText that gives users a reusable input f
 
 ### `inputProps` (view asInput component for details)
 `inputProps` specifies all of the properties that are necessary from the included `asInput` component.  Please see details for input requirements within that component. 
+
+### `type` (string; optional)
+`type` specifies the `type` attribute to apply to the input field. For this component, valid types include `text` (default) or `password`.
```
thallada: Nit: I could see this being misread as `text` and `password` are the only valid options. Maybe you could link the full list of options here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Form_%3Cinput%3E_types

PR #196, opened 2018-03-30
feat(asinput): add "isValid" and "validationMessage" props

mikix:
As discussed in IRC, we felt like this was a reasonable way to:
- Let a validator be async
- Let a container have dependencies between children validations (like invalidate child2 when child1 changes)

I decided to ignore these new properties if the older 'validator' property is set to a function.

I didn't modify the storybook, since we already had a validator and the form doesn't look any different than that. But am happy to add one if you folks like.

@arizzitano and @thallada, at least, were interested in this I believe.

### DIFF
--- a/src/asInput/index.jsx
+++ b/src/asInput/index.jsx
@@ -24,6 +24,8 @@ export const inputProps = {
   onChange: PropTypes.func,
   onBlur: PropTypes.func,
   validator: PropTypes.func,
+  isValid: PropTypes.bool,
+  validationMessage: PropTypes.string,
   className: PropTypes.arrayOf(PropTypes.string),
   themes: PropTypes.arrayOf(PropTypes.string),
   inline: PropTypes.bool,
@@ -38,6 +40,8 @@ export const defaultProps = {
   disabled: false,
   required: false,
   validator: undefined,
+  isValid: true,
+  validationMessage: '',
   className: [],
   themes: [],
   inline: false,
@@ -51,10 +55,13 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) => 
       this.handleBlur = this.handleBlur.bind(this);
 
       const id = this.props.id ? this.props.id : newId('asInput');
+      const isValid = this.props.validator ? true : this.props.isValid;
+      const validationMessage = this.props.validator ? '' : this.props.validationMessage;
       this.state = {
         id,
         value: this.props.value,
-        isValid: true,
+        isValid,
+        validationMessage,
         describedBy: [],
         errorId: `error-${id}`,
         descriptionId: `description-${id}`,
@@ -62,10 +69,23 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) => 
     }
 
     componentWillReceiveProps(nextProps) {
+      const updatedState = {};
       if (nextProps.value !== this.props.value) {
-        this.setState({
-          value: nextProps.value,
-        });
+        updatedState.value = nextProps.value;
+      }
+      if (nextProps.isValid !== this.props.isValid && !nextProps.validator) {
+        updatedState.isValid = nextProps.isValid;
+      }
+      if (nextProps.validationMessage !== this.props.validationMessage && !nextProps.validator) {
+        updatedState.validationMessage = nextProps.validationMessage;
+      }
+      // If validator goes away, revert to props
+      if (nextProps.validator !== this.props.validator && !nextProps.validator) {
+        updatedState.isValid = nextProps.isValid;
+        updatedState.validationMessage = nextProps.validationMessage;
+      }
+      if (Object.keys(updatedState).length > 0) {
+        this.setState(updatedState);
       }
     }
 
--- a/src/asInput/README.md
+++ b/src/asInput/README.md
@@ -34,5 +34,11 @@ Handles all necessary props that are related to Input typed components.
 ### `validator` (function; optional)
 `validator` specifies the function to use for validation logic if the input needs to be validated. Default is undefined.
 
+### `isValid` (boolean; optional)
+`isValid` specifies whether the current input has validated correctly. Consider updating this from an `onBlur` handler. Only used if `validator` is not specified. The default is true.
+
+### `validationMessage` (string; optional)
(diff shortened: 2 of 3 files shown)


src/asInput/index.jsx
```diff
@@ -67,6 +74,23 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) =>
           value: nextProps.value,
         });
       }
+      if (nextProps.isValid !== this.props.isValid && !nextProps.validator) {
+        this.setState({
```
jaebradley: If `nextProps.isValid`, `nextProps.validationMessage`, and `nextProps.validator` are all different and `!nextProps.validator` that could trigger up to 3 additional re-renders right?

src/asInput/index.jsx
```diff
@@ -67,6 +74,23 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) =>
...
+          isValid: nextProps.isValid,
+        });
+      }
+      if (nextProps.validationMessage !== this.props.validationMessage && !nextProps.validator) {
+        this.setState({
+          validationMessage: nextProps.validationMessage,
+        });
+      }
+      // If validator goes away, revert to props
```
jaebradley: Do we need to handle the case where a new validator is defined in `nextProps`?

jaebradley:
@mikix thanks for doing this!

I feel like we should either have validators or use props, but supporting both seems like it adds a decent amount of complexity?

I know this is (another) breaking change, but just a thought.

src/asInput/index.jsx
mikix: @jaebradley that's how React does it? Yeah I can consolidate by updating a dict then setting that as the state. I had assumed React would wait until the JS event loop got back.

src/asInput/index.jsx
mikix: @jaebradley naw. We don't do any revalidation today when a new validator is set. So validator functions currently expect to only update state when they are called onBlur.

mikix:
@jaebradley I'm fine to drop the old validator method. I'll have to read the paragon docs on breaking changes. But I can do that retool if you paragon folks prefer that.

src/asInput/index.jsx
jaebradley: I think so? It's actually pretty unclear from [the docs what the exact batching logic looks like for `setState`](https://reactjs.org/docs/react-component.html#setstate) (something something I guess I *could* read the source-code)

> Think of setState() as a request rather than an immediate command to update the component. For better perceived performance, React may delay it, and then update several components in a single pass. React does not guarantee that the state changes are applied immediately.

> setState() does not always immediately update the component. It may batch or defer the update until later.

> setState() will always lead to a re-render unless shouldComponentUpdate() returns false. If mutable objects are being used and conditional rendering logic cannot be implemented in shouldComponentUpdate(), calling setState() only when the new state differs from the previous state will avoid unnecessary re-renders.

jaebradley:
@mikix here are [the commit message instructions for breaking changes](https://github.com/edx/paragon#breaking-changes) as per our `semantic-release` automated deployment flow.

src/asInput/index.jsx
thallada: Given the additional complexity these checks incur, I'm not really sure it's worth it. Also, why shouldn't it be allowed to initialize an input as invalid but still have a validator? Or, to have a validator function, but then override the valid state in a parent later?

I think that as long as the props are well-documented we should allow updating the isValid props even if there is a validator defined. It might also help to have two stories for InputText: one example to show using just `validator` synchronously , and a second example where validation is done in the parent asynchronously and only `isValid` and `validationMessage` are passed as props.

mikix: I've edited the branch to bundle state changes in a staging dictionary.

mikix:
I fixed the state bundling issue. I'll just merge this in now since both Tyler and I are working on code that needs it. We can deprecate 'validator' separately.

src/asInput/index.jsx
```diff
@@ -51,22 +55,36 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) =>
...
 
     componentWillReceiveProps(nextProps) {
+      const updatedState = {};
       if (nextProps.value !== this.props.value) {
-        this.setState({
-          value: nextProps.value,
-        });
+        updatedState.value = nextProps.value;
       }
+      if (nextProps.isValid !== this.props.isValid && !nextProps.validator) {
```
jaebradley: All of the next three conditionals are dependent on `!nextProps.validator` - might this be simplified by adding a parent `if (!nextProps.validator)`?

mikix: In my head, I liked this better than nested conditionals, but I could be persuaded if you feel strongly.

src/asInput/index.jsx
```diff
@@ -51,22 +55,36 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) =>
...
+      }
+      if (nextProps.validationMessage !== this.props.validationMessage && !nextProps.validator) {
+        updatedState.validationMessage = nextProps.validationMessage;
+      }
+      // If validator goes away, revert to props
+      if (nextProps.validator !== this.props.validator && !nextProps.validator) {
+        updatedState.isValid = nextProps.isValid;
+        updatedState.validationMessage = nextProps.validationMessage;
+      }
+      this.setState(updatedState);
```
jaebradley: Nit: probably don't need to trigger a re-render if `updateState` is empty (I'm assuming `this.setState` is not smart like that, but it very well could be).

mikix: Alright, I'm looking at code for this. React had better not be that dumb.

src/asInput/index.jsx
jaebradley: nope, I don't feel strongly!

src/asInput/index.jsx
mikix: Ugh it is that dumb. OK, will fix.

mikix: Fixed.

mikix:
Thanks all for the help with my first Paragon commit! Wooo!

PR #197, opened 2018-03-30
fix(modal): update close button styles

dsjen:
UX requested these updates:
* Updated 'x' to use FA icon and removed background color
* Updated close button styles

screen shot 2018-03-30 at 12 44 58 pm

@edx/fedx-team @abbyluke


dsjen:
@abbyluke -- any feedback?

dsjen:
Got confirmation from @abbyluke offline that this is good!

PR #198, opened 2018-03-30
fix(asinput): allow value of type number or string

thallada:
Input elements of type "number" have a Number value. React accepts a string or a number.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#Value

FYI: @MichaelRoytman I think that this will resolve the console errors we are seeing in studio-frontend.

### DIFF
--- a/src/asInput/index.jsx
+++ b/src/asInput/index.jsx
@@ -14,7 +14,7 @@ export const inputProps = {
   label: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
   name: PropTypes.string.isRequired,
   id: PropTypes.string,
-  value: PropTypes.string,
+  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
   description: PropTypes.oneOfType([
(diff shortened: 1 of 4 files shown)


src/asInput/index.jsx
```diff
@@ -14,7 +14,7 @@ export const inputProps = {
   label: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
   name: PropTypes.string.isRequired,
   id: PropTypes.string,
-  value: PropTypes.string,
+  value: PropTypes.oneOfType(PropTypes.string, PropTypes.number),
```
jaebradley: Do you need to pass in an array like `PropTypes.oneOfType([PropTypes.string, PropTypes.number])` ?

thallada: Whoops, yes. I didn't check the JS console on the stories page for this.

I also fixed a console error about 12 text inputs having the same `id` on the "different textual inputs" story, so I defined a separate `id` for each of the components in that story.

PR #202, opened 2018-04-02
fix(asinput): add dangerIconDescription as prop

thallada:
Also, allow passing `validationMessage` as an element.

These are some additions to the [PR](https://github.com/edx/paragon/pull/196) @mikix merged last week. I only noticed these things after I tried to use `isValid` and `validationMessage` in studio-frontend.

### DIFF
--- a/.storybook/__snapshots__/Storyshots.test.js.snap
+++ b/.storybook/__snapshots__/Storyshots.test.js.snap
@@ -1135,6 +1135,42 @@ exports[`Storyshots CheckBox basic usage 1`] = `
                       <td
                         className="css-1ygfcef"
                       >
+                        dangerIconDescription
+                      </td>
+                      <td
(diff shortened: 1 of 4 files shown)


mikix:
Question: is specifying dangerIconDescription necessary for proper a11y? If so, can there be a reasonable default in the common "danger" theme case?

thallada:
@mikix I'm hesitant to give a default because users of Paragon might get English when they don't expect it. Paragon doesn't have an infrastructure to do translations and I don't think it currently has any display text. It expects users to pass in translated strings themselves.

efischer19:
@thallada see https://github.com/edx/paragon/issues/139 - I think you'd be fine to merge this as-is and make a note of the assumed English on that issue, to be dealt with at some point in the future.

PR #206, opened 2018-04-04
fix(asinput): allow validationMessage to be an element

efischer19:
More-or-less the same as https://github.com/edx/paragon/pull/133, as noted [here](https://github.com/edx/studio-frontend/pull/158#issuecomment-378643670)


efischer19:
I didn't realize #202 existed; I've replayed my test addition over there.

PR #207, opened 2018-04-04
fix(status-alert): fix add ability to set close button aria-label

AlasdairSwan:
This would allow the component to support i18n and enable the addition of context specific labels to the close button (e.g. 'Close the edX Cookie Policy alert.').

### DIFF
--- a/src/StatusAlert/StatusAlert.test.jsx
+++ b/src/StatusAlert/StatusAlert.test.jsx
@@ -34,6 +34,14 @@ describe('<StatusAlert />', () => {
       expect(statusAlertDialog.text()).toEqual(dialog);
       expect(wrapper.find('button')).toHaveLength(0);
     });
+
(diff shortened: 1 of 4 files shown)


src/StatusAlert/StatusAlert.test.jsx
```diff
@@ -34,6 +34,14 @@ describe('<StatusAlert />', () => {
       expect(statusAlertDialog.text()).toEqual(dialog);
       expect(wrapper.find('button')).toHaveLength(0);
     });
+
+    it('renders custom aria-label view', () => {
+      const customLabel = 'Dismiss this alert post-haste!';
```
jaebradley: post-haste

fysheets:
Looks like you just need to run a new snapshot

jaebradley:
👏  first Paragon commit!

first

PR #211, opened 2018-04-12
fix(input-text): add autocomplete prop

AlasdairSwan:


src/InputText/index.jsx
```diff
@@ -37,6 +38,7 @@ const textDefaultProps = {
   type: 'text',
   describedBy: '',
   isValid: true,
+  autoComplete: true,
```
jaebradley: Need to add the `PropType.bool` for this?

jaebradley: Mind adding this configuration parameter in [the README](https://github.com/edx/paragon/blob/master/src/InputText/README.md) as well? Since this is only defined in `InputText` and not in `asInput`, it should probably be explicitly noted?

src/asInput/README.md
```diff
@@ -4,6 +4,11 @@ Handles all necessary props that are related to Input typed components.
 
 ## API
 
+### `autoComplete` (string; optional)
+`autoComplete` specifies whether or not an input field should have autocomplete enabled. The default value is 'on'.
+
+Autocomplete allows the browser to predict the value. When a user starts to type in a field, the browser should display options to fill in the field, based on earlier typed valuesThe default type is true.
```
jaebradley: Nit: minor typo and formatting.

PR #214, opened 2018-04-13
fix(modal): enable closing modal by clicking outside of it

AlasdairSwan:


fysheets:
Quick test of this branch in storybook: look at the story for invoking the modal with a button

Locally the button gets pushed down the width of the modal - which I'm assuming will push all content on the page down that way. If you look at the modal in the[ production storybook](https://edx.github.io/paragon/?selectedKind=Modal&selectedStory=modal%20invoked%20via%20a%20button&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybook%2Factions%2Factions-panel) the background content stays the same.

We tried to play around with the DOM to mimic the setup from the [bootstrap docs](https://getbootstrap.com/docs/4.0/components/modal/#live-demo) (looking at the DOM from the live demo) but couldn't get it quite right.

I'll wait to check this out with studio frontend locally until this is sorted.

src/Modal/index.jsx
```diff
@@ -168,7 +167,22 @@ class Modal extends React.Component {
...
+          onClick={this.close}
+        />
+        <div
+          className={classNames(
+            styles.modal,
+            styles['modal-dialog'],
+            {
+              [styles['modal-open']]: open,
+              [styles.show]: open,
+              [styles.fade]: !open,
```
fysheets: It looks good in Storybook, but now we have the `modal-open`, `show`, and `fade` styles at two levels of the Modal tree?

AlasdairSwan: I know, it's not my favourite. The main benefit of this approach is that it simplifies the close click handler logic considerably. I will try taking another stab at it without updating the DOM, and just using move logic in the click handler.

fysheets: Ah alright, I was mostly asking to understand if that had been left in there on purpose or not. I'll keep an eye out for that next commit before trying to validate further then. 👍

AlasdairSwan:
Closed in favour of https://github.com/edx/paragon/pull/243

PR #215, opened 2018-04-17
feat(fieldset): add Fieldset and ValidationMessage

thallada:
Also, refactors the invalid-feedback div out of asInput into its own component.


src/Fieldset/README.md
```diff
@@ -0,0 +1,29 @@
+# Fieldset
+
+Wraps many inputs in a border and allows validating them as one form control.
+
+## API
+
+### `dangerIconDescription` (string or element; optional)
```
jaebradley: "danger" seems implementation (aka `Bootstrap`)-specific. This feels like it's some kind've "invalidInputDescription" or something #NamingIsHard

jaebradley: Why would somebody use an element here?

src/Fieldset/README.md
```diff
@@ -0,0 +1,29 @@
...
+`dangerIconDescription` can be used to provide a screen-reader description of the "exclamation" icon used in the validation message displayed when `isValid` is false and `themes` includes "danger". The default is an empty string.
+
+### `children` (string, element, or arrays of strings or elements; optional)
+`children` specifies the elements that the Fieldset will wrap. The default is an empty string.
```
jaebradley: Is there a reason to explicitly point out the `children` prop vs. the implicit use of the `children` prop when using `Fieldset` as a wrapper.

I could see how somebody might get confused with this `prop` and do something like

```javascript
<FieldSet children={<noblah />)>
  <blah />
</FieldSet>
```

src/Fieldset/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+### `className` (string array; optional)
+`className` specifies Bootstrap class names to apply to the input component. The default is an empty array.
+
+### `id` (string; optional)
+`id` specifies a unique identifier for the fieldset.  It will be used to create an id for the ValidationMessage that is referenced by an `aria-describedby` on the `fieldset`. The default is an auto-generated id of the form `fieldsetN` where N is an auto-incremented integer.
```
jaebradley: `fieldsetN` is typo?

jaebradley: lol no it's not, I am

src/Fieldset/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+### `isValid` (boolean; optional)
+`isValid` specifies whether the whole fieldset has validated correctly. Consider updating this from an `onSubmit` handler of a parent `form` element. The default is true.
+
+### `legend` (string or element; required)
+`legend` specifies the label to be used for the overarching fieldset. This can be a string or element type. It appears at the top of the fieldset, inline with the border.
+
+### `themes` (array of strings; optional)
+`themes` specifies the themes to apply to the input (e.g. "danger"). The default is an empty array.
```
jaebradley: Maybe clarify these are one of [the eight `Bootstrap` contextual variations](https://getbootstrap.com/docs/4.0/components/badge/#contextual-variations)?

src/Fieldset/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+### `validationMessage` (string or element; optional)
```
jaebradley: If it's only displayed when `isValid` is `false`, maybe this is more of an `invalidMessage`?

src/Fieldset/README.md
```diff
@@ -0,0 +1,29 @@
...
+### `dangerIconDescription` (string or element; optional)
+`dangerIconDescription` can be used to provide a screen-reader description of the "exclamation" icon used in the validation message displayed when `isValid` is false and `themes` includes "danger". The default is an empty string.
+
+### `children` (string, element, or arrays of strings or elements; optional)
+`children` specifies the elements that the Fieldset will wrap. The default is an empty string.
+
+### `className` (string array; optional)
+`className` specifies Bootstrap class names to apply to the input component. The default is an empty array.
+
+### `id` (string; optional)
```
jaebradley: If somebody wants to use an `id` of `1` will this result in a `PropType` error? Or will `1` get stringified in-time?

src/Fieldset/index.jsx
```diff
@@ -0,0 +1,79 @@
...
+  isValid: true,
+  themes: [],
+  validationMessage: '',
+};
+
+class Fieldset extends React.Component {
+  constructor(props) {
+    super(props);
+
+    const id = this.props.id ? this.props.id : newId('fieldset');
```
jaebradley: [`airbnb/javascript#15.7`](https://github.com/airbnb/javascript#comparison--unneeded-ternary)

Consider

```javascript
const id = this.props.id || newId('fieldset');
```

src/Fieldset/index.jsx
```diff
@@ -0,0 +1,79 @@
...
+      errorId: `error-${id}`,
+    };
+  }
+
+  hasDangerTheme() {
+    return this.props.themes.indexOf('danger') >= 0;
+  }
+
+  render() {
+    return (
```
jaebradley: consider destructuring `this.props` and `this.state`

src/ValidationMessage/index.jsx
```diff
@@ -0,0 +1,85 @@
+
```
jaebradley: Extra line

src/ValidationMessage/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+  errorId: PropTypes.string,
+  isValid: PropTypes.bool,
+  themes: PropTypes.arrayOf(PropTypes.string),
+  validationMessage: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+};
+
+const defaultProps = {
+  dangerIconDescription: '',
+  className: [],
+  errorId: undefined,
```
jaebradley: `''` vs. `undefined`?

src/ValidationMessage/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+  isValid: true,
+  themes: [],
+  validationMessage: '',
+};
+
+class ValidationMessage extends React.Component {
+  constructor(props) {
+    super(props);
+
+    const id = this.props.errorId ? this.props.errorId : `error-${newId('validationMessage')}`;
```
jaebradley: There's an implicit dependency between the `error-{id}` format here and the format that's defined in the initial state declaration in `Fieldset` (aka, if we change something in one place we probably need to change it in the other place) - this seems kind've rigid?

jaebradley: Nit: if you call this variable `errorId` then you can just do 

```javascript
this.state = { errorId };
```

on the next line

src/ValidationMessage/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+              aria-hidden
+            />
+            <span
+              className={classNames(styles['sr-only'])}
+            >
+              {this.props.dangerIconDescription}
+            </span>
+          </span>
+          ),
+          <span key="1">
```
jaebradley: Since `React` uses `key`s to determine if something has changed if the `this.props.validationMessage` changes will this not be reflected in subsequent renders?

src/Fieldset/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## API
+
+### `dangerIconDescription` (string or element; optional)
+`dangerIconDescription` can be used to provide a screen-reader description of the "exclamation" icon used in the validation message displayed when `isValid` is false and `themes` includes "danger". The default is an empty string.
+
+### `children` (string, element, or arrays of strings or elements; optional)
+`children` specifies the elements that the Fieldset will wrap. The default is an empty string.
+
+### `className` (string array; optional)
```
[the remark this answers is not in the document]
thallada: I agree that this is confusing. However, I think we should keep this as `className` but make it `string` type instead. That way it is consistent with the React API and also familiar to React programmers. If users want to specify multiple classes, then they can use the `classNames` function in the value for `className` just like any normal base React component.

But, if I make that change here, I'd also want to make sure we make the change for every `className` prop in Paragon (which would be a breaking change).

src/Fieldset/README.md
thallada: studio-frontend uses an element here because we wrap every translated string in a span that indicates the language of the string with the `lang` attribute.

It's called `dangerIconDescription` because it describes the exclamation icon that only appears when the danger theme is applied.

I'm really just copying what `asInput` is doing. I personally don't like the whole theme thing because I don't see us ever needing an input/fieldset styled with anything other than red or no-red, and it's not clear what happens if you specify multiple themes in the array.

I could completely redo the API for Fieldset, but then it would be inconsistent with asInput. We cannot also redo the API for asInput without making a breaking change release. I opted to make the least changes and inconsistencies and just copied the API for asInput. If we have a clear plan for what the API should be, then I can adopt that here and then we can update asInput to it in a separate PR.

src/Fieldset/README.md
thallada: Isn't that docs page specific to badges though? I don't see any docs pages in Bootstrap about the different conceptual themes they have (danger, warning, info).

src/Fieldset/README.md
thallada: Yes, that's probably a little clearer, but this would be another change that makes the Fieldset API inconsistent with asInput.

src/Fieldset/README.md
thallada: I think that would result in a proptype error, but I can verify that.

src/ValidationMessage/index.jsx
thallada: I'm going to switch this code over to wrap the list of elements in `<React.Fragment>`s instead to avoid this `key`s issue.

src/Fieldset/README.md
fysheets: Want to note that it could be styled with yellow warning or any of the other bootstrap coloring variant/themes as well even though we're not doing that for SFE right now.

src/Fieldset/README.md
fysheets: https://v4-alpha.getbootstrap.com/utilities/colors/

fysheets:
One more overarching comment -- with this refactor to pull validation out, can we also refactor inputs to use the `variant` naming instead of `theme` (that whole discussion we had with this and implemented with the modal) so that we can think of it as different types of inputs/fieldsets/components/etc instead of it being themed like we theme other edX related items? Or would that end up being a breaking change as well?

src/asInput/index.jsx
```diff
@@ -37,7 +37,7 @@ export const inputProps = {
 export const defaultProps = {
   onChange: () => {},
   onBlur: () => {},
-  id: newId('asInput'),
+  id: undefined,
```
fysheets: Mostly curious about why this change was made.

src/asInput/index.jsx
```diff
@@ -109,43 +109,14 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) =>
       const descriptionId = `description-${this.state.id}`;
       const desc = {};
 
-      const hasDangerTheme = this.hasDangerTheme();
-
       desc.error = (
-        <div
-          className={classNames(
-            styles['form-control-feedback'],
```
fysheets: `form-control-feedback` is a bootstrap class that adds a bit more formatting for us, curious on why it was removed? https://v4-alpha.getbootstrap.com/components/forms/ (search for the class name)

fysheets: Looking more into those docs, we weren't using it correctly is my guess on the answer -- so new question, should we be?

src/asInput/asInput.scss
```diff
@@ -3,8 +3,8 @@
 @import "~bootstrap/scss/utilities/_screenreaders.scss";
 @import '~bootstrap/scss/input-group';
 
-.fa-icon-spacing {
-    padding: 0px 5px 0px 0px;
```
fysheets: Curious about this CSS change as well, as it doesn't look to me to be a simple swap.

fysheets: Nevermind, I started from the bottom up and made it to the new scss file that this was moved to :)

src/ValidationMessage/index.jsx
jaebradley: That will make this component incompatible with anything before `React` 16 right?

Is this something we need to be concerned about? (Probably not)

src/Fieldset/README.md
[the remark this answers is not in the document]
thallada: I talked with Jae in-person. I think we are okay with making changes like this one just in Fieldset for now. Eventually, we can make the breaking changes in the rest of Paragon to resolve the inconsistencies.

src/Fieldset/README.md
thallada: Thanks Farhannah, I found the v4.0 docs for that and linked it in the docs here: https://getbootstrap.com/docs/4.0/utilities/colors/

src/ValidationMessage/index.jsx
thallada: I refactored this so that `ValidationMessage` has a required `id` prop and no state. `Fieldset` and `asInput` now create and pass the `error-{id}` themselves. I think that makes the most sense since `Fieldset` and `asInput` will have to use that same id in their `aria-describedby`.

src/ValidationMessage/index.jsx
thallada: I'm not aware of anything that's using Paragon and not on React 16 so I don't think it is something we should be concerned about.

src/asInput/index.jsx
thallada: The problem with this is that `defaultProps` is only defined once per import of this module. So every `asInput` created without an id specified would get this same id. The code in the constructor, which is called on every new `asInput` component, should call `newId` if this is `undefined` now.

src/asInput/index.jsx
thallada: I think `form-control-feedback` was removed in Bootstrap v4.0. It also looks like Bootstrap only supports green or red validation themes now (`is-valid` and `is-invalid` classes): https://getbootstrap.com/docs/4.0/components/forms/#server-side

thallada:
@jaebradley and @fysheets, I think I have addressed all of your comments. `Fieldset` and `ValidationMessage` now use `Variant` and have less confusing prop names.

src/Fieldset/README.md
thallada: screenshot-20180420143153-969x109

Yup, PropType error.

src/asInput/index.jsx
jaebradley: I was thinking about this recently - I think we should switch over to something like https://github.com/dylang/shortid ?

src/Fieldset/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+const inputProps = {
+  variantIconDescription: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+  children: PropTypes.node,
+  className: PropTypes.arrayOf(PropTypes.string),
+  id: PropTypes.string,
+  isValid: PropTypes.bool,
+  legend: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
+  invalidMessage: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+  variant: PropTypes.shape({
```
jaebradley: Are there plans to add more `variant` options in the future? A `shape` with one field feels a little over-structured...

src/Fieldset/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+import newId from '../utils/newId';
+import styles from './Fieldset.scss';
+import ValidationMessage from '../ValidationMessage/index';
+import Variant from '../utils/constants';
+
+const inputProps = {
+  variantIconDescription: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+  children: PropTypes.node,
+  className: PropTypes.arrayOf(PropTypes.string),
```
jaebradley: `classNames`?

src/Fieldset/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+  id: PropTypes.string,
+  isValid: PropTypes.bool,
+  legend: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
```
jaebradley: Nit: do you mind separating the required and optional `props`? I know I'm OCD, but it makes it a little easier to see at a glance.

src/Fieldset/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+    status: Variant.status.INFO,
+  },
+};
+
+class Fieldset extends React.Component {
+  constructor(props) {
+    super(props);
+
+    const id = props.id || newId('fieldset');
+    this.state = {
```
jaebradley: style nit - consider doing

```javascript
this.state = { id };
```

and save two lines.

or

```javascript
this.state = { id: props.id || newId('fieldset') };
```

and save three.

src/Fieldset/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+      id,
+    };
+  }
+
+  componentWillReceiveProps(nextProps) {
+    // TODO: use getDerivedStateFromProps when upgraded to React 16.3+
+    if (nextProps.id !== this.props.id) {
+      if (nextProps.id) {
```
jaebradley: Can this be simplified to

```javascript
this.setState({ id: nextProps.id || newId('fieldset') });
```

src/Fieldset/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+      } else {
+        this.setState({
+          id: newId('fieldset'),
+        });
+      }
+    }
+  }
+
+  getVariantFieldsetClassName() {
+    const { variant } = this.props;
```
jaebradley: Can this be simplified to

```javascript
if (this.props.variant.status === Variant.status.INFO) {
  return styles['is-invalid-nodanger'];
}
```

src/Fieldset/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+      <div className={styles['paragon-fieldset']}>
+        <fieldset
+          className={classNames(
+            styles['form-control'],
+            styles['p-3'],
+            { 'is-invalid': !isValid },
+            this.getVariantFieldsetClassName(),
+            className,
+          )}
+          aria-describedby={`error-${this.state.id}`}
```
jaebradley: move `error-${this.state.id}` to variable?

src/Fieldset/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+        });
+      } else {
+        this.setState({
+          id: newId('fieldset'),
+        });
+      }
+    }
+  }
+
+  getVariantFieldsetClassName() {
```
jaebradley: `getVariantClassName` vs. `getVariantFieldsetClassName` - isn't the `Fieldset` implied?

src/Fieldset/README.md
```diff
@@ -0,0 +1,41 @@
...
+* Fieldset border is always black, even when `isValid` is `false`.
+* Displays `invalidMessage` in black.
+* Does not display an icon next to `invalidMessage`.
+
+**DANGER**:
+* Fieldset border is red when `isValid` is `false` and black when `true`.
+* Displays `invalidMessage` in red.
+* Displays an "exclamation" icon next to `invalidMessage`.
+
+### `variantIconDescription` (string or element; optional)
```
jaebradley: If `variant` is going to be an object, wouldn't it make sense for it's icon description to live within that object?

src/Fieldset/index.jsx
thallada: I didn't create Variant, so I'm not 100% sure what the purpose of it is. It looks like it was designed so that there would be different Variant "types". The only one right now is `status`: `INFO`, `WARNING`, `DANGER`, `SUCCESS`. But, I don't know what other Variant types are possible and don't want to collapse this to a string if we'd want to add more variants later.

src/Fieldset/index.jsx
thallada: That is simpler, but the current structure makes it really easy to add another variant.

src/Fieldset/README.md
thallada: I think `variant` is intended just to specify one theme out of an enum. If we start putting display strings in it then I'm afraid we'll start dumping any theme-related data into `variant`.

src/Fieldset/index.jsx
jaebradley: Fair - I guess shouldn't that be something the implementer that wants to add more variants should solve? Especially since it's unclear if more variants will be needed (something something pre-mature optimization)?

Either way, not a big deal.

src/Fieldset/Fieldset.stories.jsx
```diff
@@ -0,0 +1,144 @@
...
+
+    this.state = {
+      isValid: true,
+    };
+
+    this.handleSubmit = this.handleSubmit.bind(this);
+  }
+
+  handleSubmit(event) {
+    if (this.firstInputRef.value.length === 0 && this.secondInputRef.value.length === 0) {
```
jaebradley: can this just be simplified to

```javascript
this.setState({ isValid: this.firstInputRef.value.length !== 0 || this.secondInputRef.value.length !== 0 });
```

jaebradley: Whoops, forgot this was in a story. Carry on.

src/Fieldset/index.jsx
jaebradley: What do you mean by

> don't want to collapse this to a string if we'd want to add more variants later

Couldn't the props just be `buttonVariant`, `inputVariant`, `anotherInputVariant`, etc. (assuming we want multiple variants).

src/Fieldset/index.jsx
```diff
@@ -0,0 +1,105 @@
...
+const inputProps = {
+  legend: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
+
+  children: PropTypes.node,
+  className: PropTypes.string,
+  id: PropTypes.string,
+  isValid: PropTypes.bool,
+  invalidMessage: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+  variant: PropTypes.shape({
+    status: PropTypes.string,
```
jaebradley: Should this be `PropTypes.oneOf(Variant.status)`? We probably want to have `prop` warnings if somebody puts in `foobar` or something?

src/Fieldset/README.md
```diff
@@ -0,0 +1,41 @@
+# Fieldset
+
+Wraps many inputs in a border and allows validating them as one form control.
+
+## API
+
+### `children` (node; optional)
+`children` specifies the elements that the Fieldset will wrap. The default is null.
+
+### `className` (string array; optional)
```
jaebradley: Should just be `string` not `string array`

src/Fieldset/README.md
jaebradley: Why wouldn't an `id` value of `1` be valid?

src/Fieldset/README.md
```diff
@@ -0,0 +1,41 @@
...
+`invalidMessage` specifies the message to display immediately below the fieldset outside the border when `isValid` is false. The default is an empty string.
+
+### `isValid` (boolean; optional)
+`isValid` specifies whether the whole fieldset has validated correctly. Consider updating this from an `onSubmit` handler of a parent `form` element. The default is true.
+
+### `legend` (string or element; required)
+`legend` specifies the label to be used for the overarching fieldset. This can be a string or element type. It appears at the top of the fieldset, inline with the border.
+
+### `variant` (object with a string `status` field; optional)
+`variants` specifies which type of fieldset to display. These are the Variants currently supported:
```
jaebradley: `variant` vs. `variants`

src/Fieldset/README.md
thallada: I think supporting both would unnecessarily complicate things because we later do string operations on the id because it is assumed to be a string. Also, I feel like an `id` attribute that is just a number is not very descriptive.

PR #217, opened 2018-04-18
fix(modal): make header close button optional

AlasdairSwan:
More modal tweaks.

### DIFF
--- a/.storybook/__snapshots__/Storyshots.test.js.snap
+++ b/.storybook/__snapshots__/Storyshots.test.js.snap
@@ -15688,7 +15688,7 @@ exports[`Storyshots Modal modal with element body 1`] = `
 
 exports[`Storyshots Modal modal with warning variant 1`] = `
 <div
-  aria-labelledby="id10"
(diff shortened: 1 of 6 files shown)


fysheets:
Two comments regarding focus management:
1. When you open this Modal in storybook the focus doesn't go anywhere anymore (I think I'd expect it on the close button at the bottom)
2. When you try to tab through the Modal now the following console error appears because it's trying to focus on the hidden button:
screen shot 2018-04-18 at 9 50 26 am

AlasdairSwan:
Damn, that head close button is very useful for focus management.

fysheets:
@AlasdairSwan that works for me -- I did a quick check with your previous commit for the focus errors and saw that they were gone & it worked as expected. Thanks for the clarification on focus going to the Modal itself :)

AlasdairSwan:
@fysheets I updated the Sass to add indication of focus to the modal. If you could take another look that would be great.

AlasdairSwan:
@jaebradley I addressed your comments too.

PR #218, opened 2018-04-18
fix(as-input): renders classNames to DOM

AlasdairSwan:
InputText was rendering array positions from the ```className``` array instead of the strings passed in. This PR resolves that issue.

### DIFF
--- a/src/InputText/InputText.test.jsx
+++ b/src/InputText/InputText.test.jsx
@@ -52,5 +52,15 @@ describe('<InputText />', () => {
       const input = wrapper.find('input').at(0);
       expect(input.prop('autoComplete')).toEqual('off');
     });
+
+    it('should render with custom classNames if set', () => {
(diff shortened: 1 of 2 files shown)


.storybook/__snapshots__/Storyshots.test.js.snap
```diff
@@ -20277,7 +20277,7 @@ exports[`Storyshots Textarea validation 1`] = `
   <textarea
     aria-describedby="error-asInput1 description-asInput1"
     aria-invalid={false}
-    className="form-control"
+    className="form-control "
```
jaebradley: This is why the extra space is here

image

- [RunKit](https://runkit.com/jaebradley/5ad8991ff4c25d0012f5b2c3)

will this cause any problems?

AlasdairSwan: That's annoying but it won't cause any issues.

PR #220, opened 2018-04-19
fix(modal): make footer close button optional

AlasdairSwan:
The sibling of PR#217 - which enables me to implement the use case of having the close button on the left instead of the right.


fysheets:
As discussed offline - Code here technically looks good, but I do have concerns about having both the header and footer close buttons being optional as I could see someone unwittingly creating a modal that will only close by clicking on the backdrop (other PR) just because it will be possible, which would be an inaccessible solution. Also the aim for consistency between our buttons across Modals at edX was a consideration on how these were initially set up, so might be worthwhile to have a UX/UI discussion of what that consistency is to ensure we're all on the same page.

Issue #227, opened 2018-04-26
Rename validationMessage props to invalidMessage


thallada:
[See discussion about this](https://github.com/edx/paragon/pull/215#discussion_r182223123).

yro:
All set.

PR #240, opened 2018-05-07
refactor(validationmessage): rename validationMessage props

yro:
This is meant to tackle issue #227 . I also changed the class/file names, though that might have been a glib reading of the original issue. Happy to change back if extensibility to other types of validation messages is desired in the future.

BREAKING CHANGE: The validationMessage props have been changed to
invalidMessage, including file and class name changes.


arizzitano:
@yro thanks for doing this and welcome to Paragon! I just pushed up a new branch `4.0.0-rc` -- would you mind changing the base branch of your PR to point at `4.0.0-rc` rather than `master` (so we can push out all our breaking changes in a single future release)?

Changes look good to me -- @thallada, would you say this covers the scope of #227?

yro:
@arizzitano @thallada Thanks y'all! I'll squash and merge.

src/asInput/index.jsx
```diff
@@ -5,7 +5,7 @@ import classNames from 'classnames';
 
 import newId from '../utils/newId';
 import styles from './asInput.scss';
-import ValidationMessage from '../ValidationMessage/index';
+import InvalidMessage from '../InvalidMessage/index';
```
jaebradley: Nit: I know you're just replacing the `ValidationMessage` with `InvalidMessage` but I don't think the `/index` is necessary?

src/asInput/index.jsx
```diff
@@ -27,7 +27,7 @@ export const inputProps = {
   onBlur: PropTypes.func,
   validator: PropTypes.func,
   isValid: PropTypes.bool,
-  validationMessage: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+  invalidMessage: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
```
jaebradley: Alright, I'll also be that person - I'm not sure `invalidMessage` is the most appropriate name here. 

When I see `invalidMessage` I think of a message that's not valid or a `boolean` that represents if a message is valid or not, not necessarily a message that will be displayed if a value is invalid.

Maybe `invalidValueMessage` is slightly more accurate?

Could also be the case that

alt-text

yro: @jaebradley What do you think? On the one hand it's kind of the crux of the thing, but on the other it's not a big deal to change. Happy to go either way.

src/asInput/index.jsx
yro: Makes sense 👍

src/asInput/index.jsx
jaebradley: Err I meant to make this comment about [line 67](https://github.com/edx/paragon/pull/240/files#diff-8f7eccce481920c8f772c0263869dde4R67), the internal variable called `invalidMessage` (and not the prop).

It might be worth discussing with @arizzitano, @thallada, and @edx/fedx-team ?

yro: @jaebradley You raise a valid point, but I think this is ok. Shipping as is.

PR #242, opened 2018-05-07
fix(as-input): support multiple input group addon elements

adamstankiewicz:
Per [Bootstrap 4's documention](https://getbootstrap.com/docs/4.0/components/input-group/#multiple-addons) for multiple input group addons, this PR adds the ability to pass in an array of elements to the `inputGroupPrepend` and `inputGroupAppend` props in the `asInput` component.

### DIFF
--- a/src/asInput/index.jsx
+++ b/src/asInput/index.jsx
@@ -31,8 +31,8 @@ export const inputProps = {
   className: PropTypes.arrayOf(PropTypes.string),
   themes: PropTypes.arrayOf(PropTypes.string),
   inline: PropTypes.bool,
-  inputGroupPrepend: PropTypes.element,
-  inputGroupAppend: PropTypes.element,
+  inputGroupPrepend: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.element), PropTypes.element]),
+  inputGroupAppend: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.element), PropTypes.element]),
 };
 
 export const defaultProps = {
@@ -136,6 +136,17 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) => 
       return desc;
     }
 
+    getAddons({ addonElements, type }) {
+      if (Array.isArray(addonElements)) {
+        return addonElements.map((addon, index) =>
+          React.cloneElement(
+            addon,
+            { key: this.generateInputGroupAddonKey({ prefix: type, index }) },
+          ));
+      }
+      return addonElements;
+    }
+
     getLabel() {
       return (
         // eslint-disable-next-line jsx-a11y/label-has-for
@@ -181,6 +192,10 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) => 
       );
     }
 
+    generateInputGroupAddonKey({ prefix, index }) {
+      return `${this.state.id}-${prefix}-${index}`;
+    }
+
     renderInput(describedBy) {
       const { className } = this.props;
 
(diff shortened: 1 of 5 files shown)


src/asInput/index.jsx
```diff
@@ -181,6 +181,11 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) =>
       );
     }
 
+    genInputGroupAddonKey(prefix, index) {
```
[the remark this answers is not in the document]
jaebradley: I also tend to like having my functions take objects like

```javascript
generateInputGroupAddonKey({ prefix, index })
```

so that when you call this method elsewhere like

```javascript
generateInputGroupAddonKey({ prefix: 'jae', index: 'baebae' })
```

it's clearer what the arguments mean

You can also do stuff like

```javascript
generateInputGroupAddonKey({ index: 'baebae', prefix: 'jae' })
```

and not have to worry about argument order.

Here's [a good article](https://medium.freecodecamp.org/elegant-patterns-in-modern-javascript-roro-be01e7669cbd) on this pattern.

src/asInput/index.jsx
```diff
@@ -204,6 +209,27 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) =>
       );
     }
 
+    renderInputGroupAddons(type) {
+      // Type is either inputGroupPrepend or inputGroupAppend
+      const formattedType = type.replace('inputGroup', '').toLowerCase();
+      const addons = this.props[type];
+      if (addons && Array.isArray(addons)) {
```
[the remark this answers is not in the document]
adamstankiewicz: D'oh, good catch :)

src/asInput/index.jsx
```diff
@@ -204,6 +209,30 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) =>
       );
     }
 
+    renderInputGroupAddons(type) {
+      // Type is either inputGroupPrepend or inputGroupAppend
+      const formattedType = type.replace('inputGroup', '').toLowerCase();
```
jaebradley: This logic feels brittle - if we change the prop name to `inputGroupsAppend` for example, we have to remember to change this method.

src/asInput/index.jsx
```diff
@@ -181,6 +181,11 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) =>
       );
     }
 
+    generateInputGroupAddonKey({ prefix, index }) {
+      // Prefix is either prepend or append
```
jaebradley: Why is this comment relevant? 

What I mean is, why should an implementer of this method care what the prefix has been used for? 

It seems like if this comment should live anywhere it should live where this method is actually getting implemented, where `prefix` is probably `prepend` or `append` in that implementation context.

src/asInput/index.jsx
adamstankiewicz: Yeah, I'm not a huge fan on this logic either. I was trying to have one method that can handle both `inputGroupPrepend` and `inputGroupAppend`. A better alternative might be to split it up into 2 methods for each type of input group addon.

src/asInput/index.jsx
```diff
@@ -204,6 +209,30 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) =>
...
+      const addons = this.props[type];
+      if (Array.isArray(addons)) {
+        const addonsChildren = addons.map((addon, index) => (
+          React.cloneElement(
+            addon,
+            { key: this.generateInputGroupAddonKey({ prefix: formattedType, index }) },
+          )
+        ));
+        return (
+          <div className={styles[`input-group-${formattedType}`]}>
```
jaebradley: If you create an `getAddons` method that deals with the `array` / not array cases then you can get rid of some duplicative code by doing

```javascript
return (
  <div className={style[`input-group-${formattedType}`]}
    { this.getAddons() }
  </div>
)
```

PR #244, opened 2018-05-10
fix(modal): handle overflowing content

thallada:
This PR fixes a few things:

1. Removes now-unused `.modal-open` styles (the class is now `.show`).
2. When modal body content overflows the screen height, it will now contain it to a scrollable container and still show the modal buttons at the bottom of the screen.
screenshot-20180510141310-697x616
3. In IE11, which has bugs with the flexbox layout which breaks 2, always show a scrollbar when a modal is open which will allow scrolling of the entire modal when it overflows the screen. The scrollbar will be visible even when the modal does not overflow because `overflow: auto` was not working. This should only affect IE11.
screenshot-20180510141402-649x687
4. Use `position: fixed` on the `.modal` div so that it does not push contents that are in the DOM after it down the page when it opens (the modal will now display over top of the background elements).
screenshot-20180510144446-797x403
screenshot-20180510144545-779x333
5. The warning modal variant icon was always showing up on a new line. Specify the icon column as `col-md-2` so that it only collapses to a new row when the screen is smaller than `md`.
screenshot-20180510144824-619x374
screenshot-20180510144736-621x316

### DIFF
--- a/src/Modal/Modal.scss
+++ b/src/Modal/Modal.scss
@@ -3,13 +3,27 @@
 @import "~bootstrap/scss/_grid";
 @import "~bootstrap/scss/_utilities";
 
-.modal-open {
-  display: block;
+.modal.show {
+  position: fixed;
+  background-color: transparent;
+  max-height: 100%;
+  width: 100%;
 
   &:focus {
     .modal-dialog {
       box-shadow: $btn-focus-box-shadow;
     }
+  }
+}
+
+.modal.is-ie11 {
+  // fix browser that likes to do things its own way
+  overflow-y: scroll;
+  height: auto;
+
+  .modal-content {
+    height: auto;
+    max-height: none;
   }
 }
 
@@ -19,4 +33,30 @@
   // Fade for backdrop
   &.fade { opacity: 0; }
   &.show { opacity: 1; }
+}
+
+.modal-dialog {
+  height: 100%;
+  margin: auto;
+  padding: $spacer / 2;
+
+  @media (min-width: map-get($grid-breakpoints, 'sm')) {
+    padding: $spacer;
+  }
+}
+
+.modal-content {
+  max-height: 100%;
+}
+
+.modal-header {
+  flex: 0 0 auto;
+}
+
+.modal-body {
+  overflow: auto;
+}
+
+.modal-footer {
+  flex: 0 0 auto;
 }
--- a/src/Modal/Modal.test.jsx
+++ b/src/Modal/Modal.test.jsx
@@ -127,6 +127,36 @@ describe('<Modal />', () => {
       expect(modalFooter.find('button')).toHaveLength(1);
       expect(wrapper.find('button')).toHaveLength(1);
     });
+
+    it('renders with IE11-specific styling when IE11 is detected', () => {
+      const { MSInputMethodContext } = global;
+      const { documentMode } = global.document;
+
+      // mimic IE11
+      global.MSInputMethodContext = true;
+      global.document.documentMode = true;
+      wrapper = mount(<Modal {...defaultProps} />);
+      const modal = wrapper.find('.modal');
+      expect(modal.hasClass('is-ie11')).toEqual(true);
+
+      global.MSInputMethodContext = MSInputMethodContext;
+      global.document.documentMode = documentMode;
+    });
+
+    it('renders without IE11-specific styling when IE11 is not detected', () => {
+      const { MSInputMethodContext } = global;
+      const { documentMode } = global.document;
+
+      // mimic non-IE11 browser
+      global.MSInputMethodContext = false;
+      global.document.documentMode = false;
+      wrapper = mount(<Modal {...defaultProps} />);
+      const modal = wrapper.find('.modal');
+      expect(modal.hasClass('is-ie11')).toEqual(false);
+
+      global.MSInputMethodContext = MSInputMethodContext;
+      global.document.documentMode = documentMode;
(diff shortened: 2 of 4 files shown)


src/Modal/Modal.scss
```diff
@@ -13,10 +15,45 @@
   }
 }
 
+.modal.is-ie {
+  // fix garbage browser
```
MichaelRoytman: I mean... I agree, but maybe we should remove this comment or make it a little less negative. 'fix browser that likes to things its own way' has a nice ring, no?

thallada: Heh, ok fair. I wrote that when I was a bit frustrated :stuck_out_tongue:

MichaelRoytman: It's up to you! Perhaps I am being too cautious. My rationale was that if we're supporting a particular browser and people on our platform use that browser and Paragon is a component library for that platform we might want to remain unbiased in our comments :P. But again, maybe the caution is unnecessary.

thallada: Yeah you have a point. I like your suggestion. It's snarky in a nice way :laughing:

src/Modal/index.jsx
```diff
@@ -183,6 +185,7 @@ class Modal extends React.Component {
               [styles['d-block']]: open,
               [styles.show]: open,
               [styles.fade]: !open,
+              [styles['is-ie']]: isIE11,
```
efischer19: It makes me sad that this needs to exist. I get it, but it still makes me sad

src/Modal/Modal.scss
```diff
@@ -13,10 +15,45 @@
   }
 }
 
+.modal.is-ie {
+  // fix browser that likes to do things its own way
```
efischer19: 5af0a515bd967126008b461a-750-563

src/Modal/index.jsx
```diff
@@ -10,6 +10,8 @@ import Icon from '../Icon';
 import newId from '../utils/newId';
 import Variant from '../utils/constants';
 
+const isIE11 = !!window.MSInputMethodContext && !!document.documentMode;
```
jaebradley: In case anybody from the future wants to know why this works: https://stackoverflow.com/a/22082397/5225575

src/Modal/index.jsx
jaebradley: should this really be `['is-ie-11']`? `isIE11` will be `false` for `Edge`, for example.

jaebradley: Should there be a test for this?

src/Modal/index.jsx
jaebradley: Might not be a bad idea adding the SO link in a comment

thallada: I'll add a comment here with that link actually.

src/Modal/index.jsx
thallada: How would I test this? I would have to mimic an IE11 browser and I don't think we have the ability to do that in our tests.

thallada: Actually, I found a way to mock globals in Jest. I'll push up a test in sec.

src/Modal/index.jsx
```diff
@@ -23,6 +23,8 @@ class Modal extends React.Component {
     this.el = document.createElement('div');
 
     this.state = {
+      // Sets true for IE11, false otherwise: https://stackoverflow.com/a/22082397/6620612
+      isIE11: !!window.MSInputMethodContext && !!document.documentMode,
```
jaebradley: Why should this be part of the `state`? This doesn't really change through the lifetime of the component right?

src/Modal/Modal.test.jsx
```diff
@@ -127,6 +127,25 @@ describe('<Modal />', () => {
       expect(modalFooter.find('button')).toHaveLength(1);
       expect(wrapper.find('button')).toHaveLength(1);
     });
+
+    it('renders with IE11-specific styling when IE11 is detected', () => {
```
jaebradley: 👍 on implementing this test

src/Modal/Modal.test.jsx
```diff
@@ -127,6 +127,25 @@ describe('<Modal />', () => {
...
+
+      // mimic IE11
+      global.MSInputMethodContext = true;
+      global.document.documentMode = true;
+      wrapper = mount(<Modal {...defaultProps} />);
+      let modal = wrapper.find('.modal');
+      expect(modal.hasClass('is-ie11')).toEqual(true);
+
+      // return globals back to not IE11
+      global.MSInputMethodContext = MSInputMethodContext;
```
jaebradley: Nit: this feels likes two tests jammed into one

1. has `is-ie11` class when browser is IE11
2. does not have `is-ie11` class when browser is not IE11

jaebradley: What happens if `window.MSInputMethodContext` or `document.documentMode` is `false` in the `global` context? The following assertion would fail, right? Basically, this test is assuming that it is not executed with an IE11 browser? 

Would it be safer to assign either of these to `false` (and really the appropriate tests are when `MSInputMethodContext` is `false` and when `documentMode` is `false`) and assert that `is-ie11` doesn't exist? And then reset globals as part of an `afterEach`?

src/Modal/index.jsx
thallada: I'll make it an instance variable instead. I needed it to be set on component construction instead of import because of the test.

Issue #224, opened 2018-04-26
Make sure all display string props accept element proptype


thallada:
Studio-frontend needs to be able to pass in an element wherever a component accepts a display string because it wraps all translated string in a span that specifies the language of the translation e.g. `<span lang="en">`.

jaebradley:
I would argue that this is what `props.children` should be used for

```javascript
<MyComponent>
  blahbblahblah
</MyComponent>
```

and

```javascript
<MyComponent>
 <span>blahblahblah</span>
</MyComponent>
```

thallada:
If we do that, then we will loose the ability to enforce certain accessible patterns. Like [requiring a `legend` on a `Fieldset`](https://github.com/edx/paragon/tree/master/src/Fieldset#legend-string-or-element-required), a [`label` on an `asInput`](https://github.com/edx/paragon/tree/master/src/asInput#label-string-or-element-required), and a [description for the danger theme validation icon on `asInput` and `Fieldset`](https://github.com/edx/paragon/tree/master/src/asInput#dangericondescription-string-or-element-optional). Getting rid of those props would mean that it would be up to the consumer of Paragon to make sure to include them in the `children`. Is that the direction that we want to go in?

thallada:
As Jae pointed out, we should rename the `closeText` prop on the `Modal` to something that implies that it accepts a string or an element. `closeLabel` maybe?

jaebradley:
Again, I feel like this API might be more intuitive

```javascript
<Modal>
  <Title>
    Some title
  </Title>
  <Body>
    Some body
  </Body>
  <Label>
    Some label
  </Label>
</Modal>
```

I think we should think about using `children` instead of jamming things into `props`.

fysheets:
How do we enforce certain components/aria attributes that are needed for accessibility become children then? I feel like that could lead to regressions in accessibility due to folks *needing* to know all required children instead of having them by default and providing the text (element for i18n) and more or less getting the accessibility out of the box. That sort of change feels to me like we're doing a component library, but not necessarily an accessible one which I thought was one of the main appeals of Paragon and a goal to help prevention of regressions as we move forward.

jaebradley:
Well when you iterate through the children you can check their types. 

If, for a modal, you iterate through all the children and a `Label` has not been found then you can throw / `console.log` an error, or create one by default.

fysheets:
[...]

PR #245, opened 2018-05-11
fix(modal): allow element in closeText prop

thallada:
This addresses #224 for one component: `Modal`. I decided to do this now because it started to cause some extra pain for me in SFE.

### DIFF
--- a/src/Modal/Modal.test.jsx
+++ b/src/Modal/Modal.test.jsx
@@ -59,6 +59,7 @@ const defaultProps = {
   open: true,
   onClose: () => {},
 };
+const closeText = 'GO AWAY!';
 
 let wrapper;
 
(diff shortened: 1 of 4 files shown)


src/Modal/README.md
```diff
@@ -16,7 +16,7 @@ Provides a basic modal component with customizable title, body, and footer butto
 ### `buttons` (element or shape in form of buttonPropTypes array; optional)
 `buttons` is an array of either elements or shapes that take the form of the buttonPropTypes. See the [buttonPropTypes](https://github.com/edx/paragon/blob/master/src/Button/index.jsx#L40) for a list of acceptable props to pass as part of a button.
 
-### `closeText` (string; optional)
+### `closeText` (string or element; optional)
```
jaebradley: I just want to point out that I don't think a prop named `closeText` should take an `element` as the prop name would suggest only a `string`.

I don't think it's worth a breaking change at this point, but I think it's useful to add to the next major upgrade branch / issues.

thallada: Yeah, I didn't make that change because it would be a breaking change and we'd rather have the fix in SFE sooner rather than later. I'll mention it in the issue though.

src/Modal/Modal.test.jsx
```diff
@@ -128,6 +129,26 @@ describe('<Modal />', () => {
       expect(wrapper.find('button')).toHaveLength(1);
     });
 
+    it('renders custom close button string', () => {
+      wrapper = mount(<Modal {...defaultProps} closeText={closeText} />);
+      const modalFooter = wrapper.find('.modal-footer');
+      const closeButton = modalFooter.find('button');
+
+      expect(closeButton).toHaveLength(1);
+      expect(closeButton.text()).toEqual(closeText);
```
jaebradley: These two assertions are actually `true` in both cases right? 

I think the check here is that whatever the wrapping element around `closeText` is should only contain a string and **not** an element.

jaebradley: My point wasn't that the check isn't useful - it is. My point was more that the assertions were not fully capturing the test case that was being described (which are now 👍 ) in that the assertions would've been valid for a `string` prop or an `element` prop.

I should've made it clear that I meant that the test was **missing** a test case.

Regardless, thanks for adding the extra check! 👏

src/Modal/Modal.test.jsx
```diff
@@ -128,6 +129,27 @@ describe('<Modal />', () => {
...
+    });
+
+    it('renders custom close button element', () => {
+      const closeElem = <span>{closeText}</span>;
+      wrapper = mount(<Modal {...defaultProps} closeText={closeElem} />);
+      const modalFooter = wrapper.find('.modal-footer');
+      const closeButton = modalFooter.find('button');
+
+      expect(closeButton).toHaveLength(1);
+      expect(closeButton.find('span')).toHaveLength(1);
```
jaebradley: Should there also be a check that the `children` size should also be `1`? This guarantees that the `span` would be the only `child`.

src/Modal/README.md
jaebradley: Should there be a story added as well that showcases using an `element`?

MichaelRoytman:
I don't know if this is the pain point you mentioned, Tyler, but this will help us internationalize the closeText from within studio-frontend, so I just want to mention that here so that someone remembers to do that once this merges.

PR #246, opened 2018-05-13
feat(search-field): New SearchField component

adamstankiewicz:
Adds a `SearchField` component, which makes use of the `InputText`, `Button`, and `Icon` components.

**Example**
SearchField UI

**Customizable Options**
The `SearchField` component is customizable in the following ways:
* _Callbacks_: Callback handlers for `onSubmit`, `onFocus`, `onBlur`, and `onChange` events;
* _Input label_: A string that is passed to the `InputText` component.
* _Placeholder_: Placeholder text for the `InputText` component.
* _Screenreader Text_: An object containing screenReaderText strings to use for the clear and submit search buttons.
* _Value_: An initial value to populate the input field with.

**Accessibility**
Since `SearchField` uses the `InputText` and `Button` components, most of the accessibility concerns are handled already. However, because `SearchField` can be customized with several props, it's worth mentioning how the component remains accessible:

1. The `screenReaderText` for the clear and submit search buttons can be customized  (i.e., for i18n translations).

2. The `SearchField` component displays a clear button when the input has a non-empty value. This button can be reached using the keyboard and uses the `Icon` component and its associated `screenReaderText` prop.

### DIFF
--- a/src/SearchField/README.md
+++ b/src/SearchField/README.md
@@ -1,0 +1,45 @@
+# SearchField
+
+Provides a basic search field component.
+
+## API
+
+### `onSubmit` (function; required)
+`onSubmit` specifies a callback function for when the `SearchField` is submitted by the user. For example:
+
+```jsx
+<SearchField onSubmit={(value) => { console.log(value); } />
+```
+
+### `inputLabel` (string, optional)
+`inputLabel` specifies the label to use for the input field (e.g., for i18n translations). The default is "Search:".
+
+### `onChange` (function; optional)
+`onChange` specifies a callback function for when the value in `SearchField` is changed by the user. The default is an empty function. For example:
+
+```jsx
+<SearchField onChange={(value) => { console.log(value); } />
+```
+
+### `onFocus` (function; optional)
+`onFocus` specifies a callback function for when the user focuses in the `SearchField` component. The default is an empty function. For example:
+
+```jsx
+<SearchField onFocus={(value) => { console.log(value); } />
+```
+
+### `onBlur` (function; optional)
+`onBlur` specifies a callback function for when the user loses focus in the `SearchField` component. The default is an empty function. For example:
+
+```jsx
+<SearchField onBlur={(value) => { console.log(value); }} />
+```
+
+### `placeholder` (string; optional)
+`placeholder` specifies the placeholder text for the input. The default is an empty string.
+
+### `screenReaderText` (object, optional)
+`screenReaderText` specifies the screenreader text for both the clear and search buttons (e.g., for i18n translations). The default is `{ clearButton: 'Clear search', searchButton: 'Submit search' }`.
+
+### `value` (string; optional)
+`value` specifies the initial value for the input. The default is an empty string.
--- a/src/SearchField/SearchField.scss
+++ b/src/SearchField/SearchField.scss
@@ -1,0 +1,77 @@
+@import "~bootstrap/scss/_variables";
+@import "~bootstrap/scss/utilities/_borders";
+@import "~bootstrap/scss/utilities/_spacing";
+@import "~bootstrap/scss/_buttons";
+
+$search-field-clear-btn-width: 30px;
+$search-field-border-radius: 30px;
+
+.search-field {
+    transition: $input-transition;
+    border-radius: $search-field-border-radius;
+    box-shadow: $input-box-shadow;
+
+    .form-group {
+        margin-bottom: 0;
+    }
+
+    .input-group {
+        flex: 1 1 0;
+    }
+
+    label {
+        margin: 0;
+        padding-left: 20px;
+        padding-right: 10px;
+    }
+    
+    button:last-child {
+        border-radius: $search-field-border-radius;
+    }
+}
+
+.search-field__focused {
+    border-color: $blue !important;
+    box-shadow: $input-focus-box-shadow;
+}
+
+.search-field__input {
+    background: none;
+    border: 0;
+    box-shadow: none !important;
+    padding-left: 0;
+    padding-right: 0;
+
+    &:focus {
+        background: none;
+    }
+}
+    
+.search-field__no-clear-btn {
+    padding-right: ($search-field-clear-btn-width + 1px + $border-width * 2);
+}
+
+.search-field__search-btn {
+    border-top: 0;
+    border-right: 0;
+    border-bottom: 0;
+    transition: border-color 0.15s ease-in-out;
+    padding: 10px 15px 10px 12px;
+    color: $gray-600;
+    background: none;
+    margin-left: 0;
+    
+    &:not(:disabled):hover,
+    &:not(:disabled):focus {
+        background: #F2F8FB;
+        color: $blue;
+    }
+}
+
+.search-field__clear-btn {
+    width: $search-field-clear-btn-width;
+    justify-content: center;
+    color: $gray-600;
+    border: 0;
+    background: none;
+}
--- a/src/SearchField/index.jsx
+++ b/src/SearchField/index.jsx
@@ -1,0 +1,196 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import FontAwesomeStyles from 'font-awesome/css/font-awesome.min.css';
+
+import newId from '../utils/newId';
+import Icon from '../Icon';
+import InputText from '../InputText';
+import Button from '../Button';
+
+import styles from './SearchField.scss';
+
+const defaultProps = {
+  inputLabel: 'Search:',
+  onBlur: () => {},
+  onChange: () => {},
+  onFocus: () => {},
+  placeholder: '',
+  screenReaderText: {
+    clearButton: 'Clear search',
+    searchButton: 'Submit search',
+  },
+  value: '',
+};
+
+const propTypes = {
+  onSubmit: PropTypes.func.isRequired,
+  inputLabel: PropTypes.string,
+  onBlur: PropTypes.func,
+  onChange: PropTypes.func,
+  onFocus: PropTypes.func,
+  placeholder: PropTypes.string,
+  screenReaderText: PropTypes.shape({
+    clearButton: PropTypes.string.isRequired,
+    searchButton: PropTypes.string.isRequired,
+  }),
+  value: PropTypes.string,
+};
+
+class SearchField extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isFocused: false,
+      value: this.props.value,
+    };
+
+    this.textInput = null;
+    this.searchButton = null;
+
+    this.handleFocus = this.handleFocus.bind(this);
+    this.handleBlur = this.handleBlur.bind(this);
+    this.handleChange = this.handleChange.bind(this);
+    this.handleClear = this.handleClear.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.handleKeyPress = this.handleKeyPress.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.value !== this.props.value) {
+      this.setState({ value: nextProps.value });
+    }
+  }
+
+  getSearchActionButtons() {
+    const { isFocused } = this.state;
+    const { screenReaderText } = this.props;
+    const inputTextHasValue = this.inputTextHasValue();
+    const buttons = [
+      <Button
+        className={[classNames(
+          styles['search-field__search-btn'],
+          {
+            [styles['border-left']]: !isFocused && inputTextHasValue,
+            [styles['btn-outline-primary']]: isFocused && inputTextHasValue,
+          },
+        )]}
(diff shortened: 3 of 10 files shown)


src/SearchField/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+```jsx
+<SearchField
+  screenReaderText="search"
+  onChange={(value) => { console.log(value); }}
+/>
+```
+
+### `onFocus` (function; optional)
+`onFocus` specifies a callback function for when the value in `SearchField` is changed by the user. The default is an empty function. For example:
```
jaebradley: Is the callback executed when the value is changed or when the `SearchField` receives focus?

src/SearchField/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+```jsx
+<SearchField
+  screenReaderText="search"
+  onFocus={(value) => { console.log(value); }}
+/>
+```
+
+### `onBlur` (function; optional)
+`onBlur` specifies a callback function for when the value in `SearchField` is changed by the user. The default is an empty function. For example:
```
jaebradley: Is the callback executed when the value is changed or when the `SearchField` loses focus?

src/SearchField/README.md
```diff
@@ -0,0 +1,88 @@
...
+### `placeholder` (string; optional)
+`placeholder` specifies the placeholder text for the input. The default is an empty string.
+
+### `value` (string; optional)
+`value` specifies the initial value for the input. The default is an empty string.
+
+### `hasClearButton` (boolean; optional)
+`hasClearButton` specifies whether the search field has an additional button to allow the user to clear the `input` value. The default is `true`.
+
+### `hasRoundedCorners` (boolean; optional)
```
jaebradley: Where's the boundary between the CSS that the component comes with and the CSS that the user of the API wants?

Basically, what if I want `reallyRoundedCorners` or `somewhatRoundedCorners`?

src/SearchField/README.md
```diff
@@ -0,0 +1,88 @@
...
+### `inputLabel` (string; optional)
+`inputLabel` specifies the label text to use for the `input` element. The default is an empty string, which results in a search icon to be used as the `label` element.
+
+### `searchButtonLabel` (string or element; optional)
+`searchButtonLabel` specifies the text or element for the search button. For example:
+
+```jsx
+<SearchField
+    screenReaderText="search"
+    onSubmit={action('search-submitted')}
```
jaebradley: Nit: Can you make these callback names more specific, like `handleOnSubmit`? If an end user is unfamiliar with `action`s they might not know what `action` means in this scenario.

src/SearchField/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+  value: PropTypes.string,
+  hideInputLabel: PropTypes.bool,
+  inputLabel: PropTypes.string,
+  searchButtonLabel: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+  hasClearButton: PropTypes.bool,
+  hasRoundedCorners: PropTypes.bool,
+  onChange: PropTypes.func,
+  onFocus: PropTypes.func,
+  onBlur: PropTypes.func,
+  onSubmit: PropTypes.func.isRequired,
```
jaebradley: Makes a ton of sense to group things alphabetically 👏 
But how do you feel about grouping the required props first (and alphabetically) and the optional props second (and also alphabetically)?

src/SearchField/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+  handleKeyPress(event) {
+    if (event.key === 'Enter' && this.inputTextHasValue()) {
+      this.handleSubmit();
+    }
+  }
+
+  handleSubmit() {
+    const { value } = this.state;
+    this.searchButton.focus();
+    this.props.onSubmit(value.trim());
```
jaebradley: Why `trim`? An input value that contains whitespaces is still a valid input, right? Shouldn't it be left to the caller of this API how they deal with whitespaces?

src/SearchField/README.md
adamstankiewicz: Yeah, this is something I was struggling with how to handle. I was thinking of making this prop accept a string so that the user can pass in a value like `sm`, `md`, or `lg` to change the amount of border radius. If I go that route, I think it would also make sense for the prop to accept a boolean like it does now, but if no string is passed in, it defaults to Bootstrap's rounded corners.

src/SearchField/index.jsx
adamstankiewicz: 👍

src/SearchField/README.md
adamstankiewicz: I'm not sure I follow your question here. The `action` is just for the storybook actions that get logged. I don't think changing the callback names like `handleOnSubmit` would add much clarity to an end user compared to `onSubmit`.

src/SearchField/README.md
adamstankiewicz: Good catch. Copy/paste oversight 🙃

src/SearchField/index.jsx
adamstankiewicz: What value would leaving leading/trailing whitespaces here have for the end user? I can't think of a case where the caller of this API would actually want whitespaces, in which case, doesn't it make sense for `handleSubmit` to remove that step to clean the input?

src/SearchField/README.md
jaebradley: Right - because `action` is specific to storybook actions, somebody reading the documentation might be like "what is this `action` method"? if they're unfamiliar with Storybook.

Whereas I think that the general name for "some generic callback method" (or at least, from the documentation I've read) is either `callback` or `someCallback` or `handleOn{someEventName}`.

src/SearchField/index.jsx
jaebradley: There are two reasons why

1. It can lead to inconsistencies with the user experience. Imagine this is a username input field and imagine that I can actually create usernames with whitespaces. I type in "username ". The username that's actually created is "username" because this component stripped out the whitespace and the API to create usernames used the value from the `onSubmit` prop. Here's another one - what if I want to display a max character length warning for some input value?
2. It's arbitrarily opinionated. What if I can't think of a business case (right now) for having the letter `a` in my input? Should I then strip out the letter `a` from my input? Doing formatting like this unnecessarily boxes in the API user - let the API user decide how they'd like to format the input value.

src/SearchField/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+    this.handleChange = this.handleChange.bind(this);
+    this.handleClear = this.handleClear.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.handleKeyPress = this.handleKeyPress.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const updatedState = {};
+    if (nextProps.value !== this.props.value) {
+      updatedState.value = nextProps.value;
```
jaebradley: Can't this be simplified to

```javascript
if (nextProps.value !== this.props.value) {
  this.setState({ value: nextProps.value });
}
```

src/asInput/README.md
```diff
@@ -10,6 +10,15 @@ Handles all necessary props that are related to Input typed components.
 ### `className` (string array; optional)
 `className` specifies Bootstrap class names to apply to the input component. The default is an empty array.
 
+### `formGroupClassName` (string array; optional)
+`formGroupClassName` specifies class names to apply to the `.form-group` element. The default is an empty array.
```
jaebradley: `formGroupClassNames` since it's an array?

src/SearchField/SearchField.scss
```diff
@@ -0,0 +1,73 @@
...
+}
+
+.search-field-clear-btn {
+    width: $search-field-clear-btn-width;
+    justify-content: center;
+}
+
+.search-field-input-has-clear-btn {
+    padding-right: ($search-field-clear-btn-width + 1px + $border-width * 2) !important;
+}
```
jaebradley: image

src/SearchField/README.md
```diff
@@ -0,0 +1,91 @@
...
+`placeholder` specifies the placeholder text for the input. The default is an empty string.
+
+### `value` (string; optional)
+`value` specifies the initial value for the input. The default is an empty string.
+
+### `hasClearButton` (boolean; optional)
+`hasClearButton` specifies whether the search field has an additional button to allow the user to clear the `input` value. The default is `true`.
+
+### `roundedCorners` (string; optional)
+`roundedCorners` specifies the roundedness of the `SearchField` component. The options are `sm`, `md`, or `lg`.  The default is undefined, which results in no rounded corners.
```
jaebradley: Should the default be `undefined` or `null`?

jaebradley: `roundedCorners` doesn't really imply anything about the "roundedness" of the component. If this was something like `roundedCornerSize` that might make more sense.

src/SearchField/index.jsx
```diff
@@ -0,0 +1,266 @@
...
+    return buttons;
+  }
+
+  shouldShowClearButton() {
+    return this.props.hasClearButton && this.inputTextHasValue();
+  }
+
+  inputTextHasValue() {
+    const { value } = this.state;
+    return value && value.trim().length > 0;
```
jaebradley: trim

src/SearchField/index.jsx
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  handleChange(value) {
+    this.setState({ value });
+    this.props.onChange(value);
+  }
+
+  handleClear() {
+    this.setState({ value: '' });
+    this.props.onChange('');
```
jaebradley: Should there be a separate `onClear` prop? I could kind've see some use cases where it might be useful.

src/SearchField/index.jsx
```diff
@@ -0,0 +1,266 @@
...
+            {
+              [styles['pl-0']]: !hideInputLabel,
+              [styles['pl-3']]: hideInputLabel,
+              [styles['search-field-input-has-clear-btn']]: hasClearButton && !shouldShowClearButton,
+              [styles['pr-0']]: !hasClearButton || (hasClearButton && shouldShowClearButton),
+            },
+          )]}
+          name="search"
+          label={inputLabel || this.getInputLabel()}
+          ariaLabel={hideInputLabel ? screenReaderText : ''}
```
jaebradley: `aria-label`?

src/SearchField/index.jsx
```diff
@@ -0,0 +1,266 @@
...
+        onBlur={this.handleBlur}
+      >
+        <InputText
+          className={[classNames(
+            styles['search-field-input'],
+            {
+              [styles['pl-0']]: !hideInputLabel,
+              [styles['pl-3']]: hideInputLabel,
+              [styles['search-field-input-has-clear-btn']]: hasClearButton && !shouldShowClearButton,
+              [styles['pr-0']]: !hasClearButton || (hasClearButton && shouldShowClearButton),
```
jaebradley: Can this be simplified to

```javascript
!hasClearButton || shouldShowClearButton
```

src/SearchField/index.jsx
adamstankiewicz: It turns out using `aria-label` as the prop name does not work as intended.

src/SearchField/index.jsx
```diff
@@ -0,0 +1,266 @@
...
+  onSubmit: PropTypes.func.isRequired,
+  screenReaderText: PropTypes.string.isRequired,
+  hasClearButton: PropTypes.bool,
+  hideInputLabel: PropTypes.bool,
+  inputLabel: PropTypes.string,
+  onBlur: PropTypes.func,
+  onChange: PropTypes.func,
+  onFocus: PropTypes.func,
+  placeholder: PropTypes.string,
+  roundedCorners: PropTypes.oneOf(['sm', 'md', 'lg']),
```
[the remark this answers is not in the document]
adamstankiewicz: I changed the `roundedCorners` prop to have a default value of an empty string to match the `PropType` values (i.e., strings). Even though the default value is not technically one of the `PropType` options, it does not cause an error here.

src/SearchField/index.jsx
adamstankiewicz: Hmm, I could see how it'd be useful. However, the end user of this component could watch for clearing (i.e., value becomes an empty string) by using the existing `onChange` callback prop that's called in `handleClear`, so I'm not sure if it's worth adding here.

src/asInput/index.jsx
```diff
@@ -148,16 +151,17 @@ const asInput = (WrappedComponent, inputType = undefined, labelFirst = true) =>
...
           id={`label-${this.state.id}`}
           htmlFor={this.state.id}
-          className={[classNames({
+          className={classNames({
             [styles['form-check-label']]: this.isGroupedInput(),
-          })]}
+          })}
         >
-          {this.props.label}
+          {label}
```
georgebabey: nit: label is only used once, could pull it from `this.props` here instead of pulling it out above

PR #247, opened 2018-05-14
fix(modal): make portal rendering optional

thallada:
Yet another modal PR...

Enzyme is having some issues with the `querySelector` call in `Modal`. This is because Enzyme is not a browser so it doesn't actually put anything into a DOM that is selectable with `querySelector`. Generally, it's more idiomatic to pass refs to React components, not selector strings.

Studio-frontend does not need to render the modal in a portal so I decided that it would be best to make the behavior optional. If `parentSelector` is not given, then it will revert to the old way of just rendering the `Modal` where it is defined in the DOM. As far as I know, there's no way to test the portal behavior using `querySelector` with Enzyme (it's currently not tested in Paragon).

This is a breaking change, but users should be able to pass `parentSelector="body"` to get the old default behavior. ~~I still marked this as a `fix` because SFE needs this fix soon and can't wait for `4.0.0`. Are we okay with this? I can't think of an alternative...~~

I also removed some redundant `<div>`s in favor of `<React.Fragment>`s.

### DIFF
--- a/src/Modal/index.jsx
+++ b/src/Modal/index.jsx
@@ -20,7 +20,6 @@ class Modal extends React.Component {
     this.setCloseButton = this.setCloseButton.bind(this);
 
     this.headerId = newId();
-    this.el = document.createElement('div');
 
     // Sets true for IE11, false otherwise: https://stackoverflow.com/a/22082397/6620612
     this.isIE11 = !!window.MSInputMethodContext && !!document.documentMode;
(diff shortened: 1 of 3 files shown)


src/Modal/index.jsx
```diff
@@ -53,10 +53,6 @@ class Modal extends React.Component {
     }
   }
 
-  componentWillUnmount() {
-    ReactDOM.unmountComponentAtNode(this.parentElement);
-  }
```
thallada: I don't understand why this was here before. This attempts to unmount an element that this component does not own which throws an error for me.

The modal should unmount itself from the parent element automatically.

src/Modal/index.jsx
```diff
@@ -34,11 +33,12 @@ class Modal extends React.Component {
...
-    this.parentElement = document.querySelector(this.props.parentSelector);
-    if (this.parentElement === null) {
-      throw new Error(`Modal recieved invalid parentSelector: ${this.props.parentSelector}, no matching element found`);
+    if (this.props.parentSelector) {
+      this.parentElement = document.querySelector(this.props.parentSelector);
+      if (this.parentElement === null) {
+        throw new Error(`Modal recieved invalid parentSelector: ${this.props.parentSelector}, no matching element found`);
+      }
     }
-    this.parentElement.appendChild(this.el);
```
thallada: Another somewhat major change: the modal will render directly into `this.parentElement` instead of creating `this.el`, rendering the Modal into `this.el`, and then appending it to the `this.parentElement`. I didn't see any reason to have the extra `this.el` div so I removed it. Was there any reason for it that I am not aware of?

PR #253, opened 2018-05-18
feat(listbox): add ListBox and ListBoxOption components

MichaelRoytman:


src/Option/README.md
```diff
@@ -0,0 +1,24 @@
+# Option:
```
MichaelRoytman: Perhaps a better name for this component would be ListBoxOption, to tether it to the ListBox more clearly.

src/Option/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+Provides an Option component that is passed as a child to the ListBox component.
+
+## API
+
+### `aria-selected` (boolean; optional):
+The ListBox component will handle the `aria-selected` attribute itself, but the `aria-selected` prop can be used as an override, if necessary.
```
MichaelRoytman: I anticipated this being necessary based on similar patterns we have in [asInput](https://github.com/edx/paragon/blob/master/src/asInput/index.jsx#L81) and [CheckBox](https://github.com/edx/paragon/blob/master/src/CheckBox/index.jsx#L17). An example use case would be navigating away from the ListBox (say to a different tab or page) and needing to clear its selected state.

src/Option/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+### `children` (element array; required)
+`children` specifies the list of elements that will be displayed within the Option component.
+
+### `className` (string array; optional)
+`className` specifies Bootstrap class names to apply to the Option component. The default is an empty array. (do I mention the existing classnames here?)
+
+### `tag` (string; optional)
+`tag` is used to specify the element type of the rendered Option component. The default is div. Example alternative include li, Button, etc.
```
MichaelRoytman: I have found ListBox implementations that use only divs, but I have also found ones that use list elements, so I figured this would give more flexibility in terms of what Options can be. An Option could also theoretically be a Button, for example.

src/Option/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+### `onOptionSelect` (function, optional)
+`onOptionSelect` is a function that is called when an Option component is selected, by mouse or by keyboard, to perform some additional action.
```
MichaelRoytman: The reason I'm surfacing an `onOptionSelect` prop instead of, say `onClick`, because the behavior in such a prop should happen on selection with both the mouse and the keyboard, so this would cover both cases, instead of the user needing to pass in the same function for onClick and onKeyDown (or worse, different functions).

Furthermore, the underlying implementation of the ListBox will actually have to use onMouseDown and not onClick, and I'd prefer not to surface that oddity (especially since many Paragon components expose onClick) to the consumer.

src/Option/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+### `other`
+`other` specifies any additional attributes on the option element, which can include aria-labelledby, id, key, etc.
```
MichaelRoytman: The Button component does something similar, and I like it, but I am a little concerned this will give the impression that the prop is called "other" instead of a catch-all for all other props sent into the component. Does anyone have a better way to get the meaning across?

src/ListBox/README.md
```diff
@@ -0,0 +1,128 @@
+# ListBox
+
+Provides a ListBox component that can takes Option components as children and handles selection of and navigation through the Options.
```
MichaelRoytman: TODO: `that can takes` => `that takes`

src/ListBox/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+## API
+
+### `children` (Option component array; required)
```
MichaelRoytman: Is there a better way to describe this prop so that users don't pass in an explicit `children` prop?

src/ListBox/README.md
```diff
@@ -0,0 +1,132 @@
...
+`children` specifies the list of Option components that will be displayed within the ListBox element.
+
+### `className` (string array; optional)
+`className` specifies Bootstrap class names to apply to the ListBox component. The default is an empty array.
+
+### `selectedOption` (?)
```
MichaelRoytman: I'm not sure what the type should be. An Option? The index of the Option?

MichaelRoytman:
Also, if you would like to see an example of what this component will likely look like, you can take a look at our EditImageModal in [Studio](https://studio.stage.edx.org/container/block-v1:AnalyticsX+dsjen101+2015_T3+type@vertical+block@c52c2bee0bac499c8b35d1e87453e160). Scroll down to the `Text` block with the banana, click the `Edit` button, and click the `Insert/Edit Image` button in the toolbar (second from the right). You should see the ListBox component between the search bar and the pagination.

src/Option/README.md
dsjen: I like that idea 💯

src/ListBox/README.md
```diff
@@ -0,0 +1,132 @@
...
+### Example using tag prop:
+
+```
+<ListBox
+    tag="ol"
+>
+    <Option
+        tag="li"
+    >
+        <div>Cat 1</div>
```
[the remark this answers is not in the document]
MichaelRoytman: Yes! Does that seem intuitive to you?

src/ListBox/README.md
```diff
@@ -0,0 +1,132 @@
...
+    </Option>
+    <Option
+        tag="li"
+    >
+        <div>Cat 3</div>
+    </Option>
+</ListBox>
+```
+
+### Example using onOptionSelect prop:
```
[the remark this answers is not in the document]
dsjen: I see! I think it's useful to see it here--maybe just needs a comment or something that it's part of the Option API. I was reading the READMEs separately, so might have been confused.

src/ListBox/README.md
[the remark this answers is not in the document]
dsjen: Makes sense to me

src/Option/README.md
fysheets: Would `Option` be something that's reusable with other components? If not, I think clarifying the name (similar to `RadioButtonGroup` / `RadioButton`) makes sense.

src/Option/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+Provides an Option component that is passed as a child to the ListBox component.
+
+## API
+
+### `children` (element array; required)
+`children` specifies the list of elements that will be displayed within the Option component.
+
+### `className` (string array; optional)
+`className` specifies Bootstrap class names to apply to the Option component. The default is an empty array. (do I mention the existing classnames here?)
```
thallada: Same thing about `className` here that I mentioned in ListBox. Also, I don't think it is necessary to mention the classnames applied to Option components internally by Paragon.

src/Option/README.md
```diff
@@ -0,0 +1,26 @@
...
+### `onOptionSelect` (function, optional)
+`onOptionSelect` is a function that is called when an Option component is selected, by mouse or by keyboard, to perform some additional action.
+
+### `other`
+`other` specifies any additional attributes on the Option component, which can include aria-labelledby, id, key, etc.
+
+## props that this component needs but should only come from the parent
+
+### `isSelected` (boolean; required)
+    `isSelected` specifies whether the option is selected and toggles whether the option is visually selected. The default is false. It should only be passed in as a prop from the ListBox.
```
thallada: Nit: this text is indented but other lines on the same level aren't.

src/Option/README.md
thallada: I was expecting this prop to be on the parent element: ListBox since it controls what option is selected by passing `isSelected` to children (I was thinking that ListBox would call the given function with the index of the option). Is there a reason to put this on the Option?

src/Option/README.md
```diff
@@ -0,0 +1,26 @@
+# Option:
+
+Provides an Option component that is passed as a child to the ListBox component.
+
+## API
+
+### `children` (element array; required)
+`children` specifies the list of elements that will be displayed within the Option component.
```
thallada: Can we make the type of this option be node? That way we could have options that are just strings, numbers, one element, or lists of elements.

src/Option/README.md
thallada: This makes me think that this component should be named `ListBoxOption` like you suggested, since the functionality is tied so closely to `ListBox`.

src/ListBox/README.md
MichaelRoytman: Okay, thank you.

src/ListBox/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+Provides a ListBox component that can takes Option components as children and handles selection of and navigation through the Options.
+
+## API
+
+### `children` (Option component array; required)
+`children` specifies the list of Option components that will be displayed within the ListBox element.
+
+### `className` (string array; optional)
+`className` specifies Bootstrap class names to apply to the ListBox component. The default is an empty array.
```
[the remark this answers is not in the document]
MichaelRoytman: Ah, thanks. I knew we had decided on a particular type but couldn't remember and just looked at another component for reference.

src/ListBox/README.md
MichaelRoytman: The index would be easiest for whom? The component or the user? Or both? The one hang up I have about the index is that I'm not sure how to allow the user to "clear" the selection -1? null? If a user were to pass in an actual Option, that would be a little easier to do (i.e. null or not in the children array). For some reason, it also feels a little more intuitive for me to say, "I want this Option selected" as opposed to "I want the Option at index x selected", but I guess then they'd have to keep a reference to the Option element and then we'd need to do some deep equality checking, so perhaps the index is the way to go.

MichaelRoytman: I will check on the question of the props and report back.

src/Option/README.md
MichaelRoytman: I'm not sure. I don't see it as an option (lol) for other components at the moment.

src/Option/README.md
MichaelRoytman: One reason I put this on the Option is because the clicking handling is done on the Option. If I pass in the `onOptionSelect` to the parent, then I have no way to know when the Option is clicked in order to invoke that function. I think I explain that in my pull request comment. It's one of the things I'm not sure how to handle.

The other reason is that I feel like it's more "correct". You're selecting an Option, so I would expect that the selection process be defined by the Option. I guess I can see both sides of things. If the selection function does the same thing for every Option, then, yeah, it would be less cumbersome to pass in the function into the ListBox once instead of for each Option. But, in our case, for example, the click function is specific to the Option. 

It's something like

```
assets.map((asset, index) => (<li onMouseDown={() => this.onMouseDown(asset, index)});
```

so each list item has an onClick that gets the actual asset as a parameter. Then that click function is able to call the Redux select action creator with the asset as part of the action data.  I'm not sure how I could pass in a single function to the parent that would be able to have that behavior. Does that make sense? Do you have ideas for an alternative?

thallada:
@MichaelRoytman I read your "Open Questions" comment. I think 1b makes the most sense to me except for a tweak:

How about moving the `onOptionSelect` prop to the parent `ListBox`? When rendering the `ListBoxOption`'s, it will pass an `onMouseDown` handler to each one that will update the `ListBox`'s selected option state to that index and then call the `onOptionSelect` function. The `onMouseDown` prop to `ListBoxOption` would be a prop only passed by `ListBox`, just like `isSelected`.

I think it would be okay to leave those props undocumented in the README since they are only ever meant to be used internally by `ListBox`. I think we should mention somewhere in the README that `ListBoxOption` will only work inside of a `ListBox` component instead. Also, if an independent `ListBoxOption` with no `isSelected` and `onMouseDown` props passed to it renders without errors, then we could also make those props not required to reduce confusion about them (though, I don't know why you would ever want to render one outside of a `ListBox`).

src/ListBox/README.md
thallada: Re: the index being easier, yeah I think of it like a primary key in a database. It's easier to check for equality against and you don't have to store the whole element in your redux store if you have one.

I think selectedOption as an override, but still allowing user selection to override *that* makes the most sense :+1: . I just think we should document that behavior here for clarity.

src/Option/README.md
```diff
@@ -0,0 +1,26 @@
...
+## API
+
+### `children` (node array; required)
+`children` specifies the array of elements that will be displayed within the ListBoxOption component.
+
+### `className` (string; optional)
+`className` specifies Bootstrap class names to apply to the ListBoxOption component. The default is an empty string.
+
+### `tag` (string; optional)
+`tag` is used to specify the element type of the rendered ListBoxOption component. The default is div. Example alternatives include li, Button, etc.
```
jaebradley: Is there a reason not to use `type` instead of `tag`? I get that it's referencing `HTML` tags, but `Button`, for example, is not a valid `HTML` tag (it is a valid tag for a component in `JSX`)

src/ListBox/README.md
```diff
@@ -0,0 +1,132 @@
+# ListBox
+
+Provides a ListBox component that takes ListBoxOption components as children and handles the selection of and navigation through the ListBoxOption components.
+
+## API
+
+### `children` (ListBoxOption component array; required)
```
jaebradley: Nit: It's not an `array` though? `ListBox`'s `children` should consist of zero or more `ListBoxOption` components, right?

src/Option/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+### `onOptionSelect` (function, optional)
+`onOptionSelect` is a function that is called when an ListBoxOption component is selected, by mouse or by keyboard, to perform some additional action.
+
+### `other`
+`other` specifies any additional attributes on the ListBoxOption component, which can include aria-labelledby, id, key, etc.
+
+## props that this component needs but should only come from the parent
+
+### `isSelected` (boolean; required)
```
jaebradley: I'm not sure this needs to be exposed? Since it's an implementation detail of how `ListBox` will interact with `ListBoxOption`?

src/Option/README.md
MichaelRoytman: I can change it to type. Just so that I'm sure I understand correctly, you're not advocating for supporting anything but HTML tags here, right?

jaebradley: Oh I misunderstood this - when I saw `Button` I thought that meant a `Button` component.

If that's not the case, and it's only HTML tags, then I think `tag` is fine!

src/Option/README.md
```diff
@@ -0,0 +1,26 @@
+# ListBoxOption:
+
+Provides an ListBoxOption component that is passed as a child to the ListBox component.
+
+## API
```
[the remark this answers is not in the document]
MichaelRoytman: My one concern is that focusing on the ListBox is the only way to interact with it with the keyboard. The pattern is that the ListBox gets focus and then you navigate through the ListBoxOptions with the arrow keys. The ListBox is also supposed to auto-select the first ListBoxOption if you focus on the ListBox for the first time, and I was a little worried about an onFocus prop providing the opportunity to break accessibility without the consumer knowing (e.g. they choose to select the last option onFocus with the selectedOptionIndex prop or something). I definitely see the use for onFocus and onBlur, and they seem to be common props to expore. Do you think we should just expose them and put the onus on the user to know when they're breaking accessibility patterns?

src/Option/README.md
```diff
@@ -0,0 +1,26 @@
...
+`className` specifies Bootstrap class names to apply to the ListBoxOption component. The default is an empty string.
+
+### `tag` (string; optional)
+`tag` is used to specify the element type of the rendered ListBoxOption component. The default is div. Example alternatives include li, Button, etc.
+
+### `onOptionSelect` (function, optional)
+`onOptionSelect` is a function that is called when an ListBoxOption component is selected, by mouse or by keyboard, to perform some additional action.
+
+### `other`
+`other` specifies any additional attributes on the ListBoxOption component, which can include aria-labelledby, id, key, etc.
```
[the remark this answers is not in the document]
MichaelRoytman: I think the reason I added this is because not all of our components have this behavior, and sometimes it can be very useful. The Button has it, and I only discovered it after looking through the source code. But I can remove it if it seems unnecessary.

src/Option/README.md
jaebradley: Also, for the `PropType` you could use `oneOf` https://github.com/sindresorhus/html-tags

src/ListBox/README.md
jaebradley: Are you planning on throwing any errors or warnings if an API user doesn't use `ListBoxOption`s for all the `children`?

MichaelRoytman: Yes. But the majority of Paragon components that explicitly specify the children prop refer to an array, with the exception being Fieldset, which specifies node. I'd like to make that consistent across the components.

src/ListBox/README.md
```diff
@@ -0,0 +1,132 @@
...
+`children` specifies the list of ListBoxOption components that will be displayed within the ListBox element.
+
+### `className` (string; optional)
+`className` specifies Bootstrap class names to apply to the ListBox component. The default is an empty string.
+
+### `selectedOptionIndex` (number; optional)
```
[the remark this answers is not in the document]
MichaelRoytman: Actually, a question I have is how a user could clear the selected ListBoxOption using this prop. Would they pass in a negative integer? A null? I'd like to support that case, since there have been multiple instances of us needing a way to clear a components state this way.

src/Option/README.md
MichaelRoytman: Yep, thanks for catching that.

src/Option/README.md
MichaelRoytman: Ack! Sorry, that's my mistake. I did actually mean Button, but a string "Button" is pretty useless. Hm. Is there even a way we could support a JSX component in this way?

src/Option/README.md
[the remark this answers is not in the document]
jaebradley: Definitely something to be concerned about 👍 .

I guess my opinion is that if there's a feature where somebody **could** misuse the API I'm not sure that is reason enough to not support such a feature. Particularly if it's

1. Well-documented: So maybe in the documentation for `onFocus/Blur` call out that certain `state` changes should not occur in certain circumstances.
2. Supported by the API itself: Maybe there are console warnings or something in the cases where accessibility patterns are broken - this could be too complex though.

All that being said, I also think that if there's not an initial use case, it might make sense to punt - it's (almost) always easier to add API methods vs. removing them.

src/Option/README.md
jaebradley: Check [this StackBlitz example](https://stackblitz.com/edit/dynamically-create-react-component) out

```jsx
import React, { Component } from 'react';
import { render } from 'react-dom';
import Foo from './Foo';
import Bar from './Bar';
import './style.css';

class App extends Component {
  constructor() {
    super();
    this.state = {
      name: 'React'
    };
  }

  render() {
    return (
      <div>
        <Bar component={Foo} />
      </div>
    );
  }
}

render(<App />, document.getElementById('root'));
```

src/ListBox/ListBox.stories.jsx
```diff
@@ -0,0 +1,169 @@
+/* eslint-disable import/no-extraneous-dependencies */
+/* eslint-disable react/no-multi-comp */
+// import { action } from '@storybook/addon-actions';
```
fysheets: If this is not going to be used, let's just delete it.

src/ListBox/ListBox.scss
```diff
@@ -0,0 +1 @@
+@import 'bootstrap/scss/_list-group.scss';
```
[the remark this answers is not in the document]
MichaelRoytman: I don't think it lints scss files, so I always make this mistake :(. I'll add a new line.

src/ListBox/ListBox.stories.jsx
MichaelRoytman: Oops. I didn't mean to leave that in. Thanks for catching that.

src/ListBox/ListBox.stories.jsx
```diff
@@ -0,0 +1,168 @@
...
+      >
+        {fruit}
+      </ListBoxOption>
+    ));
+
+    return (
+      <React.Fragment>
+        <span aria-live="polite">
+          Selected Fruit:
+          {this.state.selectedOptionIndex !== undefined ?
```
dsjen: Nit: I often find it's easier to read in the positive. In this case, you could flip the conditional to be `this.state.selectedOptionIndex === undefined` rather.

src/ListBox/ListBox.stories.jsx
```diff
@@ -0,0 +1,168 @@
...
+    super(props);
+
+    this.onSelect = this.onSelect.bind(this);
+    this.onButtonClick = this.onButtonClick.bind(this);
+
+    this.state = {
+      reset: true,
+    };
+
+    this.fruits = ['Apple', 'Orange', 'Strawberry', 'Banana'];
```
dsjen: Nit: this could be a little DRY-er if a `fruits` hash/object/dictionary `const` were declared outside of the classes and `this.fruits` were extracted from the keys, e.g. [Object.keys()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys).

src/ListBox/ListBox.test.jsx
```diff
@@ -0,0 +1,169 @@
...
+
+    it('should not select first ListBoxOption on focus if ListBoxOption selected', () => {
+      wrapper = shallow(listBox);
+
+      wrapper.setState({
+        selectedOptionIndex: 1,
+      });
+
+      wrapper.simulate('focus');
+      expect(wrapper.state('selectedOptionIndex')).toEqual(1);
```
dsjen: Nice test! 🥂

src/ListBox/index.jsx
```diff
@@ -0,0 +1,121 @@
...
+
+    if (this.shouldUpdateSelectedOptionIndex(selectedOptionIndex)) {
+      this.setState({
+        selectedOptionIndex,
+      });
+    }
+  }
+
+  onFocus() {
+    // if no ListBoxOption is selected, select first ListBoxOption on ListBox focus
```
dsjen: I guess this statement is mostly true. If `0` were selected, it would set the state to `0`, which is totally fine.

src/ListBox/index.jsx
```diff
@@ -0,0 +1,121 @@
...
+      this.setState({
+        selectedOptionIndex: 0,
+      });
+    }
+  }
+
+  onKeyDown(e) {
+    switch (e.key) {
+      case 'ArrowDown': {
+        // prevent scrolling entire modal body with arrow keys
```
dsjen: These comments are all super helpful!

src/ListBoxOption/ListBoxOption.scss
```diff
@@ -0,0 +1 @@
+@import 'bootstrap/scss/_list-group.scss';
```
dsjen: Nit: new line :)

src/ListBoxOption/ListBoxOption.test.jsx
```diff
@@ -0,0 +1,133 @@
...
+    it('should call onSelect on mouse down', () => {
+      wrapper = shallow(listBoxOption);
+      const onSelectSpy = jest.fn();
+
+      wrapper.setProps({
+        onSelect: onSelectSpy,
+      });
+
+      wrapper.simulate('mouseDown');
+      expect(onSelectSpy).toHaveBeenCalledTimes(1);
```
dsjen: Nice!

src/ListBoxOption/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+      tag,
+      ...other
+    } = this.props;
+
+    return React.createElement(
+      this.props.tag,
+      {
+        'aria-selected': isSelected,
+        className: classNames(styles['list-group-item'], styles['list-group-item-action'], { active: this.props.isSelected }, className),
+        id: `list-box-option-${index}`,
```
thallada: Maybe if `index` is not passed and it defaults to `undefined`, we should just not add an `id` attribute to the element. Otherwise, there might be a whole bunch of option elements with the same "`list-box-option-undefined`" ids.

src/ListBox/ListBox.stories.jsx
```diff
@@ -0,0 +1,172 @@
...
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import centered from '@storybook/addon-centered';
+import 'bootstrap/scss/utilities/_screenreaders.scss';
+
+import Button from '../Button';
+import ListBox from '../ListBox';
+import ListBoxOption from '../ListBoxOption';
+
+const fruits = {
```
adamstankiewicz: Haha, this is great 🤣

src/ListBox/index.jsx
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  shouldUpdateSelectedOptionIndex(selectedOptionIndex) {
+    return selectedOptionIndex !== this.state.selectedOptionIndex
+      && selectedOptionIndex !== undefined;
+  }
+
+  renderChildren() {
+    return React.Children.map(this.props.children, (child, index) =>
+
```
adamstankiewicz: nit: unnecessary line break

src/ListBoxOption/index.jsx
```diff
@@ -0,0 +1,73 @@
...
+
+  /**
+   * onMouseDown is used instead of onClick because onClick triggers the focus
+   * event before click event. This focus event bubbles up to the parent
+   * (since onFocus bubbles in React), which triggers the ListBox's onFocus function.
+   * This function will select the first ListBoxOption if one is not selected, and this
+   * causes the user to see the first option selected before their desired option is selected
+   * when the click event is fired shortly thereafter. The mouseDown event occurs before the
+   * focus event, which prevents this behavior.
+   */
```
adamstankiewicz: Great comment!

src/ListBoxOption/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+      isSelected,
+      tag,
+      ...other
+    } = this.props;
+
+    return React.createElement(
+      this.props.tag,
+      {
+        'aria-selected': isSelected,
+        className: classNames(styles['list-group-item'], styles['list-group-item-action'], { active: this.props.isSelected }, className),
```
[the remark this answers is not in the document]
adamstankiewicz: Agree with Tyler here. Maybe this instead:
```jsx
className: classNames(
  styles['list-group-item'],
  styles['list-group-item-action'],
  {
    active: this.props.isSelected,
  },
  className
)
```

src/ListBoxOption/index.jsx
adamstankiewicz: Agree with Tyler here.

PR #256, opened 2018-05-23
docs(checkbox-group): add storybook addons

jaebradley:
image

@fysheets any idea what the best solution for the "Ensures related <input type="checkbox"> elements have a group and that that group designation is consistent" accessibility violation is?


fysheets:
Running aXe on that page shows the following error that makes it a bit more clear:

screen shot 2018-05-23 at 10 10 02 am

Looks like it was a bad copy/paste job for the story where `name=basicCheckBox` on each of the CheckBoxes - making each name unique for the story (`basicCheckBox1`, `basicCheckBox2`, `basicCheckBox3`) will fix that accessibility error.

PR #297, opened 2018-07-11
fix(pagination): only update if currentPage state changes

adamstankiewicz:


src/Pagination/index.jsx
```diff
@@ -38,6 +38,10 @@ class Pagination extends React.Component {
     }
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
+    return nextState.currentPage !== this.state.currentPage;
```
georgebabey: nit: I'd add a comment explaining why this is required, something like `Avoids rerender when only the pageButtonSelected changes`

PR #307, opened 2018-08-03
fix(greenkeeper): greenkeeper incompatible with npm ci, use npm install

MichaelRoytman:


MichaelRoytman:
Per [this](https://github.com/greenkeeperio/greenkeeper-lockfile#npm) documentation,

```
npm ci won't work with greenkeeper pull requests because:
If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock.

Travis will use npm ci by default if lockfiles are present so you'll need to explicitly tell your CI to run npm install instead of npm ci
```

Greenkeeper is aware of the issue, it seems, and at the time of [this comment](https://github.com/greenkeeperio/greenkeeper-lockfile/issues/156#issuecomment-394146762), recommended using `npm install` instead of Travis's default `npm ci`. This may have been [fixed](https://github.com/greenkeeperio/greenkeeper-lockfile/pull/140#issuecomment-393483674) in their `2.0.0` [release](https://github.com/greenkeeperio/greenkeeper-lockfile/pull/140#issuecomment-397284158), but I haven't verified that, and it's not their latest release. I believe it's still being tested.

I have changed `.travis.yml` to use `npm install` instead, but we lose some of the benefits of `npm ci`. Below, there are two example PRs that demonstrate that this fixes the issue.

https://github.com/edx/paragon/pull/300
https://github.com/edx/paragon/pull/299

PR #313, opened 2018-08-17
feat(input-text): add readonly

bbaker6225:
This work is done as part of [LEARNER-6116](https://openedx.atlassian.net/browse/LEARNER-6116). 

Adds an optional readOnly attribute to the inputText component since Firefox does not allow users to copy text from a disabled field.

**Storyboard**
If the reviewer decides that this doesn't need its own story, I'm happy to remove it.  Added just to give an idea of the UI.
screen shot 2018-08-17 at 11 21 01 am

### DIFF
--- a/src/InputText/InputText.test.jsx
+++ b/src/InputText/InputText.test.jsx
@@ -62,5 +62,15 @@ describe('<InputText />', () => {
       expect(input.hasClass('first')).toEqual(true);
       expect(input.hasClass('last')).toEqual(true);
     });
+
+    it('should not be readOnly if the readOnly property is not set', () => {
+      const wrapper = mount(<InputText {...props} />);
+      expect(wrapper.props().readOnly).toBeUndefined();
+    });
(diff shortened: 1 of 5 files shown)


src/InputText/InputText.test.jsx
```diff
@@ -62,5 +62,10 @@ describe('<InputText />', () => {
       expect(input.hasClass('first')).toEqual(true);
       expect(input.hasClass('last')).toEqual(true);
     });
+
+    it('should render with the readOnly property if set', () => {
+      const wrapper = mount(<InputText {...props} readOnly />);
+      expect(wrapper.props().readOnly).toEqual(true);
+    });
```
AlasdairSwan: It would be great to add a test to check that it defaults to not having a readOnly prop set too.

bbaker6225:
Added new test to verify that the input is readOnly only when it is explicitly set

PR #315, opened 2018-08-22
fix(tabs): fix naviation in Tabs component

tuchfarber:
Tabs component didn't allow tab navigation or keyboard navigation. To meet accessibility best practices, it was determined to follow an implementation similar to:
http://pauljadam.com/demos/aria-tabpanel.html

### DIFF
--- a/src/Tabs/index.jsx
+++ b/src/Tabs/index.jsx
@@ -7,6 +7,7 @@ import React from 'react';
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
 
+import Button from '../Button';
 import styles from './Tabs.scss';
 import newId from '../utils/newId';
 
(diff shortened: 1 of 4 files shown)


src/Tabs/index.jsx
```diff
@@ -43,25 +43,21 @@ class Tabs extends React.Component {
       const labelId = this.genLabelId(i);
 
       return (
-        <li
-          className={styles['nav-item']}
+        <button
```
adamstankiewicz: Thoughts on using the `Button` component here?

tuchfarber: I don't have a preference, but I'm willing to convert it if that makes sense.

adamstankiewicz: Personally, my preference would be to use the `Button` component.

tuchfarber:
@adamstankiewicz Fixed button question

src/Tabs/index.jsx
```diff
@@ -43,25 +44,20 @@ class Tabs extends React.Component {
...
-          >
-            {label}
-          </a>
-        </li>
+          onClick={() => { this.toggle(i); }}
+          className={classNames(
+            styles['nav-link'],
+            styles['nav-item'],
+            { [styles.active]: selected },
+          ).split(' ')}
```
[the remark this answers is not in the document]
tuchfarber: I think if we're planning on making a breaking change to make Buttons accept strings instead of arrays (like they should), I like that this is an obvious work around. An array of 1 string with spaces in it is less likely to be caught in the refactor.

PR #319, opened 2018-08-23
fix(icon): swap wrapping div for a span

AlasdairSwan:
So Icons will be inline by default.


adamstankiewicz:
Thanks for doing this, @AlasdairSwan! 😃

PR #322, opened 2018-08-27
feat(searchfield): change SearchField to accept ref for input

tuchfarber:
SearchField can now pass a ref to it's <input> that parent container that parent container can reference.

### DIFF
--- a/src/SearchField/index.jsx
+++ b/src/SearchField/index.jsx
@@ -22,6 +22,7 @@ const defaultProps = {
     searchButton: 'Submit search',
   },
   value: '',
+  inputRef: null,
 };
 
 const propTypes = {
@@ -37,6 +38,7 @@ const propTypes = {
     searchButton: PropTypes.string.isRequired,
   }),
   value: PropTypes.string,
+  inputRef: PropTypes.string,
 };
 
 class SearchField extends React.Component {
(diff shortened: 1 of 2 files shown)


src/SearchField/index.jsx
```diff
@@ -185,7 +187,7 @@ class SearchField extends React.Component {
           onChange={this.handleChange}
           onKeyPress={this.handleKeyPress}
           inputGroupAppend={this.getSearchActionButtons()}
-          inputRef={(input) => { this.textInput = input; }}
+          inputRef={(input) => { if (!this.inputRef) { this.inputRef = input; } }}
```
georgebabey: I think this should be pulled out of the render method and dealt with in the constructor:
`this.inputRef = this.props.inputRef || React.createRef()` so that this is just a straight `inputRef={this.inputRef}`

But we should wait for @adamstankiewicz review, as I'm not sure if there was a good reason this was being set in `render()`

adamstankiewicz: No, there wasn't a good reason for the `ref` being set in the `render` (personal preference). Totally cool if it's changed to what @georgebabey suggests. Probably makes more sense now since there's additional logic around whether to set the ref or not.

adamstankiewicz:
Looks good! I'd just suggest adding the `inputRef` prop to the component's `README` and adding a Storybook story demonstrating how to pass it in (and set focus).

src/SearchField/index.jsx
tuchfarber: @georgebabey Looks like Paragon is still on React 16.2 and createRef didn't come until 16.3. Not sure if I should updated Paragon to 16.3, or if there's a different way to do the `this.inputRef = this.props.inputRef || React.createRef()`

tuchfarber:
Closing because out of date

PR #323, opened 2018-08-28
fix(package): update react to version 16.4.2

tuchfarber:


adamstankiewicz:
@tuchfarber I think it's a great idea to upgrade React! Out of curiosity, is there a particular reason for the upgrade? Also, did you happen to go through and test that this wouldn't be a breaking change for any of our existing components?

tuchfarber:
@adamstankiewicz 16.3 is when createRef was introduced and I need that for #322. I updated and flipped through the storybook and everything seemed to work. Tests all pass as well.

PR #327, opened 2018-09-13
refactor(search-field): add i18n ability

matthugs:
This adds the ability to add react elements in SearchFields' aria-label props, which is necessary for our Next Big Thing New Approach to internationalizing such messages.


matthugs:
FYI, @edx/fedx-team

matthugs:
Note: updated commit message to make sure there's a version bump. Seems like refactor may not do that as reliably, and I'll need it for some downstream diffs.

PR #335, opened 2018-09-26
feat(breadcrumb): add a breadcrumb component

AlasdairSwan:


src/Breadcrumb/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import styles from './Breadcrumb.scss';
+import Icon from '../Icon';
+
+const Breadcrumbs = ({
+  links, activeLabel, spacer, clickHandler,
+}) => {
+  const linkCount = links.length;
```
bbaker6225: It feels cleaner to me to just use links.length instead of assigning it to a variable for just one use

src/Breadcrumb/Breadcrumb.test.jsx
```diff
@@ -0,0 +1,68 @@
...
+  it('renders with just links', () => {
+    wrapper = mount(<Breadcrumb {...baseProps} />);
+
+    const list = wrapper.find('ol li');
+    expect(list.length).toEqual(5);
+    expect(list.find('a').length).toEqual(3);
+  });
+
+  it('renders with links and active label', () => {
+    const label = 'Link 4';
```
bbaker6225: Nit: I'd prefer the string was just 'Label' or something like that since `activeLabel` is not actually a link

src/Breadcrumb/Breadcrumb.test.jsx
```diff
@@ -0,0 +1,68 @@
...
+    {
+      label: 'Link 2',
+      url: '/link-2',
+    },
+    {
+      label: 'Link 3',
+      url: '/link-3',
+    },
+  ],
+  activeCourseRun: {
```
bbaker6225: This and the `uuid` property should probably be removed

src/Breadcrumb/Breadcrumb.stories.jsx
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import centered from '@storybook/addon-centered';
+import { checkA11y } from '@storybook/addon-a11y';
+import { withInfo } from '@storybook/addon-info';
+import { withReadme } from 'storybook-readme';
+import README from './README.md';
```
bbaker6225: Nit: This would be better grouped with `import Breadcrumb from './index';`

src/Breadcrumb/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+    <nav aria-label="breadcrumb">
+      <ol className={classNames(styles['list-inline'])}>
+        {links.map(({ url, label }, i) => (
+          <React.Fragment key={url}>
+            <li className={classNames(styles['list-inline-item'])}>
+              <a href={url} {...(clickHandler && { onClick: clickHandler })}>{label}</a>
+            </li>
+            {(activeLabel || ((i + 1) < linkCount)) &&
+              <li className={classNames(styles['list-inline-item'])} role="presentation" aria-label="spacer">
+                {spacer || <Icon className={['fa', 'fa-chevron-right']} />}
```
bbaker6225: This might be an issue with `Icon`, but I noticed that all the spacer Icons have the same ID in storybook

src/Breadcrumb/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+}) => {
+  const linkCount = links.length;
+
+  return (
+    <nav aria-label="breadcrumb">
+      <ol className={classNames(styles['list-inline'])}>
+        {links.map(({ url, label }, i) => (
+          <React.Fragment key={url}>
+            <li className={classNames(styles['list-inline-item'])}>
+              <a href={url} {...(clickHandler && { onClick: clickHandler })}>{label}</a>
```
bbaker6225: Could you add a simple test for the clickHandler?

src/Breadcrumb/index.jsx
AlasdairSwan: It's more performant to assign it to a variable, otherwise it has to calculate `.length` for each loop of `map`.

src/Breadcrumb/index.jsx
AlasdairSwan: It is an issue with `Icon`, but I can get around it until that issue is resolved.

PR #336, opened 2018-09-26
feat(collapsible): add collapsible component

bbaker6225:
Add collapsible component

### DIFF
--- a/src/Collapsible/README.md
+++ b/src/Collapsible/README.md
@@ -1,0 +1,43 @@
+# Collapsible Component
+
+Provides a component that can collapse to hide its child elements and
+optionally has the ability to display without the open/close button based on the
+screen size. The collapsible functionality mimics that of an accordion section,
+with the exception that multiple collapsibles can be open at the same time.
+
+## API
+
+### `children` (object; required)
+`children` are the objects that are the children of the collapsible that should
+be hidden when the collapsible is closed.
+
+### `title` (string; required)
+`title` is the string to be displayed on the collapsible button.
+
+### `expandedTitle` (element; optional)
+`expandedTitle` is the element to be displayed as the title when the collapsible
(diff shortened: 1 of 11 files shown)


src/Collapsible/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+    const { isExpanded, isOpen } = this.state;
+
+    return (
+      <div className={[classNames(
+        'collapsible',
+        { [styles.open]: isOpen && !isExpanded },
+        )]}
+      >
+        {isExpanded ? (
+          expandedTitle && expandedTitle
```
AlasdairSwan: Why not just ```expandedTitle```?

src/Collapsible/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### `expandedTitle` (element; optional)
+`expandedTitle` is the element to be displayed as the title when the collapsible
+is expanded. Defaults to undefined.
+
+### `isOpen` (boolean; optional)
+`isOpen` specifies whether the collapsible should initially be open. Defaults
+to false.
+
+### `isWindowExpanded` (function; optional)
```
AlasdairSwan: A nitpick around the name of the prop: because any function can be passed in a reference to the window being expanded seems misleading. How about something along the lines of ```isCollapsible``` or ```disableCollapse``` to make it explicit that the function determines whether the ability to collapse is active.

bbaker6225: I was worried that isCollapsible or something like that might cause confusion about what functionality is dependent on the passed in function (i.e. would it affect the open/close button), but I'm fine with this change if it makes sense to you

bbaker6225:
@AlasdairSwan Added requested style changes from ux here, will change the rest in prospectus once I add this in.  Holding off on the open/close animation for now

PR #338, opened 2018-09-27
Update enzyme to the latest version

dsjen:


dsjen:
Ok, so the enzyme 1.8 upgrade is going to be more work than just upgrading!

PR #340, opened 2018-10-10
fix(collapsible): fix open/close animation

bbaker6225:


bbaker6225:
This fix doesn't work on firefox, and looks pretty bad on safari 😞

PR #341, opened 2018-10-11
fix(collapsible): fix animation and isOpen state

bbaker6225:


src/Collapsible/index.jsx
```diff
@@ -24,6 +25,14 @@ class Collapsible extends React.Component {
     }
   }
 
+  componentDidUpdate(prevProps) {
```
bbaker6225: @AlasdairSwan this change is to allow the credentials link to be able to open the collapsible

src/Collapsible/index.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-did-update-set-state */
```
AlasdairSwan: I'm not a fan of disabling this. It should be relatively easy to refactor (see https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-did-update-set-state.md).

bbaker6225: Isn't it ok in this case since it's wrapped in a condition checking the previous props? https://reactjs.org/docs/react-component.html#componentdidupdate
```jsx
componentDidUpdate(prevProps) {
  // Typical usage (don't forget to compare props):
  if (this.props.userID !== prevProps.userID) {
    this.fetchData(this.props.userID);
  }
}
```
"You may call setState() immediately in componentDidUpdate() but note that it must be wrapped in a condition like in the example above"

bbaker6225: Added comments explaining why it's disabled

PR #343, opened 2018-10-15
fix(collapsible): add onToggle callback

AlasdairSwan:


AlasdairSwan:
A separate PR would be easier I think.

PR #346, opened 2018-10-18
add icon for dropdown

muhammad-ammar:
[ENT-1104](https://openedx.atlassian.net/browse/ENT-1104)

### DIFF
--- a/src/Dropdown/README.md
+++ b/src/Dropdown/README.md
@@ -14,3 +14,6 @@ Note: When passing in a component to `menuItems` that uses a prop other than `re
 
 ### `title` (string; required)
 `title` specifies the text that is displayed within the original dropdown button.
+
+### `iconElement` (element; optional)
+`iconElement` specifies the icon to be displayed alongside the dropdown.
--- a/src/Dropdown/index.jsx
+++ b/src/Dropdown/index.jsx
@@ -131,37 +131,74 @@ class Dropdown extends React.Component {
   }
 
   render() {
+    const { open } = this.state;
+    const {
+      buttonType,
+      iconElement,
+      menuItems,
+      title,
+    } = this.props;
+    const hasIconElement = React.isValidElement(iconElement);
+
     return (
       <div
         className={classNames([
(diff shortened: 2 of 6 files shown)


muhammad-ammar:
@georgebabey Can you please give it a quick look and see if these changes are good to merge or we should do the changes in edx-portal?

src/Dropdown/index.jsx
```diff
@@ -131,14 +131,25 @@ class Dropdown extends React.Component {
...
         className={classNames([
           styles.dropdown,
-          { [styles.show]: this.state.open },
+          { [styles.show]: this.state.open, 'has-icon': hasIcon },
         ])}
         ref={(container) => { this.container = container; }}
       >
+        {hasIcon &&
+          <div className="icon-container">
+            <img
```
georgebabey: rather than forcing this to be an img, when someone may want an `<Icon>` here, I think making a props of `iconElement` that just gets rendered here would make more sense. That gives the consumer the flexibility to pass in an image or an icon.

muhammad-ammar:
@georgebabey @adamstankiewicz I have addressed the feedback. Please continue the review.

muhammad-ammar:
@georgebabey @adamstankiewicz This is waiting for your review.

src/Dropdown/README.md
```diff
@@ -14,3 +14,9 @@ Note: When passing in a component to `menuItems` that uses a prop other than `re
 
 ### `title` (string; required)
 `title` specifies the text that is displayed within the original dropdown button.
+
+### `iconElement` (element; optional)
+`iconElement` specifies the icon to be displayed alongside the dropdown. It can be one of
```
adamstankiewicz: `iconElement` could technically be any element, not just an `` or `<Icon />`. For example, you might want to wrap an `` in a router `Link` component.

muhammad-ammar:
@adamstankiewicz Bundle of thanks for the detailed review. I have incorporated your comment. Please continue the review.

PR #349, opened 2018-10-22
`react-intl` ADR

matthugs:
This is a proposed formalization of a decision that was made on 10/4 (action item in minutes [here](https://openedx.atlassian.net/wiki/spaces/FEDX/pages/845612224/2018-10-04+Meeting+notes?focusedTaskId=3)).

I'm a little apprehensive about the process for adding ADRs with "approved" already in the file; could you weigh in on this tiny point of process, @nasthagiri?

Requesting review from @edx/fedx-team (can I do this from the "Reviewers" cog?)

### DIFF
--- a/docs/decisions/0002-react-i18n.rst
+++ b/docs/decisions/0002-react-i18n.rst
@@ -1,0 +1,78 @@
+2. ``react-intl``
+-----------------
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+We have business needs to translate or otherwise internationalize
+strings of text used within ``paragon``
+components. Internationalization is critical to the goal of extending
+access to education globally, a core aspect of the mission of the
+edX platform.
+
+The React community has demonstrated fairly widespread adoption of
+react-intl_ for addressing this need in various contexts.
+The ``edx/studio-frontend`` project has used react-intl_ for some
+time.
+
+Decision
+--------
+
+We will adopt the use of react-intl_ for all new learner- and
+educator-facing web development using React.
+We will attempt to standardize our use of internationalization
+libraries (including but not limited to react-intl_) for existing
+React-based web development.
+
+Consequences
+------------
+
+react-intl_ uses the `React context API`_ when using the
+``FormattedMessage``, ``React.Component``-based approach to using the
+API. Because of this reliance on context, folks passing components
+to ``paragon`` components (e.g. as ``children``) will expect that
+these components will be rendered within the same React component
+hierarchy as the ``paragon`` component, rather than some other,
+independent component hierarchy. This way, the underlying ``Consumer``
+will be in the same hierarchy as the corresponding ``Provider``. We,
+as developers of the ``paragon`` library, must not foil this
+assumption.
+
+Thus a corollary decision: any ``prop`` for a ``paragon`` component
+which accepts and renders inputs of type ``PropType.element`` must
+include that element within the same React component hierarchy in
+which it itself is found (and no others). `An example of issues that
+arise if this isn't done can be found here.`_ As a result,
+getTextFromElement_ (and any other utilities which rely on rendering
+such props in another React component hierarchy) are deprecated
+according to the process outlined in OEP-21_.
+
+Because getTextFromElement was used to reduce duplication of strings
+used both as a11y strings (e.g. ``aria-label``) and as html-wrapped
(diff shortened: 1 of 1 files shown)


OmarIthawi:
Thanks @mattdrayer! I have one question though, how would this translate into Transifex entries?

```jsx
<FormattedMessage
    id="welcome"
    defaultMessage={`Hello {name}, you have {unreadCount, number} {unreadCount, plural,
      one {message}
      other {messages}
    }`}
    values={{name: <b>{name}</b>, unreadCount}}
/>
```

The syntax is somewhat different that what it would be in the Python world:

```python
ungettext(
    'Hello {name}, you have one message',
    'Hello {name}, you have {unreadCount} messages',
    unreadCount
).format(
    name='<b>{}</b>'.format(name),
    unreadCount=unreadCount,
)
```

The message above would be translated into `n` Transifex entries by having one `msgid` and multiple `msgstr[]` entries.

This works similarly for simple languages that has only two plural forms and complex languages with multiple plural forms such as Arabic.

Do you know if `react-intl` has a similar level of support to what Django provides?

fysheets:
Hi @OmarIthawi, `react-intl` does support multiple plurals! Here is a [link to the `react-intl` documentation on plurals](https://github.com/yahoo/react-intl/wiki/Components#formattedplural). and here is a[ link to documentation within the `edx/studio-frontend` repo](https://github.com/edx/studio-frontend/blob/fb3515347fafdc87974bfc91bbb5d73ef192e471/src/data/i18n/README.md#pluralization) of how that repository is supporting plurals with `react-intl` and `Transifex` for our translations. Finally, you can see this [pluralization support in action within `studio-frontend` to support of the `AssetsResultsCount` component](https://github.com/edx/studio-frontend/blob/fb3515347fafdc87974bfc91bbb5d73ef192e471/src/components/AssetsResultsCount/displayMessages.jsx). I hope these resources are helpful, we can also work as a team to add some pluralization documentation within Paragon for clarity.

matthugs:
@OmarIthawi Have you had an opportunity to document this question (possibly as an issue?) with [one of the FormatJS project's core libraries](https://formatjs.io/github/)? I'm not familiar enough with the subtleties of Arabic syntax to be sure I could investigate your question. I _really_ appreciate your bringing it up here, though! For now, we're going to continue formalizing the decision to use this library since it is a de facto standard among the react community, but I welcome suggestions of alternatives! Feel free to circle back if you run into any difficulty finding the right outlet for this concern.

fysheets:
Thanks for sharing that context @OmarIthawi! Great to know :)

docs/decisions/0002-react-i18n.rst
```diff
@@ -0,0 +1,70 @@
...
+We will attempt to standardize our use of internationalization
+libraries (including but not limited to react-intl_) for existing
+React-based web development.
+
+Consequences
+------------
+
+react-intl_ uses the `React context API`_ when using the
+``FormattedMessage``, ``React.Component``-based approach to using the
+API. Because of this reliance on context, components which are passed
```
MichaelRoytman: This paragraph is kind of confusing to me. Would it be possible to get an example of what `must be rendered in
the same React component hierarchy in each of the corresponding
components` would look like, more concretely?

matthugs: I hear that. It's a little abstract, because I'm trying to phrase fundamentally negative guidance as if it were positive (i.e. I'm trying to say "do X!" instead of "don't do Y!"). Let's brainstorm a good alternative during tomorrow's FedX meeting.

docs/decisions/0002-react-i18n.rst
```diff
@@ -0,0 +1,70 @@
...
+Accepted
+
+Context
+-------
+
+We have business needs to translate or otherwise internationalize
+strings of text used within ``paragon``
+components. Internationalization is critical to the goal of extending
+access to education globally, a core aspect of the mission of the
+``edx`` platform.
```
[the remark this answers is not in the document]
matthugs: I love the suggestion feature! Not going to try to use it's "apply suggestion" dealio, since I want to keep my commit history clean, but I love how clear this is, especially for stylistic concerns!

nasthagiri:
@matthugs - The FEDx group agreed to merge this PR for you while you're away.  We can always update the ADR later in a subsequent PR if needed.

PR #361, opened 2018-11-15
fix(modal): change the positioning of the modal from relative to fixed

### LINKED ISSUE #360: Modal pushes content down or is pushed down by content when rendered in a page.
Make it so the Modal is `Fixed positioned` So it will always render at the top-center of the page.

reillz10:
This resolves #360 by making sure the Modal sits on top of all other elements in the DOM.

### DIFF
--- a/src/Modal/index.jsx
+++ b/src/Modal/index.jsx
@@ -10,7 +10,7 @@ import Icon from '../Icon';
 import newId from '../utils/newId';
 import Variant from '../utils/constants';
 
-const closeModalButtonId = 'paragonCloseModalButton';
+const closeModalButtonId = newId('paragonCloseModalButton');
 
 class Modal extends React.Component {
   constructor(props) {
@@ -38,7 +38,7 @@ class Modal extends React.Component {
     }
     this.parentElement = document.querySelector(this.props.parentSelector);
     if (this.parentElement === null) {
-      throw new Error(`Modal recieved invalid parentSelector: ${this.props.parentSelector}, no matching element found`);
+      throw new Error(`Modal received invalid parentSelector: ${this.props.parentSelector}, no matching element found`);
     }
     this.parentElement.appendChild(this.el);
   }
@@ -111,9 +111,15 @@ class Modal extends React.Component {
     );
   }
 
-  close() {
-    this.setState({ open: false });
-    this.props.onClose();
+  close(e) {
+    if (e) {
+      e.stopPropagation();
+    }
+
+    if (!e || e.target.classList.contains('js-close-modal-on-click')) {
+      this.setState({ open: false });
+      this.props.onClose();
+    }
   }
 
   handleKeyDown(e) {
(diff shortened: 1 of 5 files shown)


src/Modal/index.jsx
```diff
@@ -178,30 +188,35 @@ class Modal extends React.Component {
...
-          {...(!renderHeaderCloseButton ? { ref: this.setFirstFocusableElement } : {})}
+          id="modal"
+          className={classNames(styles.modal, {
+            [styles.show]: open,
+            [styles.fade]: !open,
+            [styles['d-block']]: open,
+            [styles['is-ie11']]: this.isIE11,
+          })}
+          role="presentation"
+          onClick={e => this.close(e)}
```
AlasdairSwan: Why not just ```onClick={this.close}```?

src/Modal/index.jsx
```diff
@@ -178,30 +188,35 @@ class Modal extends React.Component {
...
-              [styles.fade]: !open,
-              [styles['is-ie11']]: this.isIE11,
-            },
-          )}
-          role="dialog"
-          aria-modal
-          aria-labelledby={this.headerId}
-          {...(!renderHeaderCloseButton ? { tabIndex: '-1' } : {})}
-          {...(!renderHeaderCloseButton ? { ref: this.setFirstFocusableElement } : {})}
+          id="modal"
```
AlasdairSwan: Do we have to use hard-coded ids? It makes me slightly nervous given that ids need to be unique.

src/Modal/index.jsx
```diff
@@ -178,30 +188,35 @@ class Modal extends React.Component {
...
+            aria-labelledby={this.headerId}
+            {...(!renderHeaderCloseButton ? { tabIndex: '-1' } : {})}
+            {...(!renderHeaderCloseButton ? { ref: this.setFirstFocusableElement } : {})}
+          >
+            <div className={styles['modal-content']}>
+              <div className={styles['modal-header']}>
+                <h2 className={styles['modal-title']} id={this.headerId}>{this.props.title}</h2>
+                { renderHeaderCloseButton &&
                 <Button
+                  id="paragonHeaderCloseModalButton"
```
AlasdairSwan: Do we have to use hard-coded ids? It makes me slightly nervous given that ids need to be unique.

src/Modal/index.jsx
```diff
@@ -178,30 +188,35 @@ class Modal extends React.Component {
...
         >
-          <div className={styles['modal-content']}>
-            <div className={styles['modal-header']}>
-              <h2 className={styles['modal-title']} id={this.headerId}>{this.props.title}</h2>
-              { renderHeaderCloseButton &&
+          <div
+            className={classNames({
+              [styles['modal-dialog']]: open,
+            })}
+            styles={{ zIndex: '2000' }}
```
AlasdairSwan: Why use inline styles here instead of stylesheets? It limits the ability for sites with less sane ```z-index``` practices to customize the modal's CSS.

reillz10: sorry, i don't actually even need this at all. Good catch

src/Modal/index.jsx
```diff
@@ -10,7 +10,9 @@ import Icon from '../Icon';
 import newId from '../utils/newId';
 import Variant from '../utils/constants';
 
-const closeModalButtonId = 'paragonCloseModalButton';
+const closeModalButtonId = newId('paragonCloseModalButton');
```
AlasdairSwan: @reillz10 to distinguish this from the ```headerCloseModalButtonId``` it may be worth updating to ```footerCloseModalButtonId```, or moving to classes so it doesn't matter the placement of the close button, just that it is a close button.

PR #364, opened 2018-11-21
fix(pagination): make styles consistently scoped

SSPJ:
The `.disabled` class in [renderPreviousButton()](https://github.com/edx/paragon/blob/master/src/Pagination/index.jsx#L208) works, whereas the one in renderNextButton() is not scoped correctly. This effects people using the static stylesheet who are trying to target those elements for custom styling, e.g. the back arrow will grey out, but the forward one won't. This PR makes them consistent and also adds scoping to `.sr-only` in the Pagination component similar to how it is in the Icon component.

**Dependencies**: None

**Merge deadline**: None

**Testing instructions**:

1. `npm run test` verifies that this PR does not change any rendered component.

**Reviewers**
- [ ] @mtyaka
- [ ] edX reviewer[s] TBD

### DIFF
--- a/src/Pagination/index.jsx
+++ b/src/Pagination/index.jsx
@@ -203,7 +203,7 @@ class Pagination extends React.Component {
     return (
       <li
         className={classNames(
-          'page-item',
+          styles['page-item'],
           {
             [styles.disabled]: isFirstPage,
           },
@@ -258,7 +258,7 @@ class Pagination extends React.Component {
         className={classNames(
           styles['page-item'],
           {
-            disabled: isLastPage,
+            [styles.disabled]: isLastPage,
           },
         )}
       >
@@ -292,7 +292,7 @@ class Pagination extends React.Component {
     const { buttonLabels, pageCount } = this.props;
     return (
       <div
-        className="sr-only"
+        className={classNames(styles['sr-only'])}
         aria-live="polite"
         aria-relevant="text"
         aria-atomic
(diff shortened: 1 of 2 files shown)


natabene:
@georgebabey Could you take a look when you have a chance?

edx-webhook:
Thanks for the pull request, @SSPJ! I've created [OSPR-2817](https://openedx.atlassian.net/browse/OSPR-2817) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

Feel free to add as much of the following information to the ticket:
- supporting documentation
- edx-code email threads

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

src/Pagination/index.jsx
```diff
@@ -292,7 +292,7 @@ class Pagination extends React.Component {
     const { buttonLabels, pageCount } = this.props;
     return (
       <div
-        className="sr-only"
+        className={classNames(styles['sr-only'])}
```
adamstankiewicz: Is it necessary to use `classNames` here, instead of assigning `styles['sr-only']` to the `className` prop? E.g.:

```jsx
<div
  className={styles['sr-only']}
  ...
/>

SSPJ: Hi @adamstankiewicz . Not at all. I've removed it and verified that it still works.

SSPJ:
Thanks for the review, @adamstankiewicz !

PR #365, opened 2018-11-27
adding openedx file

mduboseedx:
I'd like to add openedx.yaml as pull requests from the community are coming in and this file will allow some of our tools to function.

Is edx/fedx-team an appropriate "owner" for this repo? Will this repo be part of the Open edX release? If so, is master the correct branch?


openedx.yaml
```diff
@@ -0,0 +1,8 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# https://open-edx-proposals.readthedocs.io/en/latest/oep-0002-bp-repo-metadata.html
+
+nick: para
+oeps: {}
+openedx-release: {ref: master}
```
nedbat: I think we don't need this line.

PR #366, opened 2018-12-04
fix(dropdown): fixed DropDown focus on page load

HammadAhmadWaqas:
DropDown was getting **focused** for some unwanted scenarios, 
For example: on page load, if any parent component of DropDown updated, DropDown was getting focus,  as updates of DropDown was also getting called that was causing DropDown to get focus.  

Issue reference: [Story](https://openedx.atlassian.net/browse/WL-1828)

### DIFF
--- a/src/Dropdown/Dropdown.test.jsx
+++ b/src/Dropdown/Dropdown.test.jsx
@@ -22,7 +22,7 @@ const menuOpen = (isOpen, wrapper) => {
 
 describe('<Dropdown />', () => {
   describe('renders', () => {
-    const wrapper = shallow(<Dropdown {...props} />);
+    const wrapper = mount(<Dropdown {...props} />);
     const menu = wrapper.find('.dropdown-menu');
     const button = wrapper.find('Button');
 
@@ -35,6 +35,12 @@ describe('<Dropdown />', () => {
 
     it('with menu closed', () => {
       menuOpen(false, wrapper);
+    });
+
+    it('does not get focus when updated', () => {
+      const activeElementHtml = document.activeElement.outerHTML;
+      wrapper.instance().forceUpdate();
+      expect(activeElementHtml).toEqual(document.activeElement.outerHTML);
     });
   });
 
@@ -168,6 +174,8 @@ describe('<Dropdown />', () => {
 
       it('does not toggle with invalid key', () => {
         wrapper = mount(<Dropdown {...props} />);
+        // resetting focus
+        wrapper.find('Button').getDOMNode().focus();
 
         menuOpen(false, wrapper);
         // open and close button to get focus on button
--- a/src/Dropdown/index.jsx
+++ b/src/Dropdown/index.jsx
@@ -43,10 +43,10 @@ class Dropdown extends React.Component {
     }
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps, prevState) {
     if (this.state.open) {
       this.menuItems[this.state.focusIndex].focus();
-    } else if (this.toggleElem) {
+    } else if (prevState.open && this.toggleElem) {
       this.toggleElem.focus();
     }
   }


src/Dropdown/Dropdown.test.jsx
```diff
@@ -36,6 +36,12 @@ describe('<Dropdown />', () => {
     it('with menu closed', () => {
       menuOpen(false, wrapper);
     });
+
+    it('no focus on just update', () => {
```
ziafazal: name should be something like this `does not get focus when updated`

src/Dropdown/Dropdown.test.jsx
```diff
@@ -36,6 +36,12 @@ describe('<Dropdown />', () => {
...
+      expect(document.body.outerHTML).toEqual(document.activeElement.outerHTML);
```
ziafazal: Can we simplify it like this
```
const activeElementId = document.activeElement.id;
wrapper.instance().forceUpdate();
expect(activeElementId).toEqual(document.activeElement.id);
```

src/Dropdown/Dropdown.test.jsx
```diff
@@ -22,7 +22,7 @@ const menuOpen = (isOpen, wrapper) => {
 
 describe('<Dropdown />', () => {
   describe('renders', () => {
-    const wrapper = shallow(<Dropdown {...props} />);
+    const wrapper = mount(<Dropdown {...props} />);
```
ziafazal: why we have to replace `shallow` with `mount` here?

src/Dropdown/Dropdown.test.jsx
```diff
@@ -168,6 +174,8 @@ describe('<Dropdown />', () => {
 
       it('does not toggle with invalid key', () => {
         wrapper = mount(<Dropdown {...props} />);
+        // resetting focus
+        wrapper.find('Button').getDOMNode().focus();
```
ziafazal: why we have to make this change in existing test?

src/Dropdown/Dropdown.test.jsx
HammadAhmadWaqas: `document.activeElement.id` returns `""` so, that will not verify correctly.
Instead we can write this:
`     const activeElementHtml = document.activeElement.outerHTML;
      wrapper.instance().forceUpdate();
      expect(activeElementHtml).toEqual(document.activeElement.outerHTML);
`

src/Dropdown/Dropdown.test.jsx
HammadAhmadWaqas: :+1:

src/Dropdown/Dropdown.test.jsx
HammadAhmadWaqas: `shallow` don't call `componentDidUpdate`

there is [difference](https://gist.github.com/fokusferit/e4558d384e4e9cab95d04e5f35d4f913)

src/Dropdown/Dropdown.test.jsx
HammadAhmadWaqas: this test started failing on [line](https://github.com/edx/paragon/blob/master/src/Dropdown/Dropdown.test.jsx#L180)  after updating code in this PR.
**Question**: Why it was running before?
**Answer**: Before starting the test `does not toggle with invalid key` `'Button'` lost focus.
Our previous DropDown code was focusing Button on every update and `Button` was getting focus automatically before [line](https://github.com/edx/paragon/blob/master/src/Dropdown/Dropdown.test.jsx#L180) .
As now Button Don't get focus automatically we had to reset focus.

src/Dropdown/index.jsx
```diff
@@ -43,10 +43,10 @@ class Dropdown extends React.Component {
     }
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps, prevState) {
     if (this.state.open) {
       this.menuItems[this.state.focusIndex].focus();
-    } else if (this.toggleElem) {
+    } else if (prevState.open && this.toggleElem) {
```
adamstankiewicz: More of a clarifying question, but why are we checking the previous `open` state here instead of the current `open` state?

HammadAhmadWaqas: @adamstankiewicz we already defining `this.state.open` behavior in above `if` condition.
I believe [this.toggleElem.focus();](https://github.com/edx/paragon/blob/f9076fbc0d8d8b167c9c93d2038435339b4bcac7/src/Dropdown/index.jsx#L50) should run only if we just closed the `Dropdown`, to focus the `Button` Again. 
that is way we are adding extra check `prevState.open`

PR #380, opened 2019-01-09
fix(inputselect): add aria-label

jansenk:
add aria-label to InputSelect

### DIFF
--- a/src/InputSelect/InputSelect.stories.jsx
+++ b/src/InputSelect/InputSelect.stories.jsx
@@ -104,6 +104,7 @@ storiesOf('InputSelect', module)
     <InputSelect
       name="fruits"
       label="Fruits"
+      ariaLabel="Fruits"
       value="strawberry"
       options={[
         'apple',
--- a/.storybook/__snapshots__/Storyshots.test.js.snap
+++ b/.storybook/__snapshots__/Storyshots.test.js.snap
@@ -22807,6 +22807,7 @@ exports[`Storyshots InputSelect basic usage 1`] = `
           </label>
           <select
             aria-describedby="error-asInput20"
+            aria-label={undefined}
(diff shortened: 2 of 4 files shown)


src/InputSelect/InputSelect.stories.jsx
```diff
@@ -104,6 +104,7 @@ storiesOf('InputSelect', module)
     <InputSelect
       name="fruits"
       label="Fruits"
+      ariaLabel="Fruits"
```
adamstankiewicz: Is there a reason why we don't just use the value passed into the `label` prop as the `aria-label`, instead of introducing another prop?

jansenk: We have a label that's acting as the label for two separate, related selects.
@wittjeff suggested this since while we don't need to have additional independent labels for the two selects, we do need to have them for screen readers.

screen shot 2019-01-10 at 9 45 00 am

adamstankiewicz: Ahh gotcha, that makes sense then 👍

wittjeff: If you're suggesting that this could be implemented as a flag (label is visible or not) then I think that's a good question.  In the Accessible Name calculation hierarchy, aria-label overrides a label element. So if for some reason the visible (or even class=sr) label is not ideal or not friendly, you could "correct" it with aria-label.  I'm having trouble thinking of a good exemplar though.

wittjeff: We should generally be biased in favor of visible label. Our outside auditor asserts that this is a WCAG conformance requirement but I don't read so strictly in the spec for some cases such as this one, where I think the context will not be lost when interacting with the active element.

PR #382, opened 2019-01-11
fix(table): add ability to specify row header

jansenk:
Allow users to specify a column to act as a row header


jansenk:
@reillz10 @adamstankiewicz any chance someone could take a look at this today?

PR #383, opened 2019-02-04
Added required attr passthrough to InputSelect

staubina:


src/InputSelect/index.jsx
```diff
@@ -59,6 +59,7 @@ class Select extends React.Component {
         onBlur={props.onBlur}
         ref={props.inputRef}
         disabled={props.disabled}
+        required={props.requires}
```
albemarle: Should be `props.required`, I think?

staubina:
Passes locally.  It looks like Master is failing to pass tests?

PR #386, opened 2019-02-11
Creation of three new ADRs for paragon

davidjoy:
tl;dr -

1. Arch team will take "ownership" of Paragon.  (this is more an offer of coordination than anything else)
2. We use Bootstrap.
3. We will start using Reactstrap.

All three are obviously proposals - exact details are up for discussion.  None of this is expected to be controversial or ruffle any feathers - if it does, let's talk!  I must not have talked to you about it yet.

I intend to bring this up for discussion at the next Fedx meeting to solicit feedback.

### DIFF
--- a/docs/decisions/0003-repository-ownership.rst
+++ b/docs/decisions/0003-repository-ownership.rst
@@ -1,0 +1,38 @@
+3. Repository Ownership
+--------------------
+
+Status
+------
+
+Proposed
+
+Context
+-------
+
+Paragon components are currently in use in a variety of production IDAs and continue to be updated, improved, and extended by engineers throughout the organization. At the same time, the repository itself lacks explicit ownership and review processes from potential stakeholders, such as architecture, product, accessibility, and user experience. This combination of a frequently updated, mission critical library without consistent oversight is potentially a significant risk to the stability of our front ends.
+
+Meanwhile, the architecture team has been tasked with improving our codebase's overall approachability and extensibility, as well as our time-to-market for new features and applications, and has been explicitly looking at Paragon and our usage of bootstrap with respect to creating a comprehensive design system for the organization.
+
+Decision
+--------
+
+The architecture team will own Paragon to the extent that it will:
+
+- With the help of UX, accessibility, and product, document the requirements and best practices for updates and additions to Paragon's reusable components. ("I want to help you help me help you.")
+- Ensure that UX, accessibility, and product are given the opportunity to vett and sign off on changes to the library. ("Going once, going twice...")
+- Review Paragon pull requests with an eye toward ensuring that Paragon continues to be a flexible, consistent library of reusable components that fulfill our needs across the organization. ("Have you considered...")
+- Lead the charge on improving and fleshing out Paragon with the above goals in mind. ("Here, use this!")
+
+Consequences
+------------
+
+Establishing clear ownership roles and review criteria for Paragon should help ensure that code quality remains high, and that changes to the library are less likely to result in bugs and unexpected breaking changes. In addition, this ownership structure dovetails well with the architecture team's stated goal of improving our front end systems' approachability, extensibility, and the time to market of IDAs built on Paragon.
+
+This ownership will, of course, take up some of the architecture team's time above and beyond their current responsibilities. That said, documenting and disseminating the process and considerations for updating Paragon should help alleviate some of that strain by allowing pull request authors to do much of the due diligence themselves.
+
+References
+----------
+
+* https://github.com/edx/paragon/graphs/code-frequency (Paragon is in active development)
+* https://openedx.atlassian.net/wiki/spaces/AC/pages/787087363/Architecture+Roadmap+2018-19#ArchitectureRoadmap,2018-19-ArchitectureOGSP
+* https://www.designbetter.co/design-systems-handbook
--- a/docs/decisions/0003-repository-ownership.rst
+++ b/docs/decisions/0003-repository-ownership.rst
@@ -18,8 +18,8 @@ Decision
 
 The architecture team will own Paragon to the extent that it will:
 
-- With the help of UX, accessibility, and product, document the requirements and best practices for updates and additions to Paragon's reusable components. ("I want to help you help me help you.")
-- Ensure that UX, accessibility, and product are given the opportunity to vett and sign off on changes to the library. ("Going once, going twice...")
+- With the help of UX, Accessibility, and Product, document the requirements and best practices for updates and additions to Paragon's reusable components. ("I want to help you help me help you.")
+- Ensure that UX, Accessibility, and Product are given the opportunity to vet and sign off on changes to the library. ("Going once, going twice...")
 - Review Paragon pull requests with an eye toward ensuring that Paragon continues to be a flexible, consistent library of reusable components that fulfill our needs across the organization. ("Have you considered...")
 - Lead the charge on improving and fleshing out Paragon with the above goals in mind. ("Here, use this!")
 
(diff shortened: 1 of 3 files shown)


docs/decisions/0003-repository-ownership.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Meanwhile, the architecture team has been tasked with improving our codebase's overall approachability and extensibility, as well as our time-to-market for new features and applications, and has been explicitly looking at Paragon and our usage of bootstrap with respect to creating a comprehensive design system for the organization.
+
+Decision
+--------
+
+The architecture team will own Paragon to the extent that it will:
+
+- With the help of UX, accessibility, and product, document the requirements and best practices for updates and additions to Paragon's reusable components. ("I want to help you help me help you.")
+- Ensure that UX, accessibility, and product are given the opportunity to vett and sign off on changes to the library. ("Going once, going twice...")
```
albemarle: s/vett/vet/

docs/decisions/0003-repository-ownership.rst
```diff
@@ -0,0 +1,38 @@
...
+Paragon components are currently in use in a variety of production IDAs and continue to be updated, improved, and extended by engineers throughout the organization. At the same time, the repository itself lacks explicit ownership and review processes from potential stakeholders, such as architecture, product, accessibility, and user experience. This combination of a frequently updated, mission critical library without consistent oversight is potentially a significant risk to the stability of our front ends.
+
+Meanwhile, the architecture team has been tasked with improving our codebase's overall approachability and extensibility, as well as our time-to-market for new features and applications, and has been explicitly looking at Paragon and our usage of bootstrap with respect to creating a comprehensive design system for the organization.
+
+Decision
+--------
+
+The architecture team will own Paragon to the extent that it will:
+
+- With the help of UX, accessibility, and product, document the requirements and best practices for updates and additions to Paragon's reusable components. ("I want to help you help me help you.")
```
albemarle: Maybe capitalize Accessibility and Product?

docs/decisions/0005-usage-of-reactstrap.rst
```diff
@@ -0,0 +1,55 @@
...
+------
+
+Proposed
+
+Context
+-------
+
+Paragon components exist for a variety of use cases, are React-based, and are currently styled using Bootstrap-compatible CSS class names. All known usages of Paragon are backed by Bootstrap. Reactstrap is a React-based component library that provides implementations of Bootstrap's components.
+
+Paragon's strength is found in its ability to be modifyied to meet our accessibility and functionality requirements above and beyond what any third-party component library might contain.  In addition, Paragon is the home for  edX-specific reusable components that would never be found in a generic component library.
```
albemarle: s/modifyied/modified/

docs/decisions/0005-usage-of-reactstrap.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Reactstrap's strength is found in its consistent, flexible, and comprehensive components (it's API, effectively) and its more-or-less complete set of implementations of Bootstrap's various 'component' styles.
```
albemarle: it's -> its

docs/decisions/0005-usage-of-reactstrap.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Both have something to offer us.
+
+Decision
+--------
+
+We will use Reactstrap to opportunistically extend and improve the functionality of Paragon.  This implies that some existing Reactstrap components will be exposed via Paragon directly if they meet our needs, effectively extending Paragon's component library with minimal effort.
+
+This will be accomplished via one of several methods depending on the nature and utility of particular components. For the purposes of this list, 'production-ready' means a component that would pass stakeholder review as described in ADR 0003 - Repository Ownership.
```
albemarle: Could link to ADR-3 here.

docs/decisions/0005-usage-of-reactstrap.rst
```diff
@@ -0,0 +1,55 @@
...
+
+For Reactstrap components:
+
+- Reactstrap components which are deemed production-ready will be exposed as an export of Paragon as-is, making them immediately available as "paragon components" in consuming front-ends.
```
albemarle: s/paragon/Paragon/

docs/decisions/0005-usage-of-reactstrap.rst
```diff
@@ -0,0 +1,55 @@
...
+------
+
+Proposed
+
+Context
+-------
+
+Paragon components exist for a variety of use cases, are React-based, and are currently styled using Bootstrap-compatible CSS class names. All known usages of Paragon are backed by Bootstrap. Reactstrap is a React-based component library that provides implementations of Bootstrap's components.
+
+Paragon's strength is found in its ability to be modified to meet our accessibility and functionality requirements above and beyond what any third-party component library might contain.  In addition, Paragon is the home for  edX-specific reusable components that would never be found in a generic component library.
```
nasthagiri: I'm not suggesting that we remove the `edx-specific` qualifier here, but I'm wondering what's an example of something in Paragon today that's edx-specific?  I do imagine that we will be creating reusable _organisms_ that are edx-specific - but we had been thinking of not including those in Paragon itself and keeping Paragon for low-level _atoms_ and _molecules_ only.  I realize that such a decision (however we decide) will need a separate ADR.  But in the meantime, do we already have edx-specific components in Paragon today?

docs/decisions/0005-usage-of-reactstrap.rst
```diff
@@ -0,0 +1,55 @@
...
+  - They can be augmented via a custom wrapper in Paragon, improving their functionality and making them production-ready and exportable.
+  - They can be improved via a pull request to Reactstrap - more likely if straightforward accessibility bugs are found, for instance - and included once Reactstrap updates. Obviously, this approach can be done in parallel with using a custom wrapper in some cases.
+
+For existing Paragon components:
+
+- Paragon components that are deemed to be production-ready can be left as-is for now without backing or replacing them with Reactstrap implementations. No need to change a tire that isn't flat.
+- Paragon components with accessibility, flexibility, or consistency issues that could be mitigated by backing them with Reactstrap can be opportunistically updated. The extent of this change - and whether it results in a major, minor, or patch version update, largely depends on how different the Paragon/Reactstrap components are.  Where possible, we will favor the best practices and functionality of the Reactstrap components, under the assumption that they are generally more flexible and consistent than our own home-grown components.
```
nasthagiri: Question: Do we have a way to tell which legacy Paragon components aren't up to par and should be updated/replaced?

docs/decisions/0005-usage-of-reactstrap.rst
davidjoy: It's a fair point.  When I wrote it I was assuming there were some paradigms/features that found their way into Paragon that were fulfilling particular design requirements on our site.  Looking around paragon in more detail, I can't really find anything that fits that bill at the moment.

I like the distinction as you frame it.  If we simply say that any "edx-specific components" are really organisms _because_ they're specific to our business and our particular domain of data, then that excludes them from Paragon by definition.

As an example, a reusable "certificate card" that displays a cert that a user has earned would then be classified as an organism and (practically speaking) would then need its own repository or would be pushed into some non-paragon repo.  

I think that holds together.  Yes?  (in which case I _will_ remove that language above)

docs/decisions/0005-usage-of-reactstrap.rst
davidjoy: Not yet.  I think the arch ownership thing is part of determining that.  Someone needs to coordinate with the interested parties get their feedback, as well as examine our own component interfaces to see if we think they warrant revisiting.

PR #387, opened 2019-02-22
fix(tests): tests modal error throwing in a supported way

davidjoy:
No need for an error boundary component.  Uses “shallow” to avoid full mounting in jsdom, which is what spits out an error to the console.


albemarle:
Hooray!

cheerleader cat

davidjoy:
All PRs should get kittens!

PR #394, opened 2019-03-15
feat: major upgrades of many packages and configuration

abutterworth:
This is a very large set of updates. We should verify everything is still configured as needed. The themeable and static builds still build as expected.


.npmignore
```diff
@@ -1,12 +1,9 @@
 .DS_Store
-.eslintcache
 .storybook
-node_modules
```
matthugs: Why'd we add this back into the files sent to npm on publish?

Larger question I'm wondering about: can we switch to the whitelist approach described in link.medium.com/FcQryrm4bV, while we're in here mucking with which files get packed up?

.travis.yml
```diff
@@ -1,15 +1,14 @@
...
-- 6
 cache:
   directories:
   - "~/.npm"
 notifications:
   email:
     recipients:
-    - arizzitano@edx.org
-    - jbradley@edx.org
+    - abutterworth@edx.org
```
matthugs: we can fairly easily add a google group recipient here if you'd like to not be the only one feeling the spam (should spam arise)

package.json
```diff
@@ -2,21 +2,22 @@
...
   "scripts": {
-    "build": "NODE_ENV=production webpack --config webpack.config.js",
+    "build": "webpack --mode production",
     "build-storybook": "build-storybook",
+    "commit": "commit",
     "coveralls": "cat ./coverage/lcov.info | coveralls",
     "debug-test": "node --inspect-brk node_modules/.bin/jest --runInBand --coverage",
     "deploy-storybook": "storybook-to-ghpages",
     "deploy-storybook-ci": "storybook-to-ghpages --ci",
     "gc": "commit",
```
matthugs: can we get rid of this alternative alias? This name I find very counterintuitive

package.json
```diff
@@ -26,80 +27,71 @@
...
+    "source-map-loader": "^0.2.4",
+    "storybook-readme": "^4.0.5",
+    "style-loader": "^0.23.1",
+    "travis-deploy-once": "^5.0.11",
+    "uglifyjs-webpack-plugin": "^2.1.2",
+    "webpack": "^4.29.6",
+    "webpack-cli": "^3.3.0"
   },
   "jest": {
+    "testURL": "http://localhost/",
```
matthugs: this appears to be the default https://jestjs.io/docs/en/configuration.html#testurl-string

.storybook/addons.js
```diff
@@ -1,4 +1,2 @@
-import '@storybook/addon-options/register';
 import '@storybook/addon-actions/register';
 import '@storybook/addon-a11y/register';
```
matthugs: It looks like we removed all the places the corresponding decorator is used. Should we strip this out too, or are we planning on adding them back in piecemeal?

src/ListBox/ListBox.stories.jsx
```diff
@@ -145,7 +142,7 @@ storiesOf('ListBox', module)
     </ListBox>
   ))
   .add('using tag prop', () => (
-    <React.Fragment>
+    <div>
```
matthugs: nit: why?

src/ListBox/index.jsx
```diff
@@ -104,9 +103,9 @@ export default class ListBox extends React.Component {
 }
 
 ListBox.propTypes = {
-  children: or([
-    elementType(ListBoxOption),
-    PropTypes.arrayOf(elementType(ListBoxOption)),
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.element),
+    PropTypes.element,
```
matthugs: Consider also: `PropTypes.node`

src/Modal/index.jsx
```diff
@@ -139,21 +139,17 @@ class Modal extends React.Component {
...
-      buttonElement = (<Button
-        {...buttonProps}
-        key={i}
-        onKeyDown={this.handleKeyDown}
-      />);
-
-      return buttonElement;
+    return this.props.buttons.map((button) => {
+      // button is either a Button component that we want clone or a set of props
+      const buttonProps = button.type === Button ? button.props : button;
```
matthugs: Why clone at all? I feel like this adds the kind of react hierarchy-indirection that makes me nervous after ADR-2 in this repo https://github.com/edx/paragon/blob/master/docs/decisions/0002-react-i18n.rst

.travis.yml
abutterworth: Good call I'll make that happen

.storybook/addons.js
abutterworth: The new version of this addon doesn't need to be added with addDecorator – it seems this line is all you need to get set up.

src/Modal/index.jsx
abutterworth: It seems that this was made to handle these two different ways of providing buttons. I agree we shouldn't do what we are here, but for the sake of backward compatibility I didn't want to change anything. This was more just cleaning up some code that I found hard to read.

.npmignore
abutterworth: Great! I replaced this with the whitelist

package.json
abutterworth: I think I added this due to an issue with Storyshots. I can remove and see if it breaks anything.

package.json
abutterworth: removed.

src/ListBox/index.jsx
abutterworth: Thanks that's better.

src/ListBox/ListBox.stories.jsx
abutterworth: Good catch – no good reason.

src/Modal/index.jsx
matthugs: My jaw would drop if any downstream users of this component relied on the behavior of us not using the same buttons they pass in, but prop-clones of the same.

But I can understand the desire to push off this concern; should I make a note of it to drop in the backlog once we have one set up?

package.json
matthugs: Sorry I was unclear about which line I was pointing to in github's interface; it's much clear when adding a comment

Your name makes loads more sense to me than `npm run gc`. It's a pretty bad false-cognate with `git gc`

src/Modal/index.jsx
abutterworth: Yes please do.

package.json
abutterworth: No problems removing it 👍

package.json
abutterworth: Got it, changed.

src/Modal/index.jsx
matthugs: This ended up being mooted in https://github.com/edx/paragon/pull/435 by the elimination of these buttons from the modal component's prop interface. Marking as resolved.

PR #398, opened 2019-03-20
feat: remove static build output

abutterworth:
Eliminating the static build target provides us a simpler project setup and sets the stage to eliminate css modules and some of the problems caused by its current implementation.

The static build is used only in some places in edx-platform:
https://github.com/search?q=org%3Aedx+%22paragon%2Fstatic%22&type=Code

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,3 @@
 coverage/*
 dist/
-themeable/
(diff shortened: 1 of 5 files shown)


abutterworth:
@robrap yes I do thank you

PR #399, opened 2019-03-20
feat: remove css module and namespace support

abutterworth:
BREAKING CHANGE: This changes the way scss should be imported with components. Rather than a SCSS file for each module, this prefers that consumers import the scss file (src/index.scss). It must be included after bootstrap variables are set. It's also up to the consumer to include font-awesome if needed. 

## ADR
Please read the ADR within this PR: https://github.com/edx/paragon/pull/399/files#diff-787f948bcc1ce703268ef23a813b8d16

## Usage (from the README)

Paragon requires React 16 or higher. To install Paragon into your project:

```
npm i --save @edx/paragon
```

In your JSX files:
```
import { ComponentName } from '@edx/paragon';
```

Include in your SCSS file:
```
```

```
```


src/Pagination/index.jsx
```diff
@@ -203,23 +200,23 @@ class Pagination extends React.Component {
...
+          'page-item',
           {
-            [styles.disabled]: isFirstPage,
+            disabled: isFirstPage,
           },
         )}
       >
         <Button
-          className={['previous', styles['page-link']]}
+          className={['previous', 'page-link']}
```
davidjoy: Same thing here.

davidjoy: I don't see anything in the react doc about className accepting an array...

webpack.config.js
```diff
@@ -53,15 +56,11 @@ module.exports = {
...
-            options: {
-              modules: true,
-              localIdentName: '[local]',
-            },
           },
           {
             loader: 'sass-loader',
             options: {
-              data: '@import "paragon-reset";',
+              data: '@import "bootstrap-variables";',
```
davidjoy: Is there a reason to do this, rather than just manually import those three imports in bootstrap-variables.scss into index.scss?  Not sure I understand the value of this `data` thing.

src/Collapsible/index.jsx
```diff
@@ -4,7 +4,6 @@ import PropTypes from 'prop-types';
 import Button from '../Button';
 import Icon from '../Icon';
 
-import styles from './Collapsible.scss';
```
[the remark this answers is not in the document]
abutterworth: We don't need to pull in styles for CSS module mapping anymore.

Now preferring that consumers of Paragon include ./src/index.scss in their scss files. They can still include specific scss files if they prefer, but I deleted all the scss files that had no purpose other than providing javascript with css module names.

src/Pagination/index.jsx
```diff
@@ -120,15 +117,15 @@ class Pagination extends React.Component {
...
           {
-            [styles.active]: active,
+            active,
           },
         ])}
         key={page}
       >
         <Button
-          className={[styles['page-link']]}
+          className={['page-link']}
```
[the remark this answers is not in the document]
abutterworth: Button itself uses the classnames package and expects an array for className. We will be changing this shortly!

webpack.config.js
abutterworth: The value here is that when building the minified scss we need bootstrap functions, variables, and mixins to be set, so this loader adds them. 

We still want index.scss to be a clean partial so that consuming applications can determine how or which bootstrap theme variables to include.

PR #400, opened 2019-03-21
Update packages, simplify build targets, remove css modules

abutterworth:
# Paragon 4.0.0

**This PR removes support for the static, css-namespaced build that is used in edx-platform.** See all uses here: https://github.com/search?q=org%3Aedx+%22paragon%2Fstatic%22&type=Code
In letting go of this requirement the following changes are contained in this release:

### CSS Modules have been removed.
SCSS is no longer imported into Paragon component JSX files and those that no longer serve a purpose have been deleted.

### The way to include Paragon SCSS has changed
Include in your SCSS file:
```
```

```
```

### You will see minor changes in your snapshot tests
There are no breaking changes in the usage of Paragon components, but there are a couple minor changes in html output. 
- null values will be present on scope attributes in tables (this seems to be a change in the way React handles null values)
- Dropdowns with icons have a className that is no longer outputted incorrectly as an array.


src/Breadcrumb/Breadcrumb.stories.jsx
```diff
@@ -1,10 +1,5 @@
-/* eslint-disable import/no-extraneous-dependencies */
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-import centered from '@storybook/addon-centered';
-import { checkA11y } from '@storybook/addon-a11y';
-import { withInfo } from '@storybook/addon-info';
-import { withReadme } from 'storybook-readme';
```
AlasdairSwan: @abutterworth why are these being removed? Because of the Storybook version update?

src/Breadcrumb/index.jsx
```diff
@@ -12,20 +11,20 @@ const Breadcrumbs = ({
 
   return (
     <nav aria-label="breadcrumb">
-      <ol className={classNames(styles['list-inline'])}>
+      <ol className={classNames('list-inline')}>
```
AlasdairSwan: Why not just ```<ol className="list-inline">```?

src/Breadcrumb/Breadcrumb.stories.jsx
abutterworth: We had configured the same storybook decorators on each story in the past. I don’t know if it’s new with the update but you can set the default configuration for all of them in the storybook config.

src/Breadcrumb/index.jsx
abutterworth: It should be. You're right let's get this cleaned up in subsequent improvements.

PR #404, opened 2019-03-22
feat: publish major upgrades of many packages and configuration

abutterworth:
BREAKING CHANGE: This removes the static build of Paragon.

# Paragon 4.0.0

**This PR removes support for the static, css-namespaced build that is used in edx-platform.** See all uses here: https://github.com/search?q=org%3Aedx+%22paragon%2Fstatic%22&type=Code
In letting go of this requirement the following changes are contained in this release:

### CSS Modules have been removed.
SCSS is no longer imported into Paragon component JSX files and those that no longer serve a purpose have been deleted.

### The way to include Paragon SCSS has changed
Include in your SCSS file:
```
```

```
```

### You will see minor changes in your snapshot tests
There are no breaking changes in the usage of Paragon components, but there are a couple minor changes in html output. 
- null values will be present on scope attributes in tables (this seems to be a change in the way React handles null values)
- Dropdowns with icons have a className that is no longer outputted incorrectly as an array.


abutterworth:
Two PR reviews cover code contained in this commit:

#398 feat: remove static build output
#399 feat: remove css module and namespace support

PR #407, opened 2019-04-01
Add rel="noopener" for new window Hyperlinks

### LINKED ISSUE #401: Help avoid reverse tabnabbing in Hyperlink
https://www.owasp.org/index.php/Reverse_Tabnabbing

Basically, with links that are target=_blank, we are open to a certain kind of phishing attack. The recommended solution for this is rel=noopener [1].

It would be kind of Hyperlink to add that automatically (either always, or only when using an _blank target).

[1] https://mathiasbynens.github.io/rel-noopener/

abutterworth:
Fixes #401

### DIFF
--- a/.storybook/__snapshots__/Storyshots.test.js.snap
+++ b/.storybook/__snapshots__/Storyshots.test.js.snap
@@ -1258,6 +1258,7 @@ exports[`Storyshots HyperLink with blank target 1`] = `
 <a
   href="https://www.edx.org"
   onClick={[Function]}
+  rel="noopener noreferrer"
   target="_blank"
 >
   edX.org
@@ -1296,6 +1297,7 @@ exports[`Storyshots HyperLink with onClick 1`] = `
 <a
   href="https://www.edx.org"
   onClick={[Function]}
+  rel="noopener noreferrer"
   target="_blank"
 >
   edX.org
@@ -3127,6 +3129,7 @@ exports[`Storyshots MailtoLink with blank target 1`] = `
 <a
   href="mailto:edx@example.com"
   onClick={null}
+  rel="noopener noreferrer"
   target="_blank"
 >
   edx@example.com
@@ -3166,6 +3169,7 @@ exports[`Storyshots MailtoLink with onClick 1`] = `
 <a
   href="mailto:edx@example.com"
   onClick={[Function]}
+  rel="noopener noreferrer"
   target="_blank"
 >
   edx@example.com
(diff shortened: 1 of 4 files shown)


abutterworth:
Only adding noopener makes sense to me. Also maybe it should only set `rel` if none is supplied?

```
if (target === '_blank') {
    // Add this rel attribute to prevent Reverse Tabnabbing
    other.rel = other.rel || 'noopener';
```

mikix:
I like `other.rel || 'noopener';` but lets be real - no one wants `opener` and people might accidentally not realize they should add it. Maybe `other.rel + " noopener"`?

abutterworth:
I guess I was thinking that if you specify `rel` you're taking responsibility into your own hands. Should we allow full control, or force the addition of noopener (taking the attitude that if you wanted explicitly to have an opener then you wouldn't use this component)?

mikix:
I hear ya - but I guess I'm coming at this from a "the spec is broken by even allowing this stupid phishing attack in the first place".

Talk to me about the use case of a Hyperlink user that doesn't want `noopener`. And also our institutional tolerance for allowing that person to take user security into their own hands.

abutterworth:
Correct me if I'm wrong, but reverse tabnabbing is only an issue when it opens cross domain (and to a malicious site). An opener in a window is a potentially useful tool though not one I imagine we will take advantage of. A use case (maybe this is too far out there to bother discussing deeply):

I want to make an in-course tool (maybe a timer, formula sheet, calculator) that I want to open in a new window so the user can manage the course and the tool as separate windows. Maybe the tool needs to be able to manipulate or notify the host page of something. They'd both be hosted on our domain (therefore safe).  

I think I'm with you on just forcing it. If you need reference to `window.opener` you're pretty in the weeds and are doing it _on purpose_.

mikix:
You're right that tabnabbing is only a problem cross-domain (unless the attacker is running JS on one of your pages, in which case you already have problems). But that _is_ a pretty common thing we do, e.g. in our forums.

Though, I'm guessing we don't convert <a> links there into <Hyperlinks> by any chance?

mikix:
Although even trusted external sites become a problem on an untrusted network for non-https links...

abutterworth:
I don't think the forums are written in React so no worries there. I've updated to force the presence of noopener. It's safer and if someone wants to go off the rails, they can write their own anchor tag.

abutterworth:
I've updated the readme. Thanks for the feedback @mikix!

PR #408, opened 2019-04-01
Add Reactstrap (WIP)

### LINKED ISSUE #234: Should we add Layout components?
See [reactstrap's Layout components](https://reactstrap.github.io/components/layout/). This would allow us to write:

```jsx
<Container>
  <Row>
    <Col>.col</Col>
  </Row>
</Container>
```

Instead of:

```jsx
<div className="container">
  <div className="row">
    <div className="col">.col</div>
  </div>
</div>
```

abutterworth:
- Adds Layout components. Fixes #234.


abutterworth:
Decided not to add Reactstrap in Paragon. We'll follow up with an ADR.

PR #415, opened 2019-04-05
Add transition replace

abutterworth:
# Transition Replace

transition-replace 2019-04-05 16_46_06

Manages a transition when replacing content. By default this component will transition the height and do a cross-fade. The transition can be customized.

TransitionReplace expects only one child at any time. Swap content inside the component (usually based on some state).

### Usage

```jsx
<TransitionReplace>
  {this.state.isEditing ? (
    <form key="the-form">
      <label htmlFor="name">First Name</label>
      <input type="text" value={name} />
      <button type="submit">Save</button>
    </form>
  ): (
    <div key="other">
      <h4>First Name</h4>
      <p>{name}</p>
    </div>
  )}
</TransitionReplace>
```

**IMPORTANT NOTE: You must provide the `key` prop for all children**, even when only rendering a single item. This is how React will determine that a child is new rather than just updated.

### Custom Transition

TransitionReplace uses [CSSTransition from the ReactTransitionGroup package](http://reactcommunity.org/react-transition-group/css-transition). The `transitionClassNames` prop is a pass-through to CSSTransitions's `classNames` prop. You can use this to change the crossfade animation. By default this value is `pgn__transition-replace` and the cross-fade is defined as below:

```css
.pgn__transition-replace-enter {
  opacity: 0;
}
.pgn__transition-replace-enter-active {
  opacity: 1;
  z-index: 1;
  transition: opacity 300ms ease;
}
.pgn__transition-replace-exit {
  opacity: 1;
}
.pgn__transition-replace-exit-active {
  opacity: 0;
  transition: opacity 300ms ease;
}
```

If you change the timing in CSS you should also match it using the `enterDuration` and `exitDuration` props.

### DIFF
--- a/src/TransitionReplace/README.md
+++ b/src/TransitionReplace/README.md
@@ -48,3 +48,9 @@ TransitionReplace uses [CSSTransition from the ReactTransitionGroup package](htt
 ```
 
 If you change the timing in CSS you should also match it using the `enterDuration` and `exitDuration` props.
+
+
+#### Note:
+
+Children are not required. When this component is empty, the a child inserted into it will not transition the height (from zero). No "replacement" transition will occur and new content will pop in like a normal insert. This behaviour makes it easier to work with asyncronously loaded content (for example: during the first load you don't have to do any special handling).
+
--- a/src/TransitionReplace/README.md
+++ b/src/TransitionReplace/README.md
@@ -15,7 +15,7 @@ TransitionReplace expects only one child at any time. Swap content inside the co
       <button type="submit">Save</button>
     </form>
   ): (
-    <div>
+    <div key="other">
       <h4>First Name</h4>
       <p>{name}</p>
     </div>
--- a/src/TransitionReplace/README.md
+++ b/src/TransitionReplace/README.md
@@ -46,3 +46,5 @@ TransitionReplace uses [CSSTransition from the ReactTransitionGroup package](htt
   transition: opacity 300ms ease;
 }
 ```
+
+If you change the timing in CSS you should also match it using the `enterDuration` and `exitDuration` props.
--- a/src/TransitionReplace/README.md
+++ b/src/TransitionReplace/README.md
@@ -1,1 +1,48 @@
 # Transition Replace
+
+Manages a transition when replacing content. By default this component will transition the height and do a cross-fade. The transition can be customized.
+
+TransitionReplace expects only one child at any time. Swap content inside the component (usually based on some state).
+
+### Usage
+
+```jsx
+<TransitionReplace>
+  {this.state.isEditing ? (
+    <form key="the-form">
+      <label htmlFor="name">First Name</label>
+      <input type="text" value={name} />
+      <button type="submit">Save</button>
+    </form>
+  ): (
+    <div>
+      <h4>First Name</h4>
+      <p>{name}</p>
+    </div>
+  )}
+</TransitionReplace>
+```
+
+**IMPORTANT NOTE: You must provide the `key` prop for all children**, even when only rendering a single item. This is how React will determine that a child is new rather than just updated.
+
+### Custom Transition
+
+TransitionReplace uses [CSSTransition from the ReactTransitionGroup package](http://reactcommunity.org/react-transition-group/css-transition). The `transitionClassNames` prop is a pass-through to CSSTransitions's `classNames` prop. You can use this to change the crossfade animation. By default this value is `pgn__transition-replace` and the cross-fade is defined as below:
+
+```css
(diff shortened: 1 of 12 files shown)


src/TransitionReplace/README.md
```diff
@@ -0,0 +1,50 @@
...
+  ): (
+    <div key="other">
+      <h4>First Name</h4>
+      <p>{name}</p>
+    </div>
+  )}
+</TransitionReplace>
+```
+
+**IMPORTANT NOTE: You must provide the `key` prop for all children**, even when only rendering a single item. This is how React will determine that a child is new rather than just updated.
```
albemarle: Is this also in the comments somewhere?  Or is there not a good place for it?

src/TransitionReplace/TransitionReplace.test.jsx
```diff
@@ -0,0 +1,80 @@
+/* eslint-disable no-plusplus, react/prop-types */
```
albemarle: Can we just take no-plusplus out of our linting?  It is not one of the rules I think is good.

src/TransitionReplace/index.jsx
```diff
@@ -0,0 +1,176 @@
...
+
+  onChildExited(htmlNode) {
+    this.setState({ height: null });
+    if (this.props.onChildExited) this.props.onChildExited(htmlNode);
+  }
+
+  renderChildTransition(child) {
+    if (!child.key && process.env.NODE_ENV === 'development') {
+      // eslint-disable-next-line no-console
+      console.warn("TransitionReplace: A child is missing a 'key' prop. Key's are required for any child of this component.");
```
albemarle: Key's -> Keys

This answers my question above about putting this info closer to the code.

abutterworth: doh thanks!

src/TransitionReplace/TransitionReplace.test.jsx
abutterworth: I agree. Tagged you over here: https://github.com/edx/eslint-config-edx/issues/47

robrap: I thought it leads to bugs in places? Seems simple to replace with `x += 1`.

src/TransitionReplace/index.jsx
```diff
@@ -0,0 +1,176 @@
...
+      </CSSTransition>
+    );
+  }
+
+  render() {
+    return (
+      <TransitionGroup
+        className={classNames('pgn-transition-replace-group', this.props.className)}
+        style={{
+          position: 'relative',
```
AlasdairSwan: Why not just use the util class ```position-relative```?

src/TransitionReplace/index.jsx
```diff
@@ -0,0 +1,176 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { CSSTransition, TransitionGroup } from 'react-transition-group';
+import classNames from 'classnames';
+
+
+const propTypes = {
+  children: PropTypes.element,
```
AlasdairSwan: Should this be ```required``` given that without children this component wouldn't really do anything?

abutterworth: We have a current use-case in the profile app where we don't always supply children. It works fine without it – effectively it becomes a transition in from 0 height.

AlasdairSwan: Interesting, it may be worth adding an example to storybook.

src/TransitionReplace/index.jsx
abutterworth: Per this feedback, I've swapped using style to prefer bootstrap utility classes where possible. I have no particular preference on one over the other, maybe we can discuss in person pros/cons of inline styles vs utility classes in varying situations.

src/TransitionReplace/index.jsx
abutterworth: Actually just double checked this. The reason for allowing no children was to make it easier to work with async content. It won't do a transition in height from nothing. If we want it to do that we'll need to add some more to this component in the future.

AlasdairSwan: Cool, it would be great to update the readme to note this for clarity.

src/TransitionReplace/TransitionReplace.test.jsx
davidjoy: ++ on this choice

src/TransitionReplace/index.jsx
abutterworth: All set

PR #420, opened 2019-04-10
fix: pass through unknown props for asInput components

mikix:
This way properties that we don't explicitly care about can still get through to the underlying input element. There are a lot of input props, seems unnecessary to gatekeep them.

In my immediate case, I was interested in using min and step - hence the new storybook that uses them.

I also set a version for jest-cli, which had an implicit version before at 20.x or something. I hit an error when updating the snapshot that was fixed in 22.x, so I set it's version to match the version we demand for just "jest".

### DIFF
--- a/.storybook/__snapshots__/Storyshots.test.js.snap
+++ b/.storybook/__snapshots__/Storyshots.test.js.snap
@@ -1245,7 +1245,7 @@ exports[`Storyshots Navigation|Pagination basic usage 1`] = `
           <span
             aria-hidden={true}
             className="fa fa-chevron-left mr-2"
-            id="pagination-41"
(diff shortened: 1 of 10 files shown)


mikix:
Uh, I somehow put my name on Adam's commit - will fix that up after a meeting

mikix:
Hmm, package-lock.json didn't auto update when I ran `npm install` - is there a command for that now?

mikix:
@abutterworth of course you're right about the other input components - I updated the PR to cover those too.

I did not update to avoid passing in unknown props - it wasn't clear to me the best way to do that... I guess pull out and ignore known-not-used-by-html props? I dunno, seemed a bit messy.

PR #427, opened 2019-04-18
fix: pull unused vars out of props to prevent adding them to html tag

abutterworth:
- Pulls unneeded properties out of props (and therefore out of `...other`). 
- Adds `aria-describedby={describedBy}` on checkboxes which seemed to be missing. @wittjeff do you know if this was intentional?
- Adds `asInput` prop types and then `withDeprecatedProps` to `Checkbox` to satisfy react/prop-types linting requirements

### DIFF
--- a/src/TextArea/index.jsx
+++ b/src/TextArea/index.jsx
@@ -1,45 +1,37 @@
 import React from 'react';
-import classNames from 'classnames';
+import PropTypes from 'prop-types';
 
-import asInput, { inputProps } from '../asInput';
+import asInput from '../asInput';
 import withDeprecatedProps, { DEPR_TYPES } from '../withDeprecatedProps';
+
 
 function Text(props) {
   const {
     className,
-    describedBy,
     inputRef,
-    isValid,
-    // Pull these out from props so html attrs can be passed through
-    // eslint-disable no-unused-vars
-    validator,
-    themes,
-    inline,
-    inputGroupPrepend,
-    inputGroupAppend,
-    label,
-    dangerIconDescription,
-    description,
-    validationMessage,
-    errorId,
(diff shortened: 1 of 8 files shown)


abutterworth:
If you prefer we could add an htmlAttrs prop instead. Less convenient but maybe less messy.

mikix:
Well, like a nonHtmlAttrs prop, right?

abutterworth:
I meant something like this:

```
render() {
  <input
    {...this.props.htmlAttrs}
    value={this.props.value}
  />
}
```

usage:

`<InputText htmlAttrs={{min: 0, max: 3}} value={1} type="range" />`

mikix:
Eh, of those two options, I personally prefer your original way of reproducting the list 4 times.

But my favorite way would be a list in asInput.jsx like so:
```
asInputProps = {
      onChange,
      isValid,
      validator,
      themes,
      inline,
      inputGroupPrepend,
      inputGroupAppend,
      label,
      dangerIconDescription,
      description,
      validationMessage,
      errorId,
      descriptionId,
}
```

That wrappers of asInput can then import and use, without having to know the internal details of asInput, and without having to repeat code.

But you mentioned not liking that approach when we were talking in person. Absent that, the 4-list approach is good.

src/TextArea/index.jsx
```diff
@@ -1,31 +1,37 @@
...
 
   return (
     <textarea
       {...others}
-      className={classNames(className)}
-      aria-describedby={describedBy}
-      aria-invalid={!isValid}
+      className={className}
       ref={inputRef}
-      themes={['danger']}
```
mikix: Not wanted?

PR #439, opened 2019-05-13
fix: update proptypes change disabled behavior

abutterworth:
- Change labels proptype to accept nodes instead of just strings. (helpful when using ReactIntl
- Instead throw away click events instead of using the disabled attribute to disable the button

### DIFF
--- a/src/StatefulButton/index.jsx
+++ b/src/StatefulButton/index.jsx
@@ -8,9 +8,10 @@ import Button from '../Button';
 const propTypes = {
   className: PropTypes.string,
   state: PropTypes.string,
-  labels: PropTypes.objectOf(PropTypes.string).isRequired,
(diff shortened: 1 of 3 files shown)


src/StatefulButton/index.jsx
```diff
@@ -40,13 +43,23 @@ function StatefulButton({
...
+        // Swallow clicks if the button is disabled.
+        // We do this instead of disabling the button to prevent
+        // it from losing focus (disabled elements cannot have focus).
+        if (isDisabled) {
+          e.preventDefault();
+          return;
+        }
+
+        if (onClick) onClick(e);
+      }}
```
Tj-Tracy: Looks like a good solution 👍

abutterworth:
@wittjeff do you mean `aria-disabled`?

by handling onClick we should also be covering RETURN on the keyboard no?

PR #441, opened 2019-05-22
Support custom icons passed in via props

adamstankiewicz:
We are using the `Collapsible` component in the learner portal but don't currently have a way to change the icon used in the button that toggles the collapsing behavior (other than manual styling). 

This PR adds support to pass in an `icons` prop as shown below:

```jsx
<Collapsible
  // ...
  icons={{
    expanded: <FontAwesomeIcon icon={faChevronCircleUp} />,
    collapsed: <FontAwesomeIcon icon={faChevronCircleDown} />,
  }}
>
  <p>Example content</p>
</Collapsible>
```

If the `icons` prop is not supplied, it will use the following icons as needed: `faAngleUp` & `faAngleDown`.

### DIFF
--- a/src/Collapsible/index.jsx
+++ b/src/Collapsible/index.jsx
@@ -1,9 +1,10 @@
 import React from 'react';
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
-import Button from '../Button';
-import Icon from '../Icon';
+import { faAngleUp, faAngleDown } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
+import Button from '../Button';
 
 class Collapsible extends React.Component {
   constructor(props) {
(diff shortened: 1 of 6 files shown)


src/Collapsible/index.jsx
```diff
@@ -61,11 +62,21 @@ class Collapsible extends React.Component {
...
+    }
+
+    return <FontAwesomeIcon icon={isOpen ? faAngleUp : faAngleDown} />;
+  }
+
   render() {
     const {
       children,
       expandedTitle,
-      iconId,
```
christopappas: If we remove this `iconId`, is it possible this element may no longer be referenced in the same way as before? As in, is there any harm in leaving the ID just in case there are other pieces of code pointing to this element by ID?

I'm guessing probably not, but just something that came to mind

adamstankiewicz: @christopappas Good question. Previously, this component was using Paragon's `Icon` component which accepts the optional `id` prop. If the `id` prop was not supplied, it'd generate a new one with the `newId()` utility function. However, if I remember correctly, we were experiencing an issue where the generated id's were not unique so support for providing an `iconId` instead was added here.

Now that the `Collapsible` component is no longer using the `Icon` component (instead, using `FontAwesomeIcon` fo SVG icons), I think it's safe to remove here.

davidjoy: If someone was using the icon ID to get a reference to the icon outside this component, then this will break.  I think we generally agreed a month or two back that that seems really unlikely.  If we wanted to be very safe, we'd mark this build as having a BREAKING CHANGE.

PR #452, opened 2019-06-03
Refactor Dropdown

julianajlk:
This is the new usage of paragon's Dropdown Component. Feedback is appreciated. 

Changes include: 
- Dropdown with icon element now follows the same design aesthetic as a regular Dropdown. 
- Added an option to have Dropdown with caret/chevron (arrow down). 
- Button, menu, and menu items are now subcomponents to Dropdown (children). 
- Button content and icons are no longer passed in as props. 

```
<Dropdown>
      <Dropdown.Button>
        <Icon className="fa fa-user pr-3" alt="" />
        Search Engines
      </Dropdown.Button>
      <Dropdown.Menu>
        <Dropdown.Item href="https://google.com">Google</Dropdown.Item>
        <Dropdown.Item href="https://duckduckgo.com">DuckDuckGo</Dropdown.Item>
        <Dropdown.Item href="https://yahoo.com">Yahoo</Dropdown.Item>
      </Dropdown.Menu>
</Dropdown>
````
- ClassNames are no longer baked in, a consuming application is able to specify a className prop to a Dropdown, a Dropdown.Button, or a Dropdown.Menu.
- Keyboard focus is now being handled by the browser, as opposed to React state focusIndex. 

FYI @edx/arch-squad

### CODE REVIEW

File: src/Dropdown/index.jsx [hunk 1]
```diff
@@ -1,203 +1,125 @@
...
   }
 
-  componentWillUpdate(_, nextState) {
-    if (nextState.open) {
-      this.addEvents();
-    } else {
-      this.removeEvents();
-    }
+  componentDidMount = () => {
+    document.addEventListener('click', this.handleDocumentClick, true);
```
abutterworth: We should only listen to click events on the document when we need to. Ideally we add a listener when the menu opens, remove the listener when it closes, and remove it when the component unmounts (just in case).

src/Dropdown/index.jsx
```diff
@@ -1,203 +1,125 @@
...
+      this.focusNext();
+    }
+    if (e.key === 'Escape') {
       this.toggle();
     }
   }
 
-  toggle() {
+  open = () => {
+    setTimeout(() => { this.focusFirst(); }, 1);
```
abutterworth: It may be worth moving this call to `focusFirst()` inside `componentDidUpdate`. This will remove the need for setTimeout I think

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,146 @@
...
   }
 
-  handleToggleKeyDown(e) {
-    if (!this.state.open && Dropdown.isTriggerKey('OPEN_MENU', e.key)) {
-      this.toggle();
-    } else if (this.state.open && Dropdown.isTriggerKey('CLOSE_MENU', e.key)) {
-      this.toggle();
+  handleMenuKeyDown = (e) => {
+    switch (e.key) {
+      case 'ArrowUp':
```
albemarle: Not sure this will work for IE / Edge: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

src/Dropdown/Dropdown.scss
```diff
@@ -1,11 +1,11 @@
-.dropdown.has-icon {
-  .icon-container {
-    max-width: 40px;
-    box-sizing: content-box;
+.dropdown {
+  margin-bottom: 1rem;
```
abutterworth: Let's remove this. Consumers of the dropdown should decide if a margin is needed.

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,169 @@
...
     };
+    this.menuItems = React.createRef();
   }
 
-  componentWillUpdate(_, nextState) {
-    if (nextState.open) {
-      this.addEvents();
-    } else {
-      this.removeEvents();
+  componentDidUpdate = (prevProps, prevState) => {
```
abutterworth: Let's avoid the arrow functions if we don't need the functions explicitly bound to `this`.  In this case, that would leave only event handlers as arrow functions.

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,169 @@
...
       return;
     }
-    this.toggle();
-  }
-
-  handleMenuKeyDown(e) {
-    e.preventDefault();
-    if (Dropdown.isTriggerKey('CLOSE_MENU', e.key)) {
+    if (this.state.open) {
       this.toggle();
```
abutterworth: I think `this.close()` in this case be clearer.

src/Dropdown/index.jsx
abutterworth: Looks like React has done this normalization for us: https://reactjs.org/docs/events.html#supported-events

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,169 @@
...
-          {
-            show: open,
-            'has-icon': hasIconElement,
-            rounded: hasIconElement,
-            border: hasIconElement,
-            'd-flex': hasIconElement,
-            'bg-white': hasIconElement,
-          },
-        )}
+        className={this.state.open ? 'dropdown show' : 'dropdown'}
```
abutterworth: Let's be consistent on referencing `open`. Here it's `this.state.open` but below it's `open`

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,169 @@
...
         ref={(container) => { this.container = container; }}
```
abutterworth: Any reason not to convert this to the `React.createRef();` format used for `this.menuItems`?

src/Dropdown/index.jsx
abutterworth: A consuming application should be able to specify a `className` prop to a dropdown. We should update this line to allow that.

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,169 @@
...
 
 export default Dropdown;
+
+Dropdown.Item = props => {
+  const { type, children, ...other } = props;
+  const item = React.createElement(
+    type,
+    {
+      ...other,
+      className: 'dropdown-item',
```
abutterworth: A consuming application should be able to specify a `className` prop to a DropdownItem. Let's update this line to append the 'dropdown-item' class to a className prop if it's supplied.

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,169 @@
...
+  return item;
+};
+
+Dropdown.Item.propTypes = {
+  type: PropTypes.string,
+  children: PropTypes.node,
+};
+
+Dropdown.Item.defaultProps = {
+  type: 'a',
```
abutterworth: Let's prefer `tag` as in an html tag over `type`

src/Dropdown/index.jsx
julianajlk: Should 'dropdown-item' be set as defaultProps?

src/Dropdown/index.jsx
julianajlk: Updated using [classnames](https://www.npmjs.com/package/classnames) `classNames(
          'dropdown',
          {
            show: this.state.open,
          },
        )`

src/Dropdown/index.jsx
julianajlk: Throws an error using the new syntax for creating refs.

src/Dropdown/index.jsx
julianajlk: Removed `const { open } = this.state;` to keep clear/consistent as `this.state.open`

src/Dropdown/Dropdown.scss
julianajlk: Added [bootstrap spacing](https://getbootstrap.com/docs/4.3/utilities/spacing/) `mg-3` and updated icon to bootstrap spacing as well.

src/Dropdown/index.jsx
julianajlk: Removed event listener from componentDidMount to open method.

src/Dropdown/index.jsx
abutterworth: No let's leave this one baked in.

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,198 @@
...
     }
   }
 
-  componentDidUpdate(prevProps, prevState) {
-    if (this.state.open) {
-      this.menuItems[this.state.focusIndex].focus();
-    } else if (prevState.open && this.toggleElem) {
-      this.toggleElem.focus();
-    }
+  componentWillUnmount = () => {
```
abutterworth: This doesn't need to be an arrow function right?

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,198 @@
...
+      case 'Tab':
+        e.preventDefault();
+        this.focusNext();
+        break;
+      case 'Escape':
+        e.stopPropagation();
+        this.toggle();
+        break;
+      default:
+        return null;
```
abutterworth: There's no need to `return null` here.

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,198 @@
...
+  className: PropTypes.string,
+  buttonClassName: PropTypes.string,
+  children: PropTypes.node.isRequired,
+  buttonContent: PropTypes.string.isRequired,
 };
 
 Dropdown.defaultProps = {
-  buttonType: 'light',
-  iconElement: undefined,
+  buttonClassName: 'dropdown-toggle',
```
abutterworth: Is `dropdown-toggle` needed at all? If it is let's bake in 'dropdown-toggle' into our component instead of being the default so consumers of it don't need to know this class name when they add more.

abutterworth: Also, let's set the default to `btn-light` instead

src/Dropdown/Dropdown.stories.jsx
```diff
@@ -9,50 +9,42 @@ storiesOf('Navigation|Dropdown', module)
...
+        buttonClassName="dropdown-toggle btn-light"
+        buttonContent={<span>Search Engines<Icon className="fa fa-caret-down pl-3" alt="icon-caret" /></span>}
+      >
+        <Dropdown.Item href="https://google.com">Google</Dropdown.Item>
+        <Dropdown.Item href="https://duckduckgo.com">DuckDuckGo</Dropdown.Item>
+        <Dropdown.Item href="https://yahoo.com">Yahoo</Dropdown.Item>
+      </Dropdown>
+      <Dropdown
+        buttonClassName="dropdown-toggle btn-light"
+        buttonContent={<span>Search Engines<Icon className="fa fa-angle-down pl-3" alt="icon-chevron" /></span>}
```
abutterworth: The `alt` text here should be a human readable label or not present.

src/Dropdown/Dropdown.stories.jsx
```diff
@@ -9,50 +9,42 @@ storiesOf('Navigation|Dropdown', module)
...
-    <Dropdown
-      title="Search Engines"
-      menuItems={[
-        <a href="http://www.google.com">Google</a>,
-        <a href="http://www.duckduckgo.com">DuckDuckGo</a>,
-        <a href="http://www.yahoo.com">Yahoo</a>,
-      ]}
-    />
+      buttonClassName="dropdown-toggle btn-light"
+      buttonContent={<span>Search Engines</span>}
```
abutterworth: Can use just a string here: `buttonContent="Search Engines"`

src/Dropdown/index.jsx
julianajlk: Removed both null, replaced with "break"

src/Dropdown/index.jsx
julianajlk: Correct, updated open() and close() as well.

src/Dropdown/index.jsx
julianajlk: Updated using [classnames](https://www.npmjs.com/package/classnames) `className={classNames(
            'dropdown-toggle',
            buttonClassName,
          )}`
and added 'btn-light' as default.

src/Dropdown/Dropdown.stories.jsx
```diff
@@ -9,50 +9,38 @@ storiesOf('Navigation|Dropdown', module)
...
+      <Dropdown.Item href="https://google.com">Google</Dropdown.Item>
+      <Dropdown.Item href="https://duckduckgo.com">DuckDuckGo</Dropdown.Item>
+      <Dropdown.Item href="https://yahoo.com">Yahoo</Dropdown.Item>
+    </Dropdown>
+  ))
+  .add('with caret/chevron', () => (
+    <React.Fragment>
+      <Dropdown
+        className="mb-3"
+        buttonContent={<span>Search Engines<Icon className="fa fa-caret-down pl-3" alt="" /></span>}
```
albemarle: I just really want to have class names such that we can type `fa la-la-la la`.  That is all.

src/Dropdown/README.md
```diff
@@ -1,19 +1,19 @@
...
 
-Note: When passing in a component to `menuItems` that uses a prop other than `ref` for access to a focusable DOM element (such as the `Link` component from `react-router`, which uses `innerRef`), it's recommended to wrap the component in another component that exposes it's own focus method.
+Note: When passing in a component to `Item` that uses a prop other than `ref` for access to a focusable DOM element (such as the `Link` component from `react-router`, which uses `innerRef`), it's recommended to wrap the component in another component that exposes it's own focus method.
 
 ## API
 
-### `buttonType` (string; optional)
-`buttonType` is used to determine the type of button to be rendered.  See [Bootstrap's buttons documentation](https://getbootstrap.com/docs/4.0/components/buttons/) for a list of applicable button types. The default is `buttonType="light"`.
+### `buttonClassName` (string; optional)
+`buttonClassName` is used to determine the type of button to be rendered. See [Bootstrap's buttons documentation](https://getbootstrap.com/docs/4.0/components/buttons/) for a list of applicable button types and it's classNames.
```
albemarle: it's -> its

src/Dropdown/Dropdown.stories.jsx
abutterworth: This is the only reason we've refactored this component.

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,239 @@
...
 
-export const triggerKeys = {
-  OPEN_MENU: ['ArrowDown'],
-  CLOSE_MENU: ['Escape'],
-  NAVIGATE_DOWN: ['ArrowDown', 'Tab'],
-  NAVIGATE_UP: ['ArrowUp'],
-  SELECT_MENU_ITEM: ['Enter', ' '],
-};
+const { Provider, Consumer } = React.createContext();
+// const DropdownContext = React.createContext();
```
abutterworth: Unneeded comment?

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,239 @@
...
-            show: open,
-            'has-icon': hasIconElement,
-            rounded: hasIconElement,
-            border: hasIconElement,
-            'd-flex': hasIconElement,
-            'bg-white': hasIconElement,
+            show: this.state.open,
           },
         )}
         ref={(container) => { this.container = container; }}
```
abutterworth: I know we've discussed before, but can we use the new React.createRef for this ref?

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,239 @@
...
+  iconElement: {
+    deprType: DEPR_TYPES.REMOVED,
+    message: 'It should be specified inside the buttonContent prop.',
+  },
+});
+
+DropdownWithDeprecatedProps.propTypes = Dropdown.propTypes;
+DropdownWithDeprecatedProps.defaultProps = Dropdown.defaultProps;
+DropdownWithDeprecatedProps.Item = DropdownItem;
+DropdownWithDeprecatedProps.Button = DropdownButton;
```
abutterworth: Let's also do these assignments for Dropdown (e.g. `Dropdown.Button = DropdownButton`) so that we don't need to remember to add it in when we remove this withDeprecatedProps shim.

src/Dropdown/index.jsx
```diff
@@ -1,225 +1,239 @@
...
+DropdownWithDeprecatedProps.Menu = DropdownMenu;
+
+// Dropdown.Button = DropdownButton;
+// export const DropdownConsumer = DropdownContext.Consumer;
```
abutterworth: Remove?

PR #457, opened 2019-06-06
Ekii/status alert refactor

erinakii:
WIP Refactoring status alert. Let me know what your thoughts are


src/StatusAlert/StatusAlertWrapperWithButton.jsx
```diff
@@ -0,0 +1,47 @@
...
+class StatusAlertWrapperWithButton extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { isOpen: false };
+    this.toggleAlert = this.toggleAlert.bind(this);
+  }
+
+  toggleAlert() {
+    this.setState({
+      isOpen: !this.state.isOpen,
```
Tj-Tracy: Because of the async nature of react, its generally better not to access `this.state` directly during `setState()`.

instead, use the `prevState` provided by the signature. i.e something like:
```
this.setState((prevState, props) => ({ isOpen: !prevState.isOpen});
```

that way, you're making sure you're always getting the correct state of the component at the time the toggle function is called, instead of whatever the state happens to be at the time the line executes. This case is pretty simple and I doubt this will ever break as is, but it never hurts to do it this way. (a little easier to read, too).

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "eslint.autoFixOnSave": true
+}
```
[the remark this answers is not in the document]
erinakii: Thanks for the catch. Added to gitignore and removed from git history.

src/StatusAlert/StatusAlertWrapperWithButton.jsx
erinakii: thanks! that's cool, I didn't know that was best practice so I'll definitely keep that in mind for the future as well!

PR #461, opened 2019-06-07
feat: use babel instead of webpack for build

abutterworth:
Removes webpack and prefers a simple babel transpilation instead. This preserves ES6 modules allowing for consuming applications to tree shake as necessary.


.travis.yml
```diff
@@ -25,7 +25,6 @@ script:
 - npm run lint
 - npm run test
 - npm run build
-- npm run is-es5
```
abutterworth: We are explicitly leaving ES6 modules in to allow for tree shaking in a consuming app.

davidjoy:
Closing - this was merged in via a different PR.  ATTN @abutterworth

PR #465, opened 2019-06-12
Initial documentation site using gatsby and mdx

abutterworth:
See `www/README.md` for usage info.
https://github.com/edx/paragon/blob/a37ffc54a1ae4595bb25d8af35f3fdd51898048a/www/README.md

Screen Shot 2019-06-12 at 4 40 23 PM


www/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "paragon",
+    "pattern library",
+    "design",
+    "react",
+    "component library"
+  ],
+  "license": "MIT",
+  "scripts": {
+    "build": "gatsby build",
+    "develop": "gatsby develop",
```
matthugs: I'd recommend we add a `-p <some more obscure number that nothing in devstack uses>` flag now. Will help folks who live with devstack up leverage this sooner

abutterworth: What's your favorite number?

abutterworth: Actually it looks like Gatsby will ask if you'd like to run on a different port. It defaults to 8000 and works up from there. I think that should work well enough for now.

www/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "start": "npm run develop",
+    "lint": "eslint --ext .js --ext .jsx .",
+    "test": "echo \"Error: no test specified\" && exit 1"
```
Tj-Tracy: Its not super important at the moment, but adding a `gatsby clean` to the scripts will come in handy. It can help fix a few errors when content isn't updating properly.

abutterworth: Added.

PR #469, opened 2019-06-18
docs: basic layout and nav for doc site

abutterworth:
Screen Shot 2019-06-18 at 1 12 29 PM


www/gatsby-config.js
```diff
@@ -30,5 +27,12 @@ module.exports = {
         icon: 'src/images/paragon-icon.png',
       },
     },
+    {
+      // Using the layout plugin to prevent re-renders of nav between routes
```
albemarle: Just to check, "nav" is the nav sidebar?

www/src/components/scss/_hero.scss
```diff
@@ -0,0 +1,19 @@
+
```
albemarle: Extra blank line?

PR #473, opened 2019-06-20
docs: add status page

abutterworth:
The Component Status index is generated from frontmatter on mdx files.
Screen Shot 2019-06-20 at 12 29 23 PM


www/src/pages/components/breadcrumbs.mdx
```diff
@@ -1,3 +1,16 @@
+---
+title: "Breadcrumbs"
+type: "component"
+status: "Needs Work"
+designStatus: "To Do"
+devStatus: "To Do"
+notes: |
```
albemarle: Oooh, I like this.

www/src/pages/components/modal.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+  render() {
+    return (
+      <div>
+        <Modal
+          open={this.state.open}
+          title="Modal title."
+          body={(
+            <div>
+              <p>Enter your e-mail address to receive free cat facts!</p>
```
albemarle: I love those!

www/src/pages/components/transitionreplace.mdx
```diff
@@ -1,10 +1,84 @@
...
+          </div>
+        )}
+      </TransitionReplace>
+      <Button className="btn-primary" onClick={() => setIsEditing(!isEditing)}>Toggle</Button>
+    </div>
+  );
+}
+```
+
+**IMPORTANT NOTE: You must provide the `key` prop for all children**, even when only rendering a single item. This is how React will determine that a child is new rather than just updated.
```
albemarle: "provide a unique `key` prop" ?  (Don't know if that's actually true.  :) )

PR #479, opened 2019-06-28
build(treeshaking): export es6 modules so the can be treeshook

davidjoy:
BREAKING CHANGE: This potentially requires consuming apps to update their webpack config to take advantage of treeshaking, and it hasn’t yet been tested with gatsby.


.babelrc
```diff
@@ -3,10 +3,8 @@
     [
       "@babel/preset-env",
       {
-        "targets": {
```
davidjoy: "browsers" is deprecated here.  It uses the `browserslist` in package.json.

.babelrc
```diff
@@ -3,10 +3,8 @@
...
-          "browsers": ["last 2 versions", "ie 11"]
-        }
-      }
+        "modules": false
```
davidjoy: A very important boolean.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@edx/paragon",
   "version": "0.0.0-development",
   "description": "Accessible, responsive UI component library based on Bootstrap.",
-  "main": "dist/paragon.js",
+  "main": "dist/index.js",
```
davidjoy: babel won't change this name for us.

package.json
```diff
@@ -13,8 +13,9 @@
     "/src",
     "/scss"
   ],
+  "sideEffects": false,
   "scripts": {
-    "build": "webpack --mode production --display-modules && node build-scss.js",
+    "build": "rm -rf ./dist && ./node_modules/.bin/babel src --out-dir dist --source-maps --ignore **/*.test.jsx,**/*.stories.jsx,**/__mocks__,**/setupTest.js && node build-scss.js",
```
davidjoy: Note all the exclusions - this means just the code will be included.

package.json
```diff
@@ -69,17 +73,12 @@
     "@storybook/storybook-deployer": "^2.8.1",
     "babel-eslint": "^10.0.1",
     "babel-jest": "^24.5.0",
-    "babel-loader": "^8.0.5",
+    "babel-loader": "^8.0.6",
```
davidjoy: Used by... storybook... I think.  It was used by build-docs.

src/index.js
```diff
@@ -26,53 +26,13 @@ import {
...
-import StatusAlert from './StatusAlert';
-import Table from './Table';
-import Tabs from './Tabs';
-import TextArea from './TextArea';
-import ValidationFormGroup from './ValidationFormGroup';
-import TransitionReplace from './TransitionReplace';
-import ValidationMessage from './ValidationMessage';
-import Variant from './utils/constants';
-
-export {
```
davidjoy: Getting rid of this big, single export is also very important.  You can't tree shake a single export, they need to be individual exports.

webpack.config.js
```diff
@@ -1,69 +0,0 @@
-const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
```
davidjoy: Goodnight, sweet prince.

package.json
abutterworth: is `./node_modules/.bin/babel` needed instead of just `babel`?

PR #489, opened 2019-07-02
refactor: fix tabsui

erinakii:
1. change div to ul in render of labels and div to li in buildlabels method
2. changed to have bottom border for tabs not selected
3. changed Button component to native bootstrap button


src/Tabs/index.jsx
```diff
@@ -43,19 +42,21 @@ class Tabs extends React.Component {
...
+        <li key={labelId}>
+          <button
+            role="tab"
+            aria-selected={selected}
+            aria-controls={this.genPanelId(i)}
+            id={labelId}
+            onClick={() => { this.toggle(i); }}
+            className={classNames('nav-link nav-item', {
             active: selected,
+            'border-bottom-0': !selected,
```
abutterworth: Let's try to get rid of this `border-bottom-0` class name or make it optional. Generally we should avoid enforcing a style within paragon as much as we can.

erinakii: Hi Adam! Apologies for the random PR. I was trying to fix the UI of the tabs component. I'm including the border-bottom-0 className so that a bottom border is included when the tabs are not active. Otherwise it looks as though only the active tab receives a border. This is the only way I could fix the way it looks. I think @adamstankiewicz may be able to explain a little better. Let me know if you think there is a better approach

abutterworth: Thanks @erinakii. It looks like since we're using the button element we are getting a background-color on the button that we don't want. Bootstrap uses anchor tags so they are fine. I would either swap out buttons for anchor tags (if that can be made okay from an accessibility perspective @wittjeff) or add a `@import "./Tabs/Tabs.scss"` to `src/index.scss` where the tabs SCSS would contain something like this:
```
.nav-tabs .nav-link {
  background: transparent;
}
```

erinakii: @abutterworth thanks for the comments. That definitely makes sense. I added the scss file to make the background transparent associated to classNames .nav-tabs and .nav-link. I figured this would be better than changing to anchor tabs. I believe for accessibility it makes more sense to leave them as buttons, since we are not navigating away from the page. 

Let me know what you think! 
@adamstankiewicz @wittjeff

src/index.scss
```diff
@@ -8,3 +8,4 @@
 @import './Table/Table.scss';
 @import './TransitionReplace/TransitionReplace.scss';
 @import './ValidationMessage/ValidationMessage.scss';
+@import "./Tabs/Tabs.scss";
```
abutterworth: Newline at end of file

src/Tabs/Tabs.scss
```diff
@@ -0,0 +1,3 @@
+.nav-tabs .nav-link {
+    background: transparent;
+  }
```
abutterworth: Indentation and new line at end of file

src/Tabs/index.jsx
```diff
@@ -89,23 +89,22 @@ class Tabs extends React.Component {
...
           {panels}
         </div>
       </div>
     );
   }
 }
-
 // TODO: custom validator that ensures labels and panels are the same length
 Tabs.propTypes = {
   labels: PropTypes.oneOfType([
```
abutterworth: Let's prefer using `PropTypes.node` over any instance of `PropTypes.oneOfType([PropTypes.string, PropTypes.element])`. In this case we could use: `PropTypes.arrayOf(PropTypes.node)` for labels

PR #496, opened 2019-07-19
style: remove browser default clear input search

julianajlk:
Added CSS to remove browser (Chrome and IE) default "X" to clear input search box. 
Added `margin-top` to search field input to match vertical alignment with "search:"
[ARCH-1048](https://openedx.atlassian.net/browse/ARCH-1048).

### DIFF
--- a/src/SearchField/SearchField.scss
+++ b/src/SearchField/SearchField.scss
@@ -13,7 +13,6 @@ $search-field-border-radius: 30px;
 
     .input-group {
         flex: 1 1 0;
-
         //clear browser 'x' from Internet Explorer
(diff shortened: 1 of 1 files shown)


src/SearchField/SearchField.scss
```diff
@@ -29,23 +47,24 @@ $search-field-border-radius: 30px;
         border-color: $blue !important;
         box-shadow: $input-focus-box-shadow;
     }
-    
+
     .input {
-        background: none;
+        background: gray;
```
abutterworth: Why gray?

julianajlk: Changed back to none.

PR #498, opened 2019-07-24
feat: remove roboto from theme scss

davidjoy:
Roboto must be imported manually if you want it; it’s not just a part of the theme.  This means we err on the side of increased performance by not downloading the roboto font by default.


davidjoy:
Docs have been updated. 👍

PR #500, opened 2019-07-31
fix(asinput): proptype validation to include both func and obj

erinakii:
Ensuring that proptype validation if inputRef in asInput includes both functions and objects. This is because creating a previous ref used a function, where React.createRef will create an object


erinakii:
@abutterworth I changed propType validation in other components that were passing inputRef as props

PR #504, opened 2019-08-08
feat: Collapsible Component

julianajlk:
Refactor of the Collapisble Component. It is now a composable component and set of subcomponents.

BREAKING CHANGE: The prop api has changed significantly to simplify usages and enable flexibility. See the docs for details.

BREAKING CHANGE: Related styles are included in paragon scss and no longer exist as a sibling to the component.


src/Collapsible/Collapsible.scss
```diff
@@ -1,56 +1,43 @@
-// Local Variables
-$collapsible-border: 1px solid silver;
```
abutterworth: Let's do our best to remove this file entirely and fold the core scss into /core and edx specific scss into edx/theme

src/icons/ChevronDown.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const ChevronDown = ({ height, ...other }) => (
+  <svg
+    height={height}
```
abutterworth: Any reason for height here?

julianajlk: Added height and width default to be 24px (1.5rem), the natural SVG size for these specific SVG icons

src/Collapsible/Collapsible.scss
julianajlk: To be done in the future

PR #506, opened 2019-08-16
Add component status and proptype descriptions

Tj-Tracy:
This PR puts the status at the top of each components doc page, along with descriptions for most of the propTypes. Some of the Readmes were missing descriptions for proptypes, but this should migrate 99% of documentation that was on the old website.

Thanks to prettier running amok, some formatting changes are present in this PR, but I removed all of the ones that clashed with eslint, and (hopefully) kept the ones that improve readability.

### DIFF
--- a/src/Icon/index.jsx
+++ b/src/Icon/index.jsx
@@ -24,9 +24,18 @@ function Icon(props) {
 }
 
 Icon.propTypes = {
+  // eslint-disable-next-line max-len
+  /** the `id` property of the Icon element, by default this value is generated with the `newId` function with the `prefix` of `Icon`. */
   id: PropTypes.string,
+  // eslint-disable-next-line max-len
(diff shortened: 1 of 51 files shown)


src/Icon/index.jsx
```diff
@@ -24,9 +24,18 @@ function Icon(props) {
 }
 
 Icon.propTypes = {
+  // eslint-disable-next-line max-len
```
abutterworth: Maybe we add ignoreComments to this rule for this project?
https://eslint.org/docs/rules/max-len

Tj-Tracy: Adding that causes some linting errors elsewhere, that aren't caused by comments for some reason. I don't want to break the `.eslintrc` on this PR, but I can come back and fix this.

PR #510, opened 2019-08-23
feat: delete storybook

Tj-Tracy:
This PR deletes the old docsite/storybook/storyshots.

It fixes documentation in a few places and should allow for gatsby to build the new docsite and deploy it to github pages. It also fixes the npm commands.

Part of this was removing snapshots, but coverage was only lost on `Input` and the soon-to-be-deprecated `InputSelect`. I've added tests for `Input` to make up the coverage.


package.json
```diff
@@ -16,21 +16,18 @@
   "sideEffects": false,
   "scripts": {
     "build": "make build",
-    "build-docs": "build-storybook && npm run build-gatsby && npm run move-gatsby",
-    "build-storybook": "build-storybook",
+    "build-docs": "npm run build-gatsby && npm run move-gatsby",
```
abutterworth: "deploy" is missing in these scripts.

If you remove "move-gatsby" it could be something like
`"deploy": "gh-pages -d ./www/public"`

package.json
```diff
@@ -16,21 +16,18 @@
...
-    "deploy-storybook-ci": "storybook-to-ghpages --ci",
     "is-es5": "es-check es5 ./dist/*.js",
     "lint": "eslint --ext .js --ext .jsx .",
-    "move-gatsby": "mv ./www/public ./storybook-static/v2",
+    "move-gatsby": "mv ./www/public ./",
     "prepublishOnly": "npm run build",
     "semantic-release": "semantic-release",
     "snapshot": "jest --updateSnapshot",
-    "start": "start-storybook -p 6006",
+    "start": "cd ./www && \"$npm_execpath\" start",
```
abutterworth: This will break for first time users of the repository since packages won't have been installed yet. We could just echo in terminal that /www is where the doc site lives and they must do npm install both in the top level directory and the www directory

Tj-Tracy: should we create an init command of some kind?

abutterworth: Maybe `preinstall` would be better? https://stackoverflow.com/questions/48859022/override-npm-install-script-for-npm-project

PR #534, opened 2020-04-06
Updating pagination component

srwang:
-updating pagination component to accept custom elements for chevron
-updating width so that 1140px is usable space, not total including padding

### DIFF
--- a/scss/core/_variables.scss
+++ b/scss/core/_variables.scss
@@ -121,7 +121,7 @@ $container-max-widths: (
   sm: 540px,
   md: 720px,
   lg: 960px,
-  xl: 1140px,
+  xl: 1140px
 ) !default;
 
 @include _assert-ascending($container-max-widths, "$container-max-widths");
--- a/scss/core/_variables.scss
+++ b/scss/core/_variables.scss
@@ -121,7 +121,7 @@ $container-max-widths: (
   sm: 540px,
   md: 720px,
   lg: 960px,
-  xl: 1140px
+  xl: 1140px + 2 * $grid-gutter-width,
 ) !default;
 
 @include _assert-ascending($container-max-widths, "$container-max-widths");
--- a/scss/core/_variables.scss
+++ b/scss/core/_variables.scss
@@ -121,7 +121,7 @@ $container-max-widths: (
   sm: 540px,
   md: 720px,
   lg: 960px,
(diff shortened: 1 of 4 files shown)


src/Pagination/index.jsx
```diff
@@ -417,6 +417,8 @@ The default is:
 }
 ``` */
   buttonLabels: PropTypes.shape({
+    leftChevron: PropTypes.element,
```
abutterworth, line 420:
```suggestion
    leftChevron: PropTypes.node,
```
(any renderable element)

scss/core/_variables.scss
```diff
@@ -121,7 +121,7 @@ $container-max-widths: (
   sm: 540px,
   md: 720px,
   lg: 960px,
-  xl: 1140px
+  xl: 1140px + 2 * $grid-gutter-width,
```
abutterworth, line 124: This hasn't been defined yet, best to keep this change out of `core/*` and bring it into `edx/*`

scss/core/_variables.scss
```diff
@@ -121,7 +121,7 @@ $container-max-widths: (
   sm: 540px,
   md: 720px,
   lg: 960px,
-  xl: 1140px
+  xl: 1140px,
```
abutterworth, line 124:
```suggestion
  xl: 1140px
```
left over addition

scss/edx/overrides/_grid.scss
```diff
@@ -1,6 +1,17 @@
...
 .container-fluid {
-  max-width: 1140px;
+  max-width: 1140px + 2 * $grid-gutter-width;
   padding-left: $grid-gutter-width;
   padding-right: $grid-gutter-width;
 }
+
+$container-max-widths: map-merge(
+  (
+    xl: 1140px + 2 * $grid-gutter-width,
```
abutterworth, line 14: Does this also require us to change: `$grid-breakpoints`? https://github.com/edx/paragon/pull/534/files#diff-252b35fc7cc2964fadc678c056766291R109?

srwang: grid-breakpoints seems to be set to the correct width already, so I leaned on the side of writing less code

abutterworth: 👍

abutterworth: 👍

PR #536, opened 2020-04-09
fix: overriding media query

srwang:


scss/edx/overrides/_grid.scss
```diff
@@ -3,6 +3,19 @@
   padding-right: $grid-gutter-width;
 }
 
+@media (min-width: 1200px) {
+  .container,
+  .container-sm,
+  .container-md,
+  .container-lg,
+  .container-xl {
```
abutterworth, line 11: We should avoid adding to overrides if we can get the job done through variables. I suggest adding to the `edx/` theme the grid variables:
```
$grid-gutter-width:           30px !default;

// Grid breakpoints
//
// Define the minimum dimensions at which your layout will change,
// adapting to different screen sizes, for use in media queries.

$grid-breakpoints: (
  xs: 0,
  sm: 576px,
  md: 768px,
  lg: 992px,
  xl: 1200px
);

// Grid containers
//
// Define the maximum width of `.container` for different screen sizes.

$container-max-widths: (
  sm: 540px,
  md: 720px,
  lg: 960px,
  xl: 1140px
);
```
Probably in edx/varaibles/_grid.scss. Let me know if you want to adjust or if you'd like me to edit this PR.

PR #540, opened 2020-04-27
feat(searchfield): refactor SearchField component

adamstankiewicz:
Ticket: https://openedx.atlassian.net/browse/ARCH-990

BREAKING CHANGE: the refactoring of the SearchField component adds support for
additional flexibility (e.g., optional clear button, no label displayed, etc.).

The component is broken into subcomponents and shares data between components (e.g., callback handlers, input value) via the Context API.

This PR improves the a11y of this component by wrapping the search input field with a `<form role="search">` and makes use of buttons with the types "submit" and "reset" to handle form submission and clearing of the form, respectively.


src/SearchField/SearchField.scss
```diff
@@ -1,83 +1,33 @@
...
-        border: 0;
-        box-shadow: none;
-        padding-left: 0;
-        padding-right: 0;
-        margin-top: 0.15rem;
-
+    .form-control {
         &:focus {
-            background: none;
+            box-shadow: none;
```
wittjeff, line 10: Can you add **outline: transparent** to this? It makes the outline style visible in Windows High Contrast mode.  Otherwise, it'll be visible here(if not suppressed somewhere else), or outline: none would remove it as well as the box-shadow in HC mode, leaving no focus styling.

adamstankiewicz: @wittjeff `outline: transparent` does already exist on Bootstrap `.form-control` input elements:

image

src/SearchField/SearchField.scss
```diff
@@ -1,83 +1,33 @@
-$search-field-clear-btn-width: 30px;
-$search-field-border-radius: 30px;
-
-.search-field {
+.pgn-searchfield {
```
abutterworth, line 1: Let's use `.pgn__` as the prefix. See the docs for TransitionReplace here: https://github.com/edx/paragon/tree/32773a388ddfc74d422706273ec6b06e6721d408/src/TransitionReplace It looks like this isn't fully consistent, but I think using underscores is helpful to indicate that this is somewhat private. What do you think?

src/SearchField/SearchFieldInput.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+  const {
+    inputId, value, handleChange, handleFocus, handleBlur, refs,
+  } = useContext(SearchFieldContext);
+
+  return (
+    <Input
+      {...others}
+      ref={refs.input}
+      type="search"
```
abutterworth, line 19: Should we bother using `type="search"` instead of `type="text"` if it just causes styling problems? cc @wittjeff

src/SearchField/SearchFieldAdvanced.jsx
```diff
@@ -0,0 +1,172 @@
...
+  return (
+    <div
+      className={classNames(
+        'pgn-searchfield',
+        'd-flex border rounded',
+        { 'has-focus': hasFocus },
+        className,
+      )}
+    >
+      <form
```
abutterworth, line 81: Do you have any thoughts about whether it's good to have the form in here? I'm imagining a search with filters/checkboxes type situation where those other inputs wouldn't be able to share the same form element as the search field itself.

src/SearchField/SearchFieldAdvanced.jsx
```diff
@@ -0,0 +1,172 @@
...
+
+  const handleChange = (event) => {
+    setValue(event.target.value);
+  };
+
+  return (
+    <div
+      className={classNames(
+        'pgn-searchfield',
+        'd-flex border rounded',
```
abutterworth, line 76: Let's avoid using bootstrap utilities as much as possible. `border` and `rounded` should probably be part of the default search field scss or our edx theme.

src/SearchField/SearchFieldInput.jsx
adamstankiewicz: That's a fair point. As far as I know, using `type="search"` doesn't add any functionality different than using `type="text"`; it only adds stylistic differences, which we're overriding anyways.

I'm going to switch to using `type="text"` unless @wittjeff says otherwise.

src/SearchField/SearchField.scss
adamstankiewicz: Sounds good to me; will update.

src/SearchField/SearchFieldAdvanced.jsx
adamstankiewicz: Good question 🤔

For a11y, a form that encompasses a search input field should have the `role="search"` attribute. Given that, I feel that it makes sense to include the form element within the component to guarantee a11y when using this component.

If there were a filters/checkboxes type of situation alongside the search field to further narrow down results, I'm not sure it's safe to assume those controls would be within the same form as the search field input (e.g., with the same `onSubmit` handler, etc.).

The implementation of this refactor also relies on the input field and submit/clear buttons being children of a form in order to take advantage of native form submission and resetting behaviors. I'd be worried we wouldn't be able to enforce that a form wraps the `SearchField` component if we don't provide one ourselves.

FWIW, I know Algolia's `SearchBox` component includes a form element as well.

Happy to hear any additional thoughts you might have on this, of course.

abutterworth: Makes sense to me, no objections.

src/SearchField/SearchField.scss
```diff
@@ -1,83 +1,21 @@
...
 
-    .input {
-        background: none;
-        border: 0;
-        box-shadow: none;
-        padding-left: 0;
-        padding-right: 0;
-        margin-top: 0.15rem;
+    .form-control {
+        @extend .border-0;
```
abutterworth, line 11: border-0 will get you `!important` declarations I think. Might be better not to extend.

src/SearchField/SearchFieldAdvanced.jsx
```diff
@@ -0,0 +1,176 @@
...
+    submit: PropTypes.element.isRequired,
+    clear: PropTypes.element,
+  }),
+  /** specifies the aria-label attribute on the form element. This is useful if you use the `SearchField` component more than once on a page. */
+  formAriaLabel: PropTypes.string,
+};
+
+SearchFieldAdvanced.defaultProps = {
+  className: undefined,
+  formAriaLabel: undefined,
```
abutterworth, line 159: What do you think about using just `aria-label` as the prop. As it stands the prop name kind of forces you to know a little about the inner-workings of the SearchField. ..but it could be counterintuitive if we're putting the label on a particular element not the wrapper. Curious your thoughts.

src/SearchField/SearchFieldClearButton.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+const SearchFieldClearButton = () => {
+  const {
+    screenReaderText, icons, value,
+  } = useContext(SearchFieldContext);
+
+  return (
+    <button
+      type="reset"
+      className={classNames('btn px-2', { 'd-none': !value })}
```
abutterworth, line 14: Random thought, feel free to ignore. But you could also early return `null` if there's no value instead of toggling a class and leaving the button in the DOM.

src/SearchField/SearchFieldInput.jsx
```diff
@@ -0,0 +1,44 @@
...
+    <Input
+      {...others}
+      ref={refs.input}
+      type="text"
+      role="searchbox"
+      id={inputId.current}
+      name="searchfield-input"
+      className={className}
+      value={value}
+      autoComplete="off"
```
abutterworth, line 24: Should this be a default prop value?

src/SearchField/SearchField.scss
adamstankiewicz: Yes, extending does use `!important`. In this particular case, I think it makes sense since the border you would find on a normal `.form-control` is added to the `.pgn__searchfield` wrapper. If `!important` is not included on the `.form-control`, we are potentially making it easier for a consumer of `<SearchField />` to have a double border of the wrapper plus the `.form-control` if they were to override `border: 0;` (without the `!important` declaration).

Just my two cents. Happy to not use `@extend`, too, if we feel that's more appropriate.

src/SearchField/SearchFieldSubmitButton.jsx
```diff
@@ -0,0 +1,20 @@
...
+
+import { SearchFieldContext } from './SearchFieldAdvanced';
+
+const SearchFieldSubmitButton = () => {
+  const { screenReaderText, icons, refs } = useContext(SearchFieldContext);
+
+  return (
+    <button
+      type="submit"
+      className="btn px-2"
```
abutterworth, line 11: I'm mildly concerned about using `px-2` and other padding/margin utilities, but don't have a better solution in mind. Utility classes are annoying to override so if we're putting them in we should be confident that they are not something that should be themable. In this case maybe that's true.

adamstankiewicz: Fair point. In this case, I think that's largely true. However, could also move the use of the utility classes like this one to the custom SCSS file for the component with `@extend .px-2`.

src/SearchField/SearchField.scss
abutterworth: Should it be a form-control at all? Maybe like .form-control-file we should add a new class .form-control-search or something.  !important shouldn't be needed to customize stuff I think.

src/SearchField/SearchFieldSubmitButton.jsx
abutterworth: I think it's fine here. If it gets in the way we can chagne in the future. Wouldn't be a breaking change I think

src/SearchField/SearchFieldInput.jsx
adamstankiewicz: I will remove this as a default prop. 

If consumers of the component want to turn off autocomplete (which seems pretty common, afaik), they'd have to use the `SearchField.Advanced` implementation to be able to pass this `autoComplete` prop to the input.

src/SearchField/SearchFieldAdvanced.jsx
adamstankiewicz: Yeah, I was also deciding between just `aria-label` and something more specific to the form element (i.e., `formAriaLabel`). I opted for `formAriaLabel` because it does help describe which element the `aria-label` applies to. I'm still leaning towards the more explicit prop name despite it forcing you to know a bit about the inner-workings of the component.

The only time this prop would be used is if there are multiple instances of the `SearchField` component on the same page. The `aria-label` on the form can be used to differentiate the search landmarks.

src/SearchField/SearchFieldClearButton.jsx
adamstankiewicz: True. The previous implementation of this component removes the button from the DOM which caused styling issues. However, this time around, there are no styling issues if the button is removed the DOM altogether.

I will early return `null` here instead.

PR #544, opened 2020-06-26
Adopt react-bootstrap ADR

abutterworth:
The gist: We will include react-bootstrap components in Paragon. Paragon will continue to contain custom-written components, but for any component that exists in react-bootstrap we will use the react-bootstrap version instead of building our own.

### DIFF
--- a/docs/decisions/0009-usage-of-react-bootstrap.rst
+++ b/docs/decisions/0009-usage-of-react-bootstrap.rst
@@ -55,6 +55,7 @@ This decision will replace all components in the Paragon react library except:
 - Hyperlink
 - Icon
 - MailtoLink
+- Pagination
 - Responsive
 - SearchField
 - StatefulButton
--- a/docs/decisions/0009-usage-of-react-bootstrap.rst
+++ b/docs/decisions/0009-usage-of-react-bootstrap.rst
@@ -70,6 +70,7 @@ This decision will replace all components in the Paragon react library except:
 - TransitionReplace
 
 Following this decision there will be a pull request containing 18 components that pass our initial a11y audit:
+
 - Alerts
 - Badge
 - Buttons
--- a/docs/decisions/0009-usage-of-react-bootstrap.rst
+++ b/docs/decisions/0009-usage-of-react-bootstrap.rst
@@ -15,7 +15,11 @@ Paragon is built on top of the Bootstrap CSS Framework. The Paragon React compon
 
 ADR 5 Made the case for adoption of Reactstrap. ADR 8 abandoned the idea because adopting Reactstrap "comes at the cost of abstracting away raw html and introducing yet another API to learn and remember." Since then, we have added very few homegrown components. Designers and engineers are often needing components that exist in Bootstrap but not built as Paragon components.
 
-By not building these components ourselves and not adopting third-party React components, we are wasting time and generating confusion in the edX organization.
+By not building these components ourselves and not adopting third-party React components, we are wasting time and generating confusion in the edX organization. Examples:
+
+- The Modal component in Paragon is not flexible enough to meet the needs of our squads today. It has been a known issue for over a year, but we have been unable to prioritize it. The recent Prospectus Search page contains a custom-built Modal component used at a mobile screen size for changing filters. The Learner Portal is adding a similar component. In both cases, work done to satisfy a11y requirements needs in the original component needs to be rebuilt and reconfirmed.
+- There are several components that exist in Bootstrap that are frequently asked about (Badges, Tooltips, etc) that require repeated discussion, discovery, or custom work to support. If Paragon supported them these issues would be resolved.
+
 
 React-bootstrap is a mature component library with full coverage of Bootstrap components. Component design in react-bootstrap is similar to Paragon's more recent approach to component design reflected in Dropdown and Collapsible. We judge react-bootstrap to be a high-quality library with healthy activity.
 
--- a/docs/decisions/0009-usage-of-react-bootstrap.rst
+++ b/docs/decisions/0009-usage-of-react-bootstrap.rst
@@ -13,9 +13,9 @@ Context
 
 Paragon is built on top of the Bootstrap CSS Framework. The Paragon React component library aims to mirror Bootstrap's HTML and Javascript components. To date we have not expanded the Paragon React component library to include all Bootstrap components.
 
-ADR 5 Made the case for adoption of Reactstrap. ADR 8 abandoned the idea because adopting Reactstrap "comes at the cost of abstracting away raw html and introducing yet another API to learn and remember." Since then, we have added very few homegrown components. Designers and engineers are often needing components that exist in Bootstrap but not build as Paragon components.
+ADR 5 Made the case for adoption of Reactstrap. ADR 8 abandoned the idea because adopting Reactstrap "comes at the cost of abstracting away raw html and introducing yet another API to learn and remember." Since then, we have added very few homegrown components. Designers and engineers are often needing components that exist in Bootstrap but not built as Paragon components.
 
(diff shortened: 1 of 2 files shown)


docs/decisions/0009-usage-of-react-bootstrap.rst
```diff
@@ -0,0 +1,68 @@
+5. Usage of Reactstrap
```
adamstankiewicz, line 1: Reactstrap or react-bootstrap? 😉

abutterworth: oh god

docs/decisions/0009-usage-of-react-bootstrap.rst
```diff
@@ -0,0 +1,68 @@
...
+
+
+Context
+-------
+
+Paragon is built on top of the Bootstrap CSS Framework. The Paragon React component library aims to mirror Bootstrap's HTML and Javascript components. To date we have not expanded the Paragon React component library to include all Bootstrap components.
+
+ADR 5 Made the case for adoption of Reactstrap. ADR 8 abandoned the idea because adopting Reactstrap "comes at the cost of abstracting away raw html and introducing yet another API to learn and remember." Since then, we have added very few homegrown components. Designers and engineers are often needing components that exist in Bootstrap but not built as Paragon components.
+
+By not building these components ourselves and not adopting third-party react components, we are wasting time and generating confusion in the edX organization.
```
adamstankiewicz, line 18: nit: capitalize React

adamstankiewicz, line 18: > we are wasting time and generating confusion in the edX organization.

might recommend explaining why this is the case here with a concrete example.

docs/decisions/0009-usage-of-react-bootstrap.rst
```diff
@@ -0,0 +1,68 @@
...
+--------
+
+We will include react-bootstrap components in Paragon. Paragon will continue to contain custom-written components, but for any component that exists in react-bootstrap we will use the react-bootstrap version instead of building our own. This will provide us comprehensive coverage of Bootstrap features in react and allow us to focus on adding value by creating components outside the scope of Bootstrap.
+
+
+This will be accomplished by:
+
+1. React-bootstrap will be included as a dependency of this project. We will pin to a minor version. (Right now 1.0.x).
+
+2. Components will be passed-through from react-bootstrap by adding shell components in Paragon. (e.g. Modal.js in Paragon will simply include Modal from react-bootstrap and re-export it). This will provide us flexibility to swap the implementation in the future or enforce a11y rules if needed.
```
adamstankiewicz, line 33: 🔥

docs/decisions/0009-usage-of-react-bootstrap.rst
```diff
@@ -0,0 +1,68 @@
...
+
+3. Documentation for components included from react-bootstrap will:
+
+  - Include an interactive demo of the component in use, similar to today's components.
+  - The API will not be documented on the Paragon doc-site for now. (see next bullet)
+  - Include links to the component doc on react-bootstrap documentation site. We do this, aware that some day in the future react-bootstrap will upgrade to Bootstrap 5 and change the doc site. They did this for the switch from Bootstrap 3 to Bootstrap 4. The Bootstrap 3 doc site moved urls, but is still on the web. When that happens we will update our documentation to the new location.
+
+Other notes:
+
+- As we build new components outside the scope of Bootstrap, we will likely want to use other components to compose larger ones. For example, we may use Buttons and Cards in a CourseCard component. When we do this we must import the components from Paragon's shell components and not react-bootstrap directly. Importing from local Paragon component shells and not react-bootstrap will keep dependencies on react-bootstrap components in a single place, making future architectural changes easier.
```
adamstankiewicz, line 43: 💯

docs/decisions/0009-usage-of-react-bootstrap.rst
```diff
@@ -0,0 +1,68 @@
...
+
+- Collapsible
+- Hyperlink
+- Icon
+- MailtoLink
+- Responsive
+- SearchField
+- StatefulButton
+- ValidationFormGroup
+- TransitionReplace
```
adamstankiewicz, line 63: Should `Pagination` be included here? react-bootstrap's `Pagination` is presentation only, whereas ours is functional. That said, the UI elements of Paragon's `Pagination` could be replaced by react-bootstrap's.

abutterworth: I think our Pagination should become ResponsivePagination and use react-bootstrap under the hood. For no let's say it stays. I'll add it to the list.

docs/decisions/0009-usage-of-react-bootstrap.rst
muselesscreator: :+1:

docs/decisions/0009-usage-of-react-bootstrap.rst
abutterworth: Added examples

docs/decisions/0009-usage-of-react-bootstrap.rst
```diff
@@ -0,0 +1,73 @@
+5. Usage of React-Bootstrap
+---------------------------
+
+
+Status
+------
+
+Adopted
```
nasthagiri, line 8: `Adopted` -> `Accepted`

docs/decisions/0009-usage-of-react-bootstrap.rst
```diff
@@ -0,0 +1,73 @@
+5. Usage of React-Bootstrap
```
nasthagiri, line 1: I am overall supportive of this decision.
If this ADR is accepted, as part of this PR, please update the `Status` of ADR 0008 to `Superseded by ADR-0009`.

abutterworth: Will do.

davidjoy:
It goes without saying, but this does have the effect of tying us closer to Bootstrap and further from any sort of CSS-in-JS solution.  @abutterworth, do you think CSS-in-JS is still an important option for us?  It was going to be a heavy lift if we decided to do it, but now it'll be even heavier.  For those following along, we'd discussed CSS-in-JS as being a useful way of isolating our CSS by co-locating it with components, reducing our CSS footprint, and potentially easing theming by adopting something like [emotion-theming](https://emotion.sh/docs/theming).  

This is potentially just fine, but I wanted to bring it up for the sake of completion.

AlasdairSwan:
My only concern with this approach is related to a11y. Are there plans to audit the ```react-bootstrap``` components and identify/fix any a11y gaps?

abutterworth:
@davidjoy I think this may actually help us decouple from Bootstrap. Today, many Bootstrap components are written as html in our applications – this change would reduce that by pulling those components into Paragon. These new shim components we create to pass through react-bootstrap components can then be the single point to swap technology for any future innovation that takes us beyond Bootstrap.

I think CSS-in-JS is a good future for us. If we choose to adopt it we should do it iteratively (and therefore in a way that is compatible with Bootstrap being present on the page). I imagine a future where we adopt new technology by replacing our react-bootstrap shims 1 by 1 (removing sections of bootstrap scss along the way).

abutterworth:
@AlasdairSwan I'll reach out to @wittjeff to request reviews of these components as we pull them in.

abutterworth:
@wittjeff has done a quick a11y sniff test on the react-bootstrap components here: https://docs.google.com/spreadsheets/d/1FR42auyyFPfOLenlcj6OLuYZ7NU0aBTZvYIM58iSElQ/edit#gid=0

I'll update this PR to reflect the following plan:
- For the 19 components that pass our initial audit we will pull them in
- For the 8 that do not we will make upstream changes. Or, if they are needed urgently we will enforce a11y requirements in the shim component.

PR #550, opened 2020-07-15
feat: add accessible react-bootstrap components

abutterworth:
Adds the following react-bootstrap components with jump off documentation pages:

- Alert
- Badge
- Button Group
- Card
- Carousel
- Figure
- Forms
- Input Group
- Image
- Navs
- Navbar
- Overlays
- Popover
- ProgressBar
- Spinner
- Tooltip
- Toast

Sample Doc:

Screen Shot 2020-07-16 at 5 24 01 PM


src/index.js
```diff
@@ -30,9 +30,118 @@ export { default as SearchField } from './SearchField';
...
+export { default as Badge } from 'react-bootstrap/Badge';
+/*
+  TODO: Breadcrumb
+  a11y defect: aria-current (needed semantics) is on the <li>, not on the <a>, which takes focus
+
+export { default as Breadcrumb } from 'react-bootstrap/Breadcrumb';
+export { default as BreadcrumbItem } from 'react-bootstrap/BreadcrumbItem';
+*/
+// TODO: create shim – export { default as Button } from 'react-bootstrap/Button';
+export { default as ButtonGroup } from 'react-bootstrap/ButtonGroup';
```
davidjoy, line 65: Are these Button* components compatible with our Button above, that we have yet to shim?

davidjoy:
Two thoughts - 

One, I'm curious how a lot of these interact with our existing Paragon components when they're used in sensible ways that folks might expect (our input classes inside their `Form`, for instance)

Two, documenting all this will be interesting.  In the past we chose to duplicate/wrap all the Bootstrap documentation... are we going to reproduce react-bootstrap's?  Or just reference it from the paragon documentation?  Seems like we're going to need to have a big shift in the doc around which components are deprecated... are we doing a major version change and just removing stuff?  And in reference to the first thought above, documenting any subtlety to those interactions will be fun.

davidjoy:
And also, just a reminder to check on tree-shaking to make sure they've packaged react-bootstrap in such a way that it can still be slimmed down when used through Paragon.

davidjoy:
Also - after having investigated react-bootstrap quite a bit now, how many cases do we see where we want to 'wrap' things?  Have we seen any specific benefits to exposing them _through_ paragon?  

Just thinking about whether or not it'd be a reasonable choice to make react-bootstrap a recommended library and call it a day. 🤔 (not sure I think this is a good idea)

src/index.js
abutterworth: They are! I tested them in the doc page just added.

abutterworth:
Great points David.

- Interaction between components: So far it seems like they're pretty interoperable. I know ValidationFormGroup will not do its job with the react-bootstrap components. I intend to mark the components that I know won't work together. As far as I'm aware right now, that's the only one. I may find more as I continue building the docs.
- Documentation: I'm adding one or two examples of usage and then linking to the react-bootstrap docs themselves. I think this is the best blend between convenience and not doing extra work
- Tree-shaking: I will double check in a project, but judging by the structure I'm confident it will tree-shake fine.

Why pull it in? Incorporating react-bootstrap inside paragon will help us:
1. maintain flexibility in the future by making Paragon the single source of truth. We can replace react-bootstrap components with our own in the future if we choose. 
2. By using Paragon to pull in these components we can leverage them internally in Paragon without worrying about peer deps or versions of react-bootstrap. 

In general I expect the future of Paragon to aim to create more "molecule-level" components that are composed of these react-bootstrap elements and in the further future we can swap the underlying technology with CSS-in-JS iteratively.

src/index.js
```diff
@@ -1,10 +1,10 @@
 export { default as asInput } from './asInput';
 export { default as Breadcrumb } from './Breadcrumb';
-export { default as Button } from './Button';
+export { default as Button } from './Button'; // TODO: shim from react-bootstrap
 export { default as CheckBox } from './CheckBox';
 export { default as CheckBoxGroup } from './CheckBoxGroup';
 export { default as Collapsible } from './Collapsible';
-export { default as Dropdown } from './Dropdown';
+export { default as Dropdown } from './Dropdown'; // TODO: shim from by react-bootstrap
```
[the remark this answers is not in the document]
abutterworth: It is, I'll prioritize the building the shim for this one first.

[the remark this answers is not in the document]
abutterworth: from by... smh :P.  Fixed now, thank you.

davidjoy:
Thanks for indulging the second-guessing questions, @abutterworth.  Explanations totally make sense/I'm behind this!

src/index.js
```diff
@@ -1,10 +1,10 @@
 export { default as asInput } from './asInput';
 export { default as Breadcrumb } from './Breadcrumb';
-export { default as Button } from './Button';
+export { default as Button } from './Button'; // TODO: shim from react-bootstrap
```
davidjoy, line 3: Are these TODOs something that should be completed in this PR?  Or for later?

abutterworth: For follow up PRs

abutterworth: Adding the shims is more risky in terms of inadvertently making breaking changes so I'll keep them separate.

www/src/pages/components/statusalert.mdx
```diff
@@ -1,13 +1,11 @@
 ---
 title: 'StatusAlert'
 type: 'component'
-status: 'Needs Work'
+status: 'Deprecate Soon'
```
davidjoy, line 4: Feel like we have so many variations on this alert floating around in the MFEs. :)

www/src/pages/components/image.mdx
```diff
@@ -0,0 +1,45 @@
...
+<Container>
+  <Row>
+    <Col xs={6} md={4}>
+      <Image src={generateCustomPlaceholderURL(180, 180)} rounded />
+    </Col>
+    <Col xs={6} md={4}>
+      <Image src={generateCustomPlaceholderURL(180, 180)} roundedCircle />
+    </Col>
+    <Col xs={6} md={4}>
+      <Image src={generateCustomPlaceholderURL(180, 180)} thumbnail />
```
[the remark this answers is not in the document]
abutterworth: I was trying to figure out what fluid actually does and couldn't see any change there. What's it for?

src/index.js
```diff
@@ -30,9 +30,111 @@ export { default as SearchField } from './SearchField';
...
+export { default as Nav } from 'react-bootstrap/Nav';
+export { default as Navbar } from 'react-bootstrap/Navbar';
+export { default as NavbarBrand } from 'react-bootstrap/NavbarBrand';
+export { default as NavDropdown } from 'react-bootstrap/NavDropdown';
+export { default as NavItem } from 'react-bootstrap/NavItem';
+export { default as NavLink } from 'react-bootstrap/NavLink';
+export { default as Overlay } from 'react-bootstrap/Overlay';
+export { default as OverlayTrigger } from 'react-bootstrap/OverlayTrigger';
+export { default as PageItem } from 'react-bootstrap/PageItem';
+// TODO: create shim – export { default as Pagination } from 'react-bootstrap/Pagination';
```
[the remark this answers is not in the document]
abutterworth: Good call out, with Pagination in particular. What do you think about a future breaking change that pulls in react-bootstrap's pagination and moving the purpose of the current Pagination to a PaginationGroup or something?

PR #551, opened 2020-07-17
docs: update layout docs

abutterworth:
image

### DIFF
--- a/www/src/pages/foundations/layout.mdx
+++ b/www/src/pages/foundations/layout.mdx
@@ -18,7 +18,7 @@ Paragon's layout is controlled by the Bootstrap grid system. Documentation can b
 
 <p>
   These components are pass throughs from React-Bootstrap.<br/>
-  <a href="https://react-bootstrap.github.io/components/layout/" target="_blank">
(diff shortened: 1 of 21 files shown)


www/src/pages/foundations/layout.mdx
```diff
@@ -10,20 +10,65 @@ import LayoutGenerator from '../../components/LayoutGenerator'
...
-## Grid 
+Paragon's layout is controlled by the Bootstrap grid system. Documentation can be found here:
+[https://getbootstrap.com/docs/4.5/layout/grid/](https://getbootstrap.com/docs/4.5/layout/grid/)
+
+
+## Components
+
+<p>
+  These components are pass throughs from React-Bootstrap.<br/>
+  <a href="https://react-bootstrap.github.io/components/layout/" target="_blank">
```
adamstankiewicz, line 21: I should have captured this on the main PR, but links with `_blank` should also have `rel="noopener noreferrer"`. might be worthwhile the `rel` attribute to those links in this PR, as well.

abutterworth: oh right (why is the web broken)

PR #553, opened 2020-08-03
feat: add button, dropdown, and tabs from react-bootstrap

abutterworth:
This PR contains breaking changes to the following components:

- `Button`
- `StatefulButton`
- `Dropdown`
- `Tabs`

Button, Dropdown, and Tabs offer the deprecated components they replace with a "deprecated" subcomponent:

- `<Button.Deprecated>`
- `<Dropdown.Deprecated>`
- `<Tabs.Deprecated>`

The `StatefulButton` is the only component for which **usage must be updated when upgrading to this version of Paragon**. The key change is the transition from defining bootstrap button class names directly via `className="btn-primary"` to leveraging the new `variant="primary"` prop.

---

Screen Shot 2020-08-03 at 3 27 13 PM

**BREAKING CHANGE:** `<Button />` has been replaced by the react-bootstrap `<Button />`. The old implementation is available at `<Button.Deprecated />`. In the old component the styling was controlled via className directly (e.g. `className="btn-primary"`). The new component controls this via a `variant` prop (e.g. `variant="primary"`).

Old Button usage:

`<Button className="btn-primary">Primary</Button>`

New Button usage:

`<Button variant="primary">Primary</Button>`

---

**BREAKING CHANGE:** `<StatefulButton />` is now underpinned by the react-bootstrap `<Button />`. Styling `StatefulButton` was previous controlled via className directly (e.g. `className="btn-primary"`). The new component controls this via a `variant` prop (e.g. `variant="primary"`).

Old StatefulButton usage:

`<StatefulButton className="btn-success" labels={{ default: 'Save', pending: 'Saving' }} />`

New StatefulButton usage:

`<StatefulButton variant="success" labels={{ default: 'Save', pending: 'Saving' }} />`

---

Screen Shot 2020-08-03 at 3 27 19 PM

**BREAKING CHANGE:** `<Dropdown />` has been replaced by the react-bootstrap `<Dropdown />`. The old implementation is available at `<Dropdown.Deprecated />`. The subcomponents that are used to assemble the dropdown have changed. Refer to documentation for the new usage.

Old usage for Dropdown:

```
<Dropdown>
  <Dropdown.Button>
    Search Engines
  </Dropdown.Button>
  <Dropdown.Menu>
    <Dropdown.Item href="https://google.com">Google</Dropdown.Item>
    <Dropdown.Item href="https://duckduckgo.com">DuckDuckGo</Dropdown.Item>
    <Dropdown.Item href="https://yahoo.com">Yahoo</Dropdown.Item>
  </Dropdown.Menu>
</Dropdown>
```

the new usage for Dropdown:

```
<DropdownButton variant="success" title="dropdown">
  <Dropdown.Item href="#/action-1">Action</Dropdown.Item>
  <Dropdown.Item href="#/action-2">Another action</Dropdown.Item>
  <Dropdown.Item href="#/action-3">Something else</Dropdown.Item>
</DropdownButton>
```

---

Screen Shot 2020-08-03 at 3 27 25 PM

**BREAKING CHANGE:** `<Tabs />` has been replaced by the react-bootstrap `<Tabs />`. The old implementation is available at `<Tabs.Deprecated />`. The old implementation matched the order of labels with the order of children. The new implementation uses the `Tab` child component's `title` prop to create labels dynamically.

Old usage:

```
<Tabs labels={['Panel 1', 'Panel 2', 'Panel 3']}>
  <div>Hello I am the first panel</div>
  <div>Hello I am the second panel</div>
  <div>Hello I am the third panel</div>
</Tabs>
```

New usage:

```
<Tabs defaultActiveKey="profile" id="uncontrolled-tab-example">
  <Tab eventKey="home" title="Home">Hello I am the first panel.</Tab>
  <Tab eventKey="profile" title="Profile">Hello I am the second panel.</Tab>
  <Tab eventKey="contact" title="Contact" disabled>Hello I am third first panel.</Tab>
</Tabs>
```


src/index.js
```diff
@@ -1,10 +1,31 @@
...
+import ButtonDeprecated from './Button';
+
+import Dropdown from 'react-bootstrap/Dropdown';
+import DropdownDeprecated from './Dropdown';
+
+import Tabs from 'react-bootstrap/Tabs';
+import TabsDeprecated from './Tabs';
+
+// Add deprecated components as subcomponents of their replacements
+Button.Deprecated = ButtonDeprecated;
```
davidjoy, line 14: This seems like it may cause confusion... if I want the deprecated Button, I need to import Button from this index file (rather than the sub-directory), and only then will I find the `Deprecated` implementation hanging off of it... Would it be better for the `Button` sub-directory to publish the react-bootstrap Button and in `Button/index.js` attach this Deprecated version there?  That way when it's imported here and re-exported, it just works and is consistent.

davidjoy: Put another way - the button sub-directory can be responsible for its exports - all of them.  If it wants a `Deprecated` button hanging off of it, it can do it itself and this file (or any other consumers) will be none the wiser about where _either_ implementation came from.

src/StatefulButton/index.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import Icon from '../Icon';
-import Button from '../Button';
+import { Button } from '../';
```
davidjoy, line 5: Hmm, how does this impact tree-shaking?  Since this is pulling from the top-level index.js file, won't that cause all the imports from that file to get included in the bundle?

src/StatefulButton/__snapshots__/StatefulButtontest.test.jsx.snap
```diff
@@ -5,10 +5,9 @@ exports[`StatefulButton renders basic usage 1`] = `
   <button
     aria-disabled={false}
     aria-live="assertive"
-    className="btn pgn__stateful-btn pgn__stateful-btn-state-default btn-primary mr-2"
-    onBlur={[Function]}
```
davidjoy, line 9: A little confused why the blur/key down handlers would disappear and disabled = false would appear... Doesn't that mean we're not properly passing through our handlers?

www/src/pages/components/button.mdx
```diff
@@ -21,51 +22,125 @@ import SingleComponentStatus from '../../components/SingleComponentStatus';
...
 />
 
+<p style={{ fontSize: '1.5em' }}>
+  This is a pass through component from React-Bootstrap.<br/>
+  <a href="https://react-bootstrap.github.io/components/cards/" target="_blank" rel="noopener noreferrer">
+    See React-Bootstrap for documentation.
+  </a>
+</p>
+
+##### Basic Usage
```
davidjoy, line 32: Do we need to maintain this documentation for `Button` now that it's from react-bootstrap?

www/src/pages/components/button.mdx
```diff
@@ -21,51 +22,125 @@ import SingleComponentStatus from '../../components/SingleComponentStatus';
...
+
+***
+
+## Button/Deprecated
+
 ##### Basic Usage
 
 ```jsx live=true
-<Button className="btn-primary">Hello World!</Button>
+<Button.Deprecated className="btn-primary">Hello World!</Button.Deprecated>
```
davidjoy, line 102: Ah, I see the doc for the Deprecated one here - yeah, this part should stay at least.

src/StatefulButton/__snapshots__/StatefulButtontest.test.jsx.snap
```diff
@@ -5,10 +5,9 @@ exports[`StatefulButton renders basic usage 1`] = `
...
+    className="pgn__stateful-btn pgn__stateful-btn-state-default btn-primary mr-2 btn btn-primary"
```
adamstankiewicz, line 8: btn-primary is listed twice here, as well.

src/index.js
abutterworth: Good point, I'll refactor to your suggestion.

src/StatefulButton/index.jsx
abutterworth: I don't think this is the case. The es6 exports from that index.js are atomic no? Importing it here is the same as importing Button in a project (right?).

src/StatefulButton/__snapshots__/StatefulButtontest.test.jsx.snap
abutterworth: The old button had noop functions as default prop values.

www/src/pages/components/button.mdx
abutterworth: The react-bootstrap documentation here is just for ease of consumption so we're not kicking devs out to external docs for the majority need.

src/StatefulButton/__snapshots__/StatefulButtontest.test.jsx.snap
abutterworth: Thanks, I think I know why this is the case and will fix.

www/src/pages/components/dropdown.mdx
```diff
@@ -1,11 +1,11 @@
 ---
 title: 'Dropdown'
 type: 'component'
-status: 'Needs Work'
+status: 'Stable'
```
stvstnfrd, line 4: :)

src/index.js
abutterworth: Done! Please take a look.

www/src/pages/components/dropdown.mdx
abutterworth: ;)

src/StatefulButton/index.jsx
davidjoy: That wasn't my understanding... like, I thought it basically imports the tree from whatever entry point you choose.  So if you import from Button itself, it'll only get it's dependencies, but if you import from the top-level index I thought you'd get everything that file depends on... maybe I'm mis-remembering?

abutterworth: We discussed over slack, but for posterity we think we're good here: https://webpack.js.org/guides/tree-shaking/ 

```
So, what we've learned is that in order to take advantage of tree shaking, you must...

- Use ES2015 module syntax (i.e. import and export).
- Ensure no compilers transform your ES2015 module syntax into CommonJS modules (this is the default behavior of the popular Babel preset @babel/preset-env - see the documentation for more details).
- Add a "sideEffects" property to your project's package.json file.
- Use the production mode configuration option to enable various optimizations including minification and tree shaking.

PR #555, opened 2020-08-12
Icon button

raq929:
OK! This got bigger than expected. 
https://openedx.atlassian.net/jira/software/c/projects/ENT/issues/ENT-3311
Styles: 
https://www.figma.com/file/wC5LTEbZaFqHmLKvM1owoM/Paragon-2019?node-id=73%3A386

This component came out of a component that I added to frontend-app-admin-portal, and subsequent discussions with Gabe Weinberg about a similar component that was planned.
It's an icon component with (optional) tooltip and (optional) click handler. If it's a given a click handler,  it's given the button role and responds to the user pressing spacebar when the element is focused (which is how buttons are supposed to behave for screen readers). 

This PR adds several hooks to Paragon, though for now we're only exposing useWindowSize. I'm open to exposing the others, but less immediately sure of their utility. Let me know about folder structure for hooks... 

TESTING:
I'm having trouble with testing for an element in focus. I think the next step is to mock the hooks instead of testing the focus directly? But other thoughts are welcome. (There's one skipped test in that vein.)
Something I changed is making a Modal test fail, but I'm not sure what.
I'm working on figuring out how to test the hooks.

PACKAGES ADDED:
* eslint addon for hooks that has been super useful.
* `to-style`, `change-case` and `husky` because I can't seem to get the site running without having them explicitly added. I think I may want @abutterworth to help me figure out what's going on and see if they can be removed.
* fontawesome packages were added to `www/package.json`. I'm going to investigate if some of them can be removed; I suspect yes but I'm currently dealing with a different error when trying to run the server.


.eslintrc
```diff
@@ -1,5 +1,8 @@
 {
-  "extends": "eslint-config-edx",
+  "extends": [
+    "eslint-config-edx",
+    "plugin:react-hooks/recommended"
```
abutterworth, line 4: It's worth asking in FEDX if we would like to add this to our common eslint config: https://github.com/edx/eslint-config/blob/master/.eslintrc.js

package.json
```diff
@@ -28,9 +28,9 @@
     "test": "jest --coverage"
   },
   "dependencies": {
-    "@fortawesome/fontawesome-svg-core": "^1.2.21",
-    "@fortawesome/free-solid-svg-icons": "^5.10.1",
-    "@fortawesome/react-fontawesome": "^0.1.4",
+    "@fortawesome/fontawesome-svg-core": "^1.2.30",
```
abutterworth, line 31: Any specific reason to upgrade font awesome?

src/IconButton/IconButton.scss
```diff
@@ -0,0 +1,41 @@
+$iconbutton-hover-size: 1.6em;
```
abutterworth, line 1:
```suggestion
$iconbutton-hover-size: 1.6em !default;
```

Add a default here in case a consumer of paragon would change this in a theme

src/IconButton/IconButton.scss
```diff
@@ -0,0 +1,41 @@
...
+    width: $iconbutton-hover-size;
+    height: $iconbutton-hover-size;
+    border-radius: $iconbutton-hover-size;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    text-align: center;
+    position: relative;
+
+    &:hover{
```
abutterworth, line 14:
```suggestion
    &:hover {
```

src/IconButton/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+  iconClassNames,
+  onClick,
+  popoverPlacementLg,
+  popoverPlacementSm,
+  popoverText,
+  popoverTitle,
+  popoverTrigger,
+}) => {
+  const windowSize = useWindowSize();
+  const placement = windowSize.width >= breakpoint ? popoverPlacementLg : popoverPlacementSm;
```
abutterworth, line 22: Can we get this logic done with popper config on the popover? Ref: https://popper.js.org/docs/v2/modifiers/flip/#fallbackplacements

src/hooks/useWindowSize.jsx
```diff
@@ -0,0 +1,34 @@
...
+    function handleResize() {
+      // Set window width/height to state
+      setWindowSize({
+        width: window.innerWidth,
+        height: window.innerHeight,
+      });
+    }
+
+    // Add event listener
+    window.addEventListener('resize', handleResize);
```
abutterworth, line 22: Avoid using `window` and instead use `global`. This change preserves compatibility with server-rendering or some pre-rendering done by Gatsby.

www/package.json
```diff
@@ -43,9 +44,10 @@
   "devDependencies": {
     "babel-eslint": "^10.0.1",
     "css": "^2.2.4",
-    "eslint": "^5.16.0",
+    "eslint": "^7.6.0",
     "eslint-config-edx": "^4.0.4",
-    "eslint-import-resolver-alias": "^1.1.2"
+    "eslint-import-resolver-alias": "^1.1.2",
+    "husky": "^4.2.5"
```
abutterworth, line 50: Is this needed?

src/hooks/useActiveKeyPress.jsx
```diff
@@ -0,0 +1,34 @@
...
+      }
+    }
+
+    // If released key is our target key then set to false
+    const upHandler = ({ key }) => {
+      if (key === targetKey) {
+        setKeyPressed(false);
+      }
+    };
+    const elem = document.getElementById(focusId);
```
abutterworth, line 22: The use of an id to get a reference to an element seems problematic in the context of React. We should rework this to use `ref` (https://reactjs.org/docs/refs-and-the-dom.html) if we need to attach event handlers manually like this.

That said, this work is best not abstracted into a hook. React will help us attach an remove event listeners. Let's rely on React to do the work. I'll make an alternative suggestion in Interactive button.

src/IconButton/InteractiveIcon.jsx
```diff
@@ -0,0 +1,66 @@
...
+  const spacePressed = useActiveElementKeypress(focusId, ' ');
+
+  // if the element has a click handler and is therefore acting like a button, the user should
+  // be able to be click it by pressing the space bar
+  if (onClick && spacePressed) {
+    onClick();
+  }
+
+  return (
+    <div id={focusId} className={`iconbutton-hover ${darkHoverClass}`} {...buttonProps}>
```
abutterworth, line 38: If you use the `button` element pressing space on the keyboard will trigger the onClick (right?). Any reason not to use the button element?

abutterworth: Rather than the hook `useActiveElementKeypress` this is probably a better way to reproduce the button spacebar behavior:

```
const handleKeyDown = (e) => {
  if (e.code === 'Space') {
   onClick();
 }
// If a handler is supplied, allow it to handle the event too
props.onKeyDown && props.onKeyDown(e);
};
```
and in the component
```
<div ...
  onKeyDown={handleOnKeyDown}
>
```

src/IconButton/InteractiveIcon.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import useActiveElementKeypress from '../hooks/useActiveKeyPress';
+import newId from '../utils/newId';
+
+const iconPrefix = 'interactive-icon';
+
+const InteractiveIcon = React.memo(({
```
abutterworth, line 9: Why use `React.memo` here?

package.json
raq929: No, I think it happened in one of my many many many `npm install`s, and I saw no reason why not?

src/IconButton/IconButton.scss
raq929: Oh, cool, I didn't know about `!default`. :)

src/IconButton/IconButton.scss
raq929: Oh, ha, we should maybe have a sass linter...

src/IconButton/index.jsx
raq929: oooo, yeah, that looks cool. Can I chat with you about how to use it in conjunction with Bootstrap, if that's something you've done before?

www/package.json
raq929: Probably not. I'll remove.

src/IconButton/InteractiveIcon.jsx
raq929: Oh, I was having a weird thing happen where it was re-rendering too often, and this was one of the suggestions I found for fixing it. But I'm pretty sure something else fixed it in the end; I'll remove.

PR #557, opened 2020-08-13
Improve developer documentation in readme

abutterworth:
Also adds a stub page in the documentation for contributors to use as a starting point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,8 +117,8 @@ To see your component in action, you need to run the documentation site locally 
 
 1. Make a copy of `www/src/pages/components/_my-component.mdx` and rename to match your component. Use kebab-case.
 
-2. Add your new page to the navigation by making an addition to `www/src/components/navigation.jsx`. Note that the url of your new page is determined by its file name. A file at `www/src/pages/components/my-new-component.mdx` would generate a page at the path `components/my-new-component`.
-  You should see your changes reflected at [http://localhost:8000](http://localhost:8000).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -61,27 +61,66 @@ Usage on edx.org:
...
 
-To see your component in action, you will need to create a new Storybook story for it. Create a file within your component's directory named `<ComponentName>.stories.js`. Check out [InputSelect.stories.jsx](https://github.com/edx/paragon/blob/master/src/InputSelect/InputSelect.stories.jsx) for an example of some good stories. Storybook will automatically pick up this file and serve its stories -- you should be able to see them linked from the left-hand menu.
+3. Use your new doc page as the workbench for your component. It will auto refresh as you make changes.
 
-Make sure to define PropTypes and DefaultProps on your components, using the [prop-types package](https://github.com/facebook/prop-types). PropTypes provide a clear API for your component and help consumers invoke it properly within their own code. In terms of functionality, ship an MVP component up-front. Don't make it do any more than is necessary. It's easy to add new functionality later on as needed, but it's much harder to remove functionality once we've released a component.
```
raq929, line 84: I think we're still using PropTypes and Default props, so we might want to leave this in?

abutterworth: We are, but it's standard practice and enforced by our linter. This doc was written when React was very new. I want to keep our Readme succinct where we can.

PR #558, opened 2020-08-14
docs: add adr for removal of edx assets

abutterworth:
This ADR proposes a restructuring of this project. [See the rst file for details.](https://github.com/edx/paragon/blob/835918d02d7f950e89561a7d85f707b6bc46046e/docs/decisions/0010-removal-of-edx-assets.rst)

### DIFF
--- a/docs/decisions/0010-removal-of-edx-assets.rst
+++ b/docs/decisions/0010-removal-of-edx-assets.rst
@@ -53,7 +53,10 @@ edX Brand:
 Merges to edX Brand trigger work to:
   - Build and deploy a version of the Paragon documentation with the edX.org theme at https://paragon.edx.org
 
-This structure has pros and cons as well:
+Consequences
+------------
+
+This new structure has pros and cons as well:
 
 Pros:
   - The edX theme is decoupled from Paragon. Adopting theme updates is easier for teams.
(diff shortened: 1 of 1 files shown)


docs/decisions/0010-removal-of-edx-assets.rst
```diff
@@ -0,0 +1,76 @@
...
+
+Pros:
+  - The edX theme is decoupled from Paragon. Adopting theme updates is easier for teams.
+  - The default and edX themes are both documented.
+  - There is a new common location for brand assets.
+  - Paragon is an independent open source library
+
+Cons:
+  - Two documentation sites one for edX and one for Paragon makes for some additional CI complexity.
+  - edX.org theme changes may be more difficult to see during development. We will need to determine an efficient way to do this.
```
muselesscreator, line 66: This is my biggest worry... that this will make developing components with final theme in mind more complicated...

abutterworth: Fair concern. I think this can be mitigated (https://github.com/edx/frontend-build/pull/103) and is worth the trade off. 

Also, right now it's difficult for engineers to know what the component looks like without a theme. So we have the reverse challenge (though not hard to test out).

docs/decisions/0010-removal-of-edx-assets.rst
```diff
@@ -0,0 +1,76 @@
...
+Cons:
+  - edX theme changes are coupled with Paragon component changes. This can block teams from upgrading the edX theme in their projects.
+  - The default theme is not represented in documentation.
+  - This project is open source but not for general use by default.
+
+
+Decision
+--------
+
+We will create a new project for edX.org branding. It will include the edX theme and other edX brand assets. Paragon will no longer have any specific tie to edX,org visuals. The new structure
```
adamstankiewicz, line 37: nit: edx.org

docs/decisions/0010-removal-of-edx-assets.rst
abutterworth: Updated with more detail in the consequences section

PR #568, opened 2020-09-10
docs(buttons): use variant prop for buttons in live code snippets

adamstankiewicz:


www/src/pages/components/modal.mdx
```diff
@@ -63,12 +63,12 @@ class ModalWrapper extends React.Component {
           parentSelector={this.props.parentSelector}
           onClose={this.resetModalWrapperState}
           buttons={[
-            <Button buttonType="success">Green button!</Button>
+            <Button variant="success">Green button!</Button>
```
abutterworth, line 66: Modal still uses the old button I think

abutterworth: I take it back :P it looks like it's using the new one.

abutterworth (approved):
I think if you

muselesscreator (commented):
Nice concise change.  LGTM

PR #570, opened 2020-09-18
docs: make modal documentation interactive

abutterworth:
Make the Modal doc interactive again by updating the example code's use of refs.

Please note that it's expected the the netlify/paragon-edx/deploy-preview will fail. It depends upon a new script `build:with-theme` that's being added in https://github.com/edx/paragon/pull/573


muselesscreator (approved):
:-) :ship: :it:

PR #573, opened 2020-09-18
Remove edx.org theme from docs

abutterworth:
This may be easiest to digest commit by commit.

In the Paragon package: 
- Moves mobile typography extensions from the edx theme into the Paragon core SCSS.

In the documentation site:
- Remove edx.org specific documentation or styling
- Update the documentation for including Paragon to be edX agnostic. Remove the concept of a single include `theme.scss` and instead to `theme/fonts.scss, theme/variables.scss, core.scss, theme/overrides.scss`. This helps decouple themes from Paragon and its versioning.
- Create two `develop` and `build` commands. By default the site will display no theme. To see the edx theme applied run `npm run develop:with-theme`. A follow up PR will reference the edx.org theme external to this project rather than the one in `/src`.

Note the two deploy previews:
- https://deploy-preview-573--paragon-openedx.netlify.app/ (default)
- https://deploy-preview-573--paragon-edx.netlify.app/ (edx.org theme)

Note that https://edx.github.io/paragon/ will no longer reflect the edx.org theme and we may wish to configure the edx.org themed documentation to deploy to https://paragon.edx.org.


www/src/pages/index.mdx
```diff
@@ -41,38 +41,21 @@ import { ComponentName } from '@edx/paragon';
...
 ```
 
-Usage on edx.org:
+Usage with a theme:
 
 ```scss
-@import "~@edx/paragon/scss/edx/theme.scss';
-@import '~@edx/paragon/scss/edx/fonts.scss'; // Roboto
+@import "@my-brand/fonts.scss";
+@import "@my-brand/variables.scss";
```
davidjoy, line 54: Is it worth talking about the difference between SCSS variables - and how first-one-in-wins - vs the overrides and how they go below the core?  A sentence or two might go a long way to improving and solidifying peoples' mental models.  I know I personally find a it a bit confusing.

scss/example-theme/_overrides.scss
```diff
@@ -0,0 +1,2 @@
+// Use this file to define style overrides for @edx/paragon
```
davidjoy, line 1: Do we want to include this here?  Or encourage people to make a brand repo?  Isn't adding code here sorta the 'wrong' way to do it, since they're effectively editing the paragon repository instead of creating a brand to alias?

www/src/scss/stub-theme/_overrides.scss
```diff
@@ -0,0 +1,2 @@
+// Use this file to define style overrides for @edx/paragon
```
davidjoy, line 1: Do we want to include this here? Or encourage people to make a brand repo? Isn't adding code here sorta the 'wrong' way to do it, since they're effectively editing the paragon repository instead of creating a brand to alias?

www/gatsby-node.js
```diff
@@ -29,8 +29,8 @@ function createCssUtilityClassNodes({ actions, createNodeId, createContentDigest
   // We convert to CSS first since we prefer the real values over tokens.
   const compiledCSS = sass
       .renderSync({
-        file: path.resolve(__dirname, '../scss/edx/utilities-only.scss'),
-        // Resolve tildes the way webpack would in our base npm project 
+        file: path.resolve(__dirname, '../scss/core/utilities-only.scss'),
```
raq929, line 37: I'm seeing that `edx/utilities-only.scss` was deleted in this commit, but not that it was added in core. Was it preexisting file there?

www/src/scss/stub-theme/_overrides.scss
davidjoy: Oh, I guess it's an example/stub theme... question still sorta stands, but in the sense that maybe it should just direct folks to instructions on how to pull this example/stub out and use it 'properly'...

www/README.md
```diff
@@ -18,9 +18,21 @@ npm start
...
 
-MDX is a flavor of Markdown that affords writing JSX directly inside markdown files. For more information visit the website: [https://mdxjs.com/](https://mdxjs.com/). 
+1. Run
+```sh
+npm run develop:with-theme
+```
+
+(Coming soon) Control the theme used by installing a custom brand under the npm alias `@edx/brand`.
+```sh
+npm i @edx/brand@file:~/my-brand
```
davidjoy, line 30: FWIW - we may be able to use the same mechanism frontend-build exposes for doing overrides via module.config.js, if we want to make this feel a little more friendly.

www/src/scss/stub-theme/_overrides.scss
raq929: Personally, I like having the example/stub themes; it shows me exactly where things need to be in the file structure in a way that is often less intuitive when I'm just reading the docs.

davidjoy (approved):
A few questions, but in general looks like a good improvement.  Approving, since even if my thoughts are worth doing something about, that can easily be done in follow-up PRs.

package.json
```diff
@@ -17,6 +17,7 @@
   "scripts": {
     "build": "make build",
     "build-docs": "cd ./www && \"$npm_execpath\" install && \"$npm_execpath\" run build",
+    "build-docs:with-theme": "cd ./www && \"$npm_execpath\" install && \"$npm_execpath\" run build:with-theme",
```
raq929, line 20: If this is working correctly, should we remove the `(coming soon)` comment in earlier documentation?

www/src/scss/stub-theme/_overrides.scss
abutterworth: The comment may be wrong. The reason this is here is because in order to swap themes I need to include something even if in the default case I don't want it. Maybe `dummy-theme` is a better name.

raq929 (approved):

www/README.md
abutterworth: That would be nice – at the moment Paragon isn't using frontend-build so a good improvement to keep an eye on.

www/gatsby-node.js
abutterworth: Yes it was already there.

package.json
abutterworth: Currently the theme to be used is hard-coded to use the edx theme baked into the project: https://github.com/edx/paragon/pull/573/files#diff-fb043578f52339bcb2a2d2b4df1203c8R7-R10

This will change with the creation of the @edx/brand-edx.org repository.

PR #577, opened 2020-10-01
fix(modal): focus trapping with `react-focus-on` and `tabbable`

adamstankiewicz:
This PR updates the `Modal` component to:
* better handle focus management through the use of `react-focus-on` (and `react-focus-lock` under the hood). 
* re-order the default modal footer close button with the custom `buttons` passed in via a prop.
* fixes the docs site for the `Modal` page.
* removes a deprecated lifecycle method.

[Deploy Preview](https://deploy-preview-577--paragon-edx.netlify.app/components/modal)

### DIFF
--- a/scss/edx/overrides/_modals.scss
+++ b/scss/edx/overrides/_modals.scss
@@ -4,6 +4,6 @@
 }
 
 .modal-footer {
-  padding-bottom: 1.75rem;
+  padding-bottom: 0.75rem;
   padding-top: 0.75rem;
 }
--- a/src/Modal/Modal.scss
+++ b/src/Modal/Modal.scss
@@ -41,7 +41,12 @@
 }
 
 .modal-content {
-  max-height: 100%;
+  max-height: calc(100vh - (#{$spacer} * 2));
+
+  &:focus {
+    outline: 1px dotted;
+    outline: 5px auto -webkit-focus-ring-color;
+  }
 }
 
 .modal-header {
--- a/src/Modal/Modal.test.jsx
+++ b/src/Modal/Modal.test.jsx
@@ -18,7 +18,7 @@ const defaultProps = {
   title,
   body,
   open: true,
-  onClose: () => {},
+  onClose: () => { },
 };
 const closeText = 'GO AWAY!';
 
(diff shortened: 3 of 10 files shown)


scss/edx/overrides/_modals.scss
```diff
@@ -4,6 +4,6 @@
 }
 
 .modal-footer {
-  padding-bottom: 1.75rem;
+  padding-bottom: 0.75rem;
```
adamstankiewicz, line 7: the edX theme causes more spacing on the bottom of the modal footer than should be there (e.g., below any buttons).

src/Modal/Modal.test.jsx
```diff
@@ -221,38 +221,11 @@ describe('<Modal />', () => {
       wrapper.find('button').at(0).simulate('click');
       modalOpen(false, wrapper);
       wrapper.setProps({ open: true });
+      wrapper.update();
       modalOpen(true, wrapper);
     });
   });
-  describe('invalid keystrokes do nothing', () => {
```
adamstankiewicz, line 227: now that the `Modal` uses [react-focus-on](https://www.npmjs.com/package/react-focus-on) (and react-focus-lock under the hood) in favor of custom keydown events, these tests are no longer relevant.

src/Modal/index.jsx
```diff
@@ -14,58 +16,42 @@ class Modal extends React.Component {
...
-    this.setCloseButton = this.setCloseButton.bind(this);
 
-    this.closeModalButtonId = newId('paragonCloseModalButton');
     this.headerId = newId();
-    this.el = document.createElement('div');
+    this.modalBodyRef = React.createRef();
 
-    // Sets true for IE11, false otherwise: https://stackoverflow.com/a/22082397/6620612
-    this.isIE11 = !!global.MSInputMethodContext && !!document.documentMode;
+    if (typeof document !== 'undefined') {
```
adamstankiewicz, line 23: we no longer reference `document` until the component has been hydrated by Gatsby (avoids build time errors).

scss/edx/overrides/_modals.scss
chavesj: nice. this was bothering me 😄

chavesj (approved):
lgtm!

src/Modal/Modal.scss
```diff
@@ -41,7 +41,12 @@
 }
 
 .modal-content {
-  max-height: 100%;
+  max-height: calc(100vh - (#{$spacer} * 2));
```
abutterworth, line 44: Nice

src/Modal/index.jsx
```diff
@@ -214,54 +202,60 @@ class Modal extends React.Component {
...
-      this.renderModal(),
-      this.el,
-    );
+    if (this.el) {
+      return ReactDOM.createPortal(
+        this.renderModal(),
+        this.el,
+      );
+    }
+    return null;
```
abutterworth, line 258: Just a nitpick on style: My preference is to use an early return in scenarios like these. https://medium.com/better-programming/are-early-returns-any-good-eed4b4d03866

The gist is "write functions that return the expected positive result at the end of the function." So in this case:

```
if (!this.el) {
  return null;
}

return ReactDOM.createPortal(
  this.renderModal(),
  this.el,
);

adamstankiewicz: Ah yes, I usually prefer early returns too :) I will update!

abutterworth (approved):
Looks great. Just one comment to take or leave.

PR #586, opened 2020-11-20
feat: add definitions for color variables in Paragon core scss

abutterworth:
Allow consuming applications to use variable names like $primary-200 rather than depending only on the theme-color mixin.


AlasdairSwan (approved):
:shipit:

raq929 (approved):

PR #593, opened 2020-12-07
feat: add avatar and avatar button

abutterworth:
Implementations of these component specs:
- https://openedx.atlassian.net/wiki/spaces/BPL/pages/2103083206/Avatar
- https://openedx.atlassian.net/wiki/spaces/BPL/pages/2102821019/AvatarButton 

image
image


muselesscreator (approved):
LGTM!  1 extra example requested as a minor nit for passing size options when using the `as` param to forward this model to something like a `Dropdown.Toggle`

PR #596, opened 2020-12-09
docs: deprecate checkboxgroup and radiobuttongroup

abutterworth:
- Deprecate CheckboxGroup: It has no functional purpose with the already deprecated Checkbox component.
- Deprecate RadioButtonGroup: It has no functional purpose with the already deprecated RadioButton component.
- Add a "Deprecated" section in the documentation nav and move all deprecated components to it.

We will remove these components with the other deprecated components in an upcoming release.


adamstankiewicz (approved):
:shipit:

PR #599, opened 2020-12-19
feat: deprecate previous Breadcrumb for react-bootstrap implementation

adamstankiewicz:
I'm hoping to use the `react-bootstrap` implementation of the `Breadcrumb` component so that I can begin using react-router's `Link` component along with it to avoid full page refreshes when navigating between pages.

image

### Resulting DOM

Per the comment in `src/index.js`, we held off on moving to the `react-bootstrap` shim of the `Breadcrumb` component due to a semantic a11y issue where the `aria-current` attribute was placed on the `<li>` instead of the `<a>` which takes focus. However, when you supply the `active` prop to a `Breadcrumb.Item`, it does not render the active item as a link which is why `aria-current` is on the `<li>`.

```html
<nav aria-label="breadcrumb" class="small">
  <ol class="breadcrumb">
    <li class="breadcrumb-item">
      <a href="#" role="button">Find a Course</a>
    </li>
    <li class="breadcrumb-item">
      <a href="#page2">Computer Science Courses</a>
    </li>
    <li class="breadcrumb-item active" aria-current="page">
      CS50's Introduction to Computer Science
    </li>
  </ol>
</nav>
```


adamstankiewicz:
Closing this PR as after testing this with a screen reader, I recognize the problem. There's no way for a screen reader to get a read out of the last list item without it being a focusable link.

PR #600, opened 2021-01-08
feat: add icons

abutterworth:
Adds a new `icons/` package inside this project. In the future this new `paragon-icons` module should be published independently of `paragon-react` components. Until then, this structure keeps icons split from React source code.

All icons from the design documentation have been added (63 in all): https://openedx.atlassian.net/wiki/spaces/BPL/pages/2152661212/Iconography

---

This PR also includes an update to the `<Icon />` component. A new `src` prop has been introduced for use with the newly added icons. For example:

```jsx
import { Icon } from '@edx/paragon';
import { AddCircle } from '@edx/paragon/icons';

...
  
  <Icon src={AddCircle} />

```

If no src prop is provided, `Icon` will fallback to the old behavior thus remaining backward compatible until all teams can migrate to the new usage.

image


icons/.svgrrc.js
```diff
@@ -0,0 +1,22 @@
+const path = require('path');
```
raq929, line 1: Is this file named correctly or is there an extra r in it?

icons/.svgrrc.js
```diff
@@ -0,0 +1,22 @@
...
+
+module.exports = {
+  "icon": false,
+  "expandProps": "end",
+  "replaceAttrValues": {
+    "#000": "currentColor"
```
raq929, line 7: This is curiosity, but what exactly does this line do and why? (I assume replaces something with the currentColor but I'm wondering why we need that replacement.

AlasdairSwan (approved):
:shipit:

raq929 (approved):
LGTM! some questions for my edification but you're good to go.

icons/.svgrrc.js
abutterworth: It should have two rr's: `svgr` (the package) and `rc` (run commands)

icons/.svgrrc.js
abutterworth: When I export svg files from Figma they have fill colors (black). When implemented we'd like to be able to control the color via css if we need to. `fill="currentColor"` on an svg path will make the color match the css controlled text color.

abutterworth: This config is having svgr find any of the black fills and replacing them with currentColor.

PR #601, opened 2021-01-08
feat: add inline button size

abutterworth:
A new size for buttons: Inline means the button size is dictated by surrounding text so that it can fit comfortably aligned inline.

image


raq929 (approved):
Love it.

AlasdairSwan (approved):
:shipit:

PR #605, opened 2021-01-24
feat: WIP card

adamstankiewicz:


src/Card/Card.scss
```diff
@@ -1,2 +1,53 @@
...
+// .pgn__card-section-content {}
+// .pgn__card-section-title {}
+
+/**
+ * Card Divider
+ */
+
+.pgn__card-divider {
+  height: 0;
+  margin: $card-divider-margin-y 0;
```
adamstankiewicz, line 50: `$card-divider-margin-y` and `$card-divider-bg` are new variables. Should they be added to `@edx/brand`?

src/Card/Card.scss
```diff
@@ -1,2 +1,100 @@
 @import "variables";
 @import "~bootstrap/scss/card";
+
+/**
+ * Card Header
+ */
+
+ .pgn__card-header {
```
adamstankiewicz, line 8: Is this the best approach to do styles for this set of `Card` components?

src/Card/Card.scss
```diff
@@ -1,2 +1,140 @@
...
+  }
+ }
+
+/**
+ * Card Header
+ */
+
+ .pgn__card-header {
+  @extend .card-body;
+  @extend .pb-0;
```
abutterworth, line 34: I prefer we use the SASS variables instead of extending css utils. We have $spacer-0 through 5 I think

abutterworth: The sass utils will sprinkle `!important` in our styles

src/Card/Card.scss
abutterworth: No, but let's document them in the doc site. The edx brand currently attempts have all the available SASS variables  commented out, but this is done mostly because we have no documentation of what variables exist. We'll change that soon and then it'll just be a burden to update without any added value.

src/Card/Card.scss
```diff
@@ -1,2 +1,140 @@
...
+/**
+ * Card Footer
+ */
+
+.pgn__card-footer {
+  @extend .card-body;
+  @extend .pt-0;
+
+  @extend .d-flex;
+  @extend .justify-content-between;
```
abutterworth, line 96: Same about not using css utils here. Writing regular declarations is preferred. Ideally we keep the dependency of these styles to just variables in the `core/`. This will keep component sass from being coupled with core sass in unexpected ways. Consider "core sass variables" as the API from this SASS file can use. Extending other classes or using mixins is like using the private API.

src/Card/CardFooter/CardFooter.jsx
```diff
@@ -0,0 +1,89 @@
...
+      setStacked(false);
+    }
+  };
+
+  const { ref } = useResizeDetector({
+    handleHeight: false,
+    refreshMode: 'throttle',
+    refreshRate: 100,
+    onResize: handleResize,
+  });
```
abutterworth, line 43: I wonder the performance implication of this with 100+ cards on screen

src/Card/Card.scss
adamstankiewicz: Ah, yes. I noticed that while putting this together! Good idea to use variables here instead 👍

src/Card/CardFooter/CardFooter.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+  return (
+    <div
+      ref={ref}
+      {...attrs}
+      className={classNames('pgn__card-footer', className, {
+        stacked,
+      })}
+    >
+      <CardFooterContext.Provider value={contextValue}>
```
abutterworth, line 64: My gut reaction to the CardFooter components are that they're over-engineered. What if the usage presented more like this:

```
  <Card.Footer>
    <Card.Footer.Actions>
        <Button variant="tertiary">No, thanks</Button>
        <Button as="a" href="https://edx.org" variant="primary">Learn more</Button>
    <Card.Footer.Actions>
  </Card.Footer>
```

You could control block vs inline block buttons using `cloneElement` on each child of card.footer.actions. https://reactjs.org/docs/react-api.html#cloneelement . Something similar is done in ValidationFormGroup https://github.com/edx/paragon/blob/master/src/ValidationFormGroup/index.jsx#L52. (In that use, Context may have been a better solution, but here I think Context is overkill for what needs to be done)

abutterworth: Also, I think* it's better to use Button directly instead of a new component. I don't yet know why I think this and curious on your ideas here.

src/Card/CardSection/CardSection.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import Card from 'react-bootstrap/Card';
+
+const CardSection = ({
```
abutterworth, line 6: Should we just rename Card Section to Card Body and let our card have multiple bodies?

src/Card/CardFooter/CardFooter.jsx
adamstankiewicz: Me too. Supposedly `ResizeObservers` are natively quite performant, especially when throttled. This is the most sane approach I could cobble together quickly to let the component be responsive based on its own element width rather than a media query (screen width) as `Cards` take up the full width of their container.

CSS will eventually have container queries for this use case but they are not widely supported as of now, so this sort of approach using `ResizeObservers` seems to be the next best strategy.

I'd also hope we're not showing 100+ cards on a screen at once 😉 If we had a use case for that, we might only render the `Card` components into the DOM that are in the user's viewport. Regardless, it'd worth testing the performance of this out more scientifically.

src/Card/CardSection/CardSectionTitle.jsx
```diff
@@ -0,0 +1,32 @@
...
+);
+
+CardSectionTitle.propTypes = {
+  as: PropTypes.elementType,
+  className: PropTypes.string,
+  children: PropTypes.node.isRequired,
+};
+
+CardSectionTitle.defaultProps = {
+  as: HEADING_SIZES_MAP.H5,
```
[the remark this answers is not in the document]
adamstankiewicz: Generally a fan of using constants; may be overkill here though. I created the `HEADING_SIZES_MAP` variable because I didn't want to hardcode heading values into the component itself for some conditionals.

src/Card/CardSection/CardSection.jsx
adamstankiewicz: We could; this was actually a question I had during implementation as well since I replicating many of the styles Card Body already has into Card Section (i.e., padding).

src/Card/Card.scss
adamstankiewicz: Got it, this makes sense!

src/Card/CardSection/CardSectionTitle.jsx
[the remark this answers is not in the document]
abutterworth: In this case the constants are redundant, you couldn't change the value of the constant without changing its name and they're unlikely to need to change since that part of the html spec is settled. In this case I wouldn't consider "h5" a [magic number](https://moderatemisbehaviour.github.io/clean-code-smells-and-heuristics/#g25-replace-magic-numbers-with-named-constants).

src/Card/Card.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import BaseCard from 'react-bootstrap/Card';
+import CardImg from 'react-bootstrap/CardImg';
+
+const Card = ({
```
adamstankiewicz, line 8: Should be using `React.forwardRef`

PR #606, opened 2021-01-24
Colocate Component Javascript and Styles

abutterworth:
- Move component scss and related scss variables into component directories
- Create component directories for passthrough components
- Delete outdated `edx/` and `example-theme` theme directories
- Remove the pre-compiled `paragon-edx.css` published output. It was using the old theme and is unused throughout edx.

This is a technically breaking change because paragon-edx.css has been removed. I intend to release this as a breaking change, but functionally I believe there's nothing for consumers of Paragon to do to upgrade. A gut check on that would be appreciated.

Detail from the ADR in this PR:

-------

Context
-------

It is difficult for contributing developers new to Paragon to understand where new SCSS should be added or modified. The majority of Paragon styles sit inside the `scss` directory. Most component related scss is in this directory mixed with common variables or styles.

There is an `extensions` directory and `overrides` directory in `scss/core` for extensions or overrides to Bootstrap SCSS used in Paragon. It's difficult to distinguish between what style changes in Paragon are extensions to Bootstrap or simply overrides. Often this is determined by how the Bootstrap styles are implemeneted and whether they are extensible.

Decision
--------

We will colocate component-related SCSS and javascript in `src/{componentName}` directories. SCSS variables related to a component will be moved into the component directory in a distinct file `_variables.scss`. An example directory for Button:

```
src/Button/_variables.scss
src/Button/Button.scss
src/Button/index.jsx
src/Button/Button.test.jsx
```

The root `src/index.scss` will be responsible for including these SCSS files. And `Button.scss` in the example above would be responsible for including `Button/_variables.scss`

Consequences
------------

This will involve moving component styles from the `scss` directory into the `src directory`. Moving forward all component-related styles will in component directories.

Colocating SCSS and Javascript will bring us closer to more modularized component code that is more discoverable and easier to reason. This decision will simplify our future adoption of CSS-in-JS or CSS Modules + CSS Variables – a decision to be made in the coming months.

References
----------

* https://kentcdodds.com/blog/colocation


build-scss.js
```diff
@@ -21,13 +21,3 @@ var coreResult = sass.renderSync({
...
-
-
-// edX.org theme
-var edxResult = sass.renderSync({
-  file: './scss/edx/theme.scss',
-  outputStyle: 'compressed',
-  importer: tildaImporter,
-});
-
-fs.writeFileSync('./dist/edx-paragon.css', edxResult.css);
```
abutterworth, line 33: The `edx-paragon.css` output in Paragon v7 and up is not used by any edX applications

src/Alert/Alert.scss
```diff
@@ -32,8 +34,9 @@
 //
 // Expand the right padding and account for the close button's positioning.
 
+// Baking in $close-font-size: $font-size-base * 1.5 !default; to avoid any dependency
```
raq929, line 37: 💯

src/Breadcrumb/Breadcrumb.scss
```diff
@@ -0,0 +1,2 @@
+@import "variables";
+@import "~bootstrap/scss/breadcrumb";
```
raq929, line 2: Should this order be reversed? We would want our variables to override any boostrap variables of the same name, no?

src/Button/_variables.scss
```diff
@@ -0,0 +1,76 @@
+
+// Buttons + Forms
+//
+// Shared variables that are reassigned to `$input-` and `$btn-` specific variables.
+
+$input-btn-padding-y:         .375rem !default;
```
raq929, line 6: Do these `$input-*` variables need to be here, or can they live in the Form folder?

src/Code/_variables.scss
```diff
@@ -0,0 +1,17 @@
+
+// Code
+
+$code-font-size:                    87.5% !default;
+$code-color:                        #e83e8c !default;
+
+$kbd-box-shadow:              inset 0 -.1rem 0 rgba($black, .25) !default;
```
raq929, line 7: Nit: I'd love a comment about what kbd stands for

src/Collapsible/Collapsible.scss
```diff
@@ -1,3 +1,9 @@
+@import "../Card/variables"; // TODO: Remove this dependency
+@import "variables";
+
+// TODO: Remove this dependency
```
raq929, line 4: What's up with these TODOs?

src/Form/_variables.scss
```diff
@@ -0,0 +1,263 @@
+
+// Buttons + Forms
+//
+// Shared variables that are reassigned to `$input-` and `$btn-` specific variables.
```
raq929, line 4: I don't actually see any `$btn-*` variables in this folder. Are they needed here, or do we just need the `$input-btn-*` styles?

src/IconButton/_IconButton.scss
```diff
@@ -1,6 +1,6 @@
 $icon-button-colors: (
     "primary": $primary,
-    "secondary": $secondary,
```
raq929, line 3: I worry that people who are used to Bootstrap are gonna try to use "secondary" anyway? Should we alias it to something?

src/asInput/asInput.scss
```diff
@@ -1,3 +1,6 @@
+// TODO: Remove this dependency
```
raq929, line 1: Should this TODO be done for this PR, or is there a ticket for it? Generally I like to avoid leaving TODOs in committed code if possible.

src/index.js
```diff
@@ -53,99 +94,12 @@ export { default as TableHeader } from './DataTable/TableHeaderRow';
...
-export { default as Container } from 'react-bootstrap/Container';
-export { default as Image } from 'react-bootstrap/Image';
-export { default as Figure } from 'react-bootstrap/Figure';
-export { default as InputGroup } from 'react-bootstrap/InputGroup';
-export { default as Media } from 'react-bootstrap/Media';
+// Pass through any needed whole third-party library functionality
+// useTable for example is needed to use the DataTable component seamlessly
+// rather than setting a peer dependency in this project, we opt to tightly
+// couple these dependencies by passing through needed functionality.
 export { useTable } from 'react-table';
```
raq929, line 105: We're gonna need to export more than useTable from react-table... I'll make a PR for it.

raq929 (approved):
LGTM! I have a few questions but nothing blocking.

src/Breadcrumb/Breadcrumb.scss
abutterworth: Variables in Paragon and bootstrap are declared with `!default` so first in gets priority. https://getbootstrap.com/docs/4.6/getting-started/theming/#variable-defaults

src/Button/_variables.scss
abutterworth: They are declared in both places unfortunately. They are shared variables. I could move them out to core instead, but since themes come before both they're still editable.

As I describe this is seems super wonky... I think I'll either depart from bootstrap and decouple them or move it to `core`

src/Code/_variables.scss
abutterworth: 👍  will add. (keyboard) I think

src/Collapsible/Collapsible.scss
abutterworth: Collapsible styles extend Card styles. I think this is the only component with a dependency like this. I would rather we update the component to add card classes or something, but for now figured this got the job done.

src/Form/_variables.scss
abutterworth: These are the common, shared variables between buttons and forms. Like in a comment above, these variables are defined in both places right now. The definitions in buttons technically have priority because they're included first. I'll pull these back out the `core` directory.

src/asInput/asInput.scss
abutterworth: I think I could probably delete this whole file honestly. The component itself is deprecated and I'll remove in a PR soon and these styles are doing very little.

src/Breadcrumb/Breadcrumb.scss
abutterworth: Also, Bootstrap doesn't colocate their variables with component styles so this doesn't pull in the bootstrap ones anyway.

src/Card/index.jsx
```diff
@@ -0,0 +1,5 @@
+export { default } from 'react-bootstrap/Card';
```
davidjoy, line 1: Curious if we should just leave the react-bootstrap export structure as-is rather than bundling them/changing where you import them from.  Is it that this more closely matches what we're doing elsewhere in Paragon?

src/index.js
```diff
@@ -1,25 +1,61 @@
...
 export { default as Breadcrumb } from './Breadcrumb';
-export { default as Button } from './Button';
+export { default as Button, ButtonGroup, ButtonToolbar } from './Button';
+export {
+  default as Card,
+  CardColumns,
+  CardDeck,
+  CardImg,
+  CardGroup,
+} from './Card';
```
davidjoy, line 15: Related to the comment above, feels odd that we'd bundle these four other components into the `Card` exports and then unbundle them again here. :) Rather than just, I guess, having separate subdirectories for them all that re-export the react-bootstrap exports.

davidjoy: (To be clear, I don't feel strongly about this)

davidjoy (approved):

src/Card/index.jsx
abutterworth: Today, I'm finding that if we need to modify or replace the react-bootstrap components that I have to provide very detailed instructions on how to do it since it becomes, "make a directory, import the react-bootstrap, component make your shim, rexport it.

I think it's clearer if we make components uniformly have their own index.jsx in Paragon.

src/index.js
abutterworth: I agree it's somewhat weird. The grouping is not arbitrary though. I bundled together subcomponents with their parent. `Card.CardImg` for example. It's possible that we shouldn't export the subcomponents at all. But that would be a breaking change there.

src/Card/index.jsx
raq929: Agree x 100

PR #609, opened 2021-01-26
fix: pin uncontrollable to preserve es5 output

abutterworth:
Uncontrollable, a React Bootstrap dependency released a new version 7.2.0 yesterday that publishes in es6. This is causing downstream problems in micro-frontends that need to support IE11 and non es6 browsers. Pinning uncontrollable to 7.1.1 in Paragon will likely solve the issue in the near term since React Bootstrap's package.json requires `"uncontrollable": "^7.0.0",`

Issue in uncontrollable here: https://github.com/jquense/uncontrollable/issues/51

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,8 @@
     "react-table": "^7.6.1",
     "react-transition-group": "^4.0.0",
     "sanitize-html": "^1.20.0",
-    "tabbable": "^4.0.0"
+    "tabbable": "^4.0.0",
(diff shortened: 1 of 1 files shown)


davidjoy (approved):

davidjoy:
When renovate creates a PR to update to 7.2.0 of uncontrollable, you may want to figure out how to just leave a note in it and close it or something so we don't accidentally update.

PR #612, opened 2021-02-01
[PAR-356] feat: add lightbulb icon

MatthewPiatetsky:
1. how do I check that I did this correctly?
2. The svg I used has the lightbulb right side up but in the docs it is upside down? Should I make this a different icon name? Or change the docs?
https://openedx.atlassian.net/wiki/spaces/BPL/pages/2152661212/Iconography


abutterworth:
You can check that you've done it correctly by navigating to this page in the doc site: https://paragon-edx.netlify.app/foundations/icons/ and confirming that the icon you've added is there.

abutterworth (approved):

PR #614, opened 2021-02-03
feat: CardGrid to organize collection of cards in responsive grid

adamstankiewicz:


www/src/pages/components/card.mdx
```diff
@@ -87,4 +150,48 @@ import SingleComponentStatus from '../../components/SingleComponentStatus';
...
+      }
+    }
+  `}
+  render={({ componentMetadata }) => {
+    if (componentMetadata == null) return null;
+    return <PropsTable propMetaData={componentMetadata.props} />;
+  }}
+/>
+
+##### Theme variables (SCSS)
```
adamstankiewicz, line 173: What is the convention we have for these theme variables in the docs? Do we plan to essentially copy/paste the entire _variables.scss file for each component into the docs? If so, I wonder if it might make sense to figure out how to automate that across the repo similar to the auto-documenting props 🤔

abutterworth: Yes and yes.  If we stick with sass or css variables for theming then yes we should automate this.

src/Card/CardGrid.jsx
```diff
@@ -0,0 +1,54 @@
...
+  /** The Card components to organize into a responsive grid */
+  children: PropTypes.node.isRequired,
+  /** An object containing the desired column size at each breakpoint */
+  columnSizes: PropTypes.shape({
+    xs: PropTypes.number,
+    sm: PropTypes.number,
+    md: PropTypes.number,
+    lg: PropTypes.number,
+    xl: PropTypes.number,
+  }),
```
abutterworth, line 42: Does this map to the `Col` prop api?

src/Card/CardGrid.test.jsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import CardGrid from './CardGrid';
+import Card from '.';
+
+const cardContent = (
+  <Card>
+    <Card.Img variant="top" src="https://source.unsplash.com/360x200/?nature,flower" />
```
abutterworth, line 8: Might want to use some dummy url instead of unsplash

adamstankiewicz: Ah yep, meant to swap that out after the copy/pasta 👍

src/Card/CardGrid.jsx
adamstankiewicz: It does! Technically, you could pass in whatever props here `Col` accepts.

www/src/pages/components/card.mdx
adamstankiewicz: Cool, I backlogged a Paragon ticket to automate this if we end up wanting to go that route: https://openedx.atlassian.net/browse/PAR-364

abutterworth:
It might be worth having six cards in the CardGrid example. Other than that lgtm

abutterworth (approved):

adamstankiewicz:
> It might be worth having six cards in the CardGrid example. Other than that lgtm

Ha, yeah I kept going back on forth on this one :) I opted to go with less code in the code block (little easier to interpret at a quick glance) but I agree the example is a bit more solid with six cards. Will make this change and :shipit:

PR #615, opened 2021-02-03
feat: checkbox filter

raq929:
Adds a checkbox filter and a filters example page

I decided that I never want to dig through the react-table code again so I documented all the built-in filter functions for them 😛 
And wrote some more about how to use our filters.

Screen Shot 2021-02-04 at 4 40 18 PM

### DIFF
--- a/src/DataTable/CheckboxFilter.jsx
+++ b/src/DataTable/CheckboxFilter.jsx
@@ -10,7 +10,7 @@ function CheckboxFilter({
   },
 }) {
   // creates a unique label that does not change on re-render in case there are multiple checkbox filters in the dom
-  const ariaLabel = useRef(`checkbox-filter-label-${Math.floor(Math.random() * 100)}`);
+  const ariaLabel = useRef(`checkbox-filter-label-${Math.floor(Math.random() * 100000)}`);
   const inputText = `Search ${Header}`;
   const checkedBoxes = filterValue || [];
   const changeCheckbox = (name) => {
--- a/src/DataTable/CheckboxFilter.jsx
+++ b/src/DataTable/CheckboxFilter.jsx
@@ -23,15 +23,16 @@ function CheckboxFilter({
     <Form.Group>
       <FormLabel className="pgn__checkbox-filter-label">{inputText}</FormLabel>
       {checkboxFilters.map(({ name, number }) => (
-        <div key={name} className="pgn__checkbox-filter">
+        <div key={name} className="pgn__checkbox-filter align-items-center mb-2">
           <Input
-            id={`pgn__checkbox-filter--${name}`}
             type="checkbox"
             checked={checkedBoxes.includes(name)}
             onChange={() => { changeCheckbox(name); }}
           />
-          <label htmlFor={name}>{name} {number && <Badge variant="dark">{number}</Badge>}
+          <label htmlFor={name} className="mb-0">
+            {name}
           </label>
+          {number && <Badge variant="dark">{number}</Badge>}
         </div>
       ))}
     </Form.Group>
--- a/src/DataTable/CheckboxFilter.jsx
+++ b/src/DataTable/CheckboxFilter.jsx
@@ -1,7 +1,7 @@
 import React, { useRef } from 'react';
 import PropTypes from 'prop-types';
 import {
-  Form, FormLabel, Input, Badge,
+  Form, FormLabel, Badge,
 } from '..';
 
 function CheckboxFilter({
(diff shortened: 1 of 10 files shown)


www/src/pages/table/datatable.mdx
```diff
@@ -8,7 +8,8 @@ devStatus: 'In progress'
 
 import { StaticQuery, graphql } from 'gatsby';
 import get from 'lodash/get';
-import { DataTable, TextFilter } from '~paragon-react';
+import { useMemo } from 'React';
```
abutterworth, line 11: I think this is causing the doc site build to fail

www/src/pages/table/tablefilters.mdx
```diff
@@ -0,0 +1,157 @@
...
+title: 'DataTable'
+type: 'component'
+status: 'New'
+designStatus: 'In progress'
+devStatus: 'In progress'
+---
+
+import { StaticQuery, graphql } from 'gatsby';
+import get from 'lodash/get';
+import { useMemo } from 'React';
```
abutterworth, line 11: I think this may be causing the doc site build to fail

www/src/pages/table/datatable.mdx
raq929: Yessss and thankfully it was something I thought I needed that I don't actually need.

src/DataTable/CheckboxFilter.jsx
```diff
@@ -0,0 +1,65 @@
...
+      {checkboxFilters.map(({ name, number }) => (
+        <div key={name} className="pgn__checkbox-filter align-items-center mb-2">
+          <Input
+            type="checkbox"
+            checked={checkedBoxes.includes(name)}
+            onChange={() => { changeCheckbox(name); }}
+          />
+          <label htmlFor={name} className="mb-0">
+            {name}
+          </label>
```
abutterworth, line 36: The input doesn't have an id and therefore not connected to the label (via for). It might also be simpler to use Form.Check. Even then though, we should be concerned about duplicate ids. 

Another alternative that would allow us to avoid needing an id would be to put the input inside the label.

src/DataTable/tests/CheckboxFilter.test.jsx
```diff
@@ -0,0 +1,72 @@
...
+    const checkbox = wrapper.find({ type: 'checkbox' }).find('input').at(1);
+    checkbox.simulate('change');
+    expect(setFilterMock).toHaveBeenCalledWith([]);
+  });
+  it('renders checkbox label with filter name', () => {
+    const wrapper = mount(<CheckboxFilter column={{ ...props.column, filterValue: ['roan'] }} />);
+    const label = wrapper.find({ htmlFor: roan.name });
+    expect(label.text()).toContain(roan.name);
+  });
+  it('renders chekcbox label with number', () => {
```
abutterworth, line 59:
```suggestion
  it('renders checkbox label with number', () => {
```

src/DataTable/tests/CheckboxFilter.test.jsx
```diff
@@ -0,0 +1,72 @@
...
+    const wrapper = mount(<CheckboxFilter column={{ ...props.column, filterValue: ['roan'] }} />);
+    const label = wrapper.find('.pgn__checkbox-filter').at(0);
+    const badge = label.find(Badge);
+    expect(badge).toHaveLength(1);
+    expect(badge.text()).toEqual(String(roan.number));
+  });
+  it('renders chekcbox label with number', () => {
```
abutterworth, line 66:
```suggestion
  it('renders checkbox label with number', () => {
```

www/src/pages/table/datatable.mdx
```diff
@@ -8,7 +8,7 @@ devStatus: 'In progress'
 
 import { StaticQuery, graphql } from 'gatsby';
 import get from 'lodash/get';
-import { DataTable, TextFilter } from '~paragon-react';
+import { DataTable, TextFilter, CheckboxFilter } from '~paragon-react';
```
abutterworth, line 11: You don't need to import the components here. Code in the ```jsx live code blocks have a different scope controlled by the `CodeBlock` component elsewhere in the project.

www/src/pages/table/tablefilters.mdx
```diff
@@ -0,0 +1,156 @@
...
+title: 'DataTable'
+type: 'component'
+status: 'New'
+designStatus: 'In progress'
+devStatus: 'In progress'
+---
+
+import { StaticQuery, graphql } from 'gatsby';
+import get from 'lodash/get';
+import { DataTable, TextFilter, CheckboxFilter } from '~paragon-react';
```
abutterworth, line 11: You don't need to import the components here. Code in the ```jsx live code blocks have a different scope controlled by the `CodeBlock` component elsewhere in the project.

abutterworth (changes requested):
One issue to fix around the input + label connection. Other than that 🚀 🚀 🚀

src/DataTable/CheckboxFilter.jsx
raq929: I tried input inside the label at one point but the style issues were a problem. I'm gonna try Form.Check and hopefully that will fix everything at once. :)

abutterworth (approved):
🚢

PR #620, opened 2021-02-06
feat: dependency injection of analytics from external packages (e.g., frontend-platform)

adamstankiewicz:
See https://github.com/edx/frontend-platform/pull/148 for usage of `Paragon` as a provider component in the frontend-platform example app. Otherwise, read on for the details!

This top-level `Paragon` component, containing an injected analytics service from frontend-platform, is made available to any other Paragon component via React's Context API, without having the consumers of these components do anything special beyond wrapping their entire app in the `Paragon` component, and passing the `dependencies` object that frontend-platform gives frontend devs (see linked PR above).

This PR demonstrates how `Button` would be modified in Paragon to use the injected analytics service from frontend-platform to dispatch an event to Segment to track that a Button was clicked.

A benefit to this approach using the Context API is that consumers of Paragon would only need to wrap their apps in the `Paragon` HOC once and not think about tracking usage again. It also allows tracking to be opt-in.

**Testing instructions**
* Checkout #620 and  https://github.com/edx/frontend-platform/pull/148 locally.
* Use the `module.config.js` approach from frontend-build (see docs) to alias the Paragon dependency in frontend-platform to your local copy (note: you may use the `src` directory in your module.config.js file if you want hot reloading to work on Paragon at the same time as running the frontend-platform example app!)
* Run the frontend-platform example app and click some buttons and/or links.

**Open questions**
1. Is there another viable solution that doesn't require a `Paragon` HOC to do dependency injection (playing devil's advocate on this approach)?
2. What information might be useful to send to Segment as it relates to the design system (e.g., what design system analytics would be valuable for us to know about usage of Paragon components by users across the edX ecosystem?) Would sending through a copy of the component name, its attributes, active URL, etc. make sense (e.g., could see how many clicks we get for brand variant buttons vs. primary, etc.)?
3. Is sending events to Segment the best approach to collect this data? Given the tools we have today, I think so? It's pretty convenient to query Segment events from Snowflake to generate any reports we may need. But can Segment handle significantly more events in our current plan (don't know if we have a quota)?
4. When should we log info about a component? a) when user interacts with it in some way (e.g., clicks); b) when the component mounts to the DOM.
6. Would this same approach work for other dependency injections from frontend-platform such as for logging and i18n?
7. What would it take to get frontend-platform to do preview deploys of it's example app to Netlify like Paragon?

### DIFF
--- a/src/Button/index.jsx
+++ b/src/Button/index.jsx
@@ -1,5 +1,51 @@
-import Button from 'react-bootstrap/Button';
+import React, { createRef, useMemo } from 'react';
+import PropTypes from 'prop-types';
+import ButtonBase from 'react-bootstrap/Button';
+
+import { useHandleLogClick } from '../hooks/analytics';
+
 import ButtonDeprecated from './deprecated';
+
+const Button = React.forwardRef(({
+  children,
+  analyticEvents,
+  ...attrs
+}, forwardedRef) => {
+  const ref = useMemo(() => forwardedRef || createRef(), [forwardedRef]);
+
+  const handleLogClick = useHandleLogClick({
+    event: analyticEvents?.onClick,
+    onClick: attrs?.onClick,
+    ref,
+  });
+
+  return (
+    <ButtonBase
+      ref={ref}
+      {...attrs}
+      onClick={handleLogClick}
+    >
+      {children}
+    </ButtonBase>
+  );
+});
+
+Button.propTypes = {
+  children: PropTypes.node.isRequired,
+  onClick: PropTypes.func,
+  analyticEvents: PropTypes.shape({
+    onClick: PropTypes.shape({
+      name: PropTypes.string,
+      properties: PropTypes.shape({}),
+      dispatchers: PropTypes.arrayOf(PropTypes.func),
+    }),
+  }),
+};
+
+Button.defaultProps = {
+  onClick: undefined,
+  analyticEvents: undefined,
+};
 
 Button.Deprecated = ButtonDeprecated;
 
--- a/src/Button/index.jsx
+++ b/src/Button/index.jsx
@@ -1,4 +1,4 @@
-import React, { createRef, useMemo } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import ButtonBase from 'react-bootstrap/Button';
 
@@ -9,14 +9,13 @@ import ButtonDeprecated from './deprecated';
 const Button = React.forwardRef(({
   children,
   analyticEvents,
+  onClick,
   ...attrs
 }, forwardedRef) => {
-  const ref = useMemo(() => forwardedRef || createRef(), [forwardedRef]);
-
-  const handleLogClick = useHandleLogClick({
+  const [handleLogClick, ref] = useHandleLogClick({
     event: analyticEvents?.onClick,
-    onClick: attrs?.onClick,
-    ref,
+    onClick,
+    forwardedRef,
   });
 
   return (
@@ -32,7 +31,9 @@ const Button = React.forwardRef(({
 
 Button.propTypes = {
   children: PropTypes.node.isRequired,
+  /** specifies the callback function when the button is clicked */
   onClick: PropTypes.func,
+  /** specifies the analytic events to dispatch for when the user interacts with this component, e.g. `onClick`. */
   analyticEvents: PropTypes.shape({
     onClick: PropTypes.shape({
       name: PropTypes.string,
--- a/src/Hyperlink/index.jsx
+++ b/src/Hyperlink/index.jsx
@@ -1,4 +1,4 @@
-import React, { createRef, useMemo } from 'react';
+import React from 'react';
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import isRequiredIf from 'react-proptype-conditional-require';
@@ -15,14 +15,13 @@ const Hyperlink = React.forwardRef(({
   onClick,
   externalLinkAlternativeText,
   externalLinkTitle,
+  rel,
   ...attrs
 }, forwardedRef) => {
-  const ref = useMemo(() => forwardedRef || createRef(), [forwardedRef]);
-
-  const handleLogClick = useHandleLogClick({
+  const [handleLogClick, ref] = useHandleLogClick({
     event: analyticEvents?.onClick,
     onClick,
-    ref,
+    forwardedRef,
   });
 
   let externalLinkIcon;
@@ -30,8 +29,7 @@ const Hyperlink = React.forwardRef(({
   if (target === '_blank') {
     // Add this rel attribute to prevent Reverse Tabnabbing
     Object.assign(attrs, {
-      ...attrs,
-      rel: attrs.rel ? `noopener ${attrs.rel}` : 'noopener',
+      rel: rel ? `noopener ${rel}` : 'noopener',
     });
 
     externalLinkIcon = (
@@ -65,6 +63,7 @@ const Hyperlink = React.forwardRef(({
 
 Hyperlink.defaultProps = {
   target: '_self',
+  rel: undefined,
   onClick: () => {},
   externalLinkAlternativeText: 'Opens in a new window',
   externalLinkTitle: 'Opens in a new window',
@@ -78,6 +77,10 @@ Hyperlink.propTypes = {
   /** specifies where the link should open. The default behavior is `_self`, which means that the URL will be loaded into the same browsing context as the current one. If the target is `_blank` (opening a new window) `rel='noopener'` will be added to the anchor tag to prevent any potential [reverse tabnabbing attack](https://www.owasp.org/index.php/Reverse_Tabnabbing).
    */
   target: PropTypes.string,
+  /** specifies the relationship between the current document and the linked document; the
+   * `noopener` value is automatically added if `target` is "_blank".
+   */
+  rel: PropTypes.string,
   /** specifies the callback function when the link is clicked */
   onClick: PropTypes.func,
   // eslint-disable-next-line max-len
@@ -92,6 +95,7 @@ Hyperlink.propTypes = {
     PropTypes.string,
     props => props.target === '_blank',
   ),
+  /** specifies the analytic events to dispatch for when the user interacts with this component, e.g. `onClick`. */
   analyticEvents: PropTypes.shape({
     onClick: PropTypes.shape({
       name: PropTypes.string,
--- a/src/Hyperlink/index.jsx
+++ b/src/Hyperlink/index.jsx
@@ -1,56 +1,74 @@
-import React from 'react';
+import React, { createRef, useMemo } from 'react';
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import isRequiredIf from 'react-proptype-conditional-require';
 
 import withDeprecatedProps, { DEPR_TYPES } from '../withDeprecatedProps';
 
-function Hyperlink(props) {
-  const {
-    destination,
-    children,
-    target,
+import { useHandleLogClick } from '../hooks/analytics';
+
+const Hyperlink = React.forwardRef(({
+  analyticEvents,
+  destination,
+  children,
+  target,
+  onClick,
+  externalLinkAlternativeText,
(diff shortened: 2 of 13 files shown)


src/Paragon/index.jsx
```diff
@@ -0,0 +1,30 @@
+import React, { useMemo } from 'react';
+import PropTypes from 'prop-types';
+
+export const ParagonContext = React.createContext({});
+
+const Paragon = ({ dependencies, children }) => {
+  const contextValue = useMemo(
```
adamstankiewicz, line 7: note: context values should always be memoized so the value passed into the context doesn't change it's reference on each re-render of the component.

abutterworth:
This is great Adam!

My thoughts on your open questions:

> 1. Is there another viable solution that doesn't require a Paragon HOC to do dependency injection (playing devil's advocate on this approach)?

The only other solution I can think of for injecting dependencies is a home brewed solution using singletons similar to what's in frontend-platform. I think this solution is more straightforward and react-friendly.

> 2. What information might be useful to send to Segment as it relates to the design system (e.g., what design system analytics would be valuable for us to know about usage of Paragon components by users across the edX ecosystem?) Would sending through a copy of the component name, its attributes, active URL, etc. make sense (e.g., could see how many clicks we get for brand variant buttons vs. primary, etc.)?

I didn't realize this is what you meant. I envisioned this to be used to address Dillon's analytics needs here https://openedx.atlassian.net/browse/AA-634.  I think it could serve both, but for app-level analytics I think we need something a little different than what you've outlined in this PR.

**Paragon Analytics**
One concern: If we instrument every Paragon component with analytics events what kind of performance impact, if any would we incur? I'm probably overthinking it, but worth checking on. If we proceed with these runtime design system type analytics, my thoughts:
- Usage (Log on first-render): how many times are they rendered (not the React term) on page to the user?
- Props or attributes 
- Paragon Version (could potentially grab this info from package.json)
- Clicks or other interactions: I'm not sure what use these would be right now. I think it's something we could expand into if the data above proves useful.

**Aiding app analytics**
The user of the component should have an opportunity to add to or manipulate the payload sent in the analytic event. I think usage for that case might look like:

```
<Button
  analyticsName="lms.sequence.next"
  onLogClick={e => { 
    // do something and ...
   return somePayload;
  }}
>
 Next
</Button>
```

and then somewhere in `<Button />`:

```
const onClick = (e) => {
  const payload = props.onLogClick?.(e) || {};
  analytics.sendTrackEvent({
    event: props.analyticsName,
    ...payload,
  });
}
```

> 3. Is sending events to Segment the best approach to collect this data? Given the tools we have today, I think so? It's pretty convenient to query Segment events from Snowflake to generate any reports we may need. But can Segment handle significantly more events in our current plan (don't know if we have a quota)?

Segment feels right to me.

> 4. When should we log info about a component? a) when user interacts with it in some way (e.g., clicks); b) when the component mounts to the DOM.

I think starting small is the best path forward (e.g. when component mounts)

> 5. Would this same approach work for other dependency injections from frontend-platform such as for logging and i18n?

Yes, I think so, but they might all be identical in form. I've been hoping for us to use an approach like this for i18n and theming.

For i18n it's possible that all we want is
```
<ParagonContextProvider analytics={analyticsService} locale={getLocale()}>

or

<ParagonContextProvider
  analytics={analyticsService}
  i18n={{
    pagination: {
      next: '次へ',
      previous: '前へ'
    },
    ...
  }}
>

or 

<ParagonContextProvider
  lang="en"
  intlComponents={{
    Provider: IntlProvider,
    FormattedMessage,
    FormattedNumber,
  }}
>
^ This last one might cause us issues if we upgrade react-intl in our MFEs.
```

Aside from i18n's specific concerns, it's possible that would could split out a context provider/consumer for each service Paragon needs. I'm not sure of the benefits of doing that right now vs a singular "give me all the stuff" component.

> 6. What would it take to get frontend-platform to do preview deploys of it's example app to Netlify like Paragon?

Simplest path: we move frontend-platform deploys to Netlify. If we that we'd need to up the build minutes in the account.

src/Hyperlink/index.jsx
```diff
@@ -4,53 +4,86 @@ import PropTypes from 'prop-types';
...
+import {
+  useSendTrackingLogEvent,
+  useTrackComponentOnMount,
+} from '../hooks';
 
-function Hyperlink(props) {
+const Hyperlink = (props) => {
   const {
+    analyticsEvent,
     destination,
```
adamstankiewicz, line 15: TODO: rip out destination (but this is a breaking change...)

src/Button/index.jsx
```diff
@@ -1,6 +1,58 @@
...
 
+const Button = ({
+  children,
+  analyticEvents,
+  analyticDispatchers,
+  ...attrs
+}) => {
+  useTrackComponentOnMount('edx.ui.paragon.Button.mounted', attrs);
+
+  const ref = useRef();
```
adamstankiewicz, line 17: TODO: i believe our components should always use `React.forwardRef` so consumers can pass in a custom ref to use (should confirm that react-bootstrap also does this).

src/Hyperlink/index.jsx
adamstankiewicz: not tackling right now

adamstankiewicz:
A thought on the analytics props API. Is there a chance we may need to dispatch one analytic event (e.g., `onClick`) to 2 destinations while only dispatching to 1 destination for some other analytic event.

Maybe `analyticEvents` and `analyticDispatchers` becomes a single prop? 🤔 

```
analyticEvents={{
  onClick: {
    eventName: 'edx.ui.lms.link_clicked',
    eventDispatchers: ['sendTrackEvent', 'sendTrackingLogEvent'],
    eventProperties: { someProperty: 30 },
  },
}}
```

src/hooks/analytics.jsx
```diff
@@ -0,0 +1,97 @@
...
+    if (event) {
+      // handle clicked anchor links that open in same tab by adding a slight delay
+      // to the page redirect to allow enough time for an analytics event to
+      // be dispatched.
+      if (ref?.current?.href && ref?.current?.target !== '_blank') {
+        e.preventDefault();
+        dispatchAnalyticEvents();
+        setTimeout(() => {
+          global.location.href = ref?.current?.href;
+        }, [300]);
```
adamstankiewicz, line 82: TODO: change to a constant

src/hooks/analytics.jsx
```diff
@@ -0,0 +1,97 @@
...
+   * Iterates through all analytic event dispatch functions passed in, passing the event name and
+   * properties to each.
+   */
+  const sendEventToDispatchers = useCallback(
+    () => {
+      if (!eventDispatchers) {
+        return;
+      }
+      eventDispatchers.filter(dispatcher => typeof dispatcher === 'function').forEach((dispatch) => {
+        dispatch(eventName, eventProperties);
```
adamstankiewicz, line 48: TODO: any way to make this more resilient?

src/index.js
```diff
@@ -113,11 +113,20 @@ export { default as TableFooter } from './DataTable/TableFooter';
...
+  useWindowSize,
+} from './hooks';
 
 // Pass through any needed whole third-party library functionality
 // useTable for example is needed to use the DataTable component seamlessly
 // rather than setting a peer dependency in this project, we opt to tightly
 // couple these dependencies by passing through needed functionality.
 export { useTable } from 'react-table';
+
+// ``ParagonProvider`` is a HOC that allows consumers of @edx/paragon to inject dependencies
```
davidjoy, line 129: Nit: I don't think ParagonProvider is technically a higher order component, it just uses component composition.  The doc (https://reactjs.org/docs/higher-order-components.html) is somewhat vague on it, but I think the spirit of a HOC is that it wraps the component _definition_ and returns a wrapper that instantiates the component as part of whatever else it does.  i.e., `connect(MyComponent)` or `injectIntl(MyComponent)`.

That's a bit different than a component that takes `children` expressed as JSX and drops them into its render function.  I think that's just plain component composition, and the term "Provider" seems to have been adopted for components that do nothing else than provide some functionality to the ones composed in 'em.

src/hooks/analytics.jsx
```diff
@@ -0,0 +1,99 @@
...
+    if (event) {
+      // handle clicked anchor links that open in same tab by adding a slight delay
+      // to the page redirect to allow enough time for an analytics event to
+      // be dispatched.
+      if (ref?.current?.href && ref?.current?.target !== '_blank') {
+        e.preventDefault();
+        dispatchAnalyticEvents();
+        setTimeout(() => {
+          global.location.href = ref.current.href;
+        }, [ANALYTICS_TIMEOUT_MS]);
```
davidjoy, line 84: Hm, this seems brittle... I'm pretty sure the analytics network request will get canceled if this fires before it's complete, which for slow connections might happen frequently at 300ms.  

And unfortunately for fast internet users, 300ms is sorta introducing a lot of unnecessary lag into the system. :/

davidjoy:
Apologies if I've missed it somewhere in the above, but I'm trying to understand what sort of questions we intend to answer with these runtime statistics.  

I'm assuming that any domain/business-logic-specific analytics could and _should_ be instrumented by the app consuming paragon components, and that we want to add this in here so that Paragon itself can get raw statistics on its own usage.

- What will runtime usage statistics tell us that static code analysis won't?  
- Given that, what outcomes will these stats help inform/drive for Paragon?
- Are tools like Hotjar better suited to understanding how users are interacting with our components in the wild?

src/hooks/analytics.jsx
adamstankiewicz: This is based on how Segment implements their `trackLink` method, [which also introduces a 300ms delay](https://github.com/segmentio/analytics.js-core/blob/4f3035c5e959ea92d5c31a152211654242b23e00/lib/analytics.ts#L516). This delay gives enough time for the network request to resolve without being noticeably perceived by users.

If a network request doesn't resolve before the delay, we are prioritizing the UX over ensuring the event is logged to not introduce any more of a delay than necessary.

adamstankiewicz:
> I'm assuming that any domain/business-logic-specific analytics could and should be instrumented by the app consuming paragon components, and that we want to add this in here so that Paragon itself can get raw statistics on its own usage.

@davidjoy As is, this PR is no longer introducing design system usage analytics for Paragon components like what was originally proposed. This is mainly for a technical reason, as tracking every component mount would trigger more network requests behind-the-scenes than I'm comfortable with. I think we'd want to batch events from Paragon component mounts rather than sending them individually. 

Regardless, usage analytics at runtime about Paragon components could help us understand and quantity user behavior from a UX perspective that is currently not possible. It would also help us understand what props (and their values) are actually being passed to Paragon components which our static code analysis can't necessarily do, e.g. we could quantity conversion rates for brand vs. primary buttons across the entire edX platform which tools like Hotjar wouldn't be able to provide.

That said, this PR (along with the companion frontend-platform PR) still provides the basis for analytics tracking for consumers of Paragon components to instrument an analytic event more easily. As a consumer of Paragon, it is more convenient to wrap an app with `ParagonProvider`, injecting the `sendTrackEvent` function and providing an `analyticEvents` prop to components to define the event names/properties, than creating my own `onClick` event handlers for analytics tracking throughout the application, especially if the elements I'm adding analytic events to are hyperlinks as these need to delay the location change to give a reasonable amount of time for a network request to resolve.

src/index.js
adamstankiewicz: Yep, this isn't really an HOC; I will update the comment.

src/hooks/analytics.jsx
adamstankiewicz: If an error were to occur when calling the provided `dispatch` function(s), we could either swallow the error here (e.g., with a try/catch) or let it surface the issue to the MFE.

We are opting to let the error occur to make sure it is observable/known and picked up by our tooling like NewRelic.

src/hooks/analytics.jsx
adamstankiewicz: Also, FWIW, I did try to use `trackLink` here but it wasn't usable as is because it creates duplicate `onClick` event handlers on the provided element any time a prop changes with no clean way to remove the duplicate event handlers on re-renders.

PR #622, opened 2021-02-08
DropdownFilter and MultiSelectDropdownFilter

raq929:
This adds two new filters.
It also adds the ability for multiple filters to be shown in the top row.

**OPEN QUESTIONS**

- These filters require similar props on the column. I'm wondering about consolidating all the props I'm asking people for into a `filterOptions` prop so that you can switch between filters more easily. Since other filers aren't in widespread use yet, now would be a good time to do this.
- I'm trying to think of good ways to document these; showing all the code for all of them takes up *so* much space I'm leery of it.
- The dropdown filter looks out of place because it uses the default `<select>` tag (as discussed with @abutterworth) which comes with the default select arrow. Should we change it?


abutterworth:
Do you see a possible future where filters UI is composable instead of controlled by a props api? Passing in a 'how many filters to show where' seems to walk us further down a rigid path than toward a composable one.

src/DataTable/DropdownFilters.jsx
```diff
@@ -5,21 +5,23 @@ import { requiredWhen } from './utils/propTypesUtils';
 
 /** The first filter will be as an input, additional filters will be available in a dropdown.  */
 const DropdownFilters = ({
-  columns,
+  columns, noOfBreakoutFilters,
```
abutterworth, line 8: Opposed to `numBreakoutFilters` ?

raq929: Not at all, I like it better. Naming things is hard.

src/DataTable/DropdownFilters.jsx
```diff
@@ -47,6 +53,8 @@ DropdownFilters.propTypes = {
     /** React function component to display the filter */
     Filter: requiredWhen(PropTypes.func, 'canFilter'),
   })).isRequired,
+  /** Number between one and four filters that can be shown on the top row. */
+  numBreakoutFilters: PropTypes.oneOf([1, 2, 3, 4]),
```
abutterworth, line 57: Is 0 valid?

raq929: I guess? Zero will be chosen for you if your screen is too small.

src/DataTable/filters/DropdownFilter.jsx
```diff
@@ -0,0 +1,57 @@
+import React, { useRef } from 'react';
+import PropTypes from 'prop-types';
+import {
+  Form,
+} from '../..';
+import { newId } from '../../utils';
+
+const DEFAULT_VALUE = 'dropdown-filter-default';
```
abutterworth, line 8: would an empty string be better here?

abutterworth (approved):

src/DataTable/filters/DropdownFilter.jsx
raq929: I worry that an empty string would match with zero if you're using inexact equality. Which you might want to do because of numbers getting turned into strings in the DOM.

abutterworth: You're using `===` below and 0 should be converted to "0" right?  Setting a custom string to mean 'nothing' seems like misusing the value attribute to me.

PR #624, opened 2021-02-09
feat: add sized containers

abutterworth:
Add the `size` prop to `Container`

```
<Container size="xl">
  The content in this container won't exceed the extra large width.
</Container>
<Container size="lg">
  The content in this container won't exceed the large width.
</Container>
<Container size="md">
  The content in this container won't exceed the medium width.
</Container>
<Container size="sm">
  The content in this container won't exceed the small width.
</Container>
<Container size="xs">
  The content in this container won't exceed the extra small width.
</Container>
```

Also adds max-width css utilities with these defaults
```
.mw-xs     max-width: 494px;
.mw-sm     max-width: 738px;
.mw-md     max-width: 982px;
.mw-lg     max-width: 1222px;
.mw-xl     max-width: 1470px;

```

https://openedx.atlassian.net/browse/PAR-209


muselesscreator:
Looks reasonable, but I feel like the Container component should be in its own file rather than lumped into the index.js
Also, no visual example in the example page?

abutterworth:
I can move it to its own file.

No visual in the example page since the doc site currently locks the width to a point where you wouldn't be able to tell what's going on. I want to refactor the doc site layout stuff soon. That might be a good time to make this interactive

muselesscreator (approved):
:ship: :it:

PR #625, opened 2021-02-09
feat: new sheet component

muselesscreator:
Implements a basic "Sheet" component, which serves as a basic UI container element bound to and spanning one edge of the window.
Supports light/dark theming, as well as content-blocking.

https://www.figma.com/file/eGmDp94FlqEr4iSqy1Uc1K/Paragon-2021?node-id=7520%3A16719

### DIFF
--- a/src/Sheet/Sheet.scss
+++ b/src/Sheet/Sheet.scss
@@ -9,11 +9,13 @@
   &.hidden {
     display: none;
   }
+  z-index: $zindex-sheet-backdrop;
 }
 .pgn__sheet-component {
   position: fixed;
   padding: 20px;
   background-color: white;
+  z-index: $zindex-sheet;
   &.pgn__sheet__dark {
     background-color: rgb(0,38,43);
     color: rgb(242,240,239);
--- a/src/Sheet/Sheet.scss
+++ b/src/Sheet/Sheet.scss
@@ -5,7 +5,6 @@
   position: fixed;
   top: 0px;
   left: 0px;
-  font-size: 18px;
   &.hidden {
     display: none;
   }
@@ -13,12 +12,12 @@
 }
 .pgn__sheet-component {
   position: fixed;
-  padding: 20px;
+  padding: 1.25rem;
   background-color: white;
   z-index: $zindex-sheet;
   &.pgn__sheet__dark {
-    background-color: rgb(0,38,43);
-    color: rgb(242,240,239);
+    background-color: $primary-500;
+    color: $light-300;
   }
   &.bottom {
     bottom: 0px;
--- a/src/Sheet/Sheet.scss
+++ b/src/Sheet/Sheet.scss
@@ -1,0 +1,50 @@
+.pgn__sheet-skrim {
+  width: 100%;
+  height: 100%;
+  background-color: rgba(196,196,196,0.5);
+  position: fixed;
+  top: 0px;
+  left: 0px;
+  font-size: 18px;
+  &.hidden {
+    display: none;
+  }
+}
+.pgn__sheet-component {
+  position: fixed;
+  padding: 20px;
+  background-color: white;
+  &.pgn__sheet__dark {
+    background-color: rgb(0,38,43);
+    color: rgb(242,240,239);
+  }
+  &.bottom {
+    bottom: 0px;
+    box-shadow: 0px -4px 10px 0px rgba(0,0,0,0.15);
+    box-shadow: 0px -8px 16px 0px rgba(0,0,0,0.15);
+  }
+  &.top {
+    top: 0px;
+    box-shadow: 0px 4px 10px 0px rgba(0,0,0,0.15);
+    box-shadow: 0px 8px 16px 0px rgba(0,0,0,0.15);
+  }
+  &.left {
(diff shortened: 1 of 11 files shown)


src/Sheet/SheetContainer.test.jsx
```diff
@@ -0,0 +1,33 @@
...
+    expect(rootEl.classList.contains('sheet-container')).toEqual(true);
+    const childEl = rootEl.querySelector(`#${childId1}`);
+    expect(childEl.innerHTML).toEqual(childContent1);
+  });
+
+  it('only adds itself once, when invoked multiple times', () => {
+    mount(<SheetContainer>{child1}</SheetContainer>);
+    mount(<SheetContainer>{child2}</SheetContainer>);
+    const rootEls = document.querySelectorAll('#sheet-root');
+    expect(rootEls.length).toBeGreaterThan(0);
```
abutterworth, line 27: Should this be `1`?

src/Sheet/Sheet.scss
```diff
@@ -0,0 +1,50 @@
+.pgn__sheet-skrim {
+  width: 100%;
+  height: 100%;
+  background-color: rgba(196,196,196,0.5);
+  position: fixed;
```
abutterworth, line 5: We should probably specify a z-index. Here is where z-indexes are currently set: https://github.com/edx/paragon/blob/80ce916912cc13bafe81d2e90acb97b582cbb65f/scss/core/_variables.scss#L610-L611

We might want to add:
```
$zindex-sheet-backdrop:             1031 !default;
$zindex-sheet:                      1032 !default;
```

src/Sheet/index.jsx
```diff
@@ -0,0 +1,98 @@
...
+          role="presentation"
+        />
+        <div
+          className={classNames(
+            'pgn__sheet-component',
+            `pgn__sheet__${variant}`,
+            position,
+
+          )}
+          role="alert"
```
abutterworth, line 57: @wittjeff I know there was some discussion on the sheet component using the alert role, but I'm guess this role would be odd if the sheet were used for a side nav or some use case like it. Is this something we should bake in, make configurable, or leave to child elements to define?

src/Sheet/SheetContainer.test.jsx
muselesscreator: yeah, but should also have been `toEqual`

src/Sheet/Sheet.scss
muselesscreator: reasonable.

src/Sheet/index.jsx
```diff
@@ -0,0 +1,98 @@
...
+
+class Sheet extends React.Component {
+  constructor(props) {
+    super(props);
+    this.blockClicks = this.blockClicks.bind(this);
+  }
+
+  blockClicks(event) {
+    event.stopPropagation();
+  }
```
abutterworth, line 27: The sheet doesn't currently lock focus inside the sheet if it's set to blocking. I suggest using this library: https://github.com/theKashey/react-focus-on to do this work. It does a great job of handling scrolling, clicking, and focus blocking very well.

abutterworth (changes requested):
Looks great so far. A couple things on z-index and focus locking above

src/Sheet/Sheet.scss
```diff
@@ -0,0 +1,52 @@
...
+  top: 0px;
+  left: 0px;
+  font-size: 18px;
```
abutterworth, line 8:
```suggestion
```

src/Sheet/Sheet.scss
```diff
@@ -0,0 +1,52 @@
...
+  &.hidden {
+    display: none;
+  }
+  z-index: $zindex-sheet-backdrop;
+}
+.pgn__sheet-component {
+  position: fixed;
+  padding: 20px;
```
abutterworth, line 16:
```suggestion
  padding: 1.25rem;
```

src/Sheet/Sheet.scss
```diff
@@ -0,0 +1,52 @@
...
+  background-color: white;
+  z-index: $zindex-sheet;
+  &.pgn__sheet__dark {
+    background-color: rgb(0,38,43);
+    color: rgb(242,240,239);
```
abutterworth, line 21:
```suggestion
    background-color: $primary-500;
    color: $light-300;
```

src/Sheet/index.jsx
```diff
@@ -0,0 +1,99 @@
...
+        className={classNames(
+          'pgn__sheet-component',
+          `pgn__sheet__${variant}`,
+          position,
+        )}
+        role="alert"
+        {...{
+          'aria-live': 'polite',
+          'aria-atomic': 'true',
+        }}
```
abutterworth, line 39:
```suggestion
        aria-live="polite"
        aria-atomic="true"
```
You shouldn't need to do this, the linter should accept aria attributes on jsx

abutterworth (approved):
One minor thing about syntax for getting aria attributes onto the jsx component that's minor.  Let's follow up with Jeff on the aria-role when he can see an example.

PR #627, opened 2021-02-10
docs: improve the props tables

abutterworth:


www/src/components/PropType.jsx
```diff
@@ -0,0 +1,148 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Badge } from '~paragon-react';
+
+const RequiredBadge = ({ isRequired }) => {
+  if (!isRequired) return null;
+  return <> <Badge variant="light">Required</Badge></>;
```
raq929, line 7: Yessss I like this, I wonder if it should be a yellow or red color.

raq929 (commented):
OK, so, yes, this is awesome. 🎉 

I'm finding that the text with the name of the prop is maybe too small? I'm drawn immediately to the type, when what I need to know first is the name.

Also I'm finding that when a `required` pill and a `default` pill are close together my eye gets confused. 
I wonder if we can maybe make those smaller, and/or make them colored differently from each other?

raq929 (approved):

abutterworth:
I agree on your design feedback – I'll follow up with some visual iterations after this merges.

PR #630, opened 2021-02-11
feat: add foundational modal components

abutterworth:
The components here conceive any kind of Modal Dialog as a `ModalLayer` and any kind of popup or menu as a `ModalPopup`. They are intended to be used to create any number of modal type components without requiring too much thought around accessibility or focus management.


src/ModalBase/ModalCloseButton.jsx
```diff
@@ -0,0 +1,38 @@
...
+import { Button } from '..';
+import ModalContext from './ModalContext';
+
+const ModalCloseButton = React.forwardRef(({
+  as, children, onClick, ...props
+}, ref) => {
+  const { close } = useContext(ModalContext);
+
+  return React.createElement(
+    as,
```
raq929, line 12: I'm familiar with this pattern, but for people who aren't, you wanna throw in a comment above this line explaining WTF is going on here? 😄

src/ModalBase/ModalLayer.jsx
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+/*  TODO: The first of these two disabled linters is okay.
```
raq929, line 3: Maybe @wittjeff can evaluate if this second disabled linter rule is OK?

src/ModalBase/ModalLayer.jsx
```diff
@@ -0,0 +1,89 @@
...
+  zIndex: 3050, // $zindex-modal: 1050 !default;
+  overflow: 'auto',
+  position: 'fixed',
+  top: 0,
+  left: 0,
+  width: '100vw',
+  height: '100vh',
+};
+
+const useMemoizedValues = ({ close, isOpen, isBlocking }) => useMemo(() => ({
```
raq929, line 22: Why are you pulling this out of the main component?

src/ModalBase/ModalLayer.jsx
```diff
@@ -0,0 +1,89 @@
...
+      }}
+      className="position-overlay"
+      onClick={!isBlocking && close}
+    />
+  );
+};
+
+/**
+ * The ModalBase should be used for any component that wishes to capture the
+ * user in a *modal* context where the rest of the application is non-interactive.
```
raq929, line 46: I'm not sure I fully understand what you mean by "capture the user" and "the rest of the app is non-interactive". I *think* I get it, but I'm not super clear.

src/ModalBase/ModalPopup.jsx
```diff
@@ -0,0 +1,63 @@
+import React, { useMemo } from 'react';
+import PropTypes from 'prop-types';
+import { FocusOn } from 'react-focus-on';
+import Portal from './Portal';
+import PopperElement from './PopperElement';
+import ModalContext from './ModalContext';
+
+const useMemoizedValues = ({ close, isOpen, isBlocking }) => useMemo(() => ({
```
raq929, line 8: Heh, I asked you why you pulled this out earlier. This appears to be exactly the same, so maybe share it between the components?

src/ModalBase/Portal.jsx
```diff
@@ -0,0 +1,35 @@
...
+    } else {
+      const rootElement = document.createElement('div');
+      rootElement.setAttribute('id', this.rootName);
+      this.rootElement = document.body.appendChild(rootElement);
+    }
+  }
+
+  render() {
+    if (this.rootElement) {
+      return ReactDOM.createPortal(
```
raq929, line 22: What does this do?

src/hooks/useToggle.jsx
```diff
@@ -0,0 +1,33 @@
+import { useState, useCallback } from 'react';
+
+export default function useToggle(defaultIsOn, handlers = {}) {
+  const { onToggleOn, onToggleOff, onToggle } = handlers;
```
raq929, line 4: I think `onToggleOn` and `onToggle` is confusing (and hard to say 😂 ). Could `onToggle` be `handleToggle`?

raq929 (commented):
🎉  This is super cool 🎉 
I have one naming issue and the rest of my questions are mostly for my knowledge-building.

fwiw, this is where I *really* want design guidelines along with the component, because this allows almost *too* much flexibility.

src/ModalBase/Portal.jsx
adamstankiewicz: My understanding is that React Portals essentially let you mount a component into a separate DOM node than it would otherwise render within the component tree. E.g., here, any children of `Portal` will actually get rendered as child nodes of `this.rootElement`.

src/ModalBase/ModalLayer.jsx
abutterworth: Yea, I've asked him to take a look at the deploy previews, we definitely need aria-roles and maybe aria-labelledby for these things.

src/ModalBase/ModalLayer.jsx
abutterworth: See below

src/ModalBase/ModalPopup.jsx
abutterworth: The ModalPopup and ModalLayer were in the same file as I was working. Forgot to work through what to do with this. I will DRY it up into a `ModalContextProvider`

src/ModalBase/ModalLayer.jsx
abutterworth:
```
/**
 * The ModalLayer should be used for any component that wishes to engage the user
 * in a "mode" where a layer on top of the application is interactive while the
 * rest of the application is made non-interactive. The assumption made by this
 * component is that if a modal object is visible then it is "enabled"
 */
 ```

File: src/ModalBase/Portal.jsx (hunk 6 above)
Comment (abutterworth) [reply src/ModalBase/Portal.jsx:22]: What Adam said. And the react docs https://reactjs.org/docs/portals.html.  We use this for the modals because we can't guarantee that no parent element has overflow or z-index set. Portals let us control the [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) we render our components in.

File: src/hooks/useToggle.jsx (hunk 7 above)
Comment (abutterworth) [reply src/hooks/useToggle.jsx:4]: Say this line of code 10x fast. `handle..` doesn't match the pattern React has established. imo our syntax should always be `<SomeComponent onClick={handleClick] />`. I could just write `onOn` ... :P. I'll think on this a little but, but if you have more suggestions let me know.

File: src/ModalBase/ModalCloseButton.jsx [hunk 8]
```diff
@@ -0,0 +1,36 @@
+import React, { useContext } from 'react';
+import PropTypes from 'prop-types';
+import { Button } from '..';
+import ModalContext from './ModalContext';
+
+const ModalCloseButton = React.forwardRef(({ as, children, ...props }, ref) => {
+  const { close } = useContext(ModalContext);
+  const type = as;
```
davidjoy, line 8: There a reason to assign `as` to `type` instead of just passing it through to React.createElement directly?

src/ModalBase/Portal.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+
+class Portal extends React.Component {
+  constructor(props) {
+    super(props);
+    this.rootName = 'paragon-portal-root';
```
davidjoy, line 8: Does this hard coded root imply we can only have one modal at a time?  Even for ModalPopup?  Or will it append more modals to the same root safely?

src/hooks/useToggle.jsx
davidjoy: If it's helpful, a naming scheme I try to use in MFEs:

1. When I have a reference to a handler from outside my component, I call it `toggleOnHandler`
2. When I'm defining a handler in my component, I call it `handleToggleOn`. 
3. I don't write functions as `onWhatever` - those are the names of DOM attributes where we attach handlers.

I think of it like the parent component saying "Here's a toggle on handler" vs. "here's how I handle toggle on".  ¯\_(ツ)_/¯ i.e., it helps understand what _this_ component defines vs. what someone _else_ defines, without also potentially duplicating the naming scheme of the DOM attributes we use to attach handlers.

src/ModalBase/ModalCloseButton.jsx
abutterworth: Clarity for those reading is all. happy not to assign it if below is still clear

davidjoy (approved):
Vera nice.

src/ModalBase/ModalCloseButton.jsx
davidjoy: I don't feel strongly. :)

src/ModalBase/Portal.jsx
abutterworth: It appends them safely. @muselesscreator proved this earlier with `Sheet`

src/ModalBase/ModalLayer.jsx
raq929: Yes, much clearer, I love it.

src/ModalBase/Portal.jsx
raq929: Cooooool. 
Are there accessibility considerations for using them, or do screen readers just handle them gracefully?

src/Modal/ModalLayer.jsx
```diff
@@ -0,0 +1,80 @@
...
+      <Portal>
+        <FocusOn
+          scrollLock
+          enabled={isOpen}
+          onEscapeKey={close}
+          onClickOutside={!isBlocking && close}
+          className="pgn__modal-layer"
+        >
+          <ModalContentContainer>
+            <ModalBackdrop onClick={!isBlocking && close} />
```
adamstankiewicz, line 60: Is this the bit that results in the jsx-a11y eslint error? 

This `onClick` prop gets passed to a `div` element but a `div` is not inherently interactive and would typically require a `role` attribute to be added to indicate that this element is actually interactive, e.g. `role="link"` or `role="button"`.

image

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md#case-the-event-handler-is-only-being-used-to-capture-bubbled-events

I think it's ok to disable this rule but would recommend moving the eslint-disable comment above the line which causes it with an explanation to reduce confusion.

src/Modal/ModalPopup.jsx
```diff
@@ -0,0 +1,55 @@
...
+      <RootComponent>
+        <PopperElement target={positionRef} placement={placement} {...popperProps}>
+          <FocusOn
+            scrollLock={false}
+            enabled={isOpen}
+            onEscapeKey={close}
+            onClickOutside={close}
+          >
+            {isOpen && (
+              <div>
```
adamstankiewicz, line 24: tiny nit: do we need to render this `div` or would a React.Fragment (`<>`) work here?

adamstankiewicz (approved):

src/ModalBase/Portal.jsx
abutterworth: We don't get anything like tab or reading order for free through portals. In our case we're using aria-hidden on the app root element and placing focus on the first focusable element of the modal element so no a11y is handled.

src/Modal/ModalPopup.jsx
abutterworth: looks like fragment works fine 👍

src/Modal/ModalLayer.jsx
abutterworth: 👍 moved

PR #634, opened 2021-02-16
feat: make TableFooter more composeable

raq929:
Users can pass whatever components they want to be displayed in the footer
Defaults to the current RowStatus component on the left and TablePagination on the right

### DIFF
--- a/src/DataTable/DataTable.scss
+++ b/src/DataTable/DataTable.scss
@@ -90,10 +90,7 @@ $data-table-footer-position: center !default;
   display: flex;
   justify-content: space-between;
   padding: $data-table-padding-x $data-table-padding-y;
-
-  .pgn__data-table-footer-row-status {
-    align-self: $data-table-footer-position;
-  }
+  align-items: center;
 }
 
 .pgn__data-table-pagination {
--- a/src/DataTable/TableFooter.jsx
+++ b/src/DataTable/TableFooter.jsx
@@ -1,15 +1,33 @@
 /* eslint-disable react/require-default-props */
 import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
 import RowStatus from './RowStatus';
 import TablePagination from './TablePagination';
 
-const TableFooter = () => (
-  <div className="pgn__data-table-footer">
-    <RowStatus
-      className="pgn__data-table-footer-row-status"
-    />
-    <TablePagination />
+const TableFooter = ({ className, children }) => (
+  <div className={classNames(className, 'pgn__data-table-footer')}>
+    {children}
   </div>
 );
+
+TableFooter.defaultProps = {
+  children: (
+    <>
+      <TablePagination />
+      <RowStatus />
+    </>
+  ),
+  className: null,
+};
+
+TableFooter.propTypes = {
+  className: PropTypes.string,
+  children: PropTypes.oneOfType([
+    PropTypes.func,
+    PropTypes.node,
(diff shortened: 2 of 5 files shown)


src/DataTable/DataTable.scss
```diff
@@ -91,7 +91,7 @@ $data-table-footer-position: center !default;
   justify-content: space-between;
   padding: $data-table-padding-x $data-table-padding-y;
 
-  .pgn__data-table-footer-row-status {
+  .pgn__data-table-footer-left {
     align-self: $data-table-footer-position;
   }
```
abutterworth, line 96: Could do `align-items: center` to align all child to center by default, removing the need to name or style `.pgn__data-table-footer-left`

src/DataTable/TableFooter.jsx
```diff
@@ -1,15 +1,29 @@
...
-  <div className="pgn__data-table-footer">
-    <RowStatus
-      className="pgn__data-table-footer-row-status"
+const TableFooter = ({ className, LeftComponent, RightComponent }) => (
+  <div className={classNames(className, 'pgn__data-table-footer')}>
+    <LeftComponent
+      className="pgn__data-table-footer-left"
     />
-    <TablePagination />
+    <RightComponent />
```
abutterworth, line 13: I think we should prefer children here something like:

```
const TableFooter = ({ children }) => (
  <div className="...png__data-table-footer">
    {children}
  </div>
);

TableFooter.defaultProps = {
  children: (
    <>
      <TablePagination />
      <RowStatus />
    </>
  )
}

**Review by abutterworth — APPROVED** [review]
I think we should avoid `LeftComponent` `RightComponent` style props in cases where it's possible, but if this is urgently needed I don't see a reason we can't deliver this value before thinking through another approach.

File: src/DataTable/TableFooter.jsx (hunk 2 above)
Comment (raq929) [reply src/DataTable/TableFooter.jsx:13]: OK so. 
Pros: Doing this makes it infinitely flexible. That's nice. It's also in line with how the TableData component is written.
Cons: I think part of having a design system is that you're pushing people to follow good design patterns, and I can Imagine people doing all sorts of things. Also, they can just write their own custom footer at that point?

Comment (abutterworth) [reply src/DataTable/TableFooter.jsx:13]: My take:

Make it easy to do the right thing. Make it possible to do something with didn't think of, and trust our team to use this power judiciously. If our components do the right things by default, escape hatches should be infinitely flexible via composability or render props. We should be more concerned with building unnecessarily limiting constraints in our components than with abuse of the flexibility we provide.

- By default or without configuration components should do the common, desirable thing
- Desirable variants of a pattern should be easy to create without heavy customization (not related to this particular scenario)
- Escape hatches should provide as much flexibility as possible.

File: www/src/pages/table/tablefooter.mdx [hunk 3]
```diff
@@ -18,7 +18,8 @@ import SingleComponentStatus from '../../components/SingleComponentStatus';
 <a href="/table/datatable">Main DataTable page</a>
 
 The `TableFooter` component is meant to be rendered within a `DataTableContext` (provided by `DataTable` here).
-It shows a how many items are being displayed and pagination information.
+By default, it shows a how many items are being displayed and pagination information. It is a display component,
+and will accept a `RightComponent` and a `LeftComponent` that will override it's defaults.
```
abutterworth, line 22: Is this left over?

abutterworth (approved):
One thing, but lgtm

PR #640, opened 2021-02-23
feat: Add TableActions component

raq929:
Table actions are displayed when no rows are selected
Bulk actions are now only displayed when rows are selected

Updated documentation

### DIFF
--- a/src/DataTable/ActionDisplay.jsx
+++ b/src/DataTable/ActionDisplay.jsx
@@ -1,0 +1,24 @@
+import React, { useContext } from 'react';
+import BulkActions from './BulkActions';
+import TableActions from './TableActions';
+import DataTableContext from './DataTableContext';
+
+// handles layout for filters, status, and bulk actions
+const ActionDisplay = () => {
+  const { bulkActions, tableActions, selectedFlatRows } = useContext(DataTableContext);
+  const noBulkActions = bulkActions?.length < 1;
+  const noTableActions = tableActions?.length < 1;
+  const noActionsAvailable = noBulkActions && noTableActions;
+  const noRowsSelected = selectedFlatRows?.length < 1;
+  console.log(bulkActions, tableActions)
+  // return null if there are no actions or if there are only bulk actions and no selected rows
+  if (noActionsAvailable || (!noBulkActions && noTableActions && noRowsSelected)) {
+    return null;
+  }
+  if (!noRowsSelected) {
+    return <BulkActions />;
+  }
+  return <TableActions />;
+};
+
+export default ActionDisplay;
--- a/src/DataTable/ActionDisplay.jsx
+++ b/src/DataTable/ActionDisplay.jsx
@@ -10,9 +10,9 @@ const ActionDisplay = () => {
   const noTableActions = tableActions?.length < 1;
   const noActionsAvailable = noBulkActions && noTableActions;
   const noRowsSelected = selectedFlatRows?.length < 1;
-  console.log(bulkActions, tableActions)
+
   // return null if there are no actions or if there are only bulk actions and no selected rows
-  if (noActionsAvailable || (!noBulkActions && noTableActions && noRowsSelected)) {
+  if (noActionsAvailable || (noTableActions && noRowsSelected)) {
     return null;
   }
   if (!noRowsSelected) {
--- a/src/DataTable/ActionDisplay.jsx
+++ b/src/DataTable/ActionDisplay.jsx
@@ -3,7 +3,7 @@ import BulkActions from './BulkActions';
 import TableActions from './TableActions';
 import DataTableContext from './DataTableContext';
 
-// handles layout for filters, status, and bulk actions
+// handles display logic for actions
 const ActionDisplay = () => {
   const { bulkActions, tableActions, selectedFlatRows } = useContext(DataTableContext);
   const noBulkActions = bulkActions?.length < 1;
--- a/src/DataTable/ActionDisplay.jsx
+++ b/src/DataTable/ActionDisplay.jsx
@@ -9,13 +9,13 @@ const ActionDisplay = () => {
   const noBulkActions = bulkActions?.length < 1;
   const noTableActions = tableActions?.length < 1;
   const noActionsAvailable = noBulkActions && noTableActions;
-  const noRowsSelected = selectedFlatRows?.length < 1;
+  const hasRowsSelected = selectedFlatRows?.length > 0;
 
   // return null if there are no actions or if there are only bulk actions and no selected rows
-  if (noActionsAvailable || (noTableActions && noRowsSelected)) {
+  if (noActionsAvailable || (noTableActions && !hasRowsSelected)) {
     return null;
   }
-  if (!noRowsSelected) {
+  if (!noBulkActions && hasRowsSelected) {
     return <BulkActions />;
   }
   return <TableActions />;
--- a/src/DataTable/Actions.jsx
+++ b/src/DataTable/Actions.jsx
@@ -25,7 +25,7 @@ const CustomToggle = React.forwardRef(({ children, onClick }, ref) => (
 
 const Actions = ({
   className,
-  actionRows,
+  actionData,
   actions,
 }) => {
   const { width } = useWindowSize();
@@ -43,7 +43,7 @@ const Actions = ({
     return [firstTwoActions.reverse(), extraActions];
   }, [actions, width]);
 
-  if (!actionRows) {
+  if (!actionData) {
     return null;
   }
 
@@ -63,7 +63,7 @@ const Actions = ({
             <Dropdown.Item
               className={action.className}
               key={action.buttonText}
-              onClick={() => action.handleClick(actionRows)}
+              onClick={() => action.handleClick(actionData)}
               disabled={action.disabled}
             >
               {action.buttonText}
@@ -87,7 +87,7 @@ const Actions = ({
               [action.className]: action.className,
               'ml-2': true,
             })}
-            onClick={() => action.handleClick(actionRows)}
+            onClick={() => action.handleClick(actionData)}
             key={action.buttonText}
             disabled={action.disabled}
           >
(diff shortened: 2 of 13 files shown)


www/src/pages/table/datatable.mdx
```diff
@@ -336,49 +336,62 @@ can see in the console what the fetchData function is called with.
...
 ### Bulk Actions
+Bulk actions are action that are enacted on specific table rows. The bulk action click handler is called with the selected rows.
 The first two bulk actions will be displayed as buttons, while the remaining bulk actions will be displayed in a
-overflow dropdown menu. The bulk actions are called with the selected rows.
+overflow dropdown menu.
+Bulk actions are not visible unless rows have been selected.
 
 ### Actions
-An action can be definied as an additional column on the table. The Cell property can be definied to display
+An action can also be definied as an additional column on the table. The Cell property can be definied to display
```
adamstankiewicz, line 358: nit: "definied" -> "defined"

src/DataTable/Actions.jsx
```diff
@@ -0,0 +1,120 @@
...
+import { MoreVert } from '../../icons';
+import {
+  Button, Dropdown, useWindowSize, Icon, breakpoints,
+} from '..';
+
+export const DROPDOWN_BUTTON_TEXT = 'More actions';
+export const SMALL_SCREEN_DROPDOWN_BUTTON_TEXT = 'Actions';
+
+// eslint-disable-next-line react/prop-types
+const CustomToggle = React.forwardRef(({ children, onClick }, ref) => (
```
adamstankiewicz, line 13: Do we expect `CustomToggle` to always receive an `onClick` prop? If it's not provided (i.e., `'undefined`) without a default prop, calling the function directly within the `onClick` function below will throw an error.

src/DataTable/Actions.jsx
```diff
@@ -0,0 +1,120 @@
...
+    {children}
+  </Button>
+));
+
+const Actions = ({
+  className,
+  actionData,
+  actions,
+}) => {
+  const { width } = useWindowSize();
```
adamstankiewicz, line 31: I don't think this is something to address through this PR, but rather a thought question: should the responsive behavior of `DataTable` be based on window size, or the width of the component(s) themselves?

For example, I could imagine a potential scenario where `DataTable` is used side-by-side in a two column layout, displaying different data in each table. The window width hasn't changed but the `DataTable` width has, and the component should probably update accordingly 🤔

src/DataTable/ActionDisplay.jsx
```diff
@@ -0,0 +1,24 @@
...
+  const noBulkActions = bulkActions?.length < 1;
+  const noTableActions = tableActions?.length < 1;
+  const noActionsAvailable = noBulkActions && noTableActions;
+  const noRowsSelected = selectedFlatRows?.length < 1;
+
+  // return null if there are no actions or if there are only bulk actions and no selected rows
+  if (noActionsAvailable || (noTableActions && noRowsSelected)) {
+    return null;
+  }
+  if (!noRowsSelected) {
```
adamstankiewicz, line 18: nit: got slightly confused by the double negative here. might consider renaming this variable to `hasRowsSelected` and checking for a falsey value in the above conditional?

src/DataTable/tests/ActionDisplay.test.jsx
```diff
@@ -0,0 +1,85 @@
...
+    );
+    expect(wrapper.find(ActionDisplay).text()).toEqual('');
+  });
+  it('displays tableActions', () => {
+    const wrapper = mount(
+      <ActionDisplayWrapper
+        value={{ ...instance, tableActions: [firstAction, secondAction], selectedFlatRows: [{}, {}] }}
+      />,
+    );
+    expect(wrapper.find(BulkActions)).toHaveLength(1);
```
adamstankiewicz, line 59: Is this supposed to be asserting for the existence of `BulkActions` when the test case is for displaying `tableActions`? 

Kind of curious why this is currently passing when, as far as I can tell, `ActionDisplayWrapper` should not be rendering `BulkActions`?

src/DataTable/index.jsx
```diff
@@ -171,7 +179,8 @@ DataTable.propTypes = {
   initialTableOptions: PropTypes.shape({}),
   /** Total number of items */
   itemCount: PropTypes.number.isRequired,
-  /** Actions to be performed on the table. */
+  /** Actions to be performed on selected rows of the table. Called with the selected rows.
+   *  Not displayed if rows are not selected. */
```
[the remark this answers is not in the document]
raq929: Yes, I like it. I had trouble with enough thinking through the logic that I made a truth table; it's just confusing.

src/DataTable/Actions.jsx
raq929: Yeah, it's always gonna get an onClick prop, it opens the dropdown.

src/DataTable/Actions.jsx
raq929: 🤔  Yeah, I think if we run into that use case we should update. 😄

src/DataTable/ActionDisplay.jsx
raq929: Yeah, I debated doing that, will change.

src/DataTable/tests/ActionDisplay.test.jsx
raq929: OH THAT IS TRUE AND I DO NOT LIKE IT. Will investigate.

src/DataTable/ActionDisplay.jsx
```diff
@@ -0,0 +1,24 @@
+import React, { useContext } from 'react';
+import BulkActions from './BulkActions';
+import TableActions from './TableActions';
+import DataTableContext from './DataTableContext';
+
+// handles layout for filters, status, and bulk actions
```
abutterworth, line 6: Is this comment leftover from something? Seems like it doesn't apply entirely

src/DataTable/ActionDisplay.jsx
```diff
@@ -0,0 +1,24 @@
...
+const ActionDisplay = () => {
+  const { bulkActions, tableActions, selectedFlatRows } = useContext(DataTableContext);
+  const noBulkActions = bulkActions?.length < 1;
+  const noTableActions = tableActions?.length < 1;
```
abutterworth, line 10: The `?.` syntax here isn't preventing any errors if bulkActions were undefined. It never is because of default props in this file: https://github.com/edx/paragon/blob/5e7c3e97c24e5a7e91646540b3572d0494a5f603/src/DataTable/index.jsx#L114

That and `undefined < 1` is false as well. May be worth removing to avoid any false sense of safety.

src/DataTable/Actions.jsx
abutterworth: I agree it should be responsive to the container. There's a hook in frontend-app-learning (https://github.com/edx/frontend-app-learning/blob/master/src/generic/tabs/useIndexOfLastVisibleChild.js) That can help with this, but is potentially overkill. I agree we should defer the work for a future improvement.

src/DataTable/Actions.jsx
abutterworth, line 13: Could we name this `CustomDropdownToggle`? I didn't know what this component was until I read below.

src/DataTable/Actions.jsx
```diff
@@ -0,0 +1,120 @@
...
+          variant = (idx === 1 && visibleActions.length === 2) ? 'brand' : 'outline-primary';
+        }
+
+        return (
+          <Button
+            variant={variant}
+            className={classNames({
+              [action.className]: action.className,
+              'ml-2': true,
+            })}
```
abutterworth, line 89:
```suggestion
            className={classNames({ 'ml-2': true }, action.className)}
```
Minor suggestion take it or leave it. `classNames` will ignore undefined arguments so you can chain potentially missing class names rather than the object syntax you have above.

src/DataTable/Actions.jsx
```diff
@@ -0,0 +1,120 @@
...
+    onClick={(e) => {
+      e.preventDefault();
+      onClick(e);
+    }}
+  >
+    {children}
+  </Button>
+));
+
+const Actions = ({
```
abutterworth, line 26: I don't have a better suggestion, not noting: the name of this component made its purpose a little vague to me.

src/DataTable/TableActions.jsx
```diff
@@ -0,0 +1,25 @@
+import React, { useContext } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import DataTableContext from './DataTableContext';
+import Actions from './Actions';
+
+const TableActions = ({
```
abutterworth, line 8: I know we discussed this name before, but I wonder if `Actions` above is better named `TableActions` and this component `StaticActions` or `DefaultActions`.

src/DataTable/Actions.jsx
abutterworth: Am I correct in thinking this component or some form of it is something generic outside datatable that we can add? Like `ActionRow` that can handle overflows, etc?

abutterworth (approved):
Minor thoughts, looks good

src/DataTable/ActionDisplay.jsx
raq929: IT SURE DOES NOT. Will remove.

src/DataTable/Actions.jsx
raq929: Yeah, absolutely.

src/DataTable/Actions.jsx
raq929: Oh, yeah, this code is from before I knew that.

src/DataTable/ActionDisplay.jsx
raq929: I mean, it could be undefined if someone were using this component on it's own? idk.

src/DataTable/Actions.jsx
raq929: Yeah, this form of it *is* pretty generic and we could make it more widely available. 
I don't love `ActionRow` because it doesn't take up a whole row? But I don't have a better idea.

raq929: I'm gonna call it `CollapseableButtonGroup`

PR #641, opened 2021-02-23
Add dependent usage analyzer

abutterworth:
# Analyzing Paragon Dependents

Generate Paragon component usage information within a dependent projects with this command line tool. This tool uses babel to parse a series of javascript projects as an abstract syntax tree and walks through it to gather information about usage of Paragon components (version and file line numbers).

## What will we do with this output?

Generating the usage of Paragon components in Open edX frontend projects is the first step toward creating an automated dashboard for what components at what versions are used in which Open edX projects. 

This PR fill be followed up with additional Github action workflows to automate running this script.

## Usage

Make sure you're in this `dependent-usage-analyzer` directory, and then:

```
npm install
```

```
npm run analyze path/to/projects -- --out path/to/output.json
```

### DIFF
--- a/dependent-usage-analyzer/checkout-dependents.sh
+++ b/dependent-usage-analyzer/checkout-dependents.sh
@@ -2,27 +2,27 @@
 mkdir .projects
 (
   cd .projects &&
+  git clone git@github.com:edx/frontend-app-account.git --depth 1
+  git clone git@github.com:edx/frontend-app-admin-portal.git --depth 1
+  git clone git@github.com:edx/frontend-app-authn.git --depth 1
+  git clone git@github.com:edx/frontend-app-course-authoring.git --depth 1
+  git clone git@github.com:edx/frontend-app-discussions.git --depth 1
+  git clone git@github.com:edx/frontend-app-ecommerce.git --depth 1
+  git clone git@github.com:edx/frontend-app-gradebook.git --depth 1
+  git clone git@github.com:edx/frontend-app-learner-portal-enterprise.git --depth 1
+  git clone git@github.com:edx/frontend-app-learner-portal-programs.git --depth 1
+  git clone git@github.com:edx/frontend-app-learning.git --depth 1
+  git clone git@github.com:edx/frontend-app-payment.git --depth 1
+  git clone git@github.com:edx/frontend-app-profile.git --depth 1
+  git clone git@github.com:edx/frontend-app-publisher.git --depth 1
+  git clone git@github.com:edx/frontend-app-support-tools.git --depth 1
+  git clone git@github.com:edx/frontend-component-cookie-policy-banner.git --depth 1
+  git clone git@github.com:edx/frontend-component-footer.git --depth 1
+  git clone git@github.com:edx/frontend-component-header-edx.git --depth 1
+  git clone git@github.com:edx/frontend-component-header.git --depth 1
+  git clone git@github.com:edx/frontend-component-site-header.git --depth 1
+  git clone git@github.com:edx/frontend-platform.git --depth 1
+  git clone git@github.com:edx/frontend-template-application.git --depth 1
   git clone git@github.com:edx/prospectus.git --depth 1
   git clone git@github.com:edx/studio-frontend.git --depth 1
-  git clone git@github.com:edx/frontend-app-gradebook.git --depth 1
-  git clone git@github.com:edx/frontend-app-admin-portal.git --depth 1
-  git clone git@github.com:edx/frontend-template-application.git --depth 1
-  git clone git@github.com:edx/frontend-platform.git --depth 1
-  git clone git@github.com:edx/frontend-app-publisher.git --depth 1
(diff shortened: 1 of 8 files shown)


dependent-usage-analyzer/checkout-dependents.sh
```diff
@@ -0,0 +1,28 @@
+# checkout open edx frontend repos
+mkdir .projects
+(
+  cd .projects &&
+  git clone git@github.com:edx/prospectus.git --depth 1
```
raq929, line 5: It might be worth it to put these in alphabetical order so you know you have them all? (I have a keyboard shortcut on vscode for alphabetizing, but I don't remember how I got it. 😛 )

dependent-usage-analyzer/index.js
```diff
@@ -0,0 +1,126 @@
...
+}
+
+function getPackageInfo(dir) {
+  try {
+    // Package lock contains the actual Paragon version rather than a range in package.json
+    const { dependencies } = JSON.parse(fs.readFileSync(`${dir}/package-lock.json`, { encoding: 'utf-8' }));
+    const { name, repository } = JSON.parse(fs.readFileSync(`${dir}/package.json`, { encoding: 'utf-8' }));
+
+    return {
+      version: dependencies && dependencies['@edx/paragon'] ? dependencies['@edx/paragon'].version : 'unknown',
```
raq929, line 30: probably for later, but I think we'll also want to look in `peerDependencies`.

raq929 (approved):
Yessss I love it, in the glorious future I want cool charts. 
😄

dependent-usage-analyzer/index.js
abutterworth: Right now the version we're pulling out is from package-lock so that we get the exact version used. Agreed I think it makes sense to fallback to peer dependency ranges there's no direct dependency in the package-lock.

dependent-usage-analyzer/checkout-dependents.sh
abutterworth: 👍

dependent-usage-analyzer/checkout-dependents.sh
```diff
@@ -0,0 +1,28 @@
+# checkout open edx frontend repos
+mkdir .projects
+(
+  cd .projects &&
+  git clone git@github.com:edx/frontend-app-account.git --depth 1
```
adamstankiewicz, line 5: Would this approach handle finding Paragon usage in edx-platform? Is it worth trying to find usage there?

dependent-usage-analyzer/index.js
```diff
@@ -0,0 +1,126 @@
...
+  ];
+  // Gather all js and jsx source files
+  return glob.sync(`${dir}/**/*.{js,jsx}`, { ignore });
+}
+
+function getPackageInfo(dir) {
+  try {
+    // Package lock contains the actual Paragon version rather than a range in package.json
+    const { dependencies } = JSON.parse(fs.readFileSync(`${dir}/package-lock.json`, { encoding: 'utf-8' }));
+    const { name, repository } = JSON.parse(fs.readFileSync(`${dir}/package.json`, { encoding: 'utf-8' }));
```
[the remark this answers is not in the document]
abutterworth: If there are nested package.json files (a project in a project) components will probably be double counted. I could inspect these package JSONs and only consider them worth analyzing if they have paragon as a dependency or peerDependency. That should avoid any double counts.

dependent-usage-analyzer/checkout-dependents.sh
abutterworth: It may, I can try it.

dependent-usage-analyzer/index.js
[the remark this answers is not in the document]
adamstankiewicz: Hmm, I think I was referring more to packages that are siblings  of each other, for example:

```
- packages
  - packageA
    - package.json
    - package-lock.json
  - packageB
    - package.json
    - package-lock.json
  - packageC
    - package.json
    - package-lock.json
```

Nested package.json files (project in a project) is definitely another valid scenario, too, though.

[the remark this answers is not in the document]
abutterworth: As is this package handles this case. Project roots are identified by the presence of a package.json file at any depth in the directory this tool is pointed at. See this function:
```

dependent-usage-analyzer/checkout-dependents.sh
abutterworth: With some additional resilience added around parsing javascript code with babel this works for edx-platform. It's now been added to the checkout script above.

dependent-usage-analyzer/index.js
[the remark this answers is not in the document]
abutterworth: I'll wait until nested projects become a double counting issue to add a fix.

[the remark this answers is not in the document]
adamstankiewicz: Sweet, thanks for confirming the sibling packages use case :)

PR #642, opened 2021-02-24
fix: change clear selection to sentence case

raq929:
Also allows users to pass clear selection text as a prop, to aid i11n


adamstankiewicz (approved):
:shipit:

PR #646, opened 2021-03-01
docs: update mdx docs to use onClose for modal

davidjoy:
When changing the close -> onClose function for ModalLayer and ModalPopup, I forgot to update the associated mdx docs.

### DIFF
--- a/www/src/pages/components/modal-layer.mdx
+++ b/www/src/pages/components/modal-layer.mdx
@@ -24,7 +24,7 @@ A generic component for creating accessibile modal layer objects. The modal laye
     <>
       <Button variant="outline-primary" onClick={open}>Open modal layer</Button>
 
-      <ModalLayer isOpen={isOpen} close={close}>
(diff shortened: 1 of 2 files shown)


www/src/pages/components/modal-layer.mdx
```diff
@@ -18,13 +18,13 @@ A generic component for creating accessibile modal layer objects. The modal laye
 
 ```jsx live
 () => {
-  const [isOpen, open, close] = useToggle(false);
+  const [isOpen, open, onClose] = useToggle(false);
```
abutterworth, line 21: Would prefer to keep these names not mirror onClose (setClose or close)

abutterworth (approved):

www/src/pages/components/modal-layer.mdx
davidjoy: kk, will change back!

PR #650, opened 2021-03-02
fix: update default styles to be less different from edx.org brand

abutterworth:
Give edx.org developers higher confidence that what they ship in open edx applications will look similar when the edx.org brand is applied by reducing major differences between paragon default styles and edx brand styles.:
- Match font sizes in the type ramp (though line-heights are slightly different)
- Update colors to utilize the the color system in a similar way to edx.org. Notably: adds a magenta “brand” color.

The differences we'll maintain:
- Font family (system default san-serif vs. edx.org's use of Inter)
- Border radii. (rounded buttons vs. edx.org's square)

image
image

### DIFF
--- a/scss/core/_variables.scss
+++ b/scss/core/_variables.scss
@@ -37,7 +37,7 @@ $grays: map-merge(
 
 $blue: #23419f !default;
 $red: #C32D3A !default;
-$yellow: #D69600 !default;
+$yellow: #ffd900 !default;
 $green: #178253 !default;
 $teal: #006DAA !default;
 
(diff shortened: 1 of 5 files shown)


scss/core/_variables.scss
```diff
@@ -534,27 +538,27 @@ $font-weight-bold:            700 !default;
...
-$h5-font-size:                $font-size-base * 1.25 !default;
-$h6-font-size:                $font-size-base !default;
-
-$h1-mobile-font-size:                $font-size-base * 2 !default;
-$h2-mobile-font-size:                $font-size-base * 1.75 !default;
-$h3-mobile-font-size:                $font-size-base * 1.5 !default;
-$h4-mobile-font-size:                $font-size-base * 1.375 !default;
-$h5-mobile-font-size:                $h5-font-size !default;
-$h6-mobile-font-size:                $h6-font-size !default;
+$line-height-base:            1.55555555555555 !default;
```
raq929, line 541: Wait, what? Why?

raq929 (approved):
Thaaanks, I think this will make things way easier.

scss/core/_variables.scss
abutterworth: Matttttttttthhhhhssssss.  We would like our line-height base to net out at 28px like the design spec, but as Bootstrap is today I think we're required to use a unit-less line-height because some math with another unit is done with it later. 1.5555 repeating * 18px is 28px.

PR #656, opened 2021-03-04
[BD-35] PAR-280 feat: change styles according to new dropdown component design

amalbas:
@abutterworth 
Ticket on Jira: https://openedx.atlassian.net/browse/PAR-280
Update dropdown component according to new design
dropwdown-old
dropdown-new

### DIFF
--- a/src/Dropdown/Dropdown.scss
+++ b/src/Dropdown/Dropdown.scss
@@ -1,2 +1,16 @@
 @import "variables";
 @import "~bootstrap/scss/dropdown";
+
+.dropdown-toggle::after {
+    border: 0;
+    border-style: solid;
+    border-width: 0.15rem 0.15rem 0 0;
+    content: '';
+    height: 0.45rem;
+    left: 0.25rem;
+    position: relative;
+    top: 0.15rem;
+    transform: rotate(135deg);
+    vertical-align: super;
+    width: 0.45rem;
+}
--- a/src/Dropdown/_variables.scss
+++ b/src/Dropdown/_variables.scss
@@ -3,7 +3,7 @@
 //
 // Dropdown menu container and contents.
 
-$dropdown-min-width:                10rem !default;
+$dropdown-min-width:                18rem !default;
 $dropdown-padding-x:                0 !default;
 $dropdown-padding-y:                .5rem !default;
 $dropdown-spacer:                   .125rem !default;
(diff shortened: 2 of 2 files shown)


edx-webhook:
Thanks for the pull request, @amalbas! I've created [BLENDED-784](https://openedx.atlassian.net/browse/BLENDED-784) to keep track of it in Jira.


src/Dropdown/Dropdown.scss
```diff
@@ -1,2 +1,10 @@
 @import "variables";
 @import "~bootstrap/scss/dropdown";
+
+.dropdown-toggle::after {
+    border: 0;
+    content: "\f107";
+    font-family: "fontawesome";
```
abutterworth, line 7: Paragon we should avoid fontawesome icon font dependencies. Could you bake in this svg: 
https://github.com/edx/paragon/blob/master/icons/svg/expand_more.svg

Similar to how this one form errors is done?
- https://github.com/edx/brand-edx.org/blob/master/paragon/_overrides.scss#L180
- https://github.com/edx/paragon/blob/master/src/Form/_variables.scss#L130

amalbas: @abutterworth how can we define the variable to fill the svg? because the text color comes from the button and it is defined in a mixin.

abutterworth: `fill="currentColor"` in an svg will inherit the text color of the html. https://css-tricks.com/cascading-svg-fill-color/

amalbas:
@abutterworth fill="currentColor" doesn't work in css :https://stackoverflow.com/a/20515297 only works in inline svg, so maybe Do we have to modify the react files?

src/Dropdown/Dropdown.scss
amalbas: @abutterworth fill="currentColor" doesn't work in css :https://stackoverflow.com/a/20515297 only works in inline svg, so maybe Do we have to modify the react files?

abutterworth: Hmm, I see. Let's avoid wading into react for now. Is it possible to get this done with a css only solution the looks like our chevron svg? https://codepen.io/egf/pen/jzylw

amalbas: It works, I updated the new design image with the result.

abutterworth (approved):

PR #657, opened 2021-03-08
fix: google translate issue with text nodes

waheedahmed:
Fixes `Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.`

For more details see https://github.com/facebook/react/issues/11538#issuecomment-390386520


zainab-amir (approved):

waheedahmed:
Hello @abutterworth, just a note about the change. Let me know if there is any concern. Thanks!

PR #659, opened 2021-03-10
fix: do not memoize instance

raq929:
Causes table not to rerender with new data

### DIFF
--- a/src/DataTable/index.jsx
+++ b/src/DataTable/index.jsx
@@ -73,19 +73,14 @@ function DataTable({
     // Stringifying the data gives a quick way of checking deep equality
   }, [fetchData, JSON.stringify(instance.state)]);
 
-  const enhancedInstance = useMemo(() => ({
+  const enhancedInstance = {
     ...instance,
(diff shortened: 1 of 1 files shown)


src/DataTable/index.jsx
```diff
@@ -73,19 +73,14 @@ function DataTable({
     // Stringifying the data gives a quick way of checking deep equality
   }, [fetchData, JSON.stringify(instance.state)]);
 
-  const enhancedInstance = useMemo(() => ({
+  const enhancedInstance = {
```
binodpant, line 76: lol, the `instance` here is the react table instance right?

abutterworth (approved):
no need for useMemo?

binodpant:
can you send me instructions to test locally? I think I sort of know module.config.js for this, but not 100%

src/DataTable/index.jsx
raq929: yep

binodpant:
works like magic! it's fast too!!

PR #660, opened 2021-03-12
feat: Add form components

abutterworth:
This PR updates `Form.Group` `Form.Control` `Form.Control.Feedback`, `Form.Text`, and `Form.Label`. These components are a set of interoperable, accessible components and some tools to decorate ‘box’ style inputs (text, select, textarea) with leading and trailing icons or buttons and floating labels. 

Next step is to create probably three “easy mode” components SelectField TextField TextareaField. 

image


scss/core/_variables.scss
```diff
@@ -628,7 +628,7 @@ $input-btn-padding-y:         0.5625rem !default;
 $input-btn-padding-x:         1rem !default;
 $input-btn-font-family:       null !default;
 $input-btn-font-size:         1.125rem !default;
-$input-btn-line-height:       1.33334 !default;
+$input-btn-line-height:       1.3333 !default;
```
raq929, line 631: Out of curiosity, why this change?

src/Form/FormControlDecoratorGroup.test.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+import { FormFieldContext } from './FormFieldContext';
+import FormControlDecoratorGroup from './FormControlDecoratorGroup';
+import { FORM_CONTROL_SIZES } from './constants';
+
+describe('FormFieldDecoratorGroup', () => {
+  it('renders', () => {
+    const wrapper = mount((
+      <FormControlDecoratorGroup
+        leadingElement="before"
```
raq929, line 12: I know you're still adding tests, but we should probably test different options for these?

src/Form/FormField.scss
```diff
@@ -0,0 +1,356 @@
...
+    line-height: $line-height;
+    font-size: $font-size;
+  }
+}
+select.form-control {
+  background: $custom-select-bg $custom-select-background;
+  appearance: none;
+
+}
+@mixin form-control-floating-label-float(
```
raq929, line 130: Yay for using mixins!

src/Form/FormFieldContext.jsx
```diff
@@ -0,0 +1,93 @@
...
+const FormFieldContextProvider = ({
+  children,
+  id,
+  isInvalid,
+  isValid,
+  onFocus,
+  onBlur,
+  size,
+}) => {
+  const resolvedId = React.useMemo(() => id || newId('form-field'), [id]);
```
raq929, line 45: Should we also use `newId` with a user-provided id, or are we just trusting people to not mess it up?

src/Form/FormFieldControl.jsx
```diff
@@ -0,0 +1,57 @@
...
+      isValid={isValid}
+      className={classNames(className, {
+        'has-value': hasValue,
+      })}
+      onFocus={onFocus}
+      /*
+        Force a whitespace string for a placeholder if none is defined so we can
+        use the :placeholder-shown pseudo-class when styling floating labels
+      */
+      placeholder={placeholder || ' '}
```
raq929, line 38: Will this whitespace string placeholder affect screen readers in a negative way?

src/Form/FormFieldControl.jsx
```diff
@@ -0,0 +1,57 @@
...
+  const [hasValue, checkInputEventValue] = useHasValue({ defaultValue: props.defaultValue, value: props.value });
+
+  return (
+    <FormControl
+      {...props}
+      ref={ref}
+      size={size}
+      id={id}
+      isInvalid={isInvalid}
+      isValid={isValid}
```
raq929, line 29: I'm skeptical of something that has `isInvalid` and `isValid` states. What happens if they are both true or both false at the same time? (And/or can they be renamed to be more descriptive about what they are referring to?

src/Form/fieldUtils.js
```diff
@@ -0,0 +1,53 @@
...
+const callAllHandlers = (...handlers) => {
+  const unifiedEventHandler = (event) => {
+    handlers
+      .filter(handler => typeof handler === 'function')
+      .forEach(handler => handler(event));
+  };
+  return unifiedEventHandler;
+};
+
+const useHasValue = ({ defaultValue, value }) => {
```
raq929, line 22: Tell me about why we need to do this?

www/src/pages/components/form-field.mdx
```diff
@@ -0,0 +1,202 @@
...
+    >
+      <option>1</option>
+      <option>2</option>
+      <option>3</option>
+      <option>4</option>
+      <option>5</option>
+    </SelectField>
+```
+
+
```
raq929, line 196: If it has different states for success or error, I'd love to see those in the docs. :)

raq929 (commented):
I'm super excited about this!
I didn't do a super-thorough review, and mostly I've got questions for you. :)

scss/core/_variables.scss
abutterworth: Rounding up caused an off by 1 pixel alignment issue of the floating label and where the input text would be.

src/Form/FormControlDecoratorGroup.test.jsx
abutterworth: Added another test for rendering nodes instead of strings.

src/Form/FormField.scss
abutterworth: 🌀

src/Form/FormFieldContext.jsx
abutterworth: I think trusting is better. I believe newId can cause issues for some server rendered react apps (gatsby) since the index gets reset to zero when the page is loaded. I'm not sure what prospectus has been doing about this, but this would be their escape hatch, if it is indeed and issue.

src/Form/FormFieldControl.jsx
abutterworth: I agree with you, but this props api matched the current FormControl from react bootstrap, making this backward compatible. I think we should keep it as is for now. Re: `(And/or can they be renamed to be more descriptive about what they are referring to?` what do you mean?

src/Form/fieldUtils.js
abutterworth: ultimately this is for floating labels. There is no CSS selector to select inputs or selects that have a value. (and a floating label for an input with a value should be floated above). The one way this can be done is with a `::placeholder-shown` pseudo selector that ie11 does not support. That selector also does not work for select elements since they don't have placeholders. Bootstrap 5 uses this css approach and forces all selects to always float their label above.

TLDR: This provides use the ability to change the behavior of the component based on whether it has a value.

www/src/pages/components/form-field.mdx
abutterworth: For sure :) When I add those let's make sure those examples are there.

src/Form/FormControl.jsx
```diff
@@ -0,0 +1,105 @@
...
+  isInvalid: PropTypes.bool,
+};
+
+FormControl.defaultProps = {
+  as: 'input',
+  className: undefined,
+  id: undefined,
+  controlClassName: undefined,
+  defaultValue: undefined,
+  value: undefined,
```
raq929, line 95: I'm assuming that passing undefined to all these things just works? But if not, we should probably require them?

src/Form/FormControlDecoratorGroup.jsx
```diff
@@ -0,0 +1,65 @@
...
+      {leadingElement && <FormControlDecorator location="leading">{leadingElement}</FormControlDecorator>}
+      {trailingElement && <FormControlDecorator location="trailing">{trailingElement}</FormControlDecorator>}
+      {floatingLabel && <FormControlFloatingLabel>{floatingLabel}</FormControlFloatingLabel>}
+    </div>
+  );
+};
+
+FormControlDecoratorGroup.propTypes = {
+  children: PropTypes.node.isRequired,
+  leadingElement: PropTypes.node,
```
raq929, line 48: I think I asked this before but I don't remember the answer, what happens if several of these (`leadingElement`, `trailingElement`) are true at the same time?

src/Form/FormControlFeedback.jsx
```diff
@@ -0,0 +1,96 @@
...
+  type: undefined,
+  customIcon: undefined,
+};
+
+const FormControlFeedback = ({
+  children, type, icon, className, muted, ...props
+}) => {
+  const { controlId, getNewDescriptorId } = useFormGroupContext();
+  const [id, setId] = useState();
+  useEffect(() => setId(getNewDescriptorId()), [controlId]);
```
raq929, line 56: Talk to me about what controlId is doing here and why we need to generate a new id when it changes?

src/Form/fieldUtils.js
raq929: _shakes fist at ie11_

raq929 (approved):
So, mostly questions; my only concern is the one about multiple booleans being true at the same time.

src/Form/FormControlFeedback.jsx
abutterworth: While not strictly necessary, since the controlId is used as the root of the generated description id, I figured we'd regenerated it. Maybe not a great choice.

src/Form/FormControlDecoratorGroup.jsx
abutterworth: These are elements are slotted in containers so if they are multiple nodes they'll stack horizontally. It works, but should be avoided until the form control decorator is measuring their widths. Right now we assume with the width of these elements roughly matches the width of an icon button. If we want to support anything in here, we should measure the size of the rendered objects and update the horizontal padding of the underlying form control to ensure its content doesn't run underneath.

src/Form/FormControl.jsx
abutterworth: Our linter requires defaults be specified for non-required props. IMO an `undefined` default here is the true 'optional' and is a more appropriate value than `null`.

src/Form/FormControl.jsx
```diff
@@ -0,0 +1,105 @@
...
+        isValid={isValid}
+        className={classNames(controlClassName, {
+          'has-value': hasValue,
+        })}
+        onFocus={onFocus}
+        /*
+          Force a whitespace string for a placeholder if none is defined so we can
+          use the :placeholder-shown pseudo-class when styling floating labels
+        */
+        placeholder={placeholder || ' '}
```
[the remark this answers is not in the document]
abutterworth: We can delete this actually since, due to ie11 we're using the ham-handed `has-value` class instead of this css selector.

src/Form/FormControl.jsx
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import RBFormControl from 'react-bootstrap/FormControl';
+import { useFormGroupContext } from './FormGroupContext';
+import FormControlFeedback from './FormControlFeedback';
+import FormControlDecoratorGroup from './FormControlDecoratorGroup';
+
+import { useHasValue, callAllHandlers } from './fieldUtils';
```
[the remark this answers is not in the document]
abutterworth: Ya I think so to :)

src/Form/Form.scss
```diff
@@ -3,9 +3,310 @@
...
+    &.pgn__form-control-decorator-trailing {
+      padding-left: $input-padding-x-sm / 2;
+      padding-right: calc(#{$input-padding-y-sm} - #{2 * $input-border-width});
+    }
+  }
+}
+
+// Floating Labels
+
+@mixin form-control-floating-label-initial(
```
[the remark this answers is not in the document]
abutterworth: I'll move to a separate file.

PR #663, opened 2021-03-19
[BD-35] PAR-291 feat: change alert component according to new design

amalbas:
@abutterworth 
Ticket in Jira: https://openedx.atlassian.net/browse/PAR-291
Change alert component according to new design
alert-old
alert-new

### DIFF
--- a/src/Alert/Alert.scss
+++ b/src/Alert/Alert.scss
@@ -16,19 +16,31 @@
   margin-bottom: $alert-margin-bottom;
   border: $alert-border-width solid transparent;
   @include border-radius($alert-border-radius);
+  box-shadow: $alert-box-shadow;
+
+  & > :last-child {
+    margin-bottom: 0;
+  }
+
+  .alert-icon {
+    float: left;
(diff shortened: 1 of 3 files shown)


edx-webhook:
Thanks for the pull request, @amalbas! I've created [BLENDED-800](https://openedx.atlassian.net/browse/BLENDED-800) to keep track of it in Jira.


src/Alert/Alert.scss
```diff
@@ -14,22 +14,29 @@
   position: relative;
   padding: $alert-padding-y $alert-padding-x;
   margin-bottom: $alert-margin-bottom;
-  border: $alert-border-width solid transparent;
```
abutterworth, line 17: Rather than deleting this: let's set `$alert-border-width` to `0` in Alert/_variables.scss

abutterworth (changes requested):
Could we add a rule to ensure the last child of an alert has no margin? Without it we'll end up with extra space on the bottom edge

image

abutterworth (approved):

PR #665, opened 2021-03-20
docs: overhaul doc site and move component readmes to src

abutterworth:
A major update to the doc site. The design of pages and navigation is significantly changed.

Component documentation is now colocated with component source files. The doc site scans the src directory for markdown to parse.

image


scss/core/_typography.scss
```diff
@@ -1,5 +1,10 @@
 @import "~bootstrap/scss/type";
 
+body {
```
davidjoy, line 3: This is going to change our actual stylesheet in apps, right?

www/LICENSE
```diff
@@ -0,0 +1,14 @@
+The BSD Zero Clause License (0BSD)
```
davidjoy, line 1: Is this the correct license for us to use?

www/gatsby-config.js
```diff
@@ -1,54 +1,73 @@
-const path = require('path');
+const path = require("path")
```
davidjoy, line 1: wat!

scss/core/_typography.scss
abutterworth: Yes. Good catch, I should move this to the doc site css and release this change separately if desired.

www/LICENSE
abutterworth: nope.

www/gatsby-config.js
abutterworth: unintentional

www/src/components/Menu.jsx
```diff
@@ -0,0 +1,194 @@
...
+                Component status
+              </Link>
+            </li>
+            <li>
+              <Link to="/insights">
+                Usage insights
+              </Link>
+            </li>
+            <li>
+              <a
```
davidjoy, line 127: Huh, when I ran the docs site locally, this <a> tag is a button, and it's non-functional.

davidjoy (approved):
The "Design Documentation" button doesn't seem to go anywhere.

PR #667, opened 2021-03-23
[BD-35] PAR-288 feat: update styles for tabs component according to new design

amalbas:
Ticket on jira: https://openedx.atlassian.net/browse/PAR-288
Update styles for tabs component according to new design
tabs-new

### DIFF
--- a/src/Nav/Nav.scss
+++ b/src/Nav/Nav.scss
@@ -1,2 +1,42 @@
 @import "variables";
 @import "~bootstrap/scss/nav";
+
+.nav-tabs {
+  .nav-link {
+    color: $nav-tabs-link-color;
+    transition: $transition-base;
+    margin: 0;
+
+    &:hover {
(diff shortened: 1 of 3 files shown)


edx-webhook:
Thanks for the pull request, @amalbas! I've created [BLENDED-802](https://openedx.atlassian.net/browse/BLENDED-802) to keep track of it in Jira.


abutterworth (changes requested):
The edges of the bottom border are at an angle. Is there a way to keep them square? 
image

amalbas:
@abutterworth I fixed the angle border.

abutterworth (changes requested):
The bottom border should be inline with the "tab set"'s bottom border. Screenshot from the PR: 
image 
vs the design:
image

Also, the transition animation for the border is causing jitters when switching tabs.

Is it possible to address these issues without reworking the javascript component itself?

PR #668, opened 2021-03-23
fix: add headerClassName and cellClassName

raq929:
allow users to customize cell styles easily


ciduarte (approved):
Thanks for doing this!

abutterworth (approved):
Only question is whether they should both just be named `className` instead of prefixing them.

raq929:
Because one is flexbox and the other isn't, and also because you'll sometimes want styles to apply to one and not the other, I think it makes sense to namespace them.

PR #671, opened 2021-03-24
fix: data-table border styling

ciduarte:


src/DataTable/DataTable.scss
```diff
@@ -1,13 +1,16 @@
...
 $data-table-padding-y: 14px !default;
 $data-table-padding-small: 8px !default;
 $data-table-cell-padding: 8px 14px !default;
 $data-table-footer-position: center !default;
 
 .pgn__data-table-wrapper {
-  border: $data-table-border;
+  border-radius: $border-radius;
+  box-shadow: $data-table-box-shadow;
+  overflow: hidden;
```
abutterworth, line 13: Is this necessary. It is likely to cause issues with tooltips or menus in some cases.

ciduarte: I added this to avoid adding `border-radius` to both `pgn__data-table-container` and `pgn__data-table-footer`, but I see what you're saying. The Popover doesn't seem to be cutoff by the overflow hidden (is this a z-index thing?), but the filter menu is.

I can remove the overflow hidden

abutterworth (approved):
🚀

PR #675, opened 2021-03-29
fix: remove rounded corners on SearchField

raq929:


abutterworth:
This currently causes the default theme to have a square search field.  We should use the scss variables that govern input border radius:

```
$input-border-radius
$input-border-radius-lg
$input-border-radius-sm
```

PR #677, opened 2021-03-30
fix: reintroduce list group and media css

abutterworth:
These were incorrectly removed when collocating css and javascript components: https://github.com/edx/paragon/pull/606


rgraber (approved):
Thank you!

PR #678, opened 2021-03-30
feat: add Form.Radio and Form.RadioSet

abutterworth:
RadioSet inside a Form Group will be properly labeled and Radio children will share a name.
```
  <Form.Group controlId="whichColor">
    <Form.Label>
      Which color would you like?
    </Form.Label>
    <Form.RadioSet name="color">
      <Form.Radio value="red">Red</Form.Radio>
      <Form.Radio value="green">Green</Form.Radio>
      <Form.Radio value="blue">Blue</Form.Radio>
    </Form.RadioSet>
    <Form.Control.Feedback>
      The color you choose.
    </Form.Control.Feedback>
  </Form.Group>
```


src/Form/_FormRadio.scss
```diff
@@ -0,0 +1,212 @@
...
+//     }
+//   }
+
+//   .custom-control-input:disabled {
+//     &:checked ~ .custom-control-label::before {
+//       @include gradient-bg($custom-control-indicator-checked-disabled-bg);
+//     }
+//   // }?
+
+// .pgn__form-radio-label:before {
```
raq929, line 136: Remove commented code?

src/Form/FormRadio.jsx
```diff
@@ -0,0 +1,74 @@
...
+      })}
+    >
+      <span className={classNames('pgn__form-radio-lockup', controlClassName)}>
+        <input
+          {...props}
+          {...radioInputProps}
+          type="radio"
+          ref={ref}
+        />
+        <span className={classNames('pgn__form-radio-control', controlClassName)} />
```
davidjoy, line 39: Shouldn't the `pgn__form-radio-control` class name be on the actual input?  What's it mean that it's on this empty `span` between the input and the label?

src/Form/FormRadio.jsx
```diff
@@ -0,0 +1,74 @@
...
+        <span className={classNames('pgn__form-radio-label', controlClassName)}>
+          {children}
+        </span>
+      </span>
+      {description && (
```
davidjoy, line 44: So this component automatically adds valid/invalid feedback into the DOM if the group has been marked as invalid?  I thought other form components _didn't_ do this - whichever way we do it, we should probably be consistent or folks will have feedback labels that either don't show up or show up at the wrong time. :)

src/Form/FormRadioGroupFeedback.jsx
```diff
@@ -0,0 +1,47 @@
...
+const FormRadioGroupFeedback = ({ children, ...props }) => {
+  const {
+    isInvalid,
+    isValid,
+    getNewDescriptorId,
+    addNewDescriptorId,
+    removeDescriptorId,
+  } = useRadioGroupContext();
+  const [id, setId] = useState(props.id);
+  useEffect(() => {
```
davidjoy, line 17: Would it make sense to package this useEffect block into a reusable hook?  Looks like it's done in at least this component and the group legend below.

davidjoy: (I don't feel strongly and defer to your opinion in the end)

www/src/pages/index.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import Menu from '../components/Menu';
 import Layout from '../components/PageLayout';
-import { Button } from '~paragon-react'; // eslint-disable-line
+import { Button, Container, Row, Col } from '~paragon-react'; // eslint-disable-line
```
davidjoy, line 4: Were these just missing/it was broken? O_o

davidjoy (approved):
Had some questions and thoughts, wouldn't call them blocking unless you think they are.  Aproved!

src/Form/FormRadio.jsx
abutterworth: We are unable to style the input itself so this element is the visual replacement of the control. Could name it `pgn__form-radio-faux-control` if we felt it was important for clarity

src/Form/FormRadioGroupFeedback.jsx
abutterworth: feels like a great hook to make.

www/src/pages/index.jsx
abutterworth: I need to remove

src/Form/FormGroupContext.jsx
```diff
@@ -1,61 +1,82 @@
...
-  onFocus,
-  onBlur,
   size,
 }) => {
   const resolvedId = React.useMemo(() => controlId || newId('form-field'), [controlId]);
-  const [hasFocus, setHasFocusTrue, setHasFocusFalse] = useToggle(false);
-  const [describedByIds, getNewDescriptorId, removeDescriptorId] = useIdList(resolvedId);
+  const [describedByIds, useRegisteredDescriptorId] = useIdList(resolvedId);
+  const [labelledByIds, useRegisteredLabellerId] = useIdList(resolvedId);
+  const [controlIsGroup, useSetControlIsGroupEffect] = useStateEffect(false);
```
adamstankiewicz, line 38: naming suggestion/nit: `isControlGroup` vs. `controlIsGroup` (bit more indicative of a boolean).

src/Form/_FormText.scss
```diff
@@ -0,0 +1,46 @@
...
+.pgn__form-text-criteria-valid {
+  .pgn__icon {
+    color: $form-feedback-valid-color;
+  }
+}
+.pgn__form-text-criteria-invalid {
+  .pgn__icon {
+    color: $form-feedback-invalid-color;
+  }
+}
```
adamstankiewicz, line 46: tiniest of nits: new line at end of file

src/Form/FormGroupContext.jsx
```diff
@@ -1,61 +1,82 @@
...
     size,
-    hasFocus,
-    getNewDescriptorId,
-    removeDescriptorId,
-    'aria-describedby': describedByIds.length ? describedByIds.join(' ') : undefined,
-    onFocus: callAllHandlers(setHasFocusTrue, onFocus),
-    onBlur: callAllHandlers(setHasFocusFalse, onBlur),
   };
   return (
     <FormGroupContext.Provider value={contextValue}>
```
adamstankiewicz, line 82: Should `contextValue` be memoized with `useMemo` to avoid any unnecessary rendering of children components nested within this context provider? That way, the nested components will only get re-rendered when the context values actually change vs. any time the parent `FormGroupContextProvider` is re-rendered?

adamstankiewicz (approved):
LGTM. A couple nits and a question about using `useMemo` to compute the context value (non-blocking).

src/Form/FormGroupContext.jsx
abutterworth: 👍

src/Form/_FormText.scss
abutterworth: 👍

src/Form/FormGroupContext.jsx
abutterworth: I can add in a follow up PR if this is needed. I'll bug you about it in person

PR #680, opened 2021-03-31
feat: update display font sizes and line heights

rgraber:
Update all display font sizes and line heights to match design docs
Screen Shot 2021-03-31 at 3 58 43 PM

### DIFF
--- a/scss/core/_variables.scss
+++ b/scss/core/_variables.scss
@@ -1,4 +1,5 @@
 // Variables
+// Variables
 //
 // Variables should follow the `$component-state-property-size` formula for
 // consistent naming. Ex: $nav-link-disabled-color and $modal-content-box-shadow-xs.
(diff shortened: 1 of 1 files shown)


scss/core/_variables.scss
```diff
@@ -1,4 +1,5 @@
 // Variables
+// Variables
```
ciduarte, line 2: Was this a mistake?

ciduarte (approved):
I'm not sure what these values are supposed to be, but LGTM + one question

scss/core/_variables.scss
rgraber: yes, and I couldn't figure out what about it changed

abutterworth: // Våriables ?

abutterworth (approved):

PR #682, opened 2021-04-05
feat: add Checkbox, Switch, and CheckboxSet

abutterworth:
Add `Checkbox`, `Switch`, `CheckboxSet` and `useCheckboxSetValues` to facilitate check-style options.

Individual usage:
```
() => {
  const [isChecked, setChecked] = useState(false);
  const handleChange = e => setChecked(e.target.checked);
  return (
    <Form.Checkbox checked={isChecked} onChange={handleChange}>
      I want pizza.
    </Form.Checkbox>
  );
}
```

Set usage
```
() => {

  const [colorValues, { add, remove }] = useCheckboxSetValues(['green']);

  const handleChange = e => {
    if (e.target.checked) {
      add(e.target.value);
    } else {
      remove(e.target.value);
    }
  }
  return (
    <Form.Group>
      <Form.Label>Which Color?</Form.Label>
      <Form.CheckboxSet
        name="colors"
        onChange={handleChange}
        value={colorValues}
      >
        <Form.Checkbox value="red">Red</Form.Checkbox>
        <Form.Checkbox value="green">Green</Form.Checkbox>
        <Form.Checkbox value="blue">Blue</Form.Checkbox>
        <Form.Checkbox value="cyan" disabled>Cyan</Form.Checkbox>
      </Form.CheckboxSet>
    </Form.Group>
  );
}
```

### DIFF
--- a/src/Form/FormCheckboxSet.jsx
+++ b/src/Form/FormCheckboxSet.jsx
@@ -17,13 +17,13 @@ const FormCheckboxSet = ({
 }) => {
   const { getControlProps, useSetIsControlGroupEffect } = useFormGroupContext();
   useSetIsControlGroupEffect(true);
-  const controlProps = getControlProps(props);
   const className = classNames(
     'pgn__form-control-set',
     'pgn__form-checkbox-set',
     props.className,
     { 'pgn__form-checkbox-set-inline': isInline },
   );
+  const controlProps = getControlProps({ ...props, className });
   return (
     <FormCheckboxSetContextProvider
       name={name}
--- a/src/Form/FormCheckboxSet.jsx
+++ b/src/Form/FormCheckboxSet.jsx
@@ -1,0 +1,69 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { useFormGroupContext } from './FormGroupContext';
+import { FormCheckboxSetContextProvider } from './FormCheckboxSetContext';
+
+const FormCheckboxSet = ({
+  children,
+  name,
+  value,
+  defaultValue,
+  isInline,
+  onChange,
+  onFocus,
+  onBlur,
+  ...props
+}) => {
+  const { getControlProps, useSetIsControlGroupEffect } = useFormGroupContext();
+  useSetIsControlGroupEffect(true);
+  const controlProps = getControlProps(props);
+  const className = classNames(
+    'pgn__form-control-set',
+    'pgn__form-checkbox-set',
+    props.className,
+    { 'pgn__form-checkbox-set-inline': isInline },
+  );
+  return (
+    <FormCheckboxSetContextProvider
+      name={name}
+      value={value}
+      defaultValue={defaultValue}
+      onFocus={onFocus}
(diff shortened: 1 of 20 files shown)


src/Form/_FormCheckbox.scss
```diff
@@ -0,0 +1,157 @@
+.pgn__form-checkbox {
+  font-size: $font-size-base;
+  line-height: $line-height-base;
+  display: block;
+  flex-shrink: 0;
```
raq929, line 5: I thought flex-shrink was 0 by default?

src/Form/_FormCheckbox.scss
```diff
@@ -0,0 +1,157 @@
...
+  input {
+    border: 0px;
+    clip: rect(0px, 0px, 0px, 0px);
```
raq929, line 8: So, in finding out what the `clip` property is, I also found out that it is deprecated, and we should maybe use `clip-path`? https://css-tricks.com/clipping-masking-css/#the-new-clip-path
But idk because IE11.

davidjoy (approved):

raq929 (approved):
Aaaah, this is really cool I love it. 

One thing I noticed is that I tried to add isInline to a bunch of the CheckboxSets on the docs page, and for many of them it didn't work. For some I could wrap in Form.Group to get it to work, but for others I couldn't (the one with the Indeterminate one, is what I'm thinking of.)

I think if it's a prop on CheckboxSet, it should work regardless of where CheckboxSet is. 

Also, I need to get 15 minutes with you and talk about refs again, because I keep thinking I get them and then have trouble keeping track of what's happening when I'm reading code with them.

src/Form/FormCheckboxSet.jsx
```diff
@@ -0,0 +1,69 @@
...
+    <FormCheckboxSetContextProvider
+      name={name}
+      value={value}
+      defaultValue={defaultValue}
+      onFocus={onFocus}
+      onBlur={onBlur}
+      onChange={onChange}
+    >
+      <div
+        role="group"
```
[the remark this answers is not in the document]
abutterworth: In this case: role="group" has a different purpose/meaning than Form.Group (which doesn't apply a role="group"). Reference: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_group_role

src/Form/_FormCheckbox.scss
abutterworth: Defaults to 1: https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink#values

abutterworth:
@raq929 I fixed the isInline issue. Before it wouldn't apply the style if you had applied a custom class name to the set. Now it works no matter what.

src/Form/_FormCheckbox.scss
abutterworth: clip-path doesn't look like it's well supported :/  Maybe we wait until we drop ie11 support and then style the input directly like Bootstrap 5 does.

PR #683, opened 2021-04-06
feat: add five icons

abutterworth:
Add icons `ArrowForward` `ArrowBack` `Launch` `Settings` and `Program`.

image


adamstankiewicz (approved):

wittjeff:
Is that icon actually for 'launch' or 'opens in a new window'?

abutterworth:
@wittjeff The name is `Launch` though likely used for the purpose you describe. We left the name unchanged from google material

PR #686, opened 2021-04-09
feat: add ModalDialog component family

abutterworth:
Leverages `ModalLayer` to build a family of `ModalDialog` components for use to create `FullscreenModal`, `AlertModal`, `MarketingModal`, and `StandardModal`.

Includes new components:
- `ModalDialog`
- `ModalDialog.Header`
- `ModalDialog.Title`
- `ModalDialog.Hero`
- `ModalDialog.Hero.Background`
- `ModalDialog.Hero.Content`
- `ModalDialog.Body`
- `ModalDialog.Footer`
- `ActionRow` (a layout utility, documented separately) 
- `FullscreenModal`
- `AlertModal`
- `MarketingModal`
- `StandardModal`

Also includes an update to `IconButton` to make it alternatively compatible with Paragon icons instead of FontAwesome.

```
<ModalDialog
  title="My dialog"
  isOpen
  onClose={onClose}
  size="md"
  variant="default"
  hasCloseButton
>
  <ModalDialog.Header>
    <ModalDialog.Title>The title</ModalDialog.Title>
  </ModalDialog.Header>

  <ModalDialog.Body>
    <p>The content</p>
  </ModalDialog.Body>

  <ModalDialog.Footer>
    <ModalDialog.CloseButton>Cancel</ModalDialog.CloseButton>
  </ModalDialog.Footer>
</ModalDialog>
```


src/Modal/marketing-modal.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+```jsx live
+() => {
+  const [isOpen, open, close] = useToggle(false);
+  return (
+    <>
+      <Button variant="outline-primary" onClick={open}>Open marketing modal</Button>
+
+      <MarketingModal
+        title="My dialog"
```
raq929, line 26: I think this should not be here? Unless it has a purpose I'm not aware of. I'm not seeing this title actually appear anywhere.

src/ActionRow/README.md
```diff
@@ -0,0 +1,48 @@
...
+### Stacked Usage
+
+
+```jsx live
+<ActionRow isStacked>
+  <p className="x-small">
+    Bespoke leggings yuccie, portland umami readymade craft beer vaporware sriracha.
+  </p>
+  <Button variant="tertiary">
+    Go back
```
raq929, line 42: On the netlify preview, this button is smaller in size than the continue button, which irks my soul. Also it's below it, which I wouldn't expect from the code, so I would either put them in the other order or make a note of that in the docs.

src/IconButton/index.jsx
```diff
@@ -29,22 +31,31 @@ const IconButton = ({
       onClick={onClick}
       type="button"
     >
-      <FontAwesomeIcon
+      <IconComponent
```
raq929, line 34: Will this be a breaking change for people already using this? It looks like no, but I just want to double check.

src/Modal/AlertModal.jsx
```diff
@@ -0,0 +1,44 @@
...
+);
+
+AlertModal.propTypes = {
+  children: PropTypes.node.isRequired,
+  title: PropTypes.string.isRequired,
+  onClose: PropTypes.func.isRequired,
+  isOpen: PropTypes.bool,
+  hasCloseButton: PropTypes.bool,
+  size: PropTypes.oneOf(['sm', 'md', 'lg', 'xl', 'fullscreen']),
+  variant: PropTypes.oneOf(['default', 'warning', 'danger', 'success', 'dark']),
```
raq929, line 26: Do we not have these listed in a constant somewhere? If no, should we?

src/Modal/FullscreenModal.jsx
```diff
@@ -0,0 +1,40 @@
...
+);
+
+FullscreenModal.propTypes = {
+  children: PropTypes.node.isRequired,
+  title: PropTypes.string.isRequired,
+  onClose: PropTypes.func.isRequired,
+  isOpen: PropTypes.bool,
+  hasCloseButton: PropTypes.bool,
+  size: PropTypes.oneOf(['sm', 'md', 'lg', 'xl', 'fullscreen']),
+  variant: PropTypes.oneOf(['default', 'warning', 'danger', 'success', 'dark']),
```
raq929, line 22: OK, following up on my comment above, how about yes, let's throw these and the sized in a constant somewhere, we're gonna reuse them apparently. 😉

src/Modal/ModalDialog.jsx
```diff
@@ -0,0 +1,114 @@
...
+   */
+  hasCloseButton: PropTypes.bool,
+  /**
+   * Sizes determine the maximum width of the dialog box
+   */
+  size: PropTypes.oneOf(['sm', 'md', 'lg', 'xl', 'fullscreen']),
+  /**
+   * The visual style of the dialog box
+   */
+  variant: PropTypes.oneOf(['default', 'warning', 'danger', 'success', 'dark']),
```
raq929, line 83: Again, use a constant. :)

raq929 (approved):
Looking good! A few small comments. 
Also, it wasn't part of this work, but ModalPopup has a double title on the docs site.
Screen Shot 2021-04-12 at 1 53 38 PM

src/ActionRow/README.md
raq929: Screen Shot 2021-04-12 at 1 55 19 PM

(In the ActionRow docs)

src/Modal/marketing-modal.mdx
abutterworth: Dialogs must be labeled for a11y reasons. It's applied to `aria-label` on the dialog

raq929:
The real question is, why *are* fashion axe's so limited❓❓

src/ActionRow/README.md
abutterworth: I can note that in the docs! We aim to put the last item as the "primary" action. Horizontally that means the end in our language and vertically it means the top of the stack.

As far as size, these buttons should not be the same width. They happen to be close which is unfortunate, but right now our styles are not to assign arbitrary widths to buttons (either full width or natural width)

src/IconButton/index.jsx
abutterworth: Nope, no breaking changes here.

src/Modal/AlertModal.jsx
abutterworth: We could, but our docgen can't parse constants so our props table would just say 'one of VARIANTS' instead. wdyt?

PR #689, opened 2021-04-13
fix: resolve console error during testing with react-testing-renderer

raq929:
react-testing-renderer does not handle refs in the same way as the browser does, causing a large console error. This change resolves that error.

### DIFF
--- a/src/DataTable/IndeterminateCheckBox.jsx
+++ b/src/DataTable/IndeterminateCheckBox.jsx
@@ -7,7 +7,10 @@ const IndeterminateCheckbox = React.forwardRef(
     const resolvedRef = ref || defaultRef;
 
     React.useEffect(() => {
-      resolvedRef.current.indeterminate = indeterminate;
+      // this if(resolvedRef.current) prevents console errors in testing
+      if (resolvedRef.current) {
+        resolvedRef.current.indeterminate = indeterminate;
+      }
     }, [resolvedRef, indeterminate]);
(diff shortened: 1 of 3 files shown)


src/DataTable/IndeterminateCheckBox.jsx
```diff
@@ -7,7 +7,10 @@ const IndeterminateCheckbox = React.forwardRef(
     const resolvedRef = ref || defaultRef;
 
     React.useEffect(() => {
-      resolvedRef.current.indeterminate = indeterminate;
+      // this if(resolvedRef.current) prevents console errors in testing
+      if (resolvedRef.current) {
+        resolvedRef.current.indeterminate = indeterminate;
+      }
     }, [resolvedRef, indeterminate]);
```
abutterworth, line 14: is the dependency `resolvedRef.current`?

abutterworth (approved):
One question

alex-sheehan-edx (approved):
LGTM

src/DataTable/IndeterminateCheckBox.jsx
raq929: Yeah, the error I was getting was 
`Uncaught [TypeError: Cannot set property 'indeterminate' of null]`

raq929: The ref is there, but `.current` is not.

PR #691, opened 2021-04-14
feat: update IconButton styling

ciduarte:


src/IconButton/index.jsx
```diff
@@ -11,32 +11,32 @@ const IconButton = ({
...
+        `btn-icon-${size}`,
         attrs.className,
       )}
       onClick={onClick}
       type="button"
     >
-      <IconComponent
-        className={`iconbutton iconbutton__${variant} ${invertedIconClass} ${iconClassNames}`}
-        icon={icon}
-        alt={alt}
```
ciduarte, line 37: Chose to remove the `alt` because the icon in this scenario is purely decorative, and the aria-label on the button should describe the button's function.

src/IconButton/README.md
```diff
@@ -28,16 +31,20 @@ notes: ''
...
   return (
     <div className="d-flex">
-      <IconButton icon={icon} alt='Close' onClick={() => {}} variant="primary" />
-      <IconButton icon={icon} alt='Close' onClick={() => {}} variant="secondary" />
-      <IconButton icon={icon} alt='Close' onClick={() => {}} variant="success" />
-      <IconButton icon={icon} alt='Close' onClick={() => {}} variant="warning" />
-      <IconButton icon={icon} alt='Close' onClick={() => {}} variant="danger" />
-      <IconButton icon={icon} alt='Close' onClick={() => {}} variant="light" />
-      <IconButton icon={icon} alt='Close' onClick={() => {}} variant="dark" />
+      {variants.map((variant) => (
```
raq929, line 38: For the docs, I prefer to have them written out as they were; I find it easier to use when I'm developing.

src/IconButton/index.jsx
```diff
@@ -75,6 +76,7 @@ IconButton.propTypes = {
   onClick: PropTypes.func.isRequired,
   /** Type of button (uses Bootstrap options) */
   variant: PropTypes.oneOf(['primary', 'secondary', 'success', 'warning', 'danger', 'light', 'dark']),
+  size: PropTypes.oneOf(['sm', 'inline']),
```
raq929, line 79: It looks like you've got a default size of 'md', but that's not a valid type for the `size` prop?

src/IconButton/IconButton.test.jsx
```diff
@@ -13,21 +12,15 @@ describe('<IconButton />', () => {
     icon,
     variant,
   };
-  it('passes the icon and alt text to the icon', () => {
```
raq929, line 16: We should probably still test that we're passing alt text to the button.

raq929 (commented):
I'm excited about the inline variant, that seems super cool! 
I think there should be tests regarding the new `size` property.

Also it looks like the inverse variants got square?
Screen Shot 2021-04-27 at 7 40 07 PM

src/IconButton/README.md
ciduarte: Good point!

src/IconButton/README.md
```diff
@@ -47,20 +54,21 @@ notes: ''
...
-  const variants = ["primary", "secondary", "success", "warning", "danger", "light", "dark"]
+  const icon = FontAwesome.faBars;
+  const variants = ["brand", "primary", "secondary", "success", "warning", "danger", "light", "dark"];
 
   return (
     <div className="d-flex">
       {variants.map((variant) => {
         return (
-          <div class={`bg-${variant} p-2 m-2`}>
+          <div class={`bg-${variant} p-1`}>
```
ciduarte, line 64: @raq929 re: square inverse variant, they're not square there's just a background behind them because the accent colors are white. They'd disappear otherwise

src/IconButton/index.jsx
ciduarte: Yeah I wasn't sure if I should surface the md size or not because it's technically the default.

Class name `btn-icon-md` doesn't actually do anything in the css, but I felt it was weird for the default to be `btn-icon-` (just empty).

I don't know if I should surface the `md` size, or change the default to be empty and just not append the size class name if one wasn't provided?

src/IconButton/IconButton.test.jsx
raq929: Oh, so, as we speak I'm writing a test using `react-testing-library`, and I'm using `screen.getByAltText` to get the button. I'm kind of inclined *not* to change it to `aria-described` by, because that makes it actually kind of hard for `react-testing-library` to access it. I guess I can use `data-testid`, if we're passing all props along, but I'm not sure we are right now.

src/IconButton/index.jsx
ciduarte: I'm just going to surface the `md` as a size variant unless anyone has an objection

src/IconButton/README.md
raq929: oooooh, got it.

raq929 (approved):

PR #702, opened 2021-04-28
[PAR-449] feat: add emotion

abutterworth:
This PR is the first step in the proposal outlined here: https://openedx.atlassian.net/wiki/spaces/BPL/pages/2629731077/Adopt+CSS+in+JS+for+styling

- Use emotion to replace styling of `ActionRow` as a first component. (Chosen because it's not critical to theme). https://deploy-preview-702--paragon-edx.netlify.app/components/actionrow/
- Add `ParagonProvider` and theming capability
- Brand packages now require a theme.js file: https://github.com/edx/brand-edx.org/pull/34 https://github.com/edx/brand-openedx/pull/5

----

**BREAKING CHANGE:** Paragon now requires @emotion/react as a peer dependency to be present and for consuming applications to wrap their root elements with a `ParagonProvider` component to set a theme (or provide the default theme).


src/ActionRow/index.jsx
```diff
@@ -1,22 +1,46 @@
...
+  justify-content: flex-end;
+  & > * {
+    margin: 0;
+  }
+  & > * + * {
+    margin-left: .5rem;
+  }
+`;
+
+const stackedCSS = css`
```
muselesscreator, line 18: is nit-pick, but I feel like I would prefer the paradigm of an object of style values, rather than top-level variables...
ex: 
```
const styles = { 
  stacked: css``,
  horizontal: css``,
};

File: src/ParagonProvider/README.md [hunk 2]
```diff
@@ -0,0 +1,87 @@
...
+categories:
+- Theming
+status: 'Stable'
+designStatus: 'Done'
+devStatus: 'Done'
+notes: |
+
+---
+
+Use the `ParagonProvider` at the root of a React app to set the Paragon CSS-in-JS theme and locale.
```
muselesscreator, line 16: Should there be a note here or elsewhere that this is, in fact, required in order to use associated elements?

abutterworth: Good call, I'll add to the "getting started and installation" page

src/ActionRow/index.jsx
abutterworth: updatedd

src/ParagonProvider/README.md
abutterworth: Updated

src/ActionRow/index.jsx
```diff
@@ -1,22 +1,47 @@
...
-    ...props,
-    className: classNames(props.className, {
-      'pgn__action-row': !isStacked,
-      'pgn__action-row-stacked': isStacked,
-    }),
-  },
-  children,
+}) => (
+  <Component
+    css={isStacked ? styles.stacked : styles.horizontal}
```
davidjoy, line 40: Do you have any opinions on usage of `css` vs. using the `cx` function with `className` here?

The latter would provide an API more like the `classnames` package, letting us put the common properties together here on line 40, and use the conditional to add on the ones that are different.

I'm just reading and learning, not strongly opinionated about it yet. 😉

src/theme/theme.js
```diff
@@ -0,0 +1,12 @@
+// The default theme.js for Paragon. The structure of this object follows the
+// theme specification found here: https://system-ui.com/theme/. We prefer to add
+// to this theme over time rather than all valid keys at once to ensure everything
+// present here is used in the system.
+
+const theme = {
+  space: [
+    0, 4, 8, 12, 16, 24, 32, 48, 64, 80, 104, 128,
```
davidjoy, line 8: Are these numbers... pixels?  rem?  An example of their usage in the actual component might be instructive.

src/ActionRow/index.jsx
abutterworth: I was just thinking that would make sense to do

PR #720, opened 2021-05-17
fix: make useIsVisible hook more resilient

abutterworth:
Put usage of `IntersectionObserver` in a try catch so it does not need to be mocked in tests.

### DIFF
--- a/src/hooks/useIsVisible.jsx
+++ b/src/hooks/useIsVisible.jsx
@@ -18,6 +18,10 @@ const useIsVisible = (defaultIsVisible = true) => {
         };
       }
     } catch (e) {
+      const isReferenceError = e instanceof ReferenceError;
+      if (!isReferenceError) {
+        throw e;
+      }
       // Do nothing if an intersection observer can't be created.
     }
     return () => {};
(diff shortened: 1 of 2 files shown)


raq929 (changes requested):
I like it! One small change requested. Also, tell me more about what IntersectionObserver is doing here?

abutterworth:
IntersectionObserver is used to tell if the div "Sentinel" is visible on screen or not

src/hooks/useIsVisible.jsx
```diff
@@ -5,16 +5,24 @@ const useIsVisible = (defaultIsVisible = true) => {
...
+          entries.forEach(({ isIntersecting }) => {
+            setIsVisible(isIntersecting);
+          });
+        }, {});
+        observer.observe(sentinelRef.current);
+        return () => {
+          observer.disconnect();
+        };
+      }
+    } catch (e) {
```
raq929, line 20: Can it not just be `catch(isReferenceError) `? Is that too sensible for Javascript?

raq929 (approved):

src/hooks/useIsVisible.jsx
abutterworth: No, the `(e)` is the error itself, more like a function definition than an if statement https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch

PR #725, opened 2021-05-18
docs: document callbacks on collapsible

abutterworth:
https://openedx.atlassian.net/browse/PAR-446
The basic `Collapsible` accepts onToggle onOpen and onClose callbacks. Add them to the props api table and add an example.


awaisdar001 (approved):

awaisdar001:
@abutterworth  is going to bump the version? I mean does "docs" bump version at all?

abutterworth:
Docs doesn't bump the version, but there is no change to the API – only the documentation site which has been updated.

PR #728, opened 2021-05-18
fix: enforce text alternative rules for svg icon

abutterworth:
https://openedx.atlassian.net/browse/PAR-432

An svg with an "img" role must satisfy the following a11y requirements
 * It needs a text alternative in the form of aria-label, aria-labelledby, or screen-reader only text.
 * If no label is desired, aria-label will be set to an empty string and aria-hidden to "true".
 * focusable is set to false on the svg in all cases as a workaround for an ie11 bug

### DIFF
--- a/src/Icon/index.jsx
+++ b/src/Icon/index.jsx
@@ -5,6 +5,13 @@ import classNames from 'classnames';
 import newId from '../utils/newId';
 import withDeprecatedProps, { DEPR_TYPES } from '../withDeprecatedProps';
 
+/**
+ * An svg with an "img" role must satisfy the following a11y requirements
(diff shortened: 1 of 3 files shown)


src/Icon/index.jsx
```diff
@@ -15,19 +22,32 @@ function Icon({
...
     }
+
     return (
       <span
         className={classNames('pgn__icon', className)}
         id={id}
         {...attrs}
       >
-        <Component {...mergedSvgProps} />
+        <Component
```
abutterworth, line 41: `Component` is an `svg`

wittjeff (approved):

PR #735, opened 2021-05-25
feat: add paragon entry points for parts of core scss

abutterworth:
Creating these separate files will enable edx-platform to adopt Paragon more easily since some SCSS files in that project use only mixins or only utilities for example.

### DIFF
--- a/scss/core/_mixins.scss
+++ b/scss/core/_mixins.scss
@@ -1,0 +1,1 @@
+@import "~bootstrap/scss/mixins";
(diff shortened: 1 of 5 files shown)


raq929 (approved):

abutterworth:
Variables in Paragon cannot be pulled in apart from their styles right now. All the variables are collocated with their components so it would be a somewhat messy thing to do. I'm closing this PR since it would suggest someone could pull in `variables` and get all of them, when in fact they wouldn't.  If we do end up needing this functionality we can reopen and address it.

PR #738, opened 2021-05-31
feat: Alert text aligned with heading

adeelehsan:
VAN-499
Screen Shot 2021-05-31 at 7 02 57 PM

### DIFF
--- a/src/Alert/Alert.scss
+++ b/src/Alert/Alert.scss
@@ -27,6 +27,7 @@
     margin-right: $alert-icon-space;
     width: 1.25rem;
     height: 1.25rem;
+    flex-shrink: 0;
   }
 }
 
@@ -34,6 +35,11 @@
 .alert-heading {
   // Specified to prevent conflicts of changing $headings-color
   color: $alert-title-color;
+  display: flex;
+}
+
+// added to align content and icon
+.alert-content {
   display: flex;
 }
 
--- a/src/Alert/Alert.test.jsx
+++ b/src/Alert/Alert.test.jsx
@@ -1,0 +1,23 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import Alert from './index';
+
+import { Info } from '../../icons';
+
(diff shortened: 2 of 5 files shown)


src/Alert/Alert.scss
```diff
@@ -25,7 +25,7 @@
   .alert-icon {
     float: left;
     margin-right: $alert-icon-space;
-    width: 1.25rem;
+    width: $alert-icon-width;
     height: 1.25rem;
```
abutterworth, line 29:
```suggestion
    height: $alert-icon-width;
```

Icons are square so the width and height should match in these styles.

src/Alert/Alert.scss
```diff
@@ -37,6 +37,11 @@
   display: flex;
 }
 
+// adds starting margin to align text with heading
+p {
+  margin-inline-start: $alert-icon-width + $alert-icon-space
+}
```
abutterworth, line 43: This is a brittle way to achieve the desired result:

- If someone uses a `p` tag but not an icon then the text will be indented for no reason
- If someone uses a `ul` tag for content then it will not be indented the same as a paragraph.

Let's add our own shim for this react-bootstrap component and do two things:
- Add an `icon` prop to render the icon
- Put all children inside an `alert-content` div on which you can either add padding or use flexbox on the parent container to layout the icon + content.

See https://gist.github.com/abutterworth/8ead64d686523c804fceed61a182146c for an example.

abutterworth (changes requested):

src/Alert/WrapperAlert.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Alert from 'react-bootstrap/Alert';
+import Icon from '../icon';
```
abutterworth, line 4:
```suggestion
import Icon from '..';
```

abutterworth (approved):

PR #739, opened 2021-06-08
feat: add underline prop to hyperlink

zainab-amir:
Changes:

- Added a new prop to Hyperlink component that gives us an option to update the style of hyperlink. 
-- Some designs require no underline under hyperlink component. 
- Added aria-label for spinner icon
-- This is added to for accessibility because sometime stateful button has a pending state that doesn't have a text and only spinner is shown.

### DIFF
--- a/scss/core/_typography.scss
+++ b/scss/core/_typography.scss
@@ -72,3 +72,27 @@ a.muted-link {
     }
   }
 }
+
+a.brand-link {
+  color: $brand-link-color;
+  text-decoration: $brand-link-decoration;
+
+  &:hover {
+    color: $brand-link-hover-color;
+    text-decoration: $brand-link-hover-decoration;
+  }
+
+  p > &[href]:not(.btn),
+  &.inline-link {
+    color: $brand-inline-link-color;
+    text-decoration: $brand-inline-link-decoration;
+    text-decoration-line: $brand-inline-link-decoration;
+    text-decoration-color: $brand-inline-link-decoration-color;
+    &:hover {
+      color: $brand-inline-link-hover-color;
+      text-decoration: $brand-inline-link-hover-decoration;
+      text-decoration-line: $brand-inline-link-hover-decoration;
+      text-decoration-color: $brand-inline-link-hover-decoration-color;
+    }
+  }
+}
--- a/scss/core/_variables.scss
+++ b/scss/core/_variables.scss
@@ -418,6 +418,17 @@ $muted-inline-link-hover-color:                 darken($muted-inline-link-color,
 $muted-inline-link-hover-decoration:            underline !default;
 $muted-inline-link-hover-decoration-color:      rgba($muted-inline-link-hover-color, 1) !default;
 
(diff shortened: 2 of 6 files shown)


icons/jsx/SpinnerSimple.jsx
```diff
@@ -8,6 +8,7 @@ function SvgSpinnerSimple(props) {
       viewBox="0 0 24 24"
       fill="none"
       xmlns="http://www.w3.org/2000/svg"
+      aria-label="loading"
```
abutterworth, line 11: These files are generated by code. If you want to specify a label on the loading icon you need to do it in your application. Paragon doesn't have localization built in yet.

zainab-amir: Currently if I don't provide any icon to Stateful button, it uses a default spinner. Is there a way to add aria-label to that? 
Or will we need to explicitly add icon to stateful button and then add aria-label to it?

abutterworth:
Here is the Design Documentation for links: https://www.figma.com/file/eGmDp94FlqEr4iSqy1Uc1K/Paragon-2021?node-id=789%3A42

If you need the functionality you've added in this PR please take this path:
- Add a `variant` prop of values `default`, `muted`, or `brand`
- Add a boolean prop `isInline` (default: false or "standalone" in the design documentation) This default matches the way css handles links in Paragon today. `a` tags have no underline unless they exist as a direct descendent of a `p` tag (https://github.com/edx/paragon/blob/master/scss/core/_typography.scss#L38).

The props you add to hyperlink should add or remove css classes for styling links. Some of which we already have: https://github.com/edx/paragon/blob/master/scss/core/_typography.scss#L39 What's here probably needs some expanding or reworking. Here's a rough idea of what styles could become:

```
.link-default {}
.link-muted {}
.link-brand {}
.link-inline {}
.link-standalone {}
```
Since you're PR relates only to underlining you could choose to add just `link-inline` and `link-standalone` classes if you're time strapped.

Let me know if you have any questions or concerns.

icons/jsx/SpinnerSimple.jsx
abutterworth: We could hardcode English in Paragon for now, but if you do add your label here please: https://github.com/edx/paragon/blob/master/src/StatefulButton/index.jsx#L51

wittjeff (approved):

abutterworth (changes requested):
See my comment above

icons/jsx/SpinnerSimple.jsx
zainab-amir: @abutterworth I made the changes you suggested ^ but this didn't fix the accessibility problem and the icon is not focusable for stateful buttons. I have taken another route to achieving the same thing. Kindly see changes to StatefulButton component.

abutterworth (approved):
It looks good.

Could you add examples of the different variants and isInline in the documentation site? This can be done by editing this README file: https://github.com/edx/paragon/blob/master/src/Hyperlink/README.md

PR #743, opened 2021-06-09
refactor: hyperlink updated to underline link text conditionally

mehaknasir:
- hyperlink component updated to support new link designs it will support underlined text by passing prop isUnderline and will render simple text by default.
- test case added according to the new prop 
- hyperlink documentation updated

the motivation behind this change is https://www.figma.com/file/SeNpRff5HxzdpCgoD5pmlQ/Apps-and-resources-landing-page?node-id=4306%3A0

Screenshot 2021-06-09 at 2 56 18 PM

### DIFF
--- a/src/Hyperlink/README.md
+++ b/src/Hyperlink/README.md
@@ -53,3 +53,8 @@ notes: |
   />
 </Hyperlink>
 ```
+### underlined 
+
+```jsx live
+<Hyperlink href="https://en.wikipedia.org/wiki/Hyperlink" isUnderline>edX.org</Hyperlink>
+```
--- a/src/Hyperlink/index.jsx
+++ b/src/Hyperlink/index.jsx
@@ -14,6 +14,7 @@ function Hyperlink(props) {
     onClick,
     externalLinkAlternativeText,
     externalLinkTitle,
+    isUnderline,
     ...other
   } = props;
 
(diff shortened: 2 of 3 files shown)


src/Hyperlink/index.jsx
```diff
@@ -48,6 +51,7 @@ Hyperlink.defaultProps = {
   onClick: () => {},
   externalLinkAlternativeText: 'Opens in a new window',
   externalLinkTitle: 'Opens in a new window',
+  isUnderline: false,
```
awaisdar001, line 54: Correct me if I am wrong, currently, all hyperlink compoents are underlined out of the box, so adding default prop to `false` would remove the underline and can change their behavior. What are your thoughts on this?

mehaknasir: I am not sure if the current ones are underlined I am short of context on this. I think someone from the paragon team can make this decision about what should be the default behavior of this.

src/Hyperlink/README.md
```diff
@@ -53,3 +53,8 @@ notes: |
   />
 </Hyperlink>
 ```
+### underlined
```
saadyousafarbi, line 56: `underlined` -> `with underline`

src/Hyperlink/index.jsx
abutterworth: Please see my comment on this PR: https://github.com/edx/paragon/pull/739#issuecomment-856752410

We should implement links as they are reflected in the design system docs here: https://www.figma.com/file/eGmDp94FlqEr4iSqy1Uc1K/Paragon-2021?node-id=789%3A42 

We have the concept of "inline" vs "standalone" links. In our system inline links are underlined and standalone are not. This css https://github.com/edx/paragon/blob/master/scss/core/_typography.scss#L38 implements this behavior by default as best it can.

mehaknasir:
another PR in progress for the same changes hence closing this.

PR #747, opened 2021-06-11
feat: add disclaimer text class

zainab-amir:
#### Changes
- Added a new class for disclaimer text
Screenshot 2021-06-11 at 5 44 33 PM

- Fixed the stateful button spinner alignment in cases when no text is provided to the button:

|Before|After|
------|-------
Screenshot 2021-06-10 at 12 52 34 PM|Screenshot 2021-06-11 at 5 45 15 PM

#### Tickets:

https://openedx.atlassian.net/browse/VAN-546
https://openedx.atlassian.net/browse/VAN-593

### DIFF
--- a/scss/core/_utilities.scss
+++ b/scss/core/_utilities.scss
@@ -18,6 +18,12 @@ $color-levels: 100, 200, 300, 400, 500, 600, 700, 800, 900;
   font-size: $x-small-font-size;
 }
 
+.micro {
+  font-size: $micro-font-size;
+  font-weight: normal;
+  line-height: $micro-line-height;
+}
+
 .mw-xs {
   max-width: $max-width-xs !important;
 }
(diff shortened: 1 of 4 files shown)


scss/core/_utilities.scss
```diff
@@ -18,6 +18,13 @@ $color-levels: 100, 200, 300, 400, 500, 600, 700, 800, 900;
   font-size: $x-small-font-size;
 }
 
+.disclaimer {
+  font-size: $disclaimer-font-size;
+  font-weight: normal;
+  line-height: $disclaimer-line-height;
+  color: $gray-500;
+}
```
abutterworth, line 26: I've spoken to the design team about the naming of this. Let's not use the semantic name "disclaimer" here and instead simply add an `xx-small` size.
```
.xx-small {
  font-size: $xxs-font-size;
  line-height: $xxs-line-height;
}
```

And let's keep the color out of Paragon in the app for now:
```
<p className="xx-small text-muted">Legal disclaimer...</p>

abutterworth (changes requested):
Please update the name from 'disclamer' to 'micro' and remove colors from the stylesheet

abutterworth (approved):
🚀

PR #752, opened 2021-06-16
refactor: refactored hyperlink component to hide launch Icon

mehaknasir:
Initially, Launch Icon was shown as default with every link which is external. 
Now we have some designs in which external links need to be shown without external Icon.
Here are [Figma Designs ](https://www.figma.com/file/SeNpRff5HxzdpCgoD5pmlQ/Apps-and-resources-landing-page?node-id=4334%3A442) for reference 

Now we can hide icons based on prop **hideLaunchIcon** which is not required and have false as default value which means launcher icon will be visible by default but can also be hidden based on the mentioned prop.


src/Hyperlink/index.jsx
```diff
@@ -59,6 +60,7 @@ Hyperlink.defaultProps = {
   externalLinkTitle: 'Opens in a new window',
   variant: 'default',
   isInline: false,
+  hideLaunchIcon: false,
```
asadazam93, line 63: If we can make this "icon" with default true. That would make the code a lot cleaner

src/Hyperlink/index.jsx
```diff
@@ -17,12 +17,13 @@ function Hyperlink(props) {
...
+    hideLaunchIcon,
     ...other
   } = props;
 
   let externalLinkIcon;
 
-  if (target === '_blank') {
+  if (target === '_blank' && !hideLaunchIcon) {
     // Add this rel attribute to prevent Reverse Tabnabbing
     other.rel = other.rel ? `noopener ${other.rel}` : 'noopener';
```
awaisdar001, line 28: these are important attribs that are needed don't skip them.

src/Hyperlink/index.jsx
awaisdar001, line 63: `hideExternalIcon`

src/Hyperlink/index.jsx
mehaknasir: fixed, moved condition inside if. for icon only.

src/Hyperlink/README.md
```diff
@@ -27,6 +27,14 @@ notes: |
 </Hyperlink>
 ```
 
+### blank target without Icon
+
+```jsx live
+<Hyperlink destination="https://www.edx.org" target="_blank" rel="noopener noreferrer" hideLaunchIcon>
```
asadazam93, line 33: showLaunchIcon

asadazam93 (approved):

PR #756, opened 2021-06-21
feat: support controlled table selections for server side pagination

adamstankiewicz:
### Description

This PR adds first-class support for handling row selections (individual rows, or the entire table) when using server-side pagination, where only 1 page of data is known at any given time. Previously, there was no way to distinguish between individually selected rows and whether the entire table is selected (all rows across all pages) when using server-side pagination.

This is done by keeping track of row selections in the `DataTableContext` and updating the `TableActions` and `BulkActions` click handler arguments to depict when the entire table is selected or just individual rows are selected. This is important as the API endpoints used for an action that applies to the entire table may be different from the API endpoint used for an action that applies to individually selected rows (e.g., where all the individual email addresses are already known).

This change is opt-in for consumers in the sense that the `SelectionStatusComponent` and `manualSelectColumn` must be configured using the following components:
* `DataTable.ControlledSelectionStatus`
* `DataTable.ControlledSelectWithContextHeader`
* `DataTable.ControlledSelectWithContext`

When `DataTable` is configured to use these components, the user's row selections across pages will persist even during pagination as well as appropriately marking the entire row as "selected" (i.e., light blue background on the row).

I also updated the `DataTable` docs to clarify some things about using the component with server-side pagination with an example that more closely mimics how server-side pagination should work. See the "Backend filtering and sorting" section here: https://deploy-preview-756--paragon-edx.netlify.app/components/datatable/

**Note: This is a breaking change as the bulk action and table action click handler now pass new function arguments to be able to differentiate between individually selected rows and when an entire table is selected. Basic usage of `DataTable` remains unchanged.**

Tickets: 
* [PAR-317](https://openedx.atlassian.net/browse/PAR-317)
* [PAR-319](https://openedx.atlassian.net/browse/PAR-319)


iloveagent57 (approved):

raq929 (approved):
This looks really good!
The one change I think is important is finding a way for the display components that are reused in both places to be separated from the logic, so that they can be changed without a dev having to know that there are two places to change them.
I'm so excited to have this.
Also, maybe I missed it, but there should be something in the docs telling people how to use it.

PR #757, opened 2021-06-21
fix: fixed modal box shadow above page content

mehaknasir:
Ticket: https://openedx.atlassian.net/browse/TNL-8405

Issue: The issue was in the modal component in which the box-shadow was appearing under the child component.  The issue is fixed now by removing box-shadow and adding before and after pseudo-element in the modal body class with the linear gradient.
Screenshot 2021-06-22 at 8 48 32 PM

Screenshot 2021-06-22 at 8 47 57 PM

### DIFF
--- a/src/Modal/_ModalDialog.scss
+++ b/src/Modal/_ModalDialog.scss
@@ -129,24 +129,60 @@
   flex-grow: 1;
   padding: $modal-inner-padding $modal-inner-padding $modal-inner-padding / 2;
   overflow: auto;
-  box-shadow: inset 0 10px 10px -10px rgba(black, .5), inset 0 -10px 10px -10px rgba(black, .5);
-  transition: box-shadow 150ms ease;
+  position: relative;
+
+  &::before{
+    content:"";
+    background-color: transparent;
+    background-image: linear-gradient(#605c5c, #b8bebe, transparent 50%);
+    display: block;
+    height: 20px;
+    position: sticky;
+    top: -$modal-inner-padding;
+    margin-top: -$modal-inner-padding;
+    margin-left: -$modal-inner-padding;
+    margin-right: -$modal-inner-padding;
+    opacity: .5;
+  }
+  &::after{
+    content:"";
+    background-color: transparent;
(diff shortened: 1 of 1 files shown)


src/Modal/ModalDialogBody.jsx
```diff
@@ -24,7 +24,7 @@ const ModalDialogBody = ({
     (
       <>
         <div ref={topSentinelRef} />
-        <div className="pgn__modal-body-content">
+        <div className="pgn__modal-body-content" style={{ 'z-index': '-1', position: 'relative' }}>
```
mehaknasir, line 27: @abutterworth is it okay to pass inline css or should I create a class for the with these properties and pass that here?

abutterworth (changes requested):
Looks good! But, let's add these styles to the SCSS instead of inline `.pgn__modal-body-content {...}` Maybe here: https://github.com/edx/paragon/blob/master/src/Modal/_ModalDialog.scss#L146

src/Modal/ModalDialogBody.jsx
abutterworth: Oh yes, sorry see my note below

abutterworth:
I just discovered an issue in testing: It looks like by moving the z-index lower we have made the modal body content non-interactive. I'm not sure if this strategy will work for us...

awaisdar001 (changes requested):
I don't know how these changes are going to affect the UI, can you add before/after pictures?  also link the paragon docs?

abutterworth:
@awaisdar001 fyi you can see the deploy preview here: 
image

abutterworth (approved):
LGTM! Nice work.

mehaknasir:
> I don't know how these changes are going to affect the UI, can you add before/after pictures? also link the paragon docs?

@awaisdar001 

https://deploy-preview-757--paragon-edx.netlify.app/components/stepper/

we have this link in PR deploy from where we can see details I have attached it for you. I am updating description as well.

awaisdar001 (approved):

PR #758, opened 2021-06-22
fix: added min height to stepper header

AhtishamShahid:
### https://openedx.atlassian.net/browse/TNL-8406

Added min-height to stepper header so it does not change its height in case of error. and updated the example with an error message.

### Before  : 
2 - stepper resize

### After:
Webp net-gifmaker

### DIFF
--- a/src/Stepper/Stepper.scss
+++ b/src/Stepper/Stepper.scss
@@ -13,6 +13,7 @@
   justify-content: center;
   background: $white;
   padding: .75rem 1rem;
+  min-height: 5.13rem;
 }
 
 .pgn__stepper-header-step {


src/Stepper/Stepper.scss
```diff
@@ -13,6 +13,7 @@
   justify-content: center;
   background: $white;
   padding: .75rem 1rem;
+  min-height: 4rem;
```
asadazam93, line 16: @AhtishamShahid Does this new height match the Figma design?

src/Stepper/README.md
```diff
@@ -138,7 +138,7 @@ A composition of a stepper with a `FullscreenModal`.
               <HipsterIpsum numParagraphs={2} />
             </Stepper.Step>
 
-            <Stepper.Step eventKey="choose-cats" title="Chooseth a cat">
+            <Stepper.Step eventKey="choose-cats" title="Chooseth a cat" hasError="true" description="This is error text">
```
asadazam93, line 141: Instead of changing this existing one maybe add a new one for the error state?

src/Stepper/Stepper.scss
AhtishamShahid: There is a difference of few decimal points in pixels so I guess it is negligible .

src/Stepper/README.md
AhtishamShahid: I have reverted this change because it was out of the scope of this Fix

awaisdar001 (changes requested):
Can you share the updated `gif` as well?

src/Stepper/README.md
awaisdar001: @AhtishamShahid is there any existing item in the readme that illustrates the validation error? If not, it's ok to update the docs I guess.

AhtishamShahid:
> Can you share the updated `gif` as well?

Added change GIF in description

awaisdar001 (approved):
Visually looks good to me.

asadazam93 (approved):

PR #761, opened 2021-06-24
fix: form field default icon color

mubbsharanwar:
1. set icon button color gray-500 while render in form.

**Results:**

Before
before

After
after

Related ticket:
https://openedx.atlassian.net/browse/VAN-523

2. set alert child content div width 100 %.

**Results:**

Before fix in paragon 
before
Expected result
after

Related ticket:
https://openedx.atlassian.net/browse/VAN-601

### DIFF
--- a/src/Alert/Alert.scss
+++ b/src/Alert/Alert.scss
@@ -41,6 +41,10 @@
 // added to align content and icon
 .alert-content {
   display: flex;
+
+  div {
+    flex-grow: 1;
+  }
 }
 
 // Provide class for links that match alerts
--- a/src/Form/_Form.scss
+++ b/src/Form/_Form.scss
@@ -115,6 +115,19 @@ $form-control-icon-width: 32px !default;
       padding-right: calc(#{$input-padding-y-sm} - #{2 * $input-border-width});
     }
   }
+
+  .btn-icon {
+    color: $gray-500 !important;
+    &:hover {
+      background-color: $gray-500 !important;
+      color: $white !important;
+    }
+    &:focus {
+      background-color: $white !important;
+      color: $gray-500 !important;
+      box-shadow: inset 0 0 0 2px $gray-500 !important;
+    }
+  }
 }
 
 // Floating Labels


waheedahmed:
Hello @abutterworth, could you please take a look if these changes make sense?

src/Alert/Alert.scss
```diff
@@ -41,6 +41,10 @@
 // added to align content and icon
 .alert-content {
   display: flex;
+
+  div {
+    width: 100%;
```
abutterworth, line 46:
```suggestion
    flex-grow: 1;
```

src/Form/_Form.scss
```diff
@@ -115,6 +115,20 @@ $form-control-icon-width: 32px !default;
...
+    &:hover {
+      background-color: $gray-500 !important;
+      color: $white !important;
+    }
+    &:focus {
+      background-color: $white !important;
+      color: $gray-500 !important;
+      -webkit-box-shadow: inset 0 0 0 2px $gray-500 !important;
+      box-shadow: inset 0 0 0 2px $gray-500 !important;
+    }
```
abutterworth, line 130: Why are there `!important` rules here? We should avoid using !important anywhere in Paragon.

src/Form/_Form.scss
```diff
@@ -115,6 +115,20 @@ $form-control-icon-width: 32px !default;
...
+  .btn-icon {
+    color: $gray-500 !important;
+    &:hover {
+      background-color: $gray-500 !important;
+      color: $white !important;
+    }
+    &:focus {
+      background-color: $white !important;
+      color: $gray-500 !important;
+      -webkit-box-shadow: inset 0 0 0 2px $gray-500 !important;
```
abutterworth, line 128:
```suggestion
```

We don't need the prefixed version in Paragon. Webpack SASS compilers will do the job of adding it if needed.

abutterworth (changes requested):
A couple comments. Other than the suggestions and the !important flags it looks good. Let me know if removing !important is problematic.

mubbsharanwar:
> A couple comments. Other than the suggestions and the !important flags it looks good. Let me know if removing !important is problematic.


@abutterworth Thanks for suggestions. I update PR according to your comments. But removing `!important` is problematic because `IconButton` component class override color and background color. That's why we are using `!important`. If there is any other solution better than existing let us know, we will look into that.

abutterworth (approved):
I see. This seems like a larger issue we may need to iron out down the road. Looks good for now. Thank you.

PR #765, opened 2021-06-28
feat: add breakpoints to width utility class

ciduarte:
Updated `w-` utility class to incorporate breakpoints. For example `w-sm-100`, `w-md-auto`.

**Format**
`w-{breakpoint}-{size}`

**Breakpoints**
`xs`: @media(min-width: 0)
`sm`: @media(min-width: 576px)
`md`: @media(min-width: 768px)
`lg`: @media(min-width: 992px)
`xl`: @media(min-width: 1200px)

**Sizes**
`25`: 25%
`50`: 50%
`75`: 75%
`100`: 100%
`auto`: auto


raq929 (approved):
Looks good to me! is there somewhere we can document this so that people know it exists? 
That would be nice to add if you can.

PR #774, opened 2021-07-01
fix: remove opacity on stateful button

uzairr:
Remove the opacity on pending state of the stateful button so that a crisp original color would be displayed in pending state.

VAN-605

**Before the fix:**
Screen Shot 2021-07-07 at 4 36 23 AM

**After the fix:**
Screen Shot 2021-07-07 at 4 38 03 AM

### DIFF
--- a/src/StatefulButton/StatefulButton.scss
+++ b/src/StatefulButton/StatefulButton.scss
@@ -2,3 +2,7 @@
   margin-right: 0.5em;
   margin-left: -0.25em;
 }
+
+.pgn__stateful-btn-state-pending {
+  opacity: 1 !important;
+}


awaisdar001:
@uzairr can you add before/after screenshots?

raq929:
We still need there to be an opacity change when a button is actually disabled instead of pending.
Is there another option for your pending state?
Off the top of my head, you could put a spinner icon on the button, or have the button not show up until it's available to be used.
I think it will be a breaking change for other button users if this is merged; I certainly depend on that opacity change when the button is disabled.

abutterworth (changes requested):
Please address @raq929 's point. We should not remove the style as in this PR. We could instead add opacity: 1 to a new selector put in https://github.com/edx/paragon/blob/master/src/StatefulButton/StatefulButton.scss 

See where class names are generated in Stateful Button:
https://github.com/edx/paragon/blob/master/src/StatefulButton/index.jsx#L78

uzairr:
I updated the PR as per the suggestions.Thanks everyone,
@raq929 please review the changes again.

abutterworth (approved):

PR #777, opened 2021-07-07
fix: remove last margin of elements in Alert

abutterworth:
Extraneous margin: 
image

After: 
image


adamstankiewicz (approved):
Looks great! I recall needing to override the bottom margin of `p` elements to get this same behavior before; glad this is baked into the `Alert` itself now :)

PR #779, opened 2021-07-08
feat: New Scrollable component PAR-472

binodpant:
PAR-472

The component now offers scrolling, as well as automatic shadow adornment on top and/or bottom based on scroll direction

Height if specified by wrapped component is honored, but scrolling is offered regardless with a predefined height. The wrapped element is wrapped within a `flex-grow:1` and `overflow:scroll` styling

The example to follow is in the modalDialog: https://edx.github.io/paragon/components/modal/modal-dialog/

https://github.com/edx/paragon/blob/master/src/Modal/ModalDialogBody.jsx#L11

The preview will help see this in the demo page at `/components/scrollable/`

Example usage in admin portal UI

https://user-images.githubusercontent.com/528166/129050832-836f2181-28a1-40b3-8ce1-ce39dc1d0896.mov

-- prior notes --

[2] The bottom detection is not working as intended

Adam suggested for this issue, to use absolute position instead, as he laid out in this example to make it simpler

```
 the scroll bar is on a child element rather than the parent

the way it is now, the parent element gets the scrollbar, which is making it very tricky to get the shadows to be on top of stuff and in the right place

flex box is used to ensure the child container (which is the one that actually gets the scrollbar) has a height of 100% of the parent
and no taller
this allows us to set a max height on the parent element and overflow-y: auto on the child element
and get the scroll bar there
this is a somewhat obscure behavior of flex box – but it’s intentional from the spec and something you will run into if you ever try to make a carousel with elements laid out using flexbox
```

https://codepen.io/abutterworth/pen/jOmrGjN

Once that is done, this change can be tested against the stepper by wrapping the contents of the Stepper other than the Stepper.Header and Stepper.Footer and ensuring it can offer scrollability (as long as the contained element provides its own max-height), and also offers drop shadows only when scrolling is detected toward top or bottom edges. 
intersection observer is used already to detect when the wrapped contents scroll toward the top or bottom edges


src/Scrollable/Scrollable.scss
```diff
@@ -2,7 +2,7 @@
 
 .pgn__scrollable-body {
     flex-grow: 1;
-    padding: $scrollable-inner-padding $scrollable-inner-padding $scrollable-inner-padding/2 $scrollable-inner-padding;
+    padding: $scrollable-inner-padding $scrollable-inner-padding/2 $scrollable-inner-padding/2 $scrollable-inner-padding/2;
```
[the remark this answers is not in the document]
abutterworth: I don't think padding should be part of this component. It should be the responsibility of a child element or explicitly added to the Scrollable with a css class name

abutterworth (changes requested):
Looks like some style here is causing a horizontal scroll bar to appear if there is a vertical scrollbar (and macos settings are set to show scrollbars) 
image

src/Scrollable/Scrollable.test.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+jest.mock('../hooks/useIsVisible');
+
+function rangeOfNumbers() {
+  return Array.from({ length: 50 }, (v, k) => k + 1);
+}
+
+function renderScrollable(items) {
+  render(
+    <div style={{ display: 'flex', height: '5vh' }}>
```
abutterworth, line 19: If this display: flex is required I think it should be part of `<Scrollable />`.

binodpant:
> Looks like some style here is causing a horizontal scroll bar to appear if there is a vertical scrollbar (and macos settings are set to show scrollbars)
> image

interesting I cannot get it to show on the showcase page even if I zoom in/out etc. What was the mac setting btw? I can check that

binodpant:
> > Looks like some style here is causing a horizontal scroll bar to appear if there is a vertical scrollbar (and macos settings are set to show scrollbars)
> > image
> 
> interesting I cannot get it to show on the showcase page even if I zoom in/out etc. What was the mac setting btw? I can check that

Ah I see it when I go to mobile view ... can check into it

src/Scrollable/Scrollable.test.jsx
binodpant: yeah so if I remove it and leave my padding in the scrollable body it still sort of works.. I guess I don't need it here

so making

```
.pgn__scrollable-body {
    overflow: auto;
    position: relative;
```

then removing the flex here also works, but it's edge detection now is a bit less smooth :(

src/Scrollable/Scrollable.scss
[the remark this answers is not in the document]
binodpant: yea I need to figure this out.. when I removed this padding line it no longer shows shadows

binodpant:
Alright with latest changes I fixed the x scroll, and now don't need the display:flex in the contained contents.. 

I managed to also apply the shadow to look more like the level 1 as per Jen's feedback from here

https://www.figma.com/file/eGmDp94FlqEr4iSqy1Uc1K/Paragon-Design-System?node-id=9076%3A3783

but now I lost the bottom shadow again! Something about my setup messed up the bottom shadow so have to re-figure that out after these

binodpant:
Screen Shot 2021-08-13 at 2 07 25 PM

src/Scrollable/Scrollable.test.jsx
binodpant: removed padding from scrollable stlying and display flex too

binodpant:
I am having trouble getting this to work reliably in the showase page. But it works a lot better in the admin portal inside the stepper

probably because the stepper has this styling on it

```
.sticky-footer-wrapper {
  height: 70vh;
  display: flex;
  flex-direction: column;

  .stepper-step {
    flex-grow: 1;
    overflow: scroll;
  }

  .add-courses-search-data {
    margin-left: 6px;
  }
}
```

binodpant:
updated to use flex-grow : 1 in the body. Now I dn't need to have any requirements on the content area itself, but I can only get this to work with a height and display:flex in the surrounding container. otherwise I don't get the bottom sentinel div placed correctly

this also allows me to remove most of the style in admin portal `stepper-body` and now it works quite well since this component offers the overflow-y and 

so basically I could remove this and it still now works

```
  .stepper-step {
    flex-grow: 1;
    overflow: scroll;
  }
```

Screen Shot 2021-08-16 at 4 44 49 PM
Screen Shot 2021-08-16 at 4 44 42 PM

binodpant:
here is the usage PR: https://github.com/edx/frontend-app-admin-portal/pull/603

src/Scrollable/README.md
```diff
@@ -0,0 +1,69 @@
...
+<span>
+Shadow is shown on top if you scroll upwards
+Shadow is shown on bottom if you scroll downwards
+
+If you don't specify height you will get a default height per the Scrollable component (subject to change)
+</span>
+
+#### Demo with flex container around Scrollable
+```jsx live
+<div style={{'height': '80vh','display': 'flex','flex-direction': 'column'}}>
```
abutterworth, line 28: This example needs to change right?

src/Scrollable/Scrollable.scss
```diff
@@ -0,0 +1,71 @@
+@import "variables";
+
+.pgn__scrollable-body {
+    overflow-x: hidden;
```
abutterworth, line 4: Is this needed?

src/Scrollable/_variables.scss
```diff
@@ -0,0 +1,33 @@
+
+// Scrollable
+
+// Padding applied to the scrollable body
+$scrollable-top-margin:                  0.46rem !default;
+$scrollable-bottom-margin:               5.4rem !default;
+$scrollable-inner-padding:               1.58rem !default;
+$scrollable-shadow-padding:              0.96rem !default;
```
abutterworth, line 8: I don't think this should be built into the component. Ideally scrollable body has no padding and if padding is desired the developer would write something like:
```
<Scrollable>
  <div className="p-4">
   ...
  </div>
</Scrollable>
```

I realize this may mess with the way shadows are positioned...

src/Scrollable/_variables.scss
```diff
@@ -0,0 +1,33 @@
...
+
+$scrollable-xl:                          1140px !default;
+$scrollable-lg:                          800px !default;
+$scrollable-md:                          500px !default;
+$scrollable-sm:                          400px !default;
+
+$scrollable-fade-transform:              translate(0, -50px) !default;
+$scrollable-show-transform:              none !default;
+$scrollable-transition:                  transform .3s ease-out !default;
+$scrollable-scale-transform:             scale(1.02) !default;
```
abutterworth, line 33: It looks like most of these are unused and should be removed. I don't think the transitions of the shadow fade ins and outs need to be configurable.

src/Scrollable/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+    props.className,
+    {
+      [CLASSNAME_SCROLL_TOP]: isScrolledToTop,
+      [CLASSNAME_SCROLL_BOTTOM]: isScrolledToBottom,
+    },
+  );
+  return React.createElement(
+    'div',
+    { ...props, className },
+    (
```
abutterworth, line 24: If you don't intend to offer an `as` prop to allow control of the element type. You might consider going back to jsx syntax: 

```
<div {...props} className={className}>
 ...
```

src/Scrollable/Scrollable.scss
binodpant: the overflow-x : hidden? I thought it was what fixed that horizontal scroolbar but I can double check it still needed

abutterworth: The horizontal scrollbar should be fixed in a different way. It's possible some combo of width setting + padding makes the child element wider than the parent.

src/Scrollable/index.jsx
binodpant: yeah I like this thanks

src/Scrollable/_variables.scss
binodpant: yeah ok thanks will review and clean these up

src/Scrollable/_variables.scss
binodpant: hmm ok going to try this too, my shadows were going berserk when I changed paddings ...

src/Scrollable/Scrollable.scss
binodpant: Ah I see, so yeah given more horiz spacing perhaps I can avoid the scrollbar that way, will try

src/Scrollable/README.md
binodpant: yup, did!

src/Scrollable/_variables.scss
binodpant: yeah so I am trying to remove all the margin and padding settings, does not look like I need those anyway for the shadows to work

binodpant:
I actually removed the variables file since I don't need any of those settings. I did need to add a margin-top to the bottom shadow once I remove all other settings but decided I can just hardcode that one... lmk if flags

src/Scrollable/_variables.scss
binodpant: also added an example to showcase page showing consumer can add their own padding, it still works the same :)

src/Scrollable/Scrollable.scss
binodpant: Interesting, I now removed all margins for the shadow, and now I don't se the horiz scrollbar either, even withou overflow-x:hidden so I think we are good?

src/Scrollable/Scrollable.scss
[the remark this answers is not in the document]
binodpant: removed padding, and also added test case showing how a consumer can adding their own padding

abutterworth (approved):
Rock on 🚀

PR #787, opened 2021-07-17
feat: update Alert component to support dismiss and alert action buttons in Figma spec

adamstankiewicz:
Ticket: https://openedx.atlassian.net/browse/PAR-332

**Alert Changes**
* Replaces react-bootstrap's dismissible "X" with a "Dismiss" button
* Adds support for custom call-to-action (CTA) button
* Adds `stacked` prop to allow control over positioning of "Dismiss" and CTA buttons.
* Switches to `stacked` variant at the extra small breakpoint 
* Utilizes `ActionRow` component

**Screenshot Example**
image

**Previews**
* [Open edX theme](https://deploy-preview-787--paragon-openedx.netlify.app/components/Alert/#with-buttons)
* [edX.org theme](https://deploy-preview-787--paragon-edx.netlify.app/components/Alert/#with-buttons)


src/Alert/AlertButtons.jsx
```diff
@@ -0,0 +1,73 @@
...
+          ...restButtonAttrs,
+        });
+      }
+      return [buttonText, buttonAttrs];
+    },
+    [JSON.stringify(button)],
+  );
+
+  return (
+    <div className={classNames('pgn__alert-buttons d-flex justify-content-end', className)}>
```
abutterworth, line 38: Let's use `ActionRow` here instead. Also, we should avoid using css utility classes in Paragon components since they have `!important` flags on them. See: https://github.com/edx/paragon/blob/master/docs/decisions/0012-css-styling-conventions#L21

src/Alert/index.jsx
```diff
@@ -1,37 +1,106 @@
...
+        <div className="alert-message-content">
+          {children}
+        </div>
+        {(dismissible || button) && (
+          <AlertButtons
+            className={classNames({
+              'mt-3': isStacked,
+              'ml-4': !isStacked,
+            })}
+            button={button}
```
abutterworth, line 65: It's possible that we may want to offer two actions on an alert (not including dismiss). What do you think about changing `button` to `actions` as an array of `node`s?

src/Alert/README.md
```diff
@@ -44,7 +44,62 @@ notes: |
 </>
 ```
 
-##### Variants
+### With Buttons
```
abutterworth, line 47: It would be good to update the frontmatter to include 
```
components:
- Alert
```

So that a props table is generated. See https://raw.githubusercontent.com/edx/paragon/master/src/ActionRow/README.md for an example

src/Alert/index.jsx
```diff
@@ -1,37 +1,106 @@
...
-    ref={ref}
-  >
-    {icon && <Icon src={icon} className="alert-icon" />}
-    <div className="alert-message-content">
-      {children}
-    </div>
-  </Alert>
-));
+}, ref) => {
+  const [isOpen, /* open */, close] = useToggle(show);
```
abutterworth, line 23: Is this component controlled, uncontrolled or either?

src/Alert/AlertButtons.jsx
adamstankiewicz: Ah, yes. Didn't think to use `ActionRow` here :) And +1 on avoiding the utility classes; will update.

src/Alert/index.jsx
adamstankiewicz: Having an `actions` type of prop crossed my mind but didn’t implement it that way since the Figma spec didn't explicitly showcase that as a use case. But I agree it may be worth supporting.

The caveat is what happens when there’s 2+ actions and a dismiss button? There's not much real estate for more than 2 buttons here with the right-aligned button variant. We could use an overflow for the buttons like we have started to do elsewhere but I think my suggestion would be to defer on that until there's an explicit use case for it.

I'm for the `actions` prop assuming we have a custom prop type that throws a prop type warning/error when:
* `actions` has length greater than 2
* `actions` has length greater than 1 and `dismissible` is true.

src/Alert/index.jsx
adamstankiewicz: Good catch. I've sort of made this both semi-controlled/uncontrolled simultaneously. I will update to make it controlled where the consumer must update the `show` prop when `onClose` is called on dismiss, like the react-bootstrap `Alert` does today.

package-lock.json
```diff
@@ -4220,9 +4220,9 @@
       }
     },
     "caniuse-lite": {
-      "version": "1.0.30001117",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001117.tgz",
-      "integrity": "sha512-4tY0Fatzdx59kYjQs+bNxUwZB03ZEBgVmJ1UkFPz/Q8OLiUUbjct2EdpnXj0fvFTPej2EkbPIG0w8BWsjAyk1Q==",
+      "version": "1.0.30001245",
+      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001245.tgz",
+      "integrity": "sha512-768fM9j1PKXpOCKws6eTo3RHmvTUsG9UrpT4WoREFeZgJBTi4/X9g565azS/rVUGtqb8nt7FjLeF5u4kukERnA==",
```
abutterworth, line 4225: how/why did this change?

src/Alert/index.jsx
```diff
@@ -1,40 +1,108 @@
...
+      {...props}
+      className={classNames('alert-content', props.className)}
+      ref={ref}
+    >
+      {icon && <Icon src={icon} className="alert-icon" />}
+      <ActionRow
+        isStacked={isStacked}
+        isStackedReversed={false}
+        isStackedCentered={false}
+      >
```
abutterworth, line 52: Looks like you're using action row in a different way than I expected.

I thought we should use ActionRow as a replacement for the pgn_alert-actions div.

package-lock.json
adamstankiewicz: `browserslist` can get out of date with what the latest browser versions are. Anytime I see output during build about having an out of date browsers list, I run the command it says to:

```npx browserslist@latest --update-db```

It updates the package-lock.json file to point to the latest `caniuse-lite` version as you see here.

File: src/ActionRow/index.jsx [hunk 7]
```diff
@@ -1,36 +1,57 @@
...
+  );
+};
 
 ActionRow.propTypes = {
   as: PropTypes.elementType,
   className: PropTypes.string,
   children: PropTypes.node,
   isStacked: PropTypes.bool,
+  isStackedReversed: PropTypes.bool,
+  isStackedCentered: PropTypes.bool,
```
abutterworth, line 45: Is is better to use separate values: 

```
isStacked
isReversed
justifyContent
```
?

abutterworth (approved):

src/ActionRow/index.jsx
adamstankiewicz: No longer relevant 😄

PR #789, opened 2021-07-19
feat: making "Filtered by __" text optional

kiram15:
Adding a boolean to FilterStatus in order to give the option of hiding the "Filtered by ___" text. Specifically created as a request for frontend-admin-portal [here](https://openedx.atlassian.net/browse/ENT-4677). 

[Associated paragon ticket.](https://openedx.atlassian.net/browse/PAR-555)

### DIFF
--- a/src/DataTable/FilterStatus.jsx
+++ b/src/DataTable/FilterStatus.jsx
@@ -12,7 +12,7 @@ const FilterStatus = ({
   }
 
   const filterNames = state.filters ? state.filters.map((filter) => filter.id) : [];
-  const filterTexts = (<p>Filtered by {filterNames.join(', ')}</p>);
+  const filterTexts = <p>Filtered by {filterNames.join(', ')}</p>;
 
   return (
     <div className={className}>
--- a/src/DataTable/FilterStatus.jsx
+++ b/src/DataTable/FilterStatus.jsx
@@ -4,7 +4,7 @@ import { Button } from '..';
 import DataTableContext from './DataTableContext';
 
 const FilterStatus = ({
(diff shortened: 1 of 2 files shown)


src/DataTable/FilterStatus.jsx
```diff
@@ -4,18 +4,19 @@ import { Button } from '..';
...
-  className, variant, size, clearFiltersText, buttonClassName,
+  className, variant, size, clearFiltersText, buttonClassName, showFilteredFields,
 }) => {
   const { state, setAllFilters } = useContext(DataTableContext);
   if (!setAllFilters) {
     return null;
   }
 
   const filterNames = state.filters ? state.filters.map((filter) => filter.id) : [];
+  const filterTexts = (<p>Filtered by {filterNames.join(', ')}</p>);
```
binodpant, line 15: not sure we need the parens

src/DataTable/tests/FilterStatus.test.jsx
```diff
@@ -15,6 +15,17 @@ const filterProps = {
   onClick: () => {},
   clearFiltersText: 'CLEAR ME',
   className: 'filterClass',
+  showFilteredFields: true,
+};
+
+const filterPropsNoFiltered = {
```
binodpant, line 21: I think you can just copy out the prior filterProps using spread operator and only override the one value like this 

```
const filterPropsNoFiltered = {
    ...filterProps, 
    showFilteredFields: true,
}
```

binodpant (approved):
made a couple minor suggestions but change itself looks good to me

PR #794, opened 2021-07-28
fix: remove modal dialog body before z-index for non scrollable modal

awais-ansari:
[TNL-8563](https://openedx.atlassian.net/browse/TNL-8563)

Set `z-index: 0` in modal dialog body ::before selector for non scrollable modals.  Modal dialog body top shadow `z-index` and `negative margin` were overlaying the content/button of non-scrollable. Which were affecting content onClick action. Modal shadow will work as normal for scrollable body content. 

Screenshot 2021-07-28 at 5 05 46 PM

### DIFF
--- a/src/Modal/_ModalDialog.scss
+++ b/src/Modal/_ModalDialog.scss
@@ -131,8 +131,8 @@
   overflow: auto;
   position: relative;
 
-  &::before{
-    content:"";
+  &::before {
+    content: "";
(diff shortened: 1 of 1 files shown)


mehaknasir (approved):
looks good to me.

asadazam93 (approved):
looks good

awaisdar001 (approved):

PR #803, opened 2021-08-06
feat: create page banner and add icon button black variant

ciduarte:
[Figma design spec](https://www.figma.com/file/eGmDp94FlqEr4iSqy1Uc1K/Paragon-Design-System?node-id=14061%3A20)

### DIFF
--- a/src/IconButton/README.md
+++ b/src/IconButton/README.md
@@ -15,11 +15,11 @@ notes: ''
 
 ```jsx live
 () => {
-  const variants = ["brand", "primary", "secondary", "success", "warning", "danger", "light", "dark"];
+  const variants = ["brand", "primary", "secondary", "success", "warning", "danger", "light", "dark", "black"];
   return (
     <div className="d-flex">
       {variants.map((variant) => (
-        <IconButton src={Close} iconAs={Icon} alt='Close' onClick={() => {}} variant={variant} className="mr-2" />
+        <IconButton src={Close} iconAs={Icon} alt="Close" onClick={() => {}} variant={variant} className="mr-2" />
       ))}
     </div>
   );
@@ -34,14 +34,15 @@ notes: ''
 
   return (
     <div className="d-flex">
-      <IconButton className="mr-2" icon={icon} alt='Close' onClick={() => {}} variant="brand" />
-      <IconButton className="mr-2" icon={icon} alt='Close' onClick={() => {}} variant="primary" />
-      <IconButton className="mr-2" icon={icon} alt='Close' onClick={() => {}} variant="secondary" />
-      <IconButton className="mr-2" icon={icon} alt='Close' onClick={() => {}} variant="success" />
-      <IconButton className="mr-2" icon={icon} alt='Close' onClick={() => {}} variant="warning" />
-      <IconButton className="mr-2" icon={icon} alt='Close' onClick={() => {}} variant="danger" />
-      <IconButton className="mr-2" icon={icon} alt='Close' onClick={() => {}} variant="light" />
-      <IconButton className="mr-2" icon={icon} alt='Close' onClick={() => {}} variant="dark" />
+      <IconButton className="mr-2" icon={icon} alt="Close" onClick={() => {}} variant="brand" />
+      <IconButton className="mr-2" icon={icon} alt="Close" onClick={() => {}} variant="primary" />
+      <IconButton className="mr-2" icon={icon} alt="Close" onClick={() => {}} variant="secondary" />
+      <IconButton className="mr-2" icon={icon} alt="Close" onClick={() => {}} variant="success" />
+      <IconButton className="mr-2" icon={icon} alt="Close" onClick={() => {}} variant="warning" />
+      <IconButton className="mr-2" icon={icon} alt="Close" onClick={() => {}} variant="danger" />
+      <IconButton className="mr-2" icon={icon} alt="Close" onClick={() => {}} variant="light" />
+      <IconButton className="mr-2" icon={icon} alt="Close" onClick={() => {}} variant="dark" />
+      <IconButton className="mr-2" icon={icon} alt="Close" onClick={() => {}} variant="black" />
     </div>
   );
 }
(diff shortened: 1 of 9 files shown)


src/PageBanner/PageBanner.scss
```diff
@@ -0,0 +1,47 @@
+.pgn__pageBanner-component {
+  width: 100%;
+  min-height: 36px;
+  max-height: 56px;
```
ciduarte, line 4: silly me, should these `px` be `em`?

ciduarte: same question w/ everywhere else I used `px`

src/IconButton/index.jsx
```diff
@@ -75,7 +75,7 @@ IconButton.propTypes = {
   /** Click handler for the button */
   onClick: PropTypes.func.isRequired,
   /** Type of button (uses Bootstrap options) */
-  variant: PropTypes.oneOf(['primary', 'secondary', 'success', 'warning', 'danger', 'light', 'dark']),
+  variant: PropTypes.oneOf(['primary', 'secondary', 'success', 'warning', 'danger', 'light', 'dark', 'black']),
```
muselesscreator, line 78: This should be added to the IconButton examples.
(inverted should probably as well...)

src/PageBanner/PageBanner.scss
muselesscreator: only instances I see of max-height in paragon use vh or percentage, so not sure... @abutterworth ?

src/IconButton/index.jsx
muselesscreator: Update: is now included in 2 of 3 areas for IconButton doc page where "all" variants are listed o.0

ciduarte: 👍  The black inverse variant was included, but wasn't easy to identify. Fixed the inverse variants section to list each of the individual Icon Buttons for clarity.

muselesscreator: XD

src/IconButton/README.md
```diff
@@ -52,24 +53,81 @@ notes: ''
...
-              icon={icon}
-              alt='Menu'
-              onClick={() => console.log("You clicked the menu button")}
-              variant={variant}
-              invertColors
-            />
-          </div>
-        );
-      })}
+      <div class="p-1 bg-brand">
```
muselesscreator, line 59: ++

muselesscreator (approved):
niiice 😎

src/IconButton/README.md
adamstankiewicz, line 59: nit: Should these examples be using the `className` prop?

src/PageBanner/PageBanner.scss
adamstankiewicz: Should this component have a `max-height` at all? What happens if the content of the `PageBanner` is multiple lines in length or otherwise has children that expands the height beyond 56px? 

IMO, this component should not enforce a max-height 😄

src/PageBanner/PageBanner.scss
```diff
@@ -0,0 +1,47 @@
...
+  background-color: $warning-100;
+  color: black;
+}
+
+.pgn__pageBanner-content {
+  display: flex;
+  flex-grow: 1;
+  justify-content: center;
+  align-items: center;
+  padding: 8px 12px;
```
adamstankiewicz, line 34: nit: Should the padding values come from the `$spacers` map? Example:

```
padding: map.get($spacers, 2) map.get($spacers, 2.5);
```

https://github.com/edx/paragon/blob/a861078c6e4dc452ab580dbdcfdbe46c632fa6c6/scss/core/_variables.scss#L360-L376

src/IconButton/README.md
```diff
@@ -52,24 +53,81 @@ notes: ''
...
+        <IconButton
+          icon={icon}
+          alt='Menu'
```
adamstankiewicz, line 62: nit: Some props in these examples use double quotes for their values and others use single quotes. We should align on using one or the other, but not both.

src/PageBanner/PageBanner.scss
```diff
@@ -0,0 +1,47 @@
...
+  text-align: center;
+  font-size: $x-small-font-size;
+
+  @media only screen and (min-width: 768px) {
+    font-size: $small-font-size;
+  }
```
adamstankiewicz, line 40: nit: Might consider using the Bootstrap's responsive mixins here in case the breakpoint widths ever change, it'll be captured here as well:

```
@include media-breakpoint-up(md) {
  font-size: $small-font-size;
}
```

src/PageBanner/PageBanner.scss
```diff
@@ -0,0 +1,47 @@
+.pgn__pageBanner-component {
```
adamstankiewicz, line 1: [design feedback] Should this class have some padding to account for the hover state on the icon button? As is, the hover state sits flush with the edges of this page banner container, which is a bit cleaner looking with a tiny bit of additional padding:

**Without padding**

image


**With padding**
Using `padding: map.get($spacers, 1);`  (equivalent to `.p-1` utility class)

image

src/PageBanner/PageBanner.scss
ciduarte: @adamstankiewicz I added a max height because it was a measurement that was called out in the design, but I'd agree with you here

src/PageBanner/PageBanner.scss
ciduarte: Yes I didn't think to use these! Thanks

src/IconButton/README.md
muselesscreator: The standard I tend to hold to is double-quotes for props, single quotes for strings.

adamstankiewicz: Same here :)

src/PageBanner/PageBanner.scss
ciduarte: For posterity, @ekangedx changed the design. We're now using the `inline` Icon Button rather than the `small` size, and because it's centered, it mimics that added padding suggestion.

adamstankiewicz: @ciduarte Great :) Thanks for bringing it to the Paragon Working Group for feedback. Using `inline` Icon Button is definitely preferable over adding explicit padding here.

adamstankiewicz (approved):
LGTM! :shipit:

PR #809, opened 2021-08-31
feat: create select menu

connorhaugh:
(WIP) Per [Specs](https://www.figma.com/file/eGmDp94FlqEr4iSqy1Uc1K/Paragon-Design-System?node-id=14061%3A15)

Jira Ticket: https://openedx.atlassian.net/browse/PAR-456

The default menu component is triggered on the click of a button or standalone link, and expands from the center. The menu contains a list of MenuItems, with a white background, and level 2 elevation. The menu also remembers the user’s selection and displays it as the label for the button trigger.

Any button or standalone link style can be used as a trigger, not just the secondary button in the example.

TODO: Unit Tests, Make Seperate readme for selectMenu with more detail


src/Menu/Menu.scss
```diff
@@ -0,0 +1,97 @@
...
+  }
+.pgn__menu{
+  border-radius: 0.25em;
+  box-shadow: $box-shadow;
+  background-color: $white;
+  border-color: transparent;
+  overflow: scroll;
+  max-height: 264px;
+}
+.pgn__menu-select-popup{
```
connorhaugh, line 81: If its like 20 lines, it doesn't need its own file right?

connorhaugh: Also, every impulse I have had tells me that px are the wrong unit to use here? am I correct?

src/hooks/useArrowKeyNavigation.js
```diff
@@ -16,15 +16,22 @@ function handleArrowKey({ event, currentIndex, availableElements }) {
 
   // Move the focus up or down. Wrap around ends of list.
   let nextElement;
-  if (event.key === 'ArrowDown') {
+
```
connorhaugh, line 19: get rid of this empty line

kenclary (approved):
Overall this looks fine. If paragon folks have nits, they'll have nits.

PR #810, opened 2021-09-07
fix: removes TableCell wrapper

ciduarte:
- removes TableCell wrapper from `<td>`

### DIFF
--- a/src/DataTable/TableCell.jsx
+++ b/src/DataTable/TableCell.jsx
@@ -2,13 +2,14 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-const TableCell = ({ getCellProps, render, column }) => (
-  <td {...getCellProps()}>
-    <span className={classNames('pgn__data-table-cell-wrap', column.cellClassName)}>
+const TableCell = ({ getCellProps, render, column }) => {
+  const { className, ...rest } = getCellProps();
+  return (
+    <td {...rest} className={classNames('pgn__data-table-cell-wrap', className, column.cellClassName)}>
       {render('Cell')}
-    </span>
-  </td>
-);
+    </td>
+  );
+};
 
 TableCell.propTypes = {
   /** Props for the td element */
@@ -17,6 +18,7 @@ TableCell.propTypes = {
   render: PropTypes.func.isRequired,
   /** Table column */
   column: PropTypes.shape({
+    /** Class(es) to be applied to the cells in the given column */
     cellClassName: PropTypes.string,
   }).isRequired,
 };
--- a/src/DataTable/DataTable.scss
+++ b/src/DataTable/DataTable.scss
@@ -83,7 +83,6 @@ $data-table-footer-position: center !default;
 
 .pgn__data-table-cell-wrap {
   max-width: 20vw;
-  display: block;
   white-space: normal;
   overflow: hidden;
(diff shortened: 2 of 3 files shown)


src/DataTable/TableCell.jsx
```diff
@@ -17,7 +22,10 @@ TableCell.propTypes = {
   render: PropTypes.func.isRequired,
   /** Table column */
   column: PropTypes.shape({
+    /** Class names for the cell wrapper */
     cellClassName: PropTypes.string,
+    /** Class names for the td element */
+    className: PropTypes.string,
```
mikix, line 28: I'm not fully aware of the problem here, but it might be good to name this like `dataClassName` or `tdClassName` or something?

Or pulling back, this distinction seems so minute - block level vs inline? Is that a real distinction we want to expose? Could we move the `cellClassName` up to the `td` level safely, to solve the same problem you are looking at? Or is that an API break?

ciduarte: I'm not sure what the original justification was for adding the `cellClassName` to the inline element vs the block element here, but I agree this feels a little redundant. Though, `.pgn__data-table-cell-wrap` has some styling that we may want to override (ex. `.pgn__data-table-cell-wrap` has a `max-width` and I found an edX repo that overrides it to none).

My use case for exposing the `td` element's className was to change its vertical alignment, which you can't access through the inline element.

I think given that there are use cases for both, it makes sense to expose both but update the names?

mikix: You may need someone more in tune with paragon than me.

Is moving `.pgn__data-table-cell-wrap` up to `<td>` and dropping the `<span>` altogether a bad idea?

ciduarte: I just tried to drop the span and update the styling and it didn't really go well, so I think that might be a bad idea

adamstankiewicz:
> I just tried to drop the span and update the styling and it didn't really go well, so I think that might be a bad idea

The code below seems to work locally without incorporating any breaking changes or the existing styles. 

@ciduarte I think the styling issues you were running into might have been that `.pgn__data-table-cell-wrap` has `display: block`, so when it was placed on the `td` directly, the `display: table-cell` that should be on the `td` was overridden. Removing the `display: block` from the `.pgn__data-table-cell-wrap` styles seems to work well.

image


```jsx
const TableCell = ({ getCellProps, render, column }) => (
  <td {...getCellProps({ className: classNames('pgn__data-table-cell-wrap', column.cellClassName) })}>
    {render('Cell')}
  </td>
);
```

```scss
.pgn__data-table-cell-wrap {
  max-width: 20vw;
  white-space: normal;
  overflow: hidden;
  text-overflow: ellipsis;
}
```

ciduarte:
@adamstankiewicz thanks for looking into this! I did try that yesterday, and didn't see any broken styling on the Paragon site. What I saw was specific to frontend-app-learning, and I couldn't get to the bottom of it so I assumed it would be an issue everywhere. Giving it a second look, I realized what was happening and this change feels safe to make!

For the curious, this change was impacting some cells that used `float-right`. Switching the cells to use `text-right` fixed the issue.
Screen Shot 2021-09-08 at 1 32 24 PM

adamstankiewicz (approved):
:shipit:

mikix (approved):

PR #812, opened 2021-09-14
fix: resolve PropType warning for indeterminate prop in DataTable

adamstankiewicz:
Resolves the following prop type warning regarding `indeterminate` checkboxes in the `DataTable` component.

image

### DIFF
--- a/src/DataTable/utils/useConvertIndeterminateProp.js
+++ b/src/DataTable/utils/useConvertIndeterminateProp.js
@@ -1,0 +1,23 @@
+import { useMemo } from 'react';
+
+/**
+ * A React hook to convert ``indeterminate`` prop to ``isIndeterminate``.
+ *
+ * @param {object} props An object containing props.
+ * @returns Updated props object.
+ */
+const useConvertIndeterminateProp = (props) => {
+  const updatedProps = useMemo(
+    () => {
+      const newProps = { ...props };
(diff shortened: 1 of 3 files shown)


src/DataTable/utils/useConvertIndeterminateProp.js
```diff
@@ -0,0 +1,23 @@
...
+ * A React hook to convert ``indeterminate`` prop to ``isIndeterminate``.
+ *
+ * @param {object} props An object containing props.
+ * @returns Updated props object.
+ */
+const useConvertIndeterminateProp = (props) => {
+  const updatedProps = useMemo(
+    () => {
+      const newProps = { ...props };
+      newProps.isIndeterminate = newProps.indeterminate;
```
muselesscreator, line 13: is reason why not use: `{ ...props, isIndeterminate: props.indeterminate }`?

adamstankiewicz: It'd have to be written as the following, such that `indeterminate` is no longer part of `props`, but yes... this is likely cleaner:

```jsx
const { indeterminate, ...rest } = props;
const newProps = { ...rest, isIndeterminate: indeterminate };
return newProps;
```

muselesscreator (approved):
LGTM

PR #815, opened 2021-09-17
fix: date-time input fields on blur issue

mehaknasir:
Ticket ID: https://openedx.atlassian.net/browse/TNL-8726
Date/Time Fields were not showing value when blurred out of the field due to opacity set to 0 on blur.
This PR contains a fix that opacity will only be set to 0 when the field does not contain any value.


awais-ansari (approved):

awaisdar001:
Thanks, Verified date/time component here => https://deploy-preview-815--paragon-edx.netlify.app/components/form/form-control/

saadyousafarbi (approved):

PR #808, opened 2021-08-31
feat: Added CardSelect component

irfanuddinahmad:
Description: This PR adds a radio selectable card form group.
image

JIRA: https://openedx.atlassian.net/browse/ENT-4799

PR #818, opened 2021-09-20
Revert "feat: Added CardSelect component (#808)"

adamstankiewicz:
This reverts commit d8dec978fff0a14213c0b8a7091850a8312501d3.

As discussed in a recent Paragon Working Group meeting, we'd like revert `CardSelect` because, as implemented, it's not as flexible as it should be (e.g., requires a card footer to exist, design bugs, etc.).

If we'd like this to be part of Paragon in the future, we'll ensure it goes through the proper channels to ensure the design use cases are covered and the implementation is flexible and conforms to Paragon conventions.

It also replaces Coveralls with Codecov.


muselesscreator (approved):
looks fine to me, once we get it merge-able :-p

PR #822, opened 2021-09-22
Adds ability to pass functions to DataTable actions

manny-m:
## [Closes PAR-573](https://openedx.atlassian.net/browse/PAR-573)

Adds ability to pass functions to DataTable `bulkActions` and `tableActions`. 

Why: To allow custom rendering of components inside the actions area

How: We pass functions instead of object describing buttons. For `bulkActions` we currently selected items in the first parameter of the function.


src/DataTable/BulkActions.jsx
```diff
@@ -6,7 +6,10 @@ import DataTableContext from './DataTableContext';
 import Actions from './CollapsibleButtonGroup';
 
 const BulkActions = ({ className }) => {
-  const { bulkActions } = useContext(DataTableContext);
+  const { bulkActions, selectedFlatRows } = useContext(DataTableContext);
+  if (typeof bulkActions === 'function') {
```
manny-m, line 10: Not a fan of using this method to distinguish these, I think one alternative might be using a new prop, `customBulkActions` that would override `bulkActtions`. Instead of just using a single prop to service both types of params.

long74100: I agree, a new prop that takes precedence seems preferable.

adamstankiewicz (commented):
As proposed, this looks decent. Though, I am still a bit concerned that in order to get access to the selected rows in the button text, the consumer will need to manually re-implement the logic (if relevant) for putting actions in the overflow menu that appears by default when there are more than 2 bulk actions.

Curious what you think 🤔, but what if we added support for passing an array of objects (as it does today) as well as an array of functions that are passed the selected rows to handle this case where consumers of `DataTable` want the default behavior but don't want to manually re-implement it themselves should they need access to the selected rows? This could be in addition to your proposed solution.

My biggest concern with this suggestion 🔼 is that we'd lose the prop type checking for the return value of the functions (i.e., to ensure the required attributes are included) unless we create our own custom prop type check that does that validation.

src/DataTable/filters/CheckboxFilter.jsx
```diff
@@ -33,7 +33,7 @@ function CheckboxFilter({
           id={headerBasedId}
           checked={checkedBoxes.includes(value)}
           onChange={() => { changeCheckbox(value); }}
-          label={<>{name} {number && <Badge variant="light">{number}</Badge>}</>}
+          label={<>{name} {number !== undefined && <Badge variant="light">{number}</Badge>}</>}
```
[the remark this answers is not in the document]
manny-m: The purpose of this conditional is that if we don't pass a number we don't render a badge. The problem with `number && <conition>` is that if `number` is equal to `0` the condition results in false.

The `proptype` for this component defines `number` as an optional number. We shouldn't expect a boolean to be passed number (cases like this make me wish we used TS 😢 )

 `null !== undefined` results in false so we don't need to worry about that.

manny-m:
@adamstankiewicz 
> Curious what you think 🤔, but what if we added support for passing an array of objects (as it does today) as well as an array of functions that are passed the selected rows to handle this case where consumers of `DataTable` want the default behavior but don't want to manually re-implement it themselves should they need access to the selected rows? This could be in addition to your proposed solution.

Something like this? I don't love it, but I think it would cover your use case... 

```
bulkActions = [
  ( {table} )=>{
    return {
      buttonText: "Foos: (table.selected)",
      handleClick: ()=>{
             foo(table)
      }
      disabled: {table.selected > 0}
     }
  },
  ({table})=>{.....}
]
``` 

Like you said
> we'd lose the prop type checking for the return value of the functions (i.e., to ensure the required attributes are included) unless we create our own custom prop type check that does that validation.

adamstankiewicz:
@manny-m Yeah, that's essentially what I was proposing. Would you mind if we go with it just so there's a path for consumers of `DataTable` who don't want to implement their own custom bulk actions but still need the selected row counts in the buttons? I don't love it either, but at least it's giving the data needed without sacrificing default, existing behavior. It would be good to write that custom prop type check function that validates the return value of any functions passed in (where's typescript when we need it lol).

src/DataTable/filters/CheckboxFilter.jsx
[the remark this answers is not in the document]
adamstankiewicz: Makes sense! Thanks for the clarification :)

src/DataTable/tests/BulkActions.test.jsx
```diff
@@ -276,4 +278,13 @@ describe('<BulkActions />', () => {
       expect(icon.props().screenReaderText).toEqual(SMALL_SCREEN_DROPDOWN_BUTTON_TEXT);
     });
   });
+
+  describe('with function', () => {
```
adamstankiewicz, line 282: We could probably include test case(s) for the added `PropTypes.arrayOf(PropTypes.func)` arguments.

src/DataTable/tests/TableActions.test.jsx
```diff
@@ -245,4 +247,13 @@ describe('<TableActions />', () => {
       expect(icon.props().screenReaderText).toEqual(SMALL_SCREEN_DROPDOWN_BUTTON_TEXT);
     });
   });
+
+  describe('with function', () => {
+    it('passed correct number of selected rows', () => {
```
adamstankiewicz, line 252: We could probably include test case(s) for the added `PropTypes.arrayOf(PropTypes.func)` arguments.

src/DataTable/TableActions.jsx
```diff
@@ -8,10 +8,21 @@ import Actions from './CollapsibleButtonGroup';
 const TableActions = ({ className }) => {
   const tableInstance = useContext(DataTableContext);
   const { tableActions } = tableInstance;
+  if (typeof tableActions === 'function') {
+    return <div className={classNames('pgn__table-actions', className)}>{tableActions()}</div>;
```
adamstankiewicz, line 12: Should `tableActions()` pass the `tableInstance` as an argument since the individual actions are passed `tableInstance` as well?

adamstankiewicz (approved):
Looking good. Just a couple additional minor comments, mainly to add test cases before merge for when a function defined button action is provided 😄 Should be good to go after that :shipit:

PR #824, opened 2021-09-28
fix: update iconAfter prop type definition

adamstankiewicz:
Changes the `Button` component's `iconAfter` prop type to `PropTypes.oneOfType` instead of `PropTypes.instanceOf` to match the `iconBefore` prop. Also, generally it doesn't seem `PropTypes.instanceOf` actually supports passing an array to it for multiple types per its implementation:

https://github.com/facebook/prop-types/blob/4de0644a10a554d0a556daa39f029369bc007ea5/factoryWithTypeCheckers.js#L297-L307

Also, configures Jest to ignore tests in the `dependent-usage-analyzer` directory (which may include Git checkouts of 30+ projects). Running tests with these projects checked out resulted in Jest attempting to run tests in all projects.

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,8 @@
     "testPathIgnorePatterns": [
       "/node_modules/",
       "/www/",
-      "/dist/"
+      "/dist/",
+      "/dependent-usage-analyzer/"
     ],
     "transformIgnorePatterns": [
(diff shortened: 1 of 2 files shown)


muselesscreator (approved):
LGTM

PR #851, opened 2021-10-21
fix: add aria-label and title back to hyperlink

adamstankiewicz:
https://openedx.atlassian.net/browse/PAR-565

* Adds aria-label and title back to Hyperlink
* Adds React.forwardRef to Hyperlink and MailtoLink
* Allow custom class names on Hyperlink and MailtoLink
* Don't include aria-label attribute if it's empty

**Previews**
https://deploy-preview-851--paragon-edx.netlify.app/components/hyperlink/


adamstankiewicz:
Example HTML output for `Hyperlink`:

image

PR #868, opened 2021-11-03
build: use shared browserslist config

adamstankiewicz:
* Removes custom `browserslist` configuration in favor of using a [shared configuration](https://github.com/edx/browserslist-config).

This change reduces the resultant asset bundle size.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1212,6 +1212,12 @@
         "minimist": "^1.2.0"
       }
     },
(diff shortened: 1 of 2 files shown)


adamstankiewicz:
[This PR](https://github.com/openedx/paragon/pull/1068) will unblock this browserslist update.

davidjoy (approved):

PR #871, opened 2021-11-04
feat: forward isBlocking to Alert and Marketing modals

muselesscreator:
Move requiredWhen from DataTable/utils/propTypesUtils into a top-level utils dir, and generalize into `customPropTypeRequirement`
`requiredWhen` and new `requiredWhenNot` methods wrap `customPropTypeRequirement`.

ModalDialog now forwards the `isBlocking` prop to the ModalLayer, and thus its inheritors now have that prop notated.
for MarketingModal and AlertModal, the `onClose` method is marked as required only if `isBlocking` is false-y.

### DIFF
--- a/src/utils/propTypesUtils.js
+++ b/src/utils/propTypesUtils.js
@@ -1,0 +1,57 @@
+/**
+ * Returns a the passed target PropType (targetType) if the conditionFn returns true
+ * when called with the props object.
+ * If the conditional is false and the associated prop is not included, raise an error,
+ * giving the provided filterString as the explanation for the failure.
+ * @param {func} targetType - target PropType method
+ * @param {func} isRequiredFn - function taking the props object and returning whether or
+ *   not the associated prop should be required
+ * @param {string} filterString - string explanation of the isRequiredFn condition for error
+ *   messages.
+ * @return {func} - PropType based on targetType that is required if conditionFn returns true
+ *   when called with the props object.
+ */
+export const customPropTypeRequirement = (targetType, conditionFn, filterString) => (
+  (props, propName, componentName, ...rest) => {
+    if (conditionFn(props) && props[propName] === undefined) {
+      return new Error(
+        `${componentName}: ${propName} is required when ${filterString}`,
+      );
+    }
+    return targetType(props, propName, componentName, ...rest);
+  }
+);
+
+/**
+ * Returns a PropType entry with the given propType that is required if dependentOnProp
+ * is set to true.
+ * @param {func} propType - target PropType
+ * @param {string} dependentOnProp - string name for prop that, if true, marks the
+ *   associated prop as required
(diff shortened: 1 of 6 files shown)


src/utils/propTypesUtils.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {func} propType - target PropType
+ * @param {string} dependentOnProp - string name for prop that, if true, marks the
+ *   associated prop as required
+ * @return {func} - PropType based on propType that is required if dependentOnProp is
+ *   set to true.
+ */
+export const requiredWhen = (propType, dependentOnProp) => (
+  customPropTypeRequirement(
+    propType,
+    (props) => props[dependentOnProp] === true,
```
muselesscreator, line 37:
```suggestion
    (props) => !!props[dependentOnProp],
```

src/utils/propTypesUtils.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * Returns a the passed target PropType (targetType) if the conditionFn returns true
```
adamstankiewicz, line 2: grammar nit: "Returns a the..."

src/utils/propTypesUtils.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {string} dependentOnProp - string name for prop that, if true, marks the
+ *   associated prop as required
+ * @return {func} - PropType based on propType that is required if dependentOnProp is
+ *   set to true.
+ */
+export const requiredWhen = (propType, dependentOnProp) => (
+  customPropTypeRequirement(
+    propType,
+    (props) => !!props[dependentOnProp],
+    `${dependentOnProp} is set to true`,
```
adamstankiewicz, line 38: The "set to true" is no longer relevant since we're now looking just for _any_ defined truthy value here.

src/utils/propTypesUtils.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Returns a PropType entry with the given propType that is required if dependentOnProp
+ * is false-y.
+ * @param {func} propType - target PropType
+ * @param {string} otherPropName - string name for prop that, if false-y, marks the
+ *   associated prop as required
+ * @return {func} - PropType based on propType that is required if dependentOnProp is
+*    false-y
+ */
+export const requiredWhenNot = (propType, otherPropName) => (
```
adamstankiewicz, line 51: nit: maybe be consistent with the argument name (dependentOnProp) from the other function?

adamstankiewicz (approved):
LGTM just a few tiny little clean up things :)

src/utils/propTypesUtils.js
muselesscreator: I prefered otherProp here because `dependentOn` didn't track as well.  tried to come up with an equivalent to `dependentOnNot` or something, but this felt cleaner.

PR #890, opened 2021-11-17
[BD-46] docs: improve documentation for components

viktorrusakov:
Improve documentation for Alert, AlertModal, ActionRow, Avatar, AvatarButton components

JIRA: [PAR-578](https://openedx.atlassian.net/browse/PAR-578)

### DIFF
--- a/src/ActionRow/README.md
+++ b/src/ActionRow/README.md
@@ -22,6 +22,22 @@ ActionRow assumes that its last child is the primary action and lays out actions
 ```jsx live
 <ActionRow>
   <Form.Checkbox>Don't ask me again.</Form.Checkbox>
+  <Button variant="tertiary">
+    Cancel
+  </Button>
+  <Button variant="primary">
+    Submit
+  </Button>
+</ActionRow>
+```
+
+### Usage with a Spacer
+
+ActionRow can also be used with a helper component ``ActionRow.Spacer`` to insert empty space between children
+
+```jsx live
+<ActionRow>
+  <Form.Checkbox>Don't ask me again.</Form.Checkbox>
   <ActionRow.Spacer />
   <Button variant="tertiary">
     Cancel
@@ -47,4 +63,11 @@ ActionRow assumes that its last child is the primary action and lays out actions
     Continue
   </Button>
 </ActionRow>
+```
+
+### Theme variables (SCSS)
+
+```scss
+$action-row-gap-x:                   .5rem !default;
+$action-row-gap-y:                   .5rem !default;
 ```
--- a/src/ActionRow/README.md
+++ b/src/ActionRow/README.md
@@ -21,7 +21,6 @@ ActionRow assumes that its last child is the primary action and lays out actions
 
 ```jsx live
 <ActionRow>
-  <Form.Checkbox>Don't ask me again.</Form.Checkbox>
   <Button variant="tertiary">
     Cancel
   </Button>
--- a/src/Alert/README.md
+++ b/src/Alert/README.md
@@ -3,6 +3,8 @@ title: 'Alert'
 type: 'component'
 components:
 - Alert
+- AlertHeading
+- AlertLink
 categories:
 - Status & metadata
 status: 'Stable'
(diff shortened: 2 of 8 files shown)


edx-webhook:
Thanks for the pull request, @viktorrusakov! I've created [BLENDED-1016](https://openedx.atlassian.net/browse/BLENDED-1016) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


src/ActionRow/README.md
```diff
@@ -19,6 +19,22 @@ in a row in a control bar or nav.
 ActionRow assumes that its last child is the primary action and lays out actions so that the last item is in a primary location. If horizontal, the primary action sits on the right. If stacked, the primary action sits at the top of the stack (this is done via `flex-direction: column-reverse;`).
 ### Basic Usage
 
+```jsx live
+<ActionRow>
+  <Form.Checkbox>Don't ask me again.</Form.Checkbox>
```
adamstankiewicz, line 24: Hmm, for this particular example, let's remove the "Don't ask me again" checkbox and just keep it simple with the 2 action buttons. We can keep the checkbox/label with the example specific the usage with a spacer below :)

src/Alert/README.md
```diff
@@ -12,10 +14,17 @@ notes: |
...
+- **Info**: used to convey general information or actions that aren't critical, note that info variant should not contain any icons in it
+- **Warning**: used to display information that needs attention
+- **Success**: used for success messages
+- **Danger**: used to communicate problems that have to be resolved immediately
+
+<p>
   This is a pass through component from React-Bootstrap.<br/>
   <a href="https://react-bootstrap.github.io/components/alerts/" target="_blank" rel="noopener noreferrer">
-    See React-Bootstrap for documentation.
+  See React-Bootstrap for additional documentation.
```
adamstankiewicz, line 27: Since this component adds some custom things on top of the base `Alert` component from react-boostrap, maybe instead of saying "This is a pass through component...",  we can be a bit more explicit by saying something along the lines of, "This component utilizes and extends the `Alert` component from react-bootstrap"? 🤔 

I think we should only say something is a pass through component if we use the vanilla component exported by react-bootstrap without any modifications. What do you think?

src/Alert/index.jsx
```diff
@@ -104,7 +136,8 @@ Alert.defaultProps = {
   stacked: false,
 };
 
-Alert.Link = BaseAlert.Link;
-Alert.Heading = BaseAlert.Heading;
+Alert.Heading = AlertHeading;
+Alert.Link = AlertLink;
 
+export { AlertHeading, AlertLink };
```
adamstankiewicz, line 142: Are these named exports necessary when they are accessible via `Alert.Link` and `Alert.Heading`?

src/Alert/index.jsx
```diff
@@ -74,6 +75,35 @@ const Alert = React.forwardRef(({
   );
 });
 
+// This is needed to display a default prop for Alert.Heading element
+// Copied from react-bootstrap since BaseAlert.Heading component doesn't have defaultProps,
+// so there seems to be no other way of providing correct default prop for base element
+const DivStyledAsH4 = divWithClassName('h4');
```
adamstankiewicz, line 81: I think this is a reasonable approach. Interesting `BaseAlert.Heading` doesn't have defaultProps specified despite defaultProps definitely existing lol. My one concern with this approach is dealing with future code drift between the props we specify here vs. what's in react-bootstrap but I think the value of having the props listed out in our Paragon docs is worth that risk.

When upgrading react-bootstrap, we'll just have to make sure to account for any changes in prop types for components like this to account for them on our end.

src/ActionRow/README.md
viktorrusakov: Yeah, this should be better, updated

src/Alert/README.md
viktorrusakov: I totally agree! I started thinking about that too while working on the next PR, updated phrasing here also, thanks

src/Alert/index.jsx
viktorrusakov: They are actually not, removed them now. For some reason I thought documentation site would not pick up and display props definitions if you don't explicitly export the components, but now after taking a closer look at the code I see that's not how it works :). Thank for pointing it out!

src/Alert/index.jsx
viktorrusakov: Yeah, we probably need to make a list of components that use react-bootstrap, so we can quickly check everything once we upgrade.

adamstankiewicz (approved):

PR #911, opened 2021-11-30
[BD-46] feat: implement Chip component

monteri:
added Chip component with tests, description and styles

JIRA: [PAR-437](https://openedx.atlassian.net/jira/software/projects/PAR/boards/654?selectedIssue=PAR-437)

### DIFF
--- a/src/Chip/README.md
+++ b/src/Chip/README.md
@@ -1,0 +1,35 @@
+---
+title: 'Chip'
+type: 'component'
+components:
+- Chip
+categories:
+- Buttonlike
+status: 'New'
+designStatus: 'Done'
+devStatus: 'In progress'
+notes: |
+---
+
+``Chips`` are compact elements that represent an input, attribute, or action. Similar to the Badge component, but interactive.
+
+### Basic Usage
+
+```jsx live
+<div>
+  <Chip>New</Chip>
+  <Chip active>New</Chip>
+  <Chip disabled>New</Chip>
+  <Chip variant="dark">New</Chip>
+</div>
+```
+
+### With icon before and remove icon
+
+```jsx live
+<div>
+  <Chip iconBefore={Person}>New</Chip>
+  <Chip variant="dark" iconBefore={Person} iconAfter={CloseSmall}>New</Chip>
+  <Chip variant="dark" iconBefore={Person} iconAfter={CloseSmall} disabled>New</Chip>
+</div>
+```
--- a/src/Chip/README.md
+++ b/src/Chip/README.md
@@ -11,7 +11,7 @@ devStatus: 'In progress'
 notes: |
 ---
 
-``Chips`` are compact elements that represent an input, attribute, or action. Similar to the Badge component, but interactive.
+``Chips`` are compact elements that represent an input, attribute, or action. Similar to the [Badge](/components/badge) component, but interactive.
 
 ### Basic Usage
 
--- a/src/Chip/_variables.scss
+++ b/src/Chip/_variables.scss
@@ -1,0 +1,36 @@
+
+// Chip
+
+$chip-padding-y:                    .125rem !default;
+$chip-padding-x:                    .5rem !default;
+$chip-margin-inline:                .5rem !default;
+$chip-border-radius:                .4375rem !default;
+$chip-border-width:                 .0625rem !default;
+$chip-font-size:                    .75rem !default;
+$chip-font-weight:                  400 !default;
+$chip-line-height:                  1.5rem !default;
+$chip-disable-opacity:              .3 !default;
+
+$theme-variants: ('light', 'dark');
+$theme-colors:
+        (
+                "light-background-inactive": $white,
+                "light-text-inactive":       $primary-700,
+                "light-border-inactive":     $light-300,
+                "light-background-hover":    $primary-500,
+                "light-text-hover":          $white,
+                "light-border-hover":        $white,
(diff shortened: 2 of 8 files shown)


edx-webhook:
Thanks for the pull request, @monteri! I've created [OSPR-6247](https://openedx.atlassian.net/browse/OSPR-6247) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

src/Chip/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+      {children}
+      {iconAfter && <Icon className={classNames('chip-icon-after')} src={iconAfter} />}
+    </>
+  );
+  return React.createElement(props.href ? 'a' : elementType, elementProps, elementContent);
+};
+
+Chip.propTypes = {
+  /** Specifies the base HTML element. */
+  as: PropTypes.elementType,
```
viktorrusakov, line 48: In the component implementation you only allow 'button' or 'a' elements, maybe you can define this props as `PropTypes.oneOf(['button', 'a'])` or something like this? This way you don't need to check if the prop is correct like you do [here](https://github.com/edx/paragon/pull/911/files#diff-b7460be0253af56f24ad7fb65320ffdac22f6f0f26b87ff93ca36943c3cac36dR26)

I'm also not sure if we need to restrict the element to only 'a' or 'button', I think everywhere else in Paragon `as` is always an arbitrary element, maybe we should do the same here?

src/Chip/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+  disabled: PropTypes.bool,
+  /** Providing a `href` will render an `<a>` element, styled as a button. */
+  href: PropTypes.string,
+  /**
+   * An icon component to render before the content.
+   * Example import of a Paragon icon component:
+   *
+   * `import { Check } from '@edx/paragon/icons';`
+   */
+  iconBefore: PropTypes.oneOfType([PropTypes.func, PropTypes.node]),
```
viktorrusakov, line 67: Why do you allow icons to be functions also? 🤔 
Shouldn't it just be `PropTypes.node`?

viktorrusakov (changes requested):

src/Chip/README.md
```diff
@@ -0,0 +1,35 @@
...
+- Chip
+categories:
+- Buttonlike
+status: 'New'
+designStatus: 'Done'
+devStatus: 'In progress'
+notes: |
+---
+
+``Chips`` are compact elements that represent an input, attribute, or action. Similar to the Badge component, but interactive.
```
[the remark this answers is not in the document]
monteri: Yes, I will make it is as link to the Badge

src/Chip/_variables.scss
```diff
@@ -0,0 +1,36 @@
...
+                "dark-background-inactive": $primary-300,
+                "dark-text-inactive":       $white,
+                "dark-border-inactive":     $primary-300,
+                "dark-background-hover":    $white,
+                "dark-text-hover":          $primary-700,
+                "dark-border-hover":        $primary-300,
+                "dark-background-selected": $primary-300,
+                "dark-text-selected":       $white,
+                "dark-border-selected":     $white,
+        );
```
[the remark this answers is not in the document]
monteri: Yes, I missed it, thanks

src/Chip/index.jsx
monteri: According to the task it must either `button` or `a`. But I agree, everywhere it is arbitrary and here would be good to have the same behavior. @adamstankiewicz what do you think about this?

viktorrusakov: Ohh I see, I missed that in task description, sorry.

Let's see what Adam thinks then 🙂

src/Chip/index.jsx
monteri: I copied from the Button component, but yes it should be just `PropTypes.node`. Editing, thanks

src/Chip/index.jsx
adamstankiewicz: Yes, I agree. If we can build this to keep `as` consistent with how it's used every else, we should do that :)

adamstankiewicz (approved):
This looks good to me. Just a bit of cleanup to make `as` arbitrary like everywhere else from the previous suggestion by @viktorrusakov!

src/Chip/README.md
```diff
@@ -0,0 +1,35 @@
+---
+title: 'Chip'
+type: 'component'
+components:
+- Chip
+categories:
+- Buttonlike
```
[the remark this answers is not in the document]
adamstankiewicz: On second thought, I agree. It's interactive as if its a button. Let's keep this as is!

edx-webhook:
@monteri 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/paragon&entry.867055334=https://github.com/edx/paragon/pull/911&entry.1484655318=https://github.com/monteri&entry.752974735=2021-11-30+13:58&entry.1917517419=2021-12-16+13:32&entry.2133058324=Yes) so we can improve your experience in the future.

PR #921, opened 2021-12-06
[BD-46] feat: add Annotation component

viktorrusakov:
Implement Annotation component

Preview: https://deploy-preview-921--paragon-edx.netlify.app/components/annotation/

Figma: https://www.figma.com/file/eGmDp94FlqEr4iSqy1Uc1K/?node-id=14061%3A33

JIRA: [PAR-579](https://openedx.atlassian.net/browse/PAR-579)

### DIFF
--- a/src/Annotation/Annotation.scss
+++ b/src/Annotation/Annotation.scss
@@ -1,0 +1,72 @@
+@import "variables";
+
+@mixin triangle($color, $direction) {
+  height: 0;
+  width: 0;
+  position: absolute;
+  content: '';
+  border: solid transparent;
+
+  @if $direction == top {
+    border-bottom-color: $color;
+    border-width: 0 $annotation-arrow-border-width $annotation-arrow-border-width;
+    right: 0;
+    left: 0;
+    top: -$annotation-arrow-border-width;
+    margin: 0 auto;
+  } @else if $direction == right {
+    border-left-color: $color;
+    border-width: $annotation-arrow-border-width 0 $annotation-arrow-border-width $annotation-arrow-border-width;
+    top: 0;
+    bottom: 0;
+    right: -$annotation-arrow-border-width;
+    margin: auto 0;
+  } @else if $direction == bottom {
+    border-top-color: $color;
+    border-width: $annotation-arrow-border-width $annotation-arrow-border-width 0;
+    right: 0;
+    left: 0;
+    bottom: -$annotation-arrow-border-width;
+    margin: 0 auto;
+  } @else if $direction == left {
+    border-right-color: $color;
+    border-width: $annotation-arrow-border-width $annotation-arrow-border-width $annotation-arrow-border-width 0;
+    top: 0;
+    bottom: 0;
+    left: -$annotation-arrow-border-width;
+    margin: auto 0;
+  } @else {
+    @error "Unknown direction #{$direction}.";
+  }
+}
+
+.pgn__annotation {
+  padding: $annotation-padding;
+  border-radius: $annotation-border-radius;
+  max-width: $annotation-max-width;
+  filter: $annotation-box-shadow;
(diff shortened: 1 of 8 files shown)


edx-webhook:
Thanks for the pull request, @viktorrusakov! I've created [BLENDED-1034](https://openedx.atlassian.net/browse/BLENDED-1034) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


src/Annotation/Annotation.scss
```diff
@@ -0,0 +1,72 @@
+@import "variables";
+
+@mixin triangle($color, $direction) {
```
adamstankiewicz, line 3: Is this the same triangle that's used by `Tooltip` and `Popover`? If not, would it be possible to use the same arrow as `Tooltip`/`Popover`? I believe the triangle for those components is coming from react-bootstrap.

src/Annotation/README.md
```diff
@@ -0,0 +1,78 @@
...
+  <Annotation
+    variant="light"
+    className="mb-2"
+  >
+    Light variant. By default max width is set to 300px with wrapping text.
+  </Annotation>
+</>
+```
+
+### Theme variables (SCSS)
```
adamstankiewicz, line 57: With the other recently merged PR to add these theme variables automatically, they shouldn't be necessary to include in the README anymore!

adamstankiewicz (commented):
A couple things, but looking great. Also, from a documentation perspective, I'm wondering if it'd be good to include an example of the `Annotation` positioned such that it's referring to something (same a button). With `Tooltip` and `Popover`, the popup automatically positions itself onto its target. I might be a bit concerned that it may be non-obvious how to position `Annotation` to refer to another element.

src/Annotation/README.md
viktorrusakov: Yeah, we don't, removed it, thanks!

viktorrusakov:
> I might be a bit concerned that it may be non-obvious how to position `Annotation` to refer to another element.

So am I now that I think about it. I've added some examples, please take a look and tell me what you think.

src/Annotation/Annotation.scss
viktorrusakov: No, it's not the same, I though I'd try to implement new arrow after seeing that both `Popover` and `Tooltip` use arrow from react-bootstrap for couple of reasons:
1. Both arrows from react-bootsrap contain variables specific to the component they relate to (e.g. [Tooltip arrow](https://github.com/twbs/bootstrap/blob/main/scss/_tooltip.scss)), I can't override them since they are used in Paragon's  components (I believe I would need to use `@extend` and override them directly in class definition)
2. They both don't support color variants, so I would need to modify them

I thought that if I try to use bootstrap's arrow I would end up copying a lot of stuff to customize it for `Annotation` component design spec and decided that maybe it would be better to implement new one. I can switch to react-bootstrap if you prefer it more.

adamstankiewicz (approved):

edx-webhook:
@viktorrusakov 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/paragon&entry.867055334=https://github.com/edx/paragon/pull/921&entry.1484655318=https://github.com/viktorrusakov&entry.752974735=2021-12-06+08:40&entry.1917517419=2021-12-16+13:21&entry.2133058324=Yes) so we can improve your experience in the future.

PR #924, opened 2021-12-06
[BD-46] feat: implement new searchfield component design

monteri:
- added constants for the `variant` and `submitButtonLocation` props
- added functionality for the new properties: `variant`, `disabled`, `buttonText`
- added styles according to the [Figma design](https://www.figma.com/file/eGmDp94FlqEr4iSqy1Uc1K/Paragon-Design-System?node-id=14061%3A21)
- implemented the `submitButtonLocation` feature
- updated and added tests
- added a small example in the `README.md`

JIRA: [PAR-281](https://openedx.atlassian.net/jira/software/projects/PAR/boards/654?selectedIssue=PAR-281)

### DIFF
--- a/src/SearchField/README.md
+++ b/src/SearchField/README.md
@@ -152,3 +152,20 @@ For needs that deviate from the basic usage above, use `<SearchField.Advanced />
   <SearchField.Input />
 </SearchField.Advanced>
 ```
+
+### Advanced usage with the submit button outside the input
+
+Use class `pgn__searchfield_wrapper` to group input elements apart from the submit button.
+
+```jsx live
+<SearchField.Advanced
+  onSubmit={value => console.log(`search submitted: ${value}`)}
+  submitButtonLocation="external"
+>
+  <div className="pgn__searchfield_wrapper">
+    <SearchField.Label/>
+    <SearchField.Input/>
+  </div>
+  <SearchField.SubmitButton submitButtonLocation="external" />
+</SearchField.Advanced>
+```
--- a/src/SearchField/README.md
+++ b/src/SearchField/README.md
@@ -77,6 +77,15 @@ notes:
 />
 ```
 
+### With the submit button outside the input
+
+```jsx live
+<SearchField
+  submitButtonLocation="external"
+  onSubmit={value => console.log(`search submitted: ${value}`)}
+/>
+```
+
 ### Advanced Usage
 
 For needs that deviate from the basic usage above, use `<SearchField.Advanced />`. The `children` elements must contain the `SearchField.Label` and `SearchField.Input` components at a minimum.
--- a/src/SearchField/SearchField.scss
+++ b/src/SearchField/SearchField.scss
@@ -4,6 +4,7 @@
     transition: $input-transition;
     border: $search-border-width solid $search-border-color;
 
+    &.disabled,
     &:disabled {
         opacity: $search-disabled-opacity;
         pointer-events: none;
@@ -23,8 +24,8 @@
     }
 
     .form-control {
-        @extend .border-0;
-        border-radius: $search-border-radius !important;
+        border: none;
+        border-radius: $search-border-radius;
         height: $input-height-search;
 
         &:focus {
@@ -61,15 +62,31 @@
     &.pgn__searchfield--external {
         border: none;
 
+        .btn-primary {
+            background: map-get($search-btn-variants, 'light');
+        }
+
+        .btn-brand {
+            background: map-get($search-btn-variants, 'dark');
+        }
+
         &.has-focus {
             box-shadow: none;
+
+            .btn-primary {
+                background: map-get($search-btn-variants, 'light');
+            }
+
+            .btn-brand {
+                background: map-get($search-btn-variants, 'dark');
+            }
         }
     }
 
     .pgn__searchfield_wrapper {
-        @extend .d-flex;
-        @extend .align-items-center;
-        @extend .w-100;
+        display: flex;
+        align-items: center;
+        width: 100%;
 
         border: $search-border-width solid $search-border-color;
 
@@ -83,10 +100,4 @@
 .pgn__searchfield__button.btn[type='submit'] {
     border-radius: 0;
     margin-left: $search-button-margin;
-
-    &:hover:before,
-    &:focus:before,
-    &:active:before {
-        border-radius: 0 !important;
-    }
 }
--- a/src/SearchField/SearchField.scss
+++ b/src/SearchField/SearchField.scss
@@ -1,14 +1,32 @@
-.pgn__searchfield {
-    @extend .border;
-    @extend .rounded;
-    transition: $input-transition;
+@import "variables";
 
-    &.has-focus {
-        box-shadow: $input-focus-box-shadow;
+.pgn__searchfield {
+    transition: $input-transition;
+    border: $search-border-width solid $search-border-color;
+
+    &.disabled,
+    &:disabled {
+        opacity: $search-disabled-opacity;
+        pointer-events: none;
     }
 
-    .form-control {
+    &.has-focus {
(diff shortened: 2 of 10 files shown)


edx-webhook:
Thanks for the pull request, @monteri! I've created [BLENDED-1036](https://openedx.atlassian.net/browse/BLENDED-1036) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


monteri:
On the preview the external button is displayed with the wrong background color

Screenshot:
image

This is because in theme repository the styles are overridden:

Link to the overridden styles:
https://github.com/edx/brand-edx.org/blob/7a268cabb76567873c6722efb67d7400d9b3d03f/paragon/_overrides.scss#L265

src/SearchField/README.md
```diff
@@ -77,6 +77,15 @@ notes:
...
+### With the submit button outside the input
+
+```jsx live
+<SearchField
+  submitButtonLocation="external"
+  onSubmit={value => console.log(`search submitted: ${value}`)}
+/>
+```
+
 ### Advanced Usage
```
adamstankiewicz, line 89: Is there an example of using `SearchField.Advanced with an external button location? That might be a good one to include here.

Related, is the `disabled` prop on `SearchField` and/or `SearchField.Advanced` working? As far as I can tell, adding `disabled` isn't actually disabling the form / button.

src/SearchField/SearchField.scss
```diff
@@ -1,14 +1,31 @@
...
+        }
+    }
+
+    &:hover,
+    &:active {
+        border-color: $search-border-color-interaction;
     }
 
     .form-control {
         @extend .border-0;
```
adamstankiewicz, line 26: I believe it's an anti-pattern to use `@extend` for Paragon styles because the extended rules include a `!important` which makes it hard to theme the components.

As part of this PR, let's remove the `@extend` throughout this file and use SCSS variables or vanilla CSS instead (e.g., `border: none;` here).

src/SearchField/SearchField.scss
```diff
@@ -29,14 +46,47 @@
...
+    &.pgn__searchfield--external {
+        border: none;
+
+        &.has-focus {
+            box-shadow: none;
+        }
+    }
+
+    .pgn__searchfield_wrapper {
+        @extend .d-flex;
```
adamstankiewicz, line 70: See above comment about `@extend`

src/SearchField/SearchField.scss
```diff
@@ -29,14 +46,47 @@
...
+}
+
+.pgn__searchfield__button.btn[type='submit'] {
+    border-radius: 0;
+    margin-left: $search-button-margin;
+
+    &:hover:before,
+    &:focus:before,
+    &:active:before {
+        border-radius: 0 !important;
```
adamstankiewicz, line 90: Is it possible to avoid using `!important` here?

src/SearchField/index.jsx
```diff
@@ -10,21 +10,46 @@ import SearchFieldInput from './SearchFieldInput';
 import SearchFieldClearButton from './SearchFieldClearButton';
 import SearchFieldSubmitButton from './SearchFieldSubmitButton';
 
+const STYLE_VARIANTS = [
+  'light',
+  'dark',
+];
+
+const BUTTON_LOCATION_VARIANTS = [
```
adamstankiewicz, line 18: `STYLE_VARIANTS` and `BUTTON_LOCATION_VARIANTS` are defined several times. Would it make sense to export them from a constants.js file?

src/SearchField/SearchFieldSubmitButton.jsx
```diff
@@ -1,21 +1,72 @@
...
+  if (submitButtonLocation === 'internal' && value.length) {
+    return null;
+  }
 
-  return (
+  return submitButtonLocation === 'external' ? (
+    <Button
+      type="submit"
+      ref={refs.submitButton}
+      variant={variant === 'light' ? 'dark' : 'danger'}
```
adamstankiewicz, line 34: Should these be `primary` and `brand` (per Figma)?

src/SearchField/SearchField.scss
```diff
@@ -1,14 +1,31 @@
+@import "variables";
+
 .pgn__searchfield {
-    @extend .border;
-    @extend .rounded;
     transition: $input-transition;
+    border: $search-border-width solid $search-border-color;
+
+    &:disabled {
+        opacity: $search-disabled-opacity;
```
adamstankiewicz, line 8: Does this need to be `.disabled` so it applies as a className vs. a pseudoselector? I'm also curious if this is the most appropriate way to disable these elements from an a11y perspective, as the `input` and `button` elements do take a `disabled` prop natively with their own styles like a faded opacity built in.

Does it make sense to also pass `disabled` prop to these input/button elements as well?

adamstankiewicz (commented):
This looks good! A few comments/suggestions. 

Also, on the brand-edx.org theme override issue, hmm... is there a way we can release a version of brand-edx.org and this Paragon PR such that we don't break any existing styles if these packages are automatically upgraded to by bots in consuming repos?

We could at least remove those styles from 

Another observation:

src/SearchField/README.md
monteri: Yes, right, I add `submitButtonLocation` example with the `SearchField.Advanced` and I add disabled working on the `SeachField.Advanced`.

src/SearchField/SearchField.scss
monteri: Agree, updating this

src/SearchField/index.jsx
monteri: I did this at the beginning but then in the Props API it displays variable name (e.g., `STYLE_VARIANTS`) instead of `'light' | 'dark'`. That is due to importing constants from another file. For now there is no better solution other than duplicating.

src/SearchField/SearchFieldSubmitButton.jsx
monteri: Yes, I will update it

src/SearchField/SearchField.scss
monteri: Yes, I see the disable functionality was made with the `.disabled` className and then I removed it by mistake after one of the code review.
- For the whole component `disabled` works good now.
- For input I added prop `option: { disabled: true/false }` but I investigated and it doesn't implemented yet. But I pass this prop anyway, so if Input have disabled functionality in future it will use them immediately.
- For the SubmitButton I added disabled prop from the context. Now button is properly disabled.

monteri:
> This looks good! A few comments/suggestions.
> 
> Also, on the brand-edx.org theme override issue, hmm... is there a way we can release a version of brand-edx.org and this Paragon PR such that we don't break any existing styles if these packages are automatically upgraded to by bots in consuming repos?
> 
> Could we use a different class name in some way here so the brand override doesn't apply to the updated component?

Fixed the issue by overriding `background` prop for `btn-primary` and `btn-brand`

adamstankiewicz (approved):

edx-webhook:
@monteri 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/paragon&entry.867055334=https://github.com/edx/paragon/pull/924&entry.1484655318=https://github.com/monteri&entry.752974735=2021-12-06+14:24&entry.1917517419=2021-12-16+13:15&entry.2133058324=Yes) so we can improve your experience in the future.

PR #925, opened 2021-12-06
feat: update form switch per figma

long74100:
- Support label on the left for form.switch
- Change background color for form switch (we only need to support success variant)
- Add support for helper text below the switch
- Remove outline color variations


manny-m (approved):
Looks great! I would consider adding tests for `floatLabelLeft`, but otherwise looking good!

PR #937, opened 2021-12-13
Fixed Datatable ActionDisplay to support functions

manny-m:
This change makes sure that if we pass a function it is always rendered.

In `DataTable`, when passing a function through `tableActions` and `bulkActions`, with no parameters it would not render the component. This happened because we were (unintentianlly) counting the number of parameters in a function, if the function supplied took zero arguments the code would flag `noBulkActions` or `noTableActions` as true.

### DIFF
--- a/src/DataTable/index.jsx
+++ b/src/DataTable/index.jsx
@@ -110,7 +110,6 @@ function DataTable({
   }, [fetchData, JSON.stringify(tableStateWithoutSelections)]);
 
   const selectionActions = useSelectionActions(instance, controlledTableSelections);
-
   const enhancedInstance = {
     ...instance,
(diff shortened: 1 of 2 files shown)


src/DataTable/index.jsx
```diff
@@ -245,7 +245,7 @@ DataTable.propTypes = {
           /** optional button variant; only relevant for the first two buttons */
           variant: PropTypes.string,
           /** disables button */
-          disabled: PropTypes.disabled,
+          disabled: PropTypes.bool,
```
adamstankiewicz, line 248: 🎉

adamstankiewicz (approved):
LGTM, good find.

PR #939, opened 2021-12-14
[BD-46] feat: allow using IconButton as a base element for Dropdown.Toggle

viktorrusakov:
- allow using `IconButton` as a base element for `Dropdown.Toggle`
- refactor usage of `Dropdown.Toggle` in `DataTable` as a result of the first change

JIRA: [PAR-334](https://openedx.atlassian.net/browse/PAR-334)

Preview: https://deploy-preview-939--paragon-edx.netlify.app/components/dropdown/#with-iconbutton

### DIFF
--- a/src/Dropdown/README.md
+++ b/src/Dropdown/README.md
@@ -6,6 +6,7 @@ components:
 - DropdownItem
 - DropdownMenu
 - DropdownButton
+- DropdownToggle
 categories:
 - Navigation
 status: 'Stable'
@@ -22,7 +23,7 @@ notes: |
   </a>
 </p>
 
-### basic usage
+### Basic Usage
 ```jsx live
 <DropdownButton id="dropdown-basic-button" title="Dropdown button">
   <Dropdown.Item href="#/action-1">Action</Dropdown.Item>
@@ -31,7 +32,7 @@ notes: |
 </DropdownButton>
 ```
 
-### advanced usage
+### Advanced Usage
 
 ```jsx live
 <Dropdown>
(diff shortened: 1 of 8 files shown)


edx-webhook:
Thanks for the pull request, @viktorrusakov! I've created [BLENDED-1048](https://openedx.atlassian.net/browse/BLENDED-1048) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


src/Dropdown/README.md
```diff
@@ -47,6 +47,31 @@ notes: |
 </Dropdown>
 ```
 
+### With IconButton
+
+You can use `Dropdown.Toggle` with `IconButton` component, but be aware that by default it renders an arrow which messes up the positioning of the `IconButton` element. Paragon offers a helper CSS class `pgn__dropdown-iconbutton` which removes that arrow in case you don't need it, you should pass it as a `bsPrefix` prop, so that it overrides base react-bootstrap's class which adds the arrow. Also note that all props you provide to `Dropdown.Toggle` will get passed down to the `IconButton`.
```
adamstankiewicz, line 52: Hmm, when the arrow is rendered with the `IconButton`, the positioning is a bit off per the screenshots below. I'm thinking instead of fixing that we just don't show the arrow altogether using an `IconButton` with the `as` prop. 

I'm not sure there's a use case where we'd want the default dropdown arrow to show when using an `IconButton` as the dropdown target, since the target itself is already an icon.

Could we add the `pgn__dropdown-toggle-iconbutton` class by default when the `IconButton` is passed to `as`?

image
image

src/IconButton/index.jsx
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
-const IconButton = ({
+const IconButton = React.forwardRef(({
```
adamstankiewicz, line 7: 🔥

src/Dropdown/README.md
viktorrusakov: Agree, thank you! I doubt someone will need to use that arrow with an `IconBUtton`, removed it.

adamstankiewicz (approved):

edx-webhook:
@viktorrusakov 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/paragon&entry.867055334=https://github.com/edx/paragon/pull/939&entry.1484655318=https://github.com/viktorrusakov&entry.752974735=2021-12-14+18:05&entry.1917517419=2021-12-17+13:00&entry.2133058324=Yes) so we can improve your experience in the future.

PR #942, opened 2021-12-15
[PAR-473] Pass through components from react-responsive and remove custom Responsive components

bseverino:
### BREAKING CHANGE
This removes custom Responsive components in favor of passing through components from react-responsive directly.

Components that have been removed:
- `ExtraSmall`
- `Small`
- `Medium`
- `Large`
- `ExtraLarge`
- `ExtraExtraLarge`
- `LargerThanExtraSmall`

New passthrough components/hooks:
- `MediaQuery`
- `useMediaQuery`
- `ResponsiveContext`

### Supporting information
- https://openedx.atlassian.net/browse/PAR-473
- https://github.com/openedx/frontend-wg/issues/45
- [react-responsive documentation](https://www.npmjs.com/package/react-responsive)


docs/decisions/0014-removal-of-custom-responsive-components.rst
```diff
@@ -0,0 +1,49 @@
...
+
+- ExtraSmall
+- Small
+- Medium
+- Large
+- ExtraLarge
+- ExtraExtraLarge
+- LargerThanExtraSmall
+
+We will pass through the following components from react-responsive:
```
adamstankiewicz, line 37: nit: `useMediaQuery` and `ResponsiveContext` aren't really components, as they are a hook and context respectively. Maybe this line instead says, "...the following components _and functionality_ from react-responsive:"?

docs/decisions/0014-removal-of-custom-responsive-components.rst
```diff
@@ -0,0 +1,49 @@
+14. Removal of Custom Responsive Components
```
adamstankiewicz, line 1: 🔥  Thanks for the ADR!

src/Responsive/index.jsx
```diff
@@ -1,9 +1,12 @@
-import React from 'react';
-import Responsive from 'react-responsive';
+export {
```
adamstankiewicz, line 1: Rather than having these exported from `Responsive/index.jsx`, I might lean towards having `src/index.js` be responsible for the exports from `react-responsive`.

That would leave just the `breakpoints` in this file, at which point these could probably be moved to its own standalone file `src/utils/breakpoints.js`.

The Paragon components that make use of react-responsive could then still get their responsive imports directly from `react-responsive` itself, which is likely a bit more clear and understandable.

Curious to get to get your take, of course! :)

bseverino: I think that structure makes a lot more sense - thanks!

adamstankiewicz (approved):

PR #972, opened 2021-12-17
fix: remove default outline on btn-icon-variant mixin

adamstankiewicz:
Removes the default blue outline on the `IconButton` component when focused since it has its own custom focus styles.

**Before:**
image

**After:**
image

https://deploy-preview-972--paragon-edx.netlify.app/components/iconbutton/#basic-usage-with-paragon-icon

### DIFF
--- a/src/IconButton/_IconButton.scss
+++ b/src/IconButton/_IconButton.scss
@@ -33,6 +33,7 @@ $btn-icon-accent-color: white !default;
 
     &.focus,
     &:focus {
+        outline: 0;
         background-color: $btn-icon-bg;
         color: $icon-color;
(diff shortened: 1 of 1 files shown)


binodpant (approved):
that was simple! thanks

src/IconButton/_IconButton.scss
```diff
@@ -33,6 +33,7 @@ $btn-icon-accent-color: white !default;
 
     &.focus,
     &:focus {
+        outline: 0;
```
wittjeff, line 36: Please change to 2px transparent; this will show up in Windows High Contrast mode, when box-shadow treatments are automatically removed.

PR #975, opened 2021-12-17
"Show code example" as button link, resolve errors and add spacing to docs

adamstankiewicz:
* Update the "Show code example" collapsible to be a button link
* Resolve errors when using now-deleted responsive components
* Add some spacing between docs examples
* Update `IconButton` focus outline for High Contrast mode support

image


manny-m (approved):
👍

PR #976, opened 2021-12-17
fix: native a11y support for Spinner and improve docs

adamstankiewicz:
Paragon ticket: https://openedx.atlassian.net/browse/PAR-457

* Added optional `screenReaderText` prop to provide accessible screenreader-only text and a `role="status"`.
* Added additional (pre-existing) examples to the `Spinner` docs.

PR previews:
1. [brand-edx.org](https://deploy-preview-976--paragon-edx.netlify.app/components/spinner/)
2. [core](https://deploy-preview-976--paragon-openedx.netlify.app/components/spinner/)

### DIFF
--- a/www/src/components/CodeBlock.jsx
+++ b/www/src/components/CodeBlock.jsx
@@ -1,4 +1,9 @@
-import React, { useCallback, useState, useMemo } from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useState,
+  useMemo,
+} from 'react';
 import PropTypes from 'prop-types';
 import Highlight, { defaultProps } from 'prism-react-renderer';
 import theme from 'prism-react-renderer/themes/duotoneDark';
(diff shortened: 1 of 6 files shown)


www/src/components/CodeBlock.jsx
```diff
@@ -49,6 +54,7 @@ function CodeBlock({ children, className, live }) {
             ...ParagonIcons,
             ...ParagonReact,
             useCallback,
+            useEffect,
```
manny-m, line 57: Does this just add `useEffect` to things we can use in the doc site codeblocks?

manny-m (approved):
Cool changes 🚀

www/src/components/CodeBlock.jsx
adamstankiewicz: Exactly 👍  I had an example using useEffect locally to test these changes with Voiceover (show spinner on click of a button), and just left this in there because figured it could be useful for someone using the live code editors on the documentation site.

PR #981, opened 2021-12-20
[BD-46] docs: add ADR for i18n support 

viktorrusakov:
Add ADR for i18n support.

JIRA: [PAR_412](https://openedx.atlassian.net/browse/PAR-412)

### DIFF
--- a/docs/decisions/0015-i18n-support.rst
+++ b/docs/decisions/0015-i18n-support.rst
@@ -1,0 +1,46 @@
+15. Internationalization support.
+---------------------------------
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+Paragon's components are extensively used in applications that support i18n.
+Currently consumers of Paragon need to handle i18n themselves, i.e. pass translated
+strings to components as props. Furthermore, some of the components have hardcoded strings in them which do not
+allow translation in any way which can cause a lot of headache for consumers.
+
+Decision
+--------
+
+Paragon will natively support internationalization. We will use `react-intl <https://formatjs.io/docs/react-intl/>`_ library
+to translate all string type props and move all hardcoded strings from components to respective props.
+We will setup Transifex pipeline to auto pull new translations.
+All translated messages will be exported from Paragon to allow consumers to use them throughout their apps.
+
+Consequences
+------------
+
+- This will be a breaking change for all consumers who do not support i18n in their apps since react-intl requires a specific context to be used in the app. All apps (even if they don't intend use i18n) that use Paragon's components **must** use `IntlProvider` component to provide i18n context to Paragon's components, most common and recommended usage is to wrap root component of the app:
+
+.. code-block:: javascript
+
+  ReactDOM.render(
+    <IntlProvider locale={usersLocale} messages={translationsForUsersLocale}>
+      <App />
+    </IntlProvider>,
+    document.getElementById('root')
+  )
+
+- Consumers who support i18n will no longer need to pass translated strings to Paragon's components. They will either use translations offered by Paragon or override them which will greatly reduce the time needed to add i18n to their apps.
+- Additional documentation will be created both for consumers (to explain how to enable or override Paragon's translations) and developers of Paragon (to explain how to properly enable i18n support for new components).
+
+Questions
+---------
+
+1. Which languages should we support?
+2. Should we use `wrapper for react-intl <https://github.com/edx/frontend-platform/blob/master/src/i18n/index.js>`_ created in frontend-platform or features added in the wrapper are not relevant for Paragon?
--- a/docs/decisions/0015-i18n-support.rst
+++ b/docs/decisions/0015-i18n-support.rst
@@ -10,8 +10,8 @@ Context
 -------
 
 Paragon's components are extensively used in applications that support i18n.
-Currently consumers of Paragon need to handle i18n themselves, i.e. pass translated
-strings to components as props. Furthermore, some of the components have hardcoded strings in them which do not
+Currently consumers of Paragon need to handle i18n themselves (e.g., pass translated strings to components as props,
+or pass components like ``FormattedMessage`` from ``react-intl`` as nodes). Furthermore, some of the components have hardcoded strings in them which do not
 allow translation in any way which can cause a lot of headache for consumers.
 
 Decision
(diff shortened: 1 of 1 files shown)


edx-webhook:
Thanks for the pull request, @viktorrusakov! I've created [BLENDED-1050](https://openedx.atlassian.net/browse/BLENDED-1050) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


docs/decisions/0015-i18n-support.rst
```diff
@@ -0,0 +1,46 @@
...
+------
+
+Accepted
+
+Context
+-------
+
+Paragon's components are extensively used in applications that support i18n.
+Currently consumers of Paragon need to handle i18n themselves, i.e. pass translated
+strings to components as props. Furthermore, some of the components have hardcoded strings in them which do not
```
adamstankiewicz, line 14: Might change to be (e.g., pass translated strings to components as props, or pass components like `FormattedMessage` from react-intl as nodes).

docs/decisions/0015-i18n-support.rst
```diff
@@ -0,0 +1,46 @@
...
+
+Paragon will natively support internationalization. We will use `react-intl <https://formatjs.io/docs/react-intl/>`_ library
+to translate all string type props and move all hardcoded strings from components to respective props.
+We will setup Transifex pipeline to auto pull new translations.
+All translated messages will be exported from Paragon to allow consumers to use them throughout their apps.
+
+Consequences
+------------
+
+- This will be a breaking change for all consumers who do not support i18n in their apps since react-intl requires a specific context to be used in the app. All apps (even if they don't intend use i18n) that use Paragon's components **must** use `IntlProvider` component to provide i18n context to Paragon's components, most common and recommended usage is to wrap root component of the app:
```
adamstankiewicz, line 28: Might be good to talk about how this would be used in conjunction with `frontend-platform/i18n` as well. For example, most MFEs that consume Paragon and frontend-platform, use `frontend-platform`'s `AppProvider` component, which renders the `IntlProvider` for you. Will this be compatible?

docs/decisions/0015-i18n-support.rst
```diff
@@ -0,0 +1,46 @@
...
+  )
+
+- Consumers who support i18n will no longer need to pass translated strings to Paragon's components. They will either use translations offered by Paragon or override them which will greatly reduce the time needed to add i18n to their apps.
+- Additional documentation will be created both for consumers (to explain how to enable or override Paragon's translations) and developers of Paragon (to explain how to properly enable i18n support for new components).
+
+Questions
+---------
+
+1. Which languages should we support?
+2. Should we use `wrapper for react-intl <https://github.com/edx/frontend-platform/blob/master/src/i18n/index.js>`_ created in frontend-platform or features added in the wrapper are not relevant for Paragon?
```
adamstankiewicz, line 46: Ah, I see. You did note usage with frontend-platform here. If a consuming MFE is using frontend-platform, they will generally be using `AppProvider` (and its `IntlProvider` component) vs. implementing their own top-level `IntlProvider`. If not using frontend-platform though, consumers of Paragon would implement their own top-level `IntlProvider`. Let me know if this approach seems like it will cause any issues.

viktorrusakov:
So many new checks 😮

docs/decisions/0015-i18n-support.rst
viktorrusakov: This sound better, updated, thanks!

docs/decisions/0015-i18n-support.rst
viktorrusakov: It shouldn't cause issues as long as we use the same version of `react-intl` that is used in `frontend-platform`, I've tested it locally with `frontend-app-learning`. Unfortunately, if we upgrade `react-intl` to newer version it doesn't see the context from `AppProvider`

I've updated the ADR, please take a look

davidjoy:
This seems sensible - there are a few things I'd want to sanity check:

1. That the same react-intl version and code in node_modules is being bundled into the host MFE and that we're not increasing our bundle size. (I think with tree shaking and having Paragon built preserving import statements, this will be fine)
2. That the same _instance_ of react-intl, through frontend-platform, is being used at runtime.  To my knowledge, weird inconsistent stuff can happen if there are more than one instance of react-intl initialized into the page.  

Assuming the impact here is that nearly all MFEs will just use `AppProvider` and pass the paragon messages into the app via `initialize`, that seems like a fine API/footprint for this change.

viktorrusakov:
Good questions, thanks for the feedback!

> 1. That the same react-intl version and code in node_modules is being bundled into the host MFE and that we're not increasing our bundle size. (I think with tree shaking and having Paragon built preserving import statements, this will be fine)

I believe keeping `react-intl` version in sync with frontend-platform's one makes it work correctly, I've tested the build with frontend-app-payment and having `react-intl` as a dependancy in Paragon does not increase the bundle size (tested the build with prod config).

> 2. That the same _instance_ of react-intl, through frontend-platform, is being used at runtime.  To my knowledge, weird inconsistent stuff can happen if there are more than one instance of react-intl initialized into the page.

Hmm, I'm not sure that can happen since we will have only one version of `react-intl` in the bundle, so it should be one instance if I understand correctly. I've added translations for some components and they worked correctly inside frontend-app-learning when testing it locally, I haven't noticed any issues. Would you like me to check something specific?

adamstankiewicz (approved):

edx-webhook:
@viktorrusakov 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/981&entry.1484655318=https://github.com/viktorrusakov&entry.752974735=2021-12-20+11:58&entry.1917517419=2022-02-11+12:24&entry.2133058324=Yes) so we can improve your experience in the future.

PR #983, opened 2021-12-20
docs: link to github code example of component usages

adamstankiewicz:
[PAR-394](https://openedx.atlassian.net/browse/PAR-394)

Provide links to specific code examples of usages of Paragon components.

**Demo:** https://deploy-preview-983--paragon-edx.netlify.app/insights

**Known bugs**
* Hyperlinks to monorepo packages (e.g., `catalog-search`) will lead to a 404 on Github since they are linking to the root of the repository instead of the relative package path.
* Some repositories don't have a repository url specified, so there are no hyperlinks for their code usages.
* `edx/frontend-app-ora-enhanced-staff-grader` as a Github repository doesn't exist (the actual repository name is different), so links to it will lead to a 404 on Github.

### Example

image

### DIFF
--- a/www/src/components/insights/ProjectNameWithComponentUsageModal.jsx
+++ b/www/src/components/insights/ProjectNameWithComponentUsageModal.jsx
@@ -1,0 +1,77 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  Button,
+  StandardModal,
+  Hyperlink,
+  useToggle,
+} from '~paragon-react';
+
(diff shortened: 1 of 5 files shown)


muselesscreator (approved):
LGTM, though I would like for the www components to get tested at some point, as they  feel like an important part of the paragon product.

adamstankiewicz:
> LGTM, though I would like for the www components to get tested at some point, as they feel like an important part of the paragon product.

Agreed we should add testing to the `www` components, including these new ones. I will backlog a ticket for this, thanks.

PR #988, opened 2021-12-21
feat: add dataViewToggle within DataTable PAR-577

binodpant:
Use an IconButtonToggle as a controlled component, which allows switching between multiple views, CardView and Table, using a state variable as desired by the user

There is one known bug here I could not figure out at first try, where filters if used, are cleared when switching view. Since it does not block Explore Catalog usage I am moving forward with feedback from Alena, Mariana and Adam S. But I created https://openedx.atlassian.net/browse/PAR-615 to track and fix that bug

Usage is seen in the DataTable demo page: https://deploy-preview-988--paragon-openedx.netlify.app/components/datatable/#view-switching

Did not add screenshots here since the preview page int he 'Checks' section has a live preview!

### Left placement

Screen Shot 2022-01-05 at 2 01 53 PM

### bottom placement

Screen Shot 2022-01-05 at 2 01 13 PM

### DIFF
--- a/src/DataTable/DataTable.scss
+++ b/src/DataTable/DataTable.scss
@@ -138,7 +138,7 @@ $data-table-pagination-dropdown-min-width: 6rem !default;
   justify-content: space-between;
 
   .pgn__data-table-status-left {
-    align-self: center;
+    align-self: end;
   }
 }
 
--- a/src/DataTable/DataTable.scss
+++ b/src/DataTable/DataTable.scss
@@ -28,8 +28,22 @@ $data-table-pagination-dropdown-min-width: 6rem !default;
   display: flex;
 }
 
+.pgn__data-table-dataview-toggle {
+  flex: auto;
+}
+
 .pgn__data-table-actions-right {
   display: flex;
+}
+
+.pgn__data-table-actions-right-toggle-bottom {
+  display: flex;
+  flex-direction: column-reverse;
+  align-items: flex-end;
+
+  .pgn__data-table-dataview-toggle {
+    margin-top: map_get($spacers, 4);
+  }
 }
 
 .pgn__data-table-container {
--- a/src/DataTable/DataTable.scss
+++ b/src/DataTable/DataTable.scss
@@ -28,6 +28,10 @@ $data-table-pagination-dropdown-min-width: 6rem !default;
   display: flex;
 }
 
+.pgn__data-table-actions-right {
+  display: flex;
+}
+
 .pgn__data-table-container {
   width: 100%;
   display: block;
--- a/src/DataTable/DataTable.scss
+++ b/src/DataTable/DataTable.scss
@@ -100,9 +100,6 @@ $data-table-pagination-dropdown-min-width: 6rem !default;
 
 .pgn__data-table-status-bar {
   padding: $data-table-padding-y $data-table-padding-x;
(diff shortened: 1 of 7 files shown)


src/DataTable/DataViewToggle.jsx
```diff
@@ -0,0 +1,36 @@
...
+      defaultActiveStateValue,
+    },
+  } = useContext(DataTableContext);
+
+  if (!isDataViewToggleEnabled) { return null; }
+
+  const [activeValue, setActiveValue] = useState(defaultActiveStateValue || 'card');
+  const handleOnChange = value => {
+    setActiveValue(value);
+    onDataViewToggle(value);
```
adamstankiewicz, line 21: nit: Since this is coming in via an object now, should we be defensive around when `onDataViewToggle` is not provided, e.g.:

```jsx
if (onDataViewToggle) {
  onDataViewToggle(value);
}
```

binodpant: hmm indeed, great catch

src/DataTable/DataTable.scss
```diff
@@ -24,8 +23,27 @@ $data-table-pagination-dropdown-min-width: 6rem !default;
...
+
+.pgn__data-table-dataview-toggle {
+  flex: auto;
+}
+
+.pgn__data-table-actions-right {
+  display: flex;
+}
+
+.pgn__data-table-actions-right-toggle-bottom {
```
binodpant, line 39: when applied, this class pushes the iconbutontoggle to the bottom row of the actions area, only when a user asks for it using a prop

src/DataTable/TableControlBar.jsx
```diff
@@ -6,13 +6,21 @@ import SmartStatus from './SmartStatus';
...
-}) => {
-  const { setFilter, showFiltersInSidebar } = useContext(DataTableContext);
+const TableControlBar = ({ className }) => {
+  const {
+    setFilter,
+    showFiltersInSidebar,
+    dataViewToggleOptions: { togglePlacement },
+  } = useContext(DataTableContext);
 
+  const invalidtogglePlacement = !togglePlacement || !(['left', 'bottom']).includes(togglePlacement);
```
adamstankiewicz, line 19: nit: `invalidtogglePlacement` ➡️  `invalidTogglePlacement`

src/DataTable/TableControlBar.jsx
```diff
@@ -21,7 +29,8 @@ const TableControlBar = ({
           <div className="pgn__data-table-actions-left">
             <DropdownFilters />
           </div>
-          <div className="pgn__data-table-actions-right">
+          <div className={actionsSectionClassName}>
```
adamstankiewicz, line 32: To follow-up from our Slack huddle, we'll want to address the positioning of the view toggle component such that the view toggle component is vertically centered with the "Showing X of Y" and/or "1 selected, Select all 7, Clear selection" text when `togglePlacement="bottom"`.

We may want to move the view toggle component outside of the `div` intended for the table actions?

image

binodpant: agreed, I think moving it out of actions makes sense too in general, plus it can help with the above layout fix

adamstankiewicz (approved):

PR #990, opened 2021-12-22
[BD-46] feat: add ability to expand rows in DataTable

viktorrusakov:
Add ability to add expandable rows to the `DataTable`

note that we probably need to support asynchronous actions once a row has expanded, but this PR currently does not include this feature 

JIRA: [PAR-568](https://openedx.atlassian.net/browse/PAR-568)


edx-webhook:
Thanks for the pull request, @viktorrusakov! I've created [BLENDED-1054](https://openedx.atlassian.net/browse/BLENDED-1054) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


src/DataTable/README.md
```diff
@@ -707,4 +707,220 @@ For a more desktop friendly view, you can move filters into a sidebar by providi
...
+```jsx live
+  <DataTable 
+    showExpandAll
+    itemCount={7}
+    data={[
+      {
+        name: 'Lil Bub',
+        color: 'brown tabby',
+        famous_for: 'weird tongue', 
+        subcomponent: "This is just a string"
```
adamstankiewicz, line 726: The `data` prop generally represents the set of data to display in the table, and often can be thought of a list of results from an API response. That said, it doesn't quite make sense to incorporate `subcomponent` into the `data` prop since `data` is generally dynamic.

Rather, I think we should try to incorporate more of the pattern laid out in the `react-table` documentation for expanding subcomponents, where all rows would use the same subcomponent, but with row data passed in by `react-table`. See this[ documentation example](https://react-table.tanstack.com/docs/examples/sub-components).

e.g.,

```jsx
/**
 * example custom sub component for use on all expandable rows.
 * note that it has access to the `row` data so it could display
 * additional data from the row or make any additional API call(s) using the row data.
 */
const renderRowSubComponent = ({ row }) => (
  <pre>
    <code>{JSON.stringify({ values: row.values }, null, 2)}</code>
  </pre>
);
```

```jsx
/**
 * example subcomponent exported from `DataTable` for use by consumers.
 */
export const ExpandRow = ({ row }) => {
  return (
    <span {...row.getToggleRowExpandedProps()}>
      {row.isExpanded ? '-' : '+'}
    </span>
  );
};

DataTable.ExpandRow = ExpandRow;
```

```jsx
/**
 * example usage of DataTable with expanding subcomponents
 */
<DataTable
  isExpandable
  renderRowSubComponent={renderRowSubComponent}
  columns={[
    {
      id: 'expander',
      Header: () => null, // or optionally a custom/exported "Expand all" subcomponent
      Cell: DataTable.ExpandRow,
    },
    {
      id: 'name',
      Header: 'Name',
      accessor: 'name',
    },
  ]}
  data={[
    {
      name: 'Lil Bub',
      color: 'brown tabby',
      famous_for: 'weird tongue', 
    },
    {
      name: 'Grumpy Cat',
      color: 'siamese',
      famous_for: 'serving moods',
    },
  ]}
/>
```

Then, in `DataTable's` `TableRow` component, an additional `tr` could be rendered when `row.isExpanded` is true, calling `renderRowSubComponent` which passed the `row` argument.

src/DataTable/TableRow.jsx
```diff
@@ -5,21 +5,32 @@ import classNames from 'classnames';
...
+        'pgn__data-table-row': true,
+        'is-selected': isSelected,
+      }),
+    })}
+    >
+      {cells.map(cell => <TableCell {...cell} key={`${cell.column.Header}${id}`} />)}
+    </tr>
+    {isExpanded && subcomponent ? (
+      <tr>
+        <td colSpan="100">
```
adamstankiewicz, line 22: Per the react-table docs, I think the `colSpan` may make more sense as `colSpan={visibleColumns.length}` so that the column is spanning the correct number of columns rather than an arbitrary large number.

`visibleColumns` should be accessible via the `react-table` instance.

adamstankiewicz (commented):
This is a great start. I left a suggestion on how we might change the props API for this to more in line with the `react-table` documentation. Let me know if you'd like any additional clarification on it.

viktorrusakov:
Thanks for the feedback! I I've refactored the API and it does seem that it is a much better approach, so please take a look.

src/DataTable/ExpandRow.jsx
```diff
@@ -0,0 +1,16 @@
...
+
+const ExpandRow = ({ row }) => (
+  <span {...row.getToggleRowExpandedProps()}>
+    {row.isExpanded ? <Remove /> : <Add />}
+  </span>
+);
+
+ExpandRow.propTypes = {
+  /** Row data that is received from `react-table` API. */
+  row: PropTypes.object.isRequired, // eslint-disable-line
```
adamstankiewicz, line 13: nit: typically we'd write this as `PropTypes.shape({})` to define an object prop type without really caring what's in it.

adamstankiewicz (changes requested):
This looks great! I just have one comment to address related to the icons used in `ExpandRow` and another small nit. Beyond that, this looks ready to go.

src/DataTable/ExpandRow.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Add, Remove } from '../../icons';
+
+const ExpandRow = ({ row }) => (
+  <span {...row.getToggleRowExpandedProps()}>
+    {row.isExpanded ? <Remove /> : <Add />}
```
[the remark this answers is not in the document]
viktorrusakov: Agree, it does look better with `IconButton` here, thanks! Also, updated the icons to be `ExpandMore` and `ExpandLess`

src/DataTable/ExpandRow.jsx
viktorrusakov: Updated, thanks! For some reason I forgot `PropTypes.shape` existed 😅

adamstankiewicz (approved):

edx-webhook:
@viktorrusakov 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/paragon&entry.867055334=https://github.com/edx/paragon/pull/990&entry.1484655318=https://github.com/viktorrusakov&entry.752974735=2021-12-22+14:57&entry.1917517419=2022-01-11+15:30&entry.2133058324=Yes) so we can improve your experience in the future.

PR #991, opened 2021-12-22
[BD-46] feat: add a loading state to the datatable

vlasovmichael:
Add a loading state for the table that shows a loader spinner

Preview: https://deploy-preview-991--paragon-edx.netlify.app/

JIRA: https://openedx.atlassian.net/browse/PAR-322

### DIFF
--- a/src/DataTable/Table.jsx
+++ b/src/DataTable/Table.jsx
@@ -1,9 +1,11 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import Spinner from 'react-bootstrap/Spinner';
 import TableHeaderRow from './TableHeaderRow';
 import TableRow from './TableRow';
 import { useRows } from './hooks';
+import DataTableContext from './DataTableContext';
 
 const Table = ({
   isStriped,
@@ -11,6 +13,7 @@ const Table = ({
   const {
     getTableProps, prepareRow, displayRows, headerGroups, getTableBodyProps,
   } = useRows();
+  const { isLoading } = useContext(DataTableContext);
 
   const renderRows = () => displayRows.map((row) => {
     prepareRow(row);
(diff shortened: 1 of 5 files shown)


edx-webhook:
Thanks for the pull request, @vlasovmichael! I've created [BLENDED-1055](https://openedx.atlassian.net/browse/BLENDED-1055) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


src/DataTable/Table.jsx
```diff
@@ -1,16 +1,19 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import Spinner from 'react-bootstrap/Spinner';
```
adamstankiewicz, line 4: nit: Let's use the `Spinner` component exported from Paragon itself, instead of the straight pass through from `react-bootstrap`. The benefit is that we could more easily pass in a `screenReaderText` prop for a11y.

adamstankiewicz (commented):
This looks great! I just have one minor comments about which `Spinner` component is being used here.

src/DataTable/Table.jsx
vlasovmichael: Fixed, thanks. Adam, one question. Is it ok if `screenreadertext ` is used in lowercase, because if use uppercase, like `screenReaderText ` tests are failed?

adamstankiewicz: @vlasovmichael What about the tests fail with `screenReaderText` as the prop?

Passing `screenreadertext` in all lowercase would likely try to add `screenreadertext` as an attribute to the DOM node, which isn't right, or would otherwise be unhandled.

`screenReaderText` (camelCase) prop is a [valid prop for Paragon's `Spinner` component](https://github.com/edx/paragon/blob/master/src/Spinner/index.jsx#L8) so I'm unclear what about the tests would be failing here 🤔

adamstankiewicz (approved):

src/DataTable/Table.jsx
vlasovmichael: Fixed, thanks!

edx-webhook:
@vlasovmichael 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/991&entry.1484655318=https://github.com/vlasovmichael&entry.752974735=2021-12-22+19:39&entry.1917517419=2022-01-13+13:04&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1020, opened 2022-01-11
fix: improve spacing a bit with IconButtonToggle in DataTable

adamstankiewicz:


src/DataTable/CollapsibleButtonGroup.jsx
```diff
@@ -92,7 +92,7 @@ const CollapsibleButtonGroup = ({
         return (
           <Button
             variant={variant}
-            className={classNames('ml-2', action.className)}
+            className={classNames('pgn__data-table__action-btn', action.className)}
```
adamstankiewicz, line 95: Avoids using a CSS utility class, which forces use of `!important` and makes it difficult to theme the component.

long74100 (approved):

PR #1021, opened 2022-01-11
docs: use expanding table rows for example code links

adamstankiewicz:
This PR moves component example links from a modal to an expandable table row, which is a little more efficient to use quickly.

### Collapsed
image

### Expanded
image


long74100 (approved):
lgtm!

PR #1039, opened 2022-01-19
[BD-46] feat: add theme switcher to documentation site

viktorrusakov:
- add theme switcher to the docs site, currently two themes are included: openedx theme and edx.org theme
- This PR also allows to add new theme quite easily, see README changes
- after the merge only one deploy-preview will be required instead of two for each theme

JIRA: [PAR-626](https://openedx.atlassian.net/browse/PAR-626)


edx-webhook:
Thanks for the pull request, @viktorrusakov! I've created [BLENDED-1069](https://openedx.atlassian.net/browse/BLENDED-1069) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


src/Form/_Form.scss
```diff
@@ -242,21 +242,22 @@ $select-icon-padding: .5625rem !default;
...
-    appearance: none;
-  }
   select.form-control:not(.has-value):not(:focus) {
     // color: rgba(0,0,0,0); Force the rgba syntax to appear in the output rather
     // than transparent. IE11 does not understand color: transparent here.
     color: unquote("rgba(0,0,0,0)");
   }
 }
 
+select.form-control {
```
viktorrusakov, line 252: this is done to fix arrow for selects without floating label as it is currently a bit off
image
this was causing issues for theme switcher select in the header

this change adjusts arrow to be as follows
image
which is how it looks for selects with floating label

see https://openedx.github.io/paragon/components/form/ for examples

www/README.md
```diff
@@ -18,18 +18,43 @@ npm start
...
 
-## Developing with a theme applied
+## Including a custom theme
 
-1. Run
-```sh
-npm run develop:with-theme
-```
+1. Create a separate SASS stylesheet with your theme in `src/scss`, e.g. `src/scss/my-theme.scss`.
+2. Add your theme to `theme-config.js`, e.g.
```
monteri, line 24: Maybe there is a reason to add a path to the `theme-config.js` to prevent confusions? Something like `Add your theme to `theme-config.js` (project_root/www/theme-config.js)`. What do you think?

viktorrusakov: Hmm, I actually think it's fine right now, because this README is located in `www` already and I think it's clear that `theme-config.js` is located in the same directory

adamstankiewicz:
@viktorrusakov Coming along nicely! The two things I think are worth addressing:

1. Flash of unstyled content. When flipping themes, it flashes as if no CSS were applied. Is there a moment during rendering no theme is applied? 🤔 
2. Design of the theme selector. Space is a bit tight in the header with the select form control, especially on narrow screen sizes. I'm thinking we might draw inspiration from [Material UI](https://mui.com/getting-started/installation/) by adding a "Settings" icon to the header that opens a right-aligned `Sheet` containing the theme selector. Material UI also puts in a global language selector there, too which feels like we could also do for Paragon's i18n initiative and even a toggle between left-to-right to right-to-left text direction. Thoughts?

Installation---MUI

Something along the lines of:

image

Of course, the "Direction" and "Language" options would come later. I will add a new milestone for adding the ability to change the text direction for future work as well. Both `frontend-build` and `frontend-platform` have checks for right-to-left support. I'm not sure if Paragon's documentation site has any support for it currently, so I'll include a discovery ticket as well to figure out if Paragon docs site will even support RTL today.

www/README.md
adamstankiewicz: Agreed, probably fine as is.

src/Form/_Form.scss
adamstankiewicz: Great catch! 🙌

www/build-themes.js
```diff
@@ -0,0 +1,43 @@
+const path = require('path');
+const sass = require('node-sass');
```
adamstankiewicz, line 2: FYI, now libsass was replaced with dart sass, `node-sass` can be changed to `sass` here.

www/theme-config.js
```diff
@@ -0,0 +1,16 @@
+// List of themes that are available for usage across the docs site
+// Each theme is represented by object with following attributes:
+//  1. label - Label of the theme which will be shown in the selector in the header
+//  2. Name of the theme's stylesheet (without extension), it's assumed that stylesheet
+//     is located in 'src/scss/' directory. The stylesheet will be compiled into CSS
+//     and placed into ./public/static/ directory under <stylesheet>.css name
+exports.THEMES = [
+  {
+    label: 'Openedx theme',
```
adamstankiewicz, line 9: nit: "Open edX"

www/theme-config.js
```diff
@@ -0,0 +1,16 @@
...
+    stylesheet: 'openedx-theme',
+  },
+  {
+    label: 'Edx.org theme',
```
adamstankiewicz, line 13: nit: `edX.org"

component-generator/templates/test.jsx
```diff
@@ -1,10 +1,10 @@
 import React from 'react';
 import renderer from 'react-test-renderer';
-import ComponentName from './index';
+import componentName from './index';
```
viktorrusakov, line 3: reverting the changes done in one of the other PRs, this actually caused component generator to generate incorrect tests, it's important that `componentName` is in camelCase

www/src/context/SettingsContext.jsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <SettingsContext.Provider value={contextValue}>
+      <Helmet>
+        {/* Open edX theme is the base and default theme which should always be included and active in the <head>.
+            Other themes generally inherit from it and override its values, so if they are included after the base
+            one they wil get applied to the site. This is done to avoid flickering when changing themes,
+            if you simply change href of the stylesheet there is a small window of time when the previous
+            theme gets unapplied and new one loaded which leaves whose site without styles.
+         */}
+        <link
```
viktorrusakov, line 34: I'm really not sure if this is a good approach... but it does remove the problem with showing unstyled content. Tried bunch of other stuff but could't make it work

www/build-themes.js
viktorrusakov: True, updated it, thanks!

viktorrusakov:
@adamstankiewicz I do like the idea of adding a `Settings` component! This should be much better than having everything in the header. I'm not sure though where to put the `Settings` icon 🤔 , please take a look if I chose a good place!

www/src/context/SettingsContext.jsx
adamstankiewicz: Yeah, I'm also not sure of a better approach. I think this should work well enough, though :)

adamstankiewicz:
> I'm not sure though where to put the Settings icon 🤔 , please take a look if I chose a good place!

I think where you put it makes sense for now! Let's roll with it.

adamstankiewicz (approved):

PR #1054, opened 2022-01-27
[BD-46] feat: use localStorage to persist theme choice on docs site

viktorrusakov:
store theme choice in localStorage so theme does not get reset after page refresh

### DIFF
--- a/www/src/context/SettingsContext.jsx
+++ b/www/src/context/SettingsContext.jsx
@@ -1,4 +1,4 @@
-import React, { createContext, useState } from 'react';
+import React, { createContext, useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Helmet } from 'react-helmet';
 import { THEMES } from '../../theme-config';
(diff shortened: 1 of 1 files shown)


edx-webhook:
Thanks for the pull request, @viktorrusakov! I've created [BLENDED-1077](https://openedx.atlassian.net/browse/BLENDED-1077) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


www/src/context/SettingsContext.jsx
```diff
@@ -11,12 +11,27 @@ const defaultValue = {
 const SettingsContext = createContext(defaultValue);
 
 const SettingsContextProvider = ({ children }) => {
+  // gatsby does not have access to the localStorage during the build (and first render)
+  // so sadly we cannot initialize theme with value from localStorage
```
adamstankiewicz, line 15: The result of this tradeoff is the default Open edX theme will flash briefly before the edX.org theme is applied. I think this should be OK.

Though, I wonder if cookies would help get around this? Probably not since it'd still need access to `document` after the first render, too.

www/src/context/SettingsContext.jsx
```diff
@@ -11,12 +11,27 @@ const defaultValue = {
...
   const [theme, setTheme] = useState('openedx-theme');
   const [showSettings, setShowSettings] = useState(false);
 
+  const handleThemeChange = (e) => {
+    setTheme(e.target.value);
+    localStorage.setItem('theme', e.target.value);
```
adamstankiewicz, line 21: nit: should we use `global.localStorage`?

adamstankiewicz, line 21: Maybe prefix this with `pgn__`, e.g. `pgn__theme`?

adamstankiewicz (commented):
Awesome.

www/src/context/SettingsContext.jsx
viktorrusakov: Yeah, I don't think there is a way to get around this, we can't access anything on the first render

www/src/context/SettingsContext.jsx
viktorrusakov: Sounds good, update!

viktorrusakov: Yeah, this should be better, updated, thanks!

adamstankiewicz (approved):

edx-webhook:
@viktorrusakov 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1054&entry.1484655318=https://github.com/viktorrusakov&entry.752974735=2022-01-27+07:51&entry.1917517419=2022-01-28+16:32&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1057, opened 2022-01-28
[BD-46] feat: add direction switcher to the docs site

viktorrusakov:
- add direction selector to the Settings tab
- add support for RTL direction for `Sheet` component

Note that currently Paragon doesn't have support for RTL direction for all components, not sure if we should wait for all components to support RTL before merging this...

JIRA: [PAR-647](https://openedx.atlassian.net/browse/PAR-647)


edx-webhook:
Thanks for the pull request, @viktorrusakov! I've created [BLENDED-1079](https://openedx.atlassian.net/browse/BLENDED-1079) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


src/Sheet/Sheet.scss
```diff
@@ -20,34 +31,38 @@
     color: $light-300;
   }
   &.bottom {
-    bottom: 0px;
-    box-shadow: 0px -4px 10px 0px rgba(0,0,0,0.15);
```
viktorrusakov, line 24: I believe this line gets overridden by the next one, unless I'm crazy... removed it.

adamstankiewicz (commented):
Nice, looks good! A couple additional requests:

1. I'm thinking it might be worth persisting the direction choice here as well, similar to the theme selection in localStorage?
2. I agree we should probably hold off a little bit on exposing this feature until we fix any major style issues in components. Can we add this behind a feature flag of some kind for the time being? Maybe something like[ this utility function](https://github.com/openedx/frontend-enterprise/blob/master/packages/utils/src/utils.js#L38-L48) might be an option, where you can temporarily enable a feature flag via a `?feature` query parameter. Should suffice as a simple feature flag mechanism for now. What do you think?

viktorrusakov:
@adamstankiewicz good idea with the feature flag, thanks! I've added it in the latest commit, also storing the choice in localStorage now. 

I wonder if there should be some documentation for the feature flag or is it fine as is? I guess since we already started working on RTL support we don't really need the docs since it shouldn't be long until we enable it by default, what do you think?

adamstankiewicz:
> I wonder if there should be some documentation for the feature flag or is it fine as is? I guess since we already started working on RTL support we don't really need the docs since it shouldn't be long until we enable it by default, what do you think?

Hmm, yeah. I don't think we necessarily need to document anything specific to the direction switcher feature flag but what may be helpful is a section in the README about feature flags at a high level, e.g.:

```
## Feature Flags
In some scenarios, it is helpful to put your changes to the Paragon documentation site behind a feature flag so that you have more control over how a particular feature or change is rolled out more broadly.

### Configuring feature flags
To configure feature flags for the documentation site, see the file `www/src/config.js` for instructions.

### Enabling a feature flag
A feature flag may be enabled in two ways:
1. Environment variable. This will allow you to persist the state of a feature flag at build time, with no way to disable it.
2. Query parameter. Adding `?feature=` to the URL as a query parameter, you can temporarily enable one or more features.
```

And then in `www/src/config.js`, maybe include an example in comments of how to configure a feature flag with `process.env.*` or `hasFeatureFlagEnabled`?

viktorrusakov:
@adamstankiewicz yeah, updating README should be good here, thank you 🙂  I've added your suggestion and updated `config.js` with comments about how to configure a feature flag

not sure what's going on with the coverage...

adamstankiewicz (approved):
LGTM.

adamstankiewicz:
@viktorrusakov Also confused by the coverage report on this PR. Running tests and seeing coverage locally for this branch stays the same, so I'm inclined to merge this PR and hope that coverage indeed stays the same. If it changes, we can revert and figure out why.

edx-webhook:
@viktorrusakov 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1057&entry.1484655318=https://github.com/viktorrusakov&entry.752974735=2022-01-28+08:02&entry.1917517419=2022-02-04+12:53&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1066, opened 2022-02-01
docs: recalculate MeasuredItem measurements on theme change

adamstankiewicz:
Currently, the [color foundations page](https://paragon-openedx.netlify.app/foundations/colors/) does not update its hex color values upon theme change. This PR addresses this issue by:

* Adding openedx theme stylesheet to HTML `<head>` at build time instead of runtime via `gatsby-ssr.js`. This prevents a flash of unstyled content when developing locally.
* Refactored `MeasuredItem` class component to use functional hooks.
* Recalculate measurements when the `theme` from `SettingsContext` changes, with an increased timeout to allow for enough time to load new styles in the DOM.
* Ensures we hide the hex value until its measurements are re-calculated upon page load or theme change.

**Deploy preview**

https://deploy-preview-1066--paragon-openedx.netlify.app/foundations/colors/

### DIFF
--- a/www/src/components/MeasuredItem.jsx
+++ b/www/src/components/MeasuredItem.jsx
@@ -1,43 +1,50 @@
-import React from 'react';
+import React, {
+  useContext,
+  useState,
+  useRef,
+  useEffect,
+} from 'react';
 import PropTypes from 'prop-types';
 
-export default class MeasuredItem extends React.Component {
-  constructor(props) {
-    super(props);
+import { SettingsContext } from '../context/SettingsContext';
 
-    const initialMeasurements = props.properties.reduce((acc, property) => {
-      acc[property] = null;
-      return acc;
-    }, {});
+const initialMeasurements = {};
 
-    this.state = initialMeasurements;
-    this.item = React.createRef();
-  }
+const MeasuredItem = ({
+  properties,
+  renderBefore,
+  renderAfter,
+  children,
+}) => {
+  const { theme } = useContext(SettingsContext);
+  const [measurements, setMeasurements] = useState(initialMeasurements);
+  const itemRef = useRef();
 
-  componentDidMount() {
-    // Needs a moment to render children to DOM first.
-    setTimeout(this.measure.bind(this), 10);
-  }
+  useEffect(
+    () => {
+      const measure = () => {
+        const computedStyle = getComputedStyle(itemRef.current);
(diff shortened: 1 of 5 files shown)


www/src/context/SettingsContext.jsx
```diff
@@ -46,11 +46,6 @@ const SettingsContextProvider = ({ children }) => {
             if you simply change href of the stylesheet there is a small window of time when the previous
             theme gets unapplied and new one loaded which leaves whose site without styles.
          */}
-        <link
-          href="/static/openedx-theme.css"
-          rel="stylesheet"
-          type="text/css"
-        />
```
adamstankiewicz, line 53: [inform] Rendering this here means the theme isn't present in `<head>` until after the first render of `SettingsContextProvider`, which causes a quick flash of unstyled content locally. This was instead moved to gatsby-ssr.js so the stylesheet gets added at build time instead.

long74100 (approved):

www/src/components/MeasuredItem.jsx
```diff
@@ -1,43 +1,50 @@
...
+    () => {
+      const measure = () => {
+        const computedStyle = getComputedStyle(itemRef.current);
+        const measurements = properties.reduce((acc, property) => {
+          acc[property] = computedStyle.getPropertyValue(property);
+          return acc;
+        }, {});
+        setMeasurements(measurements);
+      };
+      // Needs a moment to finish switching theme and re-render children to DOM first.
```
manny-m, line 33: Does a set timeout work on very slow browsers?

I wonder if you can work around using `setTimeout` by using another `useEffect` that handles calling `measure`. Using the cleanup feature of `useEffect` might also help? 

This feels like a super small edge case I don't feel very strongly about it.

manny-m (approved):
Cool change 🚀

www/src/components/MeasuredItem.jsx
adamstankiewicz: @manny-m If the new styles don't apply within a second, the `setTimeout` indeed will not work as expected. I tried different approaches like you mentioned, using `onload`/`onreadystatechange`, react-helmet's `onChangeClientState`... all without luck unfortunately lol.

I'm going to say the `setTimeout` will suffice for now but will file a ticket to see if we can refactor to avoid the `setTimeout` in the Paragon backlog.

PR #1068, opened 2022-02-03
feat: Added support for node v16

Jawayria:
JIRA: https://openedx.atlassian.net/browse/BOM-3260

Adds support for Node 16 and introduced some PropType changes to fix to fix eslint errors.


binodpant:
@Jawayria you can update the snapshots and update the PR as per @adamstankiewicz we are okay to update those failed snapshots in the tests

you can run `fedx-scripts jest --updateSnapshot` or sth like it to update the jest snapshots and submit the changes

adamstankiewicz (commented):
In addition to @binodpant's comment above, I added a nit comment.

Jawayria:
> @Jawayria you can update the snapshots and update the PR as per @adamstankiewicz we are okay to update those failed snapshots in the tests
> 
> you can run `fedx-scripts jest --updateSnapshot` or sth like it to update the jest snapshots and submit the changes

Updated the snapshots. Now tests are passing with Node 12 and 14 but one test failure with Node 16

binodpant:
hmm I saw this same error in the explore catalog node16 node

I looked last week and have not figured out why yet... it seems there is an error about MutationObserver, which probably has changed api and some lib is still using it?

```
node:internal/process/promises:265
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "TypeError: MutationObserver is not a constructor".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

binodpant:
maybe this is useful: https://github.com/testing-library/dom-testing-library/issues/477

binodpant:
actually for me it was due to a different reason but the same path of failures 

```
[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "Error: Warning: An update to %s inside a test was not wrapped in act(...).

When testing, code that causes React state updates should be wrapped into act(...):
```

package.json
```diff
@@ -77,16 +77,16 @@
...
     "eslint-plugin-jsx-a11y": "^6.3.1",
     "eslint-plugin-react": "^7.20.6",
-    "eslint-plugin-react-hooks": "^1.7.0",
+    "eslint-plugin-react-hooks": "^4.0.0",
     "husky": "^2.0.0",
     "identity-obj-proxy": "^3.0.0",
-    "jest": "^24.5.0",
-    "jest-cli": "^24.7.1",
+    "jest": "^25.0.0",
+    "jest-cli": "^25.0.0",
```
Jawayria, line 89: Updated jest version to fix `TypeError: MutationObserver is not a constructor"` when running tests with Node 16.
[Reference](https://github.com/testing-library/dom-testing-library/issues/477#issuecomment-617652033)

src/Button/deprecated/Button.test.jsx
```diff
@@ -11,9 +11,11 @@ describe('<Button />', () => {
   let button;
 
   beforeEach(() => {
+    const app = document.createElement('div');
+    document.body.appendChild(app);
     wrapper = mount(<Button
       {...defaultProps}
-    />);
+    />, { attachTo: app });
```
Jawayria, line 18: Jest v25 uses jsdom v15 ([link](https://github.com/facebook/jest/pull/8851)) and all the tests related to `focus` started to fail due to the changes introduced in jsdom v15 ([link](https://github.com/jsdom/jsdom/issues/2924)). So I have updated the tests and introduced `attachTo` in order to fix this issue following these references: [1](https://github.com/enzymejs/enzyme/issues/2337#issuecomment-586583502) [2](https://ttfb.test.traveloka.com/testing-react-component-using-enzyme-jest-part-3/)

mraarif (approved):

src/Button/deprecated/Button.test.jsx
binodpant: added this to troubleshooting section of doc https://openedx.atlassian.net/wiki/spaces/AC/pages/3318054984/Node+16+Upgrade

PR #1071, opened 2022-02-04
fix: remove aria-label from span and move alt text to icon

adamstankiewicz:
See the following Lighthouse a11y error regarding `Hyperlink` with external URLs:

image

This PR fixes the a11y error by removing `aria-label` from the `span` and instead simply passing its contents to the `screenReaderText` prop of `Icon` instead which has the same result.

### DIFF
--- a/src/Hyperlink/Hyperlink.test.jsx
+++ b/src/Hyperlink/Hyperlink.test.jsx
@@ -38,7 +38,7 @@ describe('correct rendering', () => {
   it('renders external Hyperlink', () => {
     const wrapper = mount(<Hyperlink {...externalLinkProps} />);
 
-    expect(wrapper.find('span')).toHaveLength(2);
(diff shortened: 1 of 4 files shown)


src/Hyperlink/Hyperlink.test.jsx
```diff
@@ -38,7 +38,7 @@ describe('correct rendering', () => {
   it('renders external Hyperlink', () => {
     const wrapper = mount(<Hyperlink {...externalLinkProps} />);
 
-    expect(wrapper.find('span')).toHaveLength(2);
+    expect(wrapper.find('span')).toHaveLength(3);
```
adamstankiewicz, line 41: There are now 3 spans, since `Icon` now renders its own `span` for the `screenReaderText`.

manny-m (approved):

PR #1072, opened 2022-02-04
[PAR-633] Add aria-selected to IconButtonToggle children

bseverino:
Update `IconButtonToggle` children to use `aria-selected=true|false` based on the current `activeValue`.

- https://github.com/openedx/frontend-wg/issues/50
- https://openedx.atlassian.net/browse/PAR-633


adamstankiewicz (approved):
Awesome, looks good to me! Thanks for picking this up.

PR #1077, opened 2022-02-09
fix: text wrapping issue for Toast

zainab-amir:
On mobile screen the text is not wrapping correctly. 

|Before|After|
|-------|-----|
Screenshot 2022-02-09 at 1 45 14 PM|Screenshot 2022-02-09 at 1 44 27 PM|

Ticket: https://openedx.atlassian.net/browse/VAN-847


adamstankiewicz (approved):
Thanks for the contribution/fix! This change looks good to me.

PR #1087, opened 2022-02-11
feat: new product tour component

mikix:
This is a new branch name for https://github.com/openedx/paragon/pull/1045 - moved so that the CLA checker knows it's me, not Carla. She was an employee when she wrote the base of this PR, so the copyright for this PR is fine, but the bot doesn't know that.


adamstankiewicz (approved):
Awesome, thanks for addressing those last few comments! I think this is good to merge; LGTM ✅  🚀  :shipit:

PR #1096, opened 2022-02-16
[BD-46] feat: implement Sticky component

monteri:
Created new Sticky component:
- `position` prop allows to set `top`/`bottom`
- `offset` prop based on `spacing` adds offset depending on `position`
- added shadows using `IntersectionObserver`
- added shadows support for the `Sticky` with `offset`

JIRA: [PAR-471](https://openedx.atlassian.net/browse/PAR-471)


edx-webhook:
Thanks for the pull request, @monteri! I've created [OSPR-6464](https://openedx.atlassian.net/browse/OSPR-6464) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

src/Sticky/Sticky.scss
```diff
@@ -0,0 +1,37 @@
...
+      &.pgn__sticky-offset--#{$level} {
+        bottom: $offset;
+      }
+    }
+
+    &.pgn__sticky-shadow {
+      box-shadow: $sticky-shadow-top;
+    }
+  }
+}
```
vlasovmichael, line 37: Please add an empty line and in another files too

src/Sticky/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+  position,
+  children,
+  offset,
+  className,
+  style,
+}) => {
+  const [isSticky, setIsSticky] = useState(false);
+  const ref = useRef();
+
+  useEffect(() => {
```
adamstankiewicz, line 20: I believe we may want to be using `useLayoutEffect` here instead of a normal `useEffect`. See [documentation](https://reactjs.org/docs/hooks-reference.html#uselayouteffect) and [this summary](https://kentcdodds.com/blog/useeffect-vs-uselayouteffect#uselayouteffect) for more details.

image

Because we're taking measurements of offset for a DOM element, I believe this may be one of those rare times to utilize `useLayoutEffect`. What do you think? The main difference with `useLayoutEffect` is that the DOM measurements would be taken before the updates are painted to the screen, which might mean a more smooth shadow while scrolling, for example.

src/Sticky/README.md
```diff
@@ -0,0 +1,54 @@
...
+    <div style={{ background: '#aa0000', width: '100%', height: '150px', color: '#ffffff' }}>
+      Sticky to the bottom
+    </div>
+  </Sticky>
+</div>
+```
+
+### With offset
+
+Valid `Offset` values are the same as for [the spacing classes](/foundations/spacing).
```
adamstankiewicz, line 44: nit: lowercase `offset` since it's not a component name.

src/Sticky/README.md
```diff
@@ -0,0 +1,54 @@
...
+designStatus: 'Done'
+devStatus: 'In progress'
+notes: |
+---
+
+Makes elements remain at the top or bottom of the viewport, like a sticky navbar.
+
+### Basic Usage
+
+### Top positioning
```
adamstankiewicz, line 18: These secondary headings should be one level below the primary "Basic Usage" heading level.

src/Sticky/_variables.scss
```diff
@@ -0,0 +1,3 @@
+$sticky-offset:                       0 !default;
+$sticky-shadow-top:                   0 -0.5rem 1rem rgba(0, 0, 0, 0.15), 0 -0.25rem 0.625rem rgba(0, 0, 0, 0.15) !default;
+$sticky-shadow-bottom:                0 0.5rem 1rem rgba(0, 0, 0, 0.15), 0 0.25rem 0.625rem rgba(0, 0, 0, 0.15) !default;
```
adamstankiewicz, line 3: nit: newline at end of file

adamstankiewicz, line 3: Should these be using preexisting elevation SCSS variables, e.g. `$level-1-box-shadow` or `$box-shadow-sm`?

src/Sticky/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+    // getComputedStyle is used to get real top/bottom
+    // values on the page for proper shadows display
+    const elementStyles = window.getComputedStyle(ref.current);
+    const elementOffset = elementStyles[position || 'top'];
+    // Margin calculations according to the offset.
+    // 1 pixel above/bellow + offset pixels that determines
+    // when callback function is called
+    const elementWithOffset = 1 + (parseInt(elementOffset, 10) || 0);
+    const observer = new IntersectionObserver(
+      ([entry]) => setIsSticky(entry.intersectionRatio < 1),
```
adamstankiewicz, line 30: This is a cool solution! 🚀

src/Sticky/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+
+const Sticky = ({
+  position,
+  children,
+  offset,
+  className,
+  style,
+}) => {
+  const [isSticky, setIsSticky] = useState(false);
+  const ref = useRef();
```
adamstankiewicz, line 18: Curious if this component should also be using `React.forwardRef` in order for a consumer to pass an existing `ref` to `Sticky` like we do in other components, and otherwise fallback to its own `ref`?

src/Sticky/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+  'top',
+  'bottom',
+];
+
+const Sticky = ({
+  position,
+  children,
+  offset,
+  className,
+  style,
```
adamstankiewicz, line 15: [clarification] Should `style` be explicitly handled by this component? Generally, using inline styles are avoided so I'd be make `style` be part of the official API for this component.

Instead, could `style` be replaced with `...rest` and update below by spreading on `...rest` like `<div {...rest} />`?

src/Sticky/index.jsx
monteri: Yes, it makes sense, I fix it

src/Sticky/index.jsx
monteri: Thank you, I'm glad to hear!

src/Sticky/_variables.scss
monteri: I have tried those two variables but they are almost invisible. As it mentioned in the task I needed to copy shadows from the `Scrollable`. But there are custom shadows and they are not put into variables. Then I went to the `core/scss` and didn't find shadows for the `top`. There are only shadows by leveling (they are for the bottom of the block) and no for the positioning (left, top, right). So I went to the figma foundations and found there [Shadows level 3 for top and bottom](https://www.figma.com/file/eGmDp94FlqEr4iSqy1Uc1K/Paragon-Design-System?node-id=9076%3A3783). Transferred them to variables. I think maybe we should move them to the `core/scss`?

src/Sticky/index.jsx
monteri: Yes, right, added `React.forwardRef`

src/Sticky/index.jsx
monteri: right, `{...rest}` is more appropriate. `Style` was added as for the case when user would want to add an offset programmatically. And `{...rest}` allows this, cool.

adamstankiewicz (approved):

edx-webhook:
@monteri 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1096&entry.1484655318=https://github.com/monteri&entry.752974735=2022-02-16+10:53&entry.1917517419=2022-03-11+13:14&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1120, opened 2022-03-01
[BD-46] feat: add insights tabs routing

monteri:
Insights routing update:
- tabs on the `insights` pages is linked with corresponding routes
- used search params for route changing
- added support for back button behavior

JIRA: [PAR-724](https://openedx.atlassian.net/browse/PAR-724)

### DIFF
--- a/www/src/pages/insights.jsx
+++ b/www/src/pages/insights.jsx
@@ -14,6 +14,7 @@ import ProjectUsageExamples from '../components/insights/ProjectUsageExamples';
 import ComponentUsageExamples from '../components/insights/ComponentUsageExamples';
 import getGithubProjectUrl from '../utils/getGithubProjectUrl';
 import dependentProjectsAnalysis from '../../../dependent-usage.json';
+import { INSIGHTS_TABS, INSIGHTS_PAGES } from '../config';
 
 const {
   lastModified: analysisLastUpdated,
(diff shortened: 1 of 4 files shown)


www/src/pages/insights.jsx
```diff
@@ -171,7 +173,16 @@ const ComponentsUsage = () => (
...
 );
 
-export default function InsightsPage() {
+export default function InsightsPage({ location }) {
+  const tabs = ['/insights', '/insights?tab=projects', '/insights?tab=components'];
+  const path = `${location.pathname}${location.search}`;
+
+  const handleOnSelect = (value) => {
+    if (value !== path) {
+      navigate(value);
```
adamstankiewicz, line 182: I believe we also will need to get the current tab, if any, from the URL and navigate to it upon page load, so that if I hard refresh on `/insights?tab=components`, I will see the contents of the "Components" tab already.

monteri: Yes, I also saw this in the preview. Locally it works fine. On preview it is broken for some reason. I thought it could be due to the forward slashes (`/insights?tab=component` changed `/insights/?tab=component`) but it is not. Currently I am studying the issue.
Also I am interested if you have any ideas why this can happen 🤔

viktorrusakov: @adamstankiewicz I think we were able to fix it, I believe gatsby just served the default page version that was created during the build and javascript just didn't get a chance to run on first render to update selected tab based on the url (also I think gatsby doesn't like query parameters in the url...)

Manually building separate page for each tab solves the issue

adamstankiewicz (approved):

openedx-webhooks:
@monteri 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1120&entry.1484655318=https://github.com/monteri&entry.752974735=2022-03-01+15:01&entry.1917517419=2022-05-13+13:41&entry.2133058324=Yes) so we can improve your experience in the future.

adamstankiewicz:
@viktorrusakov Thanks for looking into the issue and getting a fix pushed up! 🙌

PR #1131, opened 2022-03-09
docs: add TextEditor ADR

connorhaugh:
This ADR proposes a Paragon TextEditor component as a pass through of Tinymce, an existing WYSIWIG text editor, with standardization. This follows the suggested reverse process for engineers to submit things to pwg

### DIFF
--- a/docs/decisions/0017-Text-Editor-Using-TinyMce
+++ b/docs/decisions/0017-Text-Editor-Using-TinyMce
@@ -1,0 +1,22 @@
+Add TextEditor Component using wysiwyg TinyMCE 
+==========================
+Synopsis
+--------
+Text Editing, beyond the scope of simple text input fields in form components, is a neglected user experience in Paragon. Already, numerous openedX experiences use TinyMCE. Standardizing styling, accessibility, and versioning for text editing in Paragon will benefit users, designers, and developers alike.
+Decision
+------
+Paragon will expose a TextEditor Component, which uses TinyMCE’s React library as a foundation. The TextEditor component will have a standardized list of plugins that will also be supported by Paragon, and will not need to be managed by the consumer. A default experience will also be provided for out-of-the-box usage. Any plugins beyond those already managed by paragon, paid or otherwise, will be up to the consumer to maintain. Additional extensions, like custom plugins for specific use cases, will live elsewhere, and not within Paragon. 
+Status
+------
+Proposed
+Context
+-------
+Already, there are two simultaneous ongoing implementations of React-based html editors using tinyMCE. One is the Text Xblock editor, the other is the instructor bulk email tool found in the communications MFE. Already, edX also uses tinyMCE in edx-platform, and hosts its own image of the service there. Presumably, this experience too will be moved from the ball of mud into an independent MFE, and other use cases for text editing will arise. Therefore, it seems obvious to standardize across implementations.
+TinyMCE itself is “based on WCAG [2] and Section 508, to ensure that the TinyMCE editor is accessible to content authors.” Other alternatives do not provide this assurance. In addition to being a familiar entity, it also accomplishes our main goals for content authoring.
+Consequences
+------------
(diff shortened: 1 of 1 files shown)


docs/decisions/0017-Text-Editor-Using-TinyMce
```diff
@@ -0,0 +1,22 @@
...
+Decision
+------
+Paragon will expose a TextEditor Component, which uses TinyMCE’s React library as a foundation. The TextEditor component will have a standardized list of plugins that will also be supported by Paragon, and will not need to be managed by the consumer. A default experience will also be provided for out-of-the-box usage. Any plugins beyond those already managed by paragon, paid or otherwise, will be up to the consumer to maintain. Additional extensions, like custom plugins for specific use cases, will live elsewhere, and not within Paragon. 
+Status
+------
+Proposed
+Context
+-------
+Already, there are two simultaneous ongoing implementations of React-based html editors using tinyMCE. One is the Text Xblock editor, the other is the instructor bulk email tool found in the communications MFE. Already, edX also uses tinyMCE in edx-platform, and hosts its own image of the service there. Presumably, this experience too will be moved from the ball of mud into an independent MFE, and other use cases for text editing will arise. Therefore, it seems obvious to standardize across implementations.
```
xitij2000, line 14: Make that three :-) The discussions MFE also uses it for post editing: https://github.com/openedx/frontend-app-discussions/blob/master/src/components/TinyMCEEditor.jsx

docs/decisions/0017-Text-Editor-Using-TinyMce
```diff
@@ -0,0 +1,22 @@
...
+Status
+------
+Proposed
+Context
+-------
+Already, there are Three simultaneous ongoing implementations of React-based html editors using tinyMCE. One is the Text Xblock editor, the other is the instructor bulk email tool found in the communications MFE. The final is the discussions MFE. Already, edX also uses tinyMCE in edx-platform, and hosts its own image of the service there. Presumably, this experience too will be moved from the ball of mud into an independent MFE, and other use cases for text editing will arise. Therefore, it seems obvious to standardize across implementations. Paragon will serve as the source of truth for TinyMce versioning, much in the same way the Data table component works.
```
adamstankiewicz, line 14: One nit might be consistency with how TinyMCE is spelled throughout :)

docs/decisions/0017-Text-Editor-Using-TinyMce
```diff
@@ -0,0 +1,22 @@
...
+TinyMCE itself is “based on WCAG [2] and Section 508, to ensure that the TinyMCE editor is accessible to content authors.” Other alternatives do not provide this assurance. In addition to being a familiar entity, it also accomplishes our main goals for content authoring.
```
adamstankiewicz, line 15: Is there a link that would be handy to include here?

adamstankiewicz (approved):
LGTM with a couple nits. Feel free to change the status to "Approved" before you merge.

muselesscreator:
I believe this makes sense in the long term, but for the short-term, the team needs w/rt the tinyMCE installation are going to leave them with a much less configurable option than we would want for Paragon consumption, and the TnL squad will not have the bandwidth to continue pressing on this from a component standardization perspective, so that work will need to get forwarded to the FWG or PWG for future ownership.

adamstankiewicz:
Following up on the previous offline discussions in Frontend/Paragon Working Groups, we decided we'd defer on considering a `TextEditor` component in Paragon, at least one that consumes TinyMCE (especially needing to support paid plugins).

That said, it may be interesting if Paragon one day supported its own custom text editor component that didn't rely on TinyMCE, but instead something like https://github.com/facebook/lexical. 

As a result, I'm inclined to close the PR for now, though we'll definitely be considering whether a text editor makes sense on the long term roadmap for Paragon. 

@connorhaugh If you disagree, please feel to re-open the issue 😄 Thanks!

PR #1133, opened 2022-03-09
[BD-46] feat: add Segment

monteri:
Segment is installed into Paragon docs using [gatsby-plugin-segment-js](https://www.gatsbyjs.com/plugins/gatsby-plugin-segment-js/)
- added `SEGMENT_KEY` inside `.env.development` (according to docs it should be at least 10 symbols otherwise it won't be launched)
- in the `gatsby-config.js` I added `process.env.SEGMENT_KEY` for both 'prod' and 'dev'. Should it be changed? Maybe different names for the variables?
- `trackPage: true` allows automatically call `window.analytics.page()` on the route change
- Added custom events: 
  1. opening/closing the "Settings" sheet
  2. switching between tabs in Usage Insights
  3. changing the theme
  4. changing the direction
  5. ❗ changing language will be added when the feature will be added
  
 Events are tested for correct firing


www/gatsby-config.js
```diff
@@ -62,5 +62,13 @@ module.exports = {
         ],
       },
     },
+    {
+      resolve: `gatsby-plugin-segment-js`,
+      options: {
+        prodKey: process.env && process.env.SEGMENT_KEY || '',
+        devKey: process.env && process.env.SEGMENT_KEY || '',
```
adamstankiewicz, line 69: I think we can keep these as both the same `process.env.SEGMENT_KEY` as you have it. Not sure the immediate value of having separate variables here.

www/.env.development
```diff
@@ -1 +1,2 @@
 FEATURE_DIRECTION_SWITCHER=true
+SEGMENT_KEY=0123456789
```
adamstankiewicz, line 2: During development, I think it would make sense to have it be `SEGMENT_KEY=''` so that Segment doesn't get configured locally unless explicitly supplying your own `SEGMENT_KEY` here for testing/QA purposes. I might worry that using an incorrect Segment key will throw errors when the script configures Segment.

www/gatsby-config.js
```diff
@@ -62,5 +62,13 @@ module.exports = {
...
+        trackPage: true,
```
adamstankiewicz, line 70: Nice, this is handy :)

www/src/context/SettingsContext.jsx
```diff
@@ -23,11 +23,13 @@ const SettingsContextProvider = ({ children }) => {
     document.body.setAttribute('dir', e.target.value);
     setDirection(e.target.value);
     global.localStorage.setItem('pgn__direction', e.target.value);
+    window.analytics.track('Direction change');
```
adamstankiewicz, line 26: We should use `global` instead of `window` here (similar to `global.localStorage` above). Also, I think it would make sense to also include the direction (`e.target.value`) as an event property here?

www/src/context/SettingsContext.jsx
```diff
@@ -23,11 +23,13 @@ const SettingsContextProvider = ({ children }) => {
...
   };
 
   const handleThemeChange = (e) => {
     setTheme(e.target.value);
     global.localStorage.setItem('pgn__theme', e.target.value);
+    window.analytics.track('Theme change');
```
adamstankiewicz, line 32: We should use global instead of window here (similar to global.localStorage above). Also, I think it would make sense to also include the theme (e.target.value) as an event property here?

www/src/context/SettingsContext.jsx
```diff
@@ -43,6 +45,10 @@ const SettingsContextProvider = ({ children }) => {
     }
   }, []);
 
+  useEffect(() => {
+    window.analytics.track('Toggle Settings');
```
adamstankiewicz, line 49: We should use global instead of window here. Also, I think it would make sense to also include state of `showSettings` here as well in the event properties.

adamstankiewicz (commented):
When I load [the Netlify PR deploy preview](https://deploy-preview-1133--paragon-edx.netlify.app/), it seems that it results in a JS error that results in a blank, white screen:

image

www/.env.development
monteri: Right, I leave it empty

www/gatsby-config.js
monteri: Thank you! 🔥

www/src/context/SettingsContext.jsx
monteri: slightly changed this, to rid of `useEffect` because of initial render. Added  `value ? 'show' : 'hide'`

www/src/context/SettingsContext.jsx
monteri: Right, added `e.target.value`

www/src/context/SettingsContext.jsx
monteri: Right, added `e.target.value`

www/.env.development
```diff
@@ -1,2 +1,2 @@
 FEATURE_DIRECTION_SWITCHER=true
-SEGMENT_KEY=0123456789
+SEGMENT_KEY=
```
adamstankiewicz, line 2: nit: default to an empty string

www/src/context/SettingsContext.jsx
```diff
@@ -23,13 +23,18 @@ const SettingsContextProvider = ({ children }) => {
     document.body.setAttribute('dir', e.target.value);
     setDirection(e.target.value);
     global.localStorage.setItem('pgn__direction', e.target.value);
-    window.analytics.track('Direction change');
+    global.analytics.track(`Direction change (${e.target.value})`);
```
adamstankiewicz, line 26: I think it would make sense to pass these values as event properties rather than a distinct event, e.g.:

```js
global.analytics.track('Direction change', { direction: e.target.value });
```

This would be similar feedback for most of the other events as well, e.g. "Toggle settings" might pass a `value` property.

www/src/pages/insights.jsx
```diff
@@ -171,7 +173,29 @@ const ComponentsUsage = () => (
...
+export default function InsightsPage({ location }) {
+  const tabs = ['/insights', '/insights/?tab=projects', '/insights/?tab=components'];
+  const path = `${location.pathname}${location.search}`;
+
+  const handleOnSelect = (value) => {
+    if (value !== path) {
+      navigate(value);
+      switch (value) {
+        case tabs[0]:
+          window.analytics.track('Usage Insights (Summary)');
```
adamstankiewicz, line 185: When the SEGMENT_KEY env variable is not set, are you seeing `global.analytics` as undefined? I'm getting errors locally calling `.track()` on undefined when SEGMENT_KEY isn't set. Any ideas on how to get around this?

www/src/pages/insights.jsx
```diff
@@ -171,7 +173,29 @@ const ComponentsUsage = () => (
...
+          break;
+        case tabs[1]:
+          window.analytics.track('Usage Insights (Projects)');
```
adamstankiewicz, line 188: These events could also benefit from using event properties here, e.g. 

```js
global.analytics.track('Usage Insights', { tab: 'projects' });

File: www/src/pages/insights.jsx [hunk 11]
```diff
@@ -181,18 +205,29 @@ export default function InsightsPage() {
           <h1>Usage Insights</h1>
           <p>Last updated: {new Date(analysisLastUpdated).toLocaleDateString()}</p>
         </header>
-        <Tabs defaultActiveKey="summary" id="uncontrolled-tab-example">
-          <Tab eventKey="summary" title="Summary">
+        <Tabs
+          activeKey={tabs.includes(path) ? path : tabs[0]}
```
adamstankiewicz, line 209: I do think we'll need to figure out a solution to the tabs routing here so that it's consistently functioning. If I'm on `/insights?tab=components` and refresh the page, for example, the URL still says I'm on "Components" tab though I'm actually now on the "Summary" tab. Then, when trying to navigate to the "Components" tab again, I'm unable to.

www/.env.development
monteri: leaving as it is, adding conditional plugin connection

www/src/context/SettingsContext.jsx
monteri: Added event properties for all occurrences

www/src/pages/insights.jsx
monteri: Adding conditional plugin connection depending on the `SEGMENT_KEY `

www/src/pages/insights.jsx
monteri: Right, also fixed here

adamstankiewicz (approved):

edx-webhook:
@monteri 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1133&entry.1484655318=https://github.com/monteri&entry.752974735=2022-03-09+15:33&entry.1917517419=2022-03-25+04:43&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1136, opened 2022-03-10
[BD-46] feat: navigation for page headings

monteri:
Added navigation for the documentation pages:

- added h2-h6 headers autolinking
- processing of link click. When user clicks on the anchor tag he is not scrolled exactly to the heading. After the route change whole page is rerendered that causes this behavior. Seems like a top offset for each heading has wrong values.
- added the Table of content with styles and tree structure
- added markup for the different screen width using Row and Col but the previous display is saved for the non mdx pages.

JIRA: [PAR-648](https://openedx.atlassian.net/browse/PAR-648)

Fixed


edx-webhook:
Thanks for the pull request, @monteri! I've created [BLENDED-1140](https://openedx.atlassian.net/browse/BLENDED-1140) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


www/src/components/Toc.scss
```diff
@@ -0,0 +1,38 @@
+.pgn-doc__toc {
+  position: sticky;
```
adamstankiewicz, line 2: I'm wondering if it would make sense to start using the new `Sticky` component here to manage the sticky positioning of the ToC?

adamstankiewicz (commented):
Just one comment, now that `Sticky` is merged, does it make sense to start using it for the ToC sidebar?

www/src/components/Toc.scss
monteri: yes, updated with Sticky usage

adamstankiewicz (approved):

PR #1163, opened 2022-03-23
fix: simplify DataTable action component to no longer depend on `as` prop

adamstankiewicz:
Updates the `DataTable` actions (e.g., bulk actions, table actions) to use a `ModalPopup` for its overflow actions instead of a `Dropdown`. This allows consumers to retain icons and variants on the action buttons the same as if they were part of the visible actions (i.e., not in the overflow).

Also, fixes some miscellaneous styles regarding the actions button.

image


src/Dropdown/Dropdown.scss
```diff
@@ -24,3 +24,28 @@
...
+  .pgn__dropdown-item__icon-after {
+    margin-inline-start: 0.5rem;
+    margin-right: -0.25em;
+
+    [dir="rtl"] & {
+      margin-right: 0.5rem;
+      margin-left: -0.25em;
+    }
+  }
+}
```
long74100, line 51: nit: new line here as well

long74100 (approved):
🥇

PR #1192, opened 2022-03-29
[BD-46] feat: fixes for Button strokes, when in-focus state

PKulkoRaccoonGang:
- fixes for Button strokes, when in-focus state;
- fixes border radius in the Search-Field Button component.

JIRA: [PAR-764](https://openedx.atlassian.net/browse/PAR-764)

### DIFF
--- a/src/Button/Button.scss
+++ b/src/Button/Button.scss
@@ -209,8 +209,6 @@
 .btn-group-lg>.btn-primary.btn,
 .btn-group-sm>.btn-primary.btn,
 body .page-link {
-  .btn:focus::before,
-  .focus:before,
   &:focus.btn-lg::before,
   &:focus.btn-sm::before,
   &:focus::before {
--- a/src/Button/Button.scss
+++ b/src/Button/Button.scss
@@ -204,6 +204,20 @@
   }
 }
 
+.btn,
+.btn-primary,
+.btn-group-lg>.btn-primary.btn,
+.btn-group-sm>.btn-primary.btn,
+body .page-link {
+  .btn:focus::before,
+  .focus:before,
+  &:focus.btn-lg::before,
+  &:focus.btn-sm::before,
+  &:focus::before {
+    @include button-size($btn-padding-y, $btn-padding-x, $btn-font-size, $btn-line-height, $btn-border-radius);
+  }
+}
+
 // Future-proof disabling of clicks on `<a>` elements
 a.btn.disabled,
 fieldset:disabled a.btn {
--- a/src/SearchField/SearchField.scss
+++ b/src/SearchField/SearchField.scss
@@ -98,6 +98,5 @@
 }
 
 .pgn__searchfield__button.btn[type='submit'] {
-    border-radius: 0;
     margin-inline-start: $search-button-margin;
 }


edx-webhook:
Thanks for the pull request, @PKulkoRaccoonGang! I've created [OSPR-6550](https://openedx.atlassian.net/browse/OSPR-6550) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

src/Button/Button.scss
```diff
@@ -204,6 +204,20 @@
   }
 }
 
+.btn,
+.btn-primary,
+.btn-group-lg>.btn-primary.btn,
+.btn-group-sm>.btn-primary.btn,
+body .page-link {
+  .btn:focus::before,
```
ihor-romaniuk, line 212: Is it OK, the `.btn` in the `.btn` class?

PKulkoRaccoonGang: Fixed, thanks

ihor-romaniuk (approved):

src/Button/Button.scss
```diff
@@ -204,6 +204,18 @@
   }
 }
 
+.btn,
+.btn-primary,
+.btn-group-lg>.btn-primary.btn,
+.btn-group-sm>.btn-primary.btn,
+body .page-link {
```
adamstankiewicz, line 211: [clarification] Do you mind shining some light on how you arrived at these selectors? For example, why does `.page-link` use the `button-size` mixin and whether each button variant will need to be included beyond just `btn-primary`?

PKulkoRaccoonGang: I did some research on the whole project and found out that using only one `btn-primary` selector does not produce the required focus state. All these selectors were found during the check of all components and their use is mandatory. I also decided to use the `button-size` mixin which we apply to the `before` pseudo-element and have the ability to get a dynamic change in the border-radius parameters when the Paragon theme changes.

adamstankiewicz: @PKulkoRaccoonGang I'm still not entirely following the approach here. It's not just the one `.btn-primary` selector without the focus state. All the variants have a rounded border radius:

image

Because `.btn` is being used as a selector, it applies to all buttons which is likely why your solution technically works, even though it doesn't need the specificity you are providing with regard to a specific variant.

Could the following be added in the already existing selector for `.btn` above?

```scss
.btn {
  @include button-size($btn-padding-y, $btn-padding-x, $btn-font-size, $btn-line-height, $btn-border-radius);
  
  // ...

  &:focus:before {
    @include button-size($btn-padding-y, $btn-padding-x, $btn-font-size, $btn-line-height, $btn-border-radius);
  }
}
```

This solves the issue locally for me.

adamstankiewicz (changes requested):

openedx-webhooks:
@PKulkoRaccoonGang Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1192&entry.1484655318=https://github.com/PKulkoRaccoonGang&entry.752974735=2022-03-29+11:19&entry.1917517419=2022-04-29+12:07&entry.2133058324=No) so we can improve your experience in the future.

PR #1200, opened 2022-03-29
fix: no longer use math.div

adamstankiewicz:
## Description

Use of `math.div` appears to be causing issues for at least one consumer. This fix replaces `math.div` with an equivalent `calc`. Note because of the negative values, we are multiplying the value by -1 to force it to be negative.

### Deploy Preview

n/a

## Merge Checklist

n/a

## Post-merge Checklist

### DIFF
--- a/www/src/components/Settings.jsx
+++ b/www/src/components/Settings.jsx
@@ -7,7 +7,6 @@ import {
 } from '~paragon-react';
 import { Close } from '~paragon-icons';
 
-import { FEATURES } from '../config';
 import SettingsContext from '../context/SettingsContext';
 import { THEMES } from '../../theme-config';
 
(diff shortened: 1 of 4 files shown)


www/src/components/Settings.jsx
```diff
@@ -7,7 +7,6 @@ import {
 } from '~paragon-react';
 import { Close } from '~paragon-icons';
 
-import { FEATURES } from '../config';
```
adamstankiewicz, line 10: eslint warning seen locally. not sure why not caught by CI without further investigation.

www/src/config.js
```diff
@@ -1,6 +1,6 @@
-import hasFeatureFlagEnabled from './utils/hasFeatureFlagEnabled';
+// import hasFeatureFlagEnabled from './utils/hasFeatureFlagEnabled';
```
adamstankiewicz, line 1: unused var discovered by eslint warning seen locally. not sure why not caught by CI without further investigation.

PR #1207, opened 2022-04-01
[BD-46] feat: update peerDependencies

PKulkoRaccoonGang:
## Description
- updated React to version 18,
- updated React-dom to version 18,
- update Prop-types to version 15.8.1.

JIRA: [PAR-772](https://openedx.atlassian.net/browse/PAR-772)

## Merge Checklist

## Post-merge Checklist


edx-webhook:
Thanks for the pull request, @PKulkoRaccoonGang! I've created [BLENDED-1180](https://openedx.atlassian.net/browse/BLENDED-1180) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


ihor-romaniuk (approved):

package.json
```diff
@@ -56,9 +56,9 @@
     "uncontrollable": "7.2.1"
   },
   "peerDependencies": {
-    "prop-types": "^15.7.2",
-    "react": "^16.8.6",
-    "react-dom": "^16.8.6"
+    "prop-types": "^15.8.1",
+    "react": "^18.0.0",
```
adamstankiewicz, line 60: @PKulkoRaccoonGang To confirm, all of Paragon works when using React 18? 🤔 

Also, as is, this change would cause a peer dependency warning with itself, as it expects React 18, but React 16 is still installed in `devDependencies`.

Most of the Open edX platform is on React 16 in addition to some consumers already using React 17, so we will need to continue to support these versions via a peer dependency range and in the components themselves, e.g.

```
{
  "react": "^16.8.6 || ^17.0.0",
  "react-dom": "^16.8.6 || ^17.0.0"
}
```

Also, `prop-types` as a peer dependency is... odd. I feel like we should probably move it to be in `devDependencies` only most likely.

PKulkoRaccoonGang: Thanks, done. I checked all components after the new changes. Paragon components work well.

package.json
```diff
@@ -88,8 +87,9 @@
     "jest": "^25.0.0",
     "jest-cli": "^25.0.0",
     "markdown-loader-jest": "^0.1.1",
-    "react": "^16.14.0",
-    "react-dom": "^16.14.0",
+    "prop-types": "^15.8.1",
+    "react": "^16.8.6 || ^17.0.0",
+    "react-dom": "^16.8.6 || ^17.0.0",
```
adamstankiewicz, line 92: We should only specify one version in the `devDependencies`, the latest version of React that Paragon will support. The range as you have it here should be reflected in the `peerDependencies`, as that is what tells consumers which versions of React this library supports.

As it is right now, it's saying Paragon relies only on React 18, yet installing... both (?) React 16 and 17.

adamstankiewicz (changes requested):

package.json
PKulkoRaccoonGang: Yes, you are right
Made changes and continued working on the [new pull request](https://github.com/openedx/paragon/pull/1248)

openedx-webhooks:
@PKulkoRaccoonGang Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1207&entry.1484655318=https://github.com/PKulkoRaccoonGang&entry.752974735=2022-04-01+16:01&entry.1917517419=2022-04-21+20:43&entry.2133058324=No) so we can improve your experience in the future.

PR #1210, opened 2022-04-04
feat: Add className prop to Toast

zainab-amir:
## Description
The issue is related to https://openedx.atlassian.net/browse/VAN-906 where two toasts are appearing (one for the component that renders when desktop view is on and other for Mobile view)

The classes added to parent components in prospectus that dynamically show and hide the content `d-block d-md-none` and `d-none d-md-block` don't work on any children Toast elements. Adding a `div` with these classes around `Toast` also doesn't work. 

Solution:
Added a span around `BaseToast` which takes className as props. This doesn't change the appearance of Toast. 

### Deploy Preview

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Toast/index.jsx
+++ b/src/Toast/index.jsx
@@ -1,4 +1,5 @@
 import React, { useState } from 'react';
+import classNames from 'classnames';
 import PropTypes from 'prop-types';
 
 import BaseToast from 'react-bootstrap/Toast';
@@ -7,21 +8,21 @@ import { Button, IconButton, Icon } from '..';
 import { Close } from '../../icons';
 
 function Toast({
-  action, children, closeLabel, delay, onClose, show,
+  action, children, className, closeLabel, onClose, show, ...rest
 }) {
   const [autoHide, setAutoHide] = useState(true);
   return (
     <ToastContainer>
       <BaseToast
         autohide={autoHide}
-        className="toast"
-        delay={delay}
+        className={classNames('pgn__toast', className)}
         onClose={onClose}
         onBlur={() => setAutoHide(true)}
         onFocus={() => setAutoHide(false)}
         onMouseOut={() => setAutoHide(true)}
         onMouseOver={() => setAutoHide(false)}
         show={show}
+        {...rest}
       >
         <div
           className="toast-header"
@@ -33,7 +34,7 @@ function Toast({
               alt={closeLabel}
               className="align-self-start"
               src={Close}
-              onClick={() => (onClose())}
+              onClick={onClose}
               variant="primary"
               invertColors
             />
@@ -57,6 +58,7 @@ function Toast({
 
 Toast.defaultProps = {
   action: null,
+  className: undefined,
   closeLabel: 'Close',
   delay: 5000,
 };
@@ -90,6 +92,8 @@ Toast.propTypes = {
   closeLabel: PropTypes.string,
   /** Time in milliseconds for which the `Toast` will display. */
   delay: PropTypes.number,
+  /** Class names for the `BaseToast` component */
+  className: PropTypes.string,
 };
 
 export default Toast;


src/Toast/index.jsx
```diff
@@ -7,56 +8,59 @@ import { Button, IconButton, Icon } from '..';
...
-        )}
-      </BaseToast>
+            <p className="small">{children}</p>
+            <div className="toast-header-btn-container">
+              <IconButton
+                iconAs={Icon}
+                alt={closeLabel}
+                className="align-self-start"
+                src={Close}
+                onClick={() => (onClose())}
```
adamstankiewicz, line 38: nit / clean up: `onClick={onClose}`

src/Toast/index.jsx
```diff
@@ -90,6 +94,8 @@ Toast.propTypes = {
   closeLabel: PropTypes.string,
   /** Time in milliseconds for which the `Toast` will display. */
   delay: PropTypes.number,
+  /** class names for the span wrapping the BaseToast component */
```
adamstankiewicz, line 97: "Class names for the ``BaseToast`` component" (no mention of `span`, since that can be removed)

adamstankiewicz (commented):
Thanks for your contribution, @zainab-amir. I left some feedback.

It'd also be helpful if you could link to the code where the issue with `Toast` is being rendered, as using utility classes as described to control whether a `Toast` is shown or not feels like an anti-pattern, given that the `Toast` has a `show` prop? 🤔

zainab-amir:
@adamstankiewicz [This is code ](https://github.com/edx/prospectus/blob/58d569925d32ca4e20143265bdce4a54383b6c34/src/templates/courseDetail.jsx#L202-L207)where both mobile and desktop version of course about pages are rendered in the DOM at the same time. They are dynamically shown/hidden based on the classes `d-block d-md-none`, `d-none d-md-block` added to these components. However the Toast in the children of these components doesn't adhere by these classes and appears twice.
<p align="center">
Screenshot 2022-03-28 at 12 23 17 PM
</p>

src/Toast/index.jsx
```diff
@@ -7,14 +7,14 @@ import { Button, IconButton, Icon } from '..';
...
-  action, children, closeLabel, delay, onClose, show,
+  action, children, className, closeLabel, delay, onClose, show,
 }) {
   const [autoHide, setAutoHide] = useState(true);
   return (
     <ToastContainer>
       <BaseToast
         autohide={autoHide}
-        className="toast"
+        className={className}
```
adamstankiewicz, line 17: As suggested in previous comment, I think we should include `pgn__toast` here, i.e. `className={classNames('pgn__toast', className)}`.

Similarly, we should also adopt the prop spreading strategy here that I previously mentioned as well, with regard to `{...rest}`:

```jsx
function Toast({
  className,
  ...rest
}) {
  return (
    <ToastContainer>
      <BaseToast
        className={classNames('pgn__toast', className)
        {...rest}
      />
    </ToastContainer>
  );
}
```

src/Toast/index.jsx
```diff
@@ -57,6 +57,7 @@ function Toast({
 
 Toast.defaultProps = {
   action: null,
+  className: '',
```
adamstankiewicz, line 60: `''` -> `undefined`

adamstankiewicz:
> @adamstankiewicz [This is code ](https://github.com/edx/prospectus/blob/58d569925d32ca4e20143265bdce4a54383b6c34/src/templates/courseDetail.jsx#L202-L207)where both mobile and desktop version of course about pages are rendered in the DOM at the same time. They are dynamically shown/hidden based on the classes `d-block d-md-none`, `d-none d-md-block` added to these components. However the Toast in the children of these components doesn't adhere by these classes and appears twice.

@zainab-amir Great, thanks for this context.

While I think this PR is worthwhile regardless, I think a potentially more optimal solution to your problem might be to not rely on `display` properties to determine whether `CoursePage` or `CoursePageMobile` gets rendered. Each is still rendered in the DOM (and doing all its underlying logic, like showing `Toast`) even if it's not visible in the UI.

I might recommend exploring conditionally rendered the `CoursePage` or `CoursePageMobile` component based on a screen width as it relates to a breakpoint instead. That way, only one of them is truly rendered in the DOM at any given time.

Example:

```jsx
import { useMediaQuery, breakpoints } from '@edx/paragon';

const CoursePageResponsive = () => {
  const isDesktop = useMediaQuery({
    query: `(min-width: ${breakpoints.medium.minWidth}px)`
  })
  if (isDesktop) {
    return <CoursePage />;
  }
  return <CoursePageMobile />;
};
```

src/Toast/index.jsx
```diff
@@ -7,21 +8,20 @@ import { Button, IconButton, Icon } from '..';
 import { Close } from '../../icons';
 
 function Toast({
-  action, children, closeLabel, delay, onClose, show,
```
adamstankiewicz, line 10: We shouldn't remove the `show` prop as it is required via `PropTypes.bool.isRequired` below.

adamstankiewicz (commented):
Just one more nit, @zainab-amir.

adamstankiewicz (approved):

PR #1230, opened 2022-04-14
fix: Toast cross icon on small screens

syedsajjadkazmii:
## Description

Toast message width is adjusted according to the screen size 
that was larger on small screens causing the cross icon on the 
toast to be unclickable.

[VAN-911](https://openedx.atlassian.net/browse/VAN-911)

**Before**

Screen Shot 2022-04-14 at 1 59 37 PM

**After**

Screen Shot 2022-04-19 at 11 48 38 AM


zainab-amir (approved):

src/Toast/_Toast.scss
```diff
@@ -5,7 +5,6 @@
   background-color: $toast-background-color;
   box-shadow: $toast-box-shadow;
   margin: 0;
-  min-width: $toast-max-width;
```
shafqatfarhan, line 8: So this `min-width` on the screen size of less than 768px was resulting in Toast to fall beyond the visible screen (towards right) and the default `.toast-container` class's `right: 0.625rem;` had no effect. 

In my opinion, we can move this inside media query with the rule `@media only screen and (min-width: 768px)` so it may work as-is for screen sizes larger than 768px. And we could override `.toast-container` class inside `@media only screen and (max-width: 768px)` and add `right: 1.25rem;`. It will work after removing this min-width property. I have also tested it and it works fine

src/Toast/_Toast.scss
```diff
@@ -46,6 +45,6 @@
   }
 
   @media only screen and (min-width: 768px) {
-    max-width: $toast-max-width;
```
shafqatfarhan, line 49: We can revert this and add max-width here as well

src/Toast/_Toast.scss
```diff
@@ -46,6 +45,6 @@
   }
```
shafqatfarhan, line 45: Try this:

```suggestion
@media only screen and (max-width: 768px)
    max-width: 100%;

    .toast-container {
      right: 1.25rem;
    }
  }
```

shafqatfarhan (commented):
Edited:

We can update the value for the `right` property [here](https://github.com/openedx/paragon/blob/master/src/Toast/_ToastContainer.scss#L14) to `1.25rem` (same as the value for `left` property) and it will reflect in the media query for screen sizes less than 768px

src/Toast/_Toast.scss
syedsajjadkazmii: Yes, updating the right property [here](https://github.com/openedx/paragon/blob/master/src/Toast/_ToastContainer.scss#L14) will centralize the toast, thank you!

shafqatfarhan (approved):

adamstankiewicz (approved):

PR #1233, opened 2022-04-15
[BD-46] feat: generate component track

monteri:
Add additional function in the `generate-component` script:
- add `dotenv` and `axios` as `devDependencies`
- indicate path to the `.env` in the `www` directory to have single `.env` file
- created request with `componentName` param

JIRA: [PAR-779](https://openedx.atlassian.net/browse/PAR-779)


openedx-webhooks:
Thanks for the pull request, @monteri! I've created [BLENDED-1205](https://openedx.atlassian.net/browse/BLENDED-1205) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


component-generator/utils.js
```diff
@@ -16,6 +18,25 @@ function validateComponentName(value) {
...
 }
 
+/**
+ * Sends request to the Netify function to inform about generate-component usage.
+ * @param {string} componentName - component name
+ */
+function sendTrackInfo(componentName) {
+  const { BASE_URL, TRACK_ANONYMOUS_ANALYTICS } = process.env;
+  if (TRACK_ANONYMOUS_ANALYTICS) {
+    const url = `${BASE_URL}/.netlify/functions/track-generate-component`;
```
adamstankiewicz, line 28: This should handle the calling of a Netlify function API endpoint, but we will also need to actually create that function/endpoint as well.

Following [the docs here](https://docs.netlify.com/functions/build-with-javascript/) and [this playground](https://functions.netlify.com/playground/), we should be able to add a file for this `netlify/functions` inside of the `www` module, e.g. `www/netlify/functions/trackGenerateComponent.js`. 
This file would be responsible for calling the Segment API to actually dispatch a new event. Something along the lines of:

```js
// www/netlify/functions/trackGenerateComponent.js

exports.handler = async function (event, context) {
  // Only allow POST
  if (event.httpMethod !== "POST") {
    return { statusCode: 405, body: "Method Not Allowed" };
  }
  const { componentName } = event.body; // may need to parse `event.body` first
  // dispatch event to Segment
  analytics.track({
    anonymousId: uuidv4(),
    event: 'edx.paragon.generate-component', // example event name
    properties: { componentName },
  });

  return {
    statusCode: 200,
    body: JSON.stringify({ succes: true }),
  };
}
```

There is [some information on the tools available](https://docs.netlify.com/functions/build-with-javascript/#tools) to help run these functions locally as well, e.g. `netlify dev`.

component-generator/utils.js
```diff
@@ -16,6 +18,25 @@ function validateComponentName(value) {
...
+    axios.get(url, { params: { componentName } })
```
adamstankiewicz, line 29: I think we'll want to treat it as a `POST`?

component-generator/utils.js
```diff
@@ -1,7 +1,9 @@
 const { InvalidOptionArgumentError } = require('commander');
 const fs = require('fs');
 const path = require('path');
+const axios = require('axios');
 const { exec } = require('child_process');
+require('dotenv').config({ path: path.resolve(__dirname, '../www/.env.development') });
```
adamstankiewicz, line 6: The `BASE_URL` and `TRACK_ANONYMOUS_ANALYTICS` are defined in `www/.env.development` but they are only related to the component generator. Should these environment variables be moved to a .env file in `component-generator` directory?

package.json
```diff
@@ -76,6 +76,8 @@
     "babel-jest": "^24.5.0",
     "babel-loader": "^8.2.4",
     "commander": "^7.2.0",
+    "axios": "^0.26.1",
+    "dotenv": "^16.0.0",
```
adamstankiewicz, line 80: I think we may want to create a new package.json file in `component-generator` directory to house its specific dependencies rather than adding them to the global Paragon (library) package.json 🤔

component-generator/utils.js
monteri: Yes, it is moved

component-generator/utils.js
monteri: Changed to `POST`

package.json
monteri: Initialized `package.json` in the `component-generator` directory. Updated command to launch the script inside root directory. Moved dependencies to `component-generator`

README.md
```diff
@@ -67,7 +67,7 @@ npm start
 To create new component run
 
 ```
-npm run generate-component -- --componentName=MyComponent
+npm run generate-component MyComponent
```
monteri, line 70: Probably it will simplify usage

monteri:
@adamstankiewicz I added installations of the `component-generator` packages to the ci script. Is it good idea or what will be the appropriate solution for this? Because if I don't install dependencies from the `component-generator` tests fail

adamstankiewicz:
> Author

adamstankiewicz (approved):
Nice work! Using `netlify dev`, I was able to run this locally and see an event come through in the Segment event debugger. Open edX platform can say it uses serverless functions now haha 😉 

image

adamstankiewicz:
(Accidentally closed and then reopened!)

> I added installations of the component-generator packages to the ci script. Is it good idea or what will be the appropriate solution for this? Because if I don't install dependencies from the component-generator tests fail

@monteri I think this should be OK? I'm going to merge it as is, but let me know if you have concerns!

openedx-webhooks:
@monteri 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1233&entry.1484655318=https://github.com/monteri&entry.752974735=2022-04-15+08:45&entry.1917517419=2022-04-29+04:42&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1257, opened 2022-04-28
[BD-46] feat: create consumer app

monteri:
Consumer app in the  `example` folder:
- used guidance and code from the [frontend-platform](https://github.com/openedx/frontend-platform) (`example` directory) and configuration from the [frontend-build](https://github.com/openedx/frontend-build#local-module-configuration-for-webpack) 
- tested with devstack launched. Set up `.env.development` variables that are required for login
- set up `ESLint` for `example/` and added to the `npm run lint` root command
- added simple component with paragon components

JIRA: https://openedx.atlassian.net/browse/PAR-798

Questions:
- Probably we would need add something to `README.md`?
- Is usage in the `MyComponent` okay?


openedx-webhooks:
Thanks for the pull request, @monteri! I've created [BLENDED-1220](https://openedx.atlassian.net/browse/BLENDED-1220) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


example/module.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  localModules: [
+    { moduleName: '@edx/paragon/scss/core', dir: '..', dist: 'scss/core' },
+    { moduleName: '@edx/paragon/icons', dir: '..', dist: 'icons' },
+    { moduleName: '@edx/paragon', dir: '..', dist: 'dist' },
```
adamstankiewicz, line 5: We could probably use `dist: 'src'` here so the example app gets hot reloading of file changes in the `src` directory without requiring a full `npm run build` to get files in `dist`?

.eslintignore
```diff
@@ -6,3 +6,4 @@ icons/
 dependent-usage-analyzer/
 build-scss.js
 component-generator/templates/test.jsx
+example/
```
adamstankiewicz, line 9: [curious] If we're adding `example` to .eslintignore, is the NPM script `example-lint` necessary?

monteri: I think we can remove. I added to have opportunity to run `eslint` from root but I see it is not necessary. The origin intention was to add `eslint` for root directory but then I should also add `npm install` for the `example`. This takes some additional time so I decided it as not good idea. Tests will last longer although `example` is not what should be tested every time.

monteri:
@adamstankiewicz Added `npm run install-all` command also

monteri:
> @adamstankiewicz Added `npm run install-all` command also

JIRA: https://openedx.atlassian.net/browse/PAR-810

PR #1267, opened 2022-05-05
[BD-46] feat: add typescript support

viktorrusakov:
## Description

Add TypeScript support

### Deploy Preview

## Merge Checklist

## Post-merge Checklist


src/Bubble/index.tsx
```diff
@@ -1,15 +1,19 @@
...
+import * as PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-const STYLE_VARIANTS = [
-  'primary',
-  'success',
-  'error',
-  'warning',
-];
+const STYLE_VARIANTS = ['primary', 'success', 'error', 'warning'] as const;
```
monteri, line 5: Is it good idea to change to `enum`?
```
export enum STYLE_VARIANTS {
  PRIMARY = 'primary',
  SUCCESS = 'success',
  ...
}
```
and then in the `interface` we can
```
...
variant?: STYLE_VARIANTS,
...
```

adamstankiewicz: I agree, an enum is cleaner for the types, but `STYLE_VARIANTS` is still used in the prop types. That said, once a component has types for the props as `Bubble` now does, the `PropType` definitions become less meaningful. They'd really only used for the purpose of forming the "Props API" tables on the component pages in the documentation site, I believe.

Will we need to keep both types and prop types moving forward or I wonder if we should do some technical discovery around migrating away from prop types while keeping the same functionality with generating the props api tables.

tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "noEmit": false,
+    "jsx": "react-jsx"
+  },
+  "include": [
+    "src"
+  ]
+}
```
adamstankiewicz, line 26: nit: newline

src/withDeprecatedProps.tsx
```diff
@@ -69,13 +81,14 @@ function withDeprecatedProps(WrappedComponent, componentName, deprecatedProps) {
...
-        <WrappedComponent {...transformedProps}>
-          {this.props.children || children /* eslint-disable-line react/prop-types */}
+        <WrappedComponent {...transformedProps as T}>
+          {this.props.children || children}
         </WrappedComponent>
       );
     }
   }
 
+  // @ts-ignore
```
adamstankiewicz, line 91: [curious] why does this need to be ts-ignore'd

adamstankiewicz (commented):
Looking good! Left a couple clarifying questions.

src/withDeprecatedProps.tsx
viktorrusakov: TypeScript complained that `displayName` property does not exist in the component, so you can't assign it.

I force-pushed, sou you can't see anymore, but `ts-ignore` complained about line
```js
WithDeprecatedProps.displayName = `withDeprecatedProps(${componentName})`;
```

src/Bubble/index.tsx
viktorrusakov: Yeah, I didn't use enum because it conflicts with propTypes, I didn't find a good way to reuse it in propTypes definition, so that it correctly displays in Props API table...

I think we will need to keep propTypes even when we move to TypeScript types, because they have different purpose - TypeScript types work only during compile type while propTypes during runtime, this means if we remove propTypes and consumers pass some variant that is not in `STYLE_VARIANTS` they will not get a warning that this prop is invalid. I guess they would if they have TypeScript setup inn their projects... but I believe most MFEs (if not all of them) do not use TypeScript

Makefile
```diff
@@ -1,8 +1,11 @@
 build:
 	rm -rf ./dist
-	./node_modules/.bin/babel src --config-file ./babel.config.json --out-dir dist --source-maps --ignore **/*.test.jsx,**/__mocks__,**/__snapshots__,**/setupTest.js --copy-files
+	tsc --emitDeclarationOnly
+	./node_modules/.bin/babel src --config-file ./babel.config.json --out-dir dist --source-maps --ignore **/*.test.jsx,**/*.test.tsx,**/__mocks__,**/__snapshots__,**/setupTest.js --copy-files --extensions ".tsx,.jsx"
 	# --copy-files will bring in everything else that wasn't processed by babel. Remove what we don't want.
-	rm -rf dist/**/*.test.jsx
```
viktorrusakov, line 5: This didn't work as expected I think, build still had tests files inside `tests` folder (e.g. `Datatable/tests/` is still in the `dist` directory with tests files in there), I think it deleted files only one level deep not going into all subdirectories.

Rewrote this to make sure we delete all tests and md files from the build

src/Bubble/index.tsx
adamstankiewicz: > I guess they would if they have TypeScript setup inn their projects... but I believe most MFEs (if not all of them) do not use TypeScript

This is correct. The TypeScript in this PR is the first TypeScript introduced in the Open edX platform, AFAIK. It's a great callout for build time vs. runtime validation in TypeScript and propTypes; I agree it makes sense we should probably use both for the time being, until we learn otherwise.

package.json
```diff
@@ -41,7 +41,10 @@
     "generate-component-lint": "cd component-generator && npm run lint",
     "generate-changelog": "node generate-changelog.js",
     "i18n_compile": "formatjs compile-folder --format transifex ./src/i18n/messages ./src/i18n/messages",
-    "i18n_extract": "formatjs extract 'src/**/*.{jsx,js}' --out-file ./src/i18n/transifex_input.json --format transifex"
+    "i18n_extract": "formatjs extract 'src/**/*.{jsx,js,tsx,ts}' --out-file ./src/i18n/transifex_input.json --format transifex",
+    "type-check": "tsc --noEmit",
+    "type-check:watch": "npm run type-check -- --watch",
+    "build-types": "tsc --emitDeclarationOnly"
```
adamstankiewicz, line 47: It might be a good idea to add documentation about what these commands do and when/why we should use them somewhere in the repo. It very well could be the starting documentation for how to work with TypeScript for all of Open edX.

Related, did we write an ADR (Architectural Decision Record) about TypeScript? If not, that's probably something we should do. Is this something you might want to take on? Let's treat that as a task separate from this PR, though 😄

adamstankiewicz (approved):
This PR looks good to me! 🚀 I added just one comment about a potential follow-on ADR/documentation task for TypeScript, but let's get this PR in first. 

Likely due to the recently merged security vulnerability package upgrades PR, there's bound to be quite a few merge conflicts on this PR unfortunately, though 😬

viktorrusakov:
@adamstankiewicz rebased, should be good to go!

openedx-webhooks:
@viktorrusakov 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1267&entry.1484655318=https://github.com/viktorrusakov&entry.752974735=2022-05-05+13:15&entry.1917517419=2022-08-05+12:32&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1288, opened 2022-05-18
[BD-46] docs: add 'Responsive' page

viktorrusakov:
## Description

Implement single source of truth for breakpoints used in Paragon, this is the improved version of [reverted PR](https://github.com/openedx/paragon/pull/1242) that fixes issues with MFE's

JIRA: [PAR-813](https://openedx.atlassian.net/browse/PAR-813)

### Deploy Preview

https://deploy-preview-1288--paragon-openedx.netlify.app/foundations/responsive

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/www/gatsby-config.js
+++ b/www/gatsby-config.js
@@ -13,7 +13,16 @@ const segmentPlugin = {
 };
 
 const plugins = [
-  "gatsby-plugin-sass",
+  {
+    resolve: `gatsby-plugin-sass`,
+    options: {
+     cssLoaderOptions: {
+       modules: {
+         namedExport: false,
+       },
+     },
+   },
+  },
   `gatsby-plugin-react-helmet`,
   {
     resolve: `gatsby-plugin-manifest`,
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -25,7 +25,10 @@
     "at-rule-no-unknown": null,
     "scss/at-rule-no-unknown": true,
     "string-quotes": "double",
-    "scss/at-import-partial-extension": "never",
-    "scss/comment-no-empty": null
+    "scss/at-import-partial-extension": null,
+    "scss/comment-no-empty": null,
+    "property-no-unknown": [true, {
+      "ignoreProperties": ["xs", "sm", "md", "lg", "xl", "xxl"]
+    }]
   }
 }
--- a/scss/core/_exports.module.scss
+++ b/scss/core/_exports.module.scss
@@ -4,19 +4,19 @@
 // adapting to different screen sizes, for use in media queries.
 
 $grid-breakpoints: (
-        xs: 0,
-        sm: 576px,
-        md: 768px,
-        lg: 992px,
-        xl: 1200px,
-        xxl: 1400px
+  xs: 0,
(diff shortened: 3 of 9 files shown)


openedx-webhooks:
Thanks for the pull request, @viktorrusakov! I've created [BLENDED-1232](https://openedx.atlassian.net/browse/BLENDED-1232) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


viktorrusakov:
@adamstankiewicz not sure if we need to merge this actually... I believe with this implementation consumers won't be able to override breakpoints used in `.js` files because they are always imported from Paragon's `.scss` file. I don't think there is a way to fix it 🤔 Sadly we can't use scss file as a single source of truth it seems 

Another idea I have is to create a context for breakpoints and export if from Paragon so consumers could override those values in all component they use in their apps, this would require them to override both css value and values used in component but I don't think there is another way
What do you think?

www/gatsby-config.js
```diff
@@ -13,7 +13,16 @@ const segmentPlugin = {
 };
 
 const plugins = [
-  "gatsby-plugin-sass",
+  {
+    resolve: `gatsby-plugin-sass`,
+    options: {
+     cssLoaderOptions: {
+       modules: {
+         namedExport: false,
```
adamstankiewicz, line 21: A'ha!  🪄

adamstankiewicz (approved):
LGTM! 🚀

openedx-webhooks:
@viktorrusakov 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1288&entry.1484655318=https://github.com/viktorrusakov&entry.752974735=2022-05-18+08:16&entry.1917517419=2022-05-20+10:44&entry.2133058324=Yes) so we can improve your experience in the future.

adamstankiewicz:
> @adamstankiewicz not sure if we need to merge this actually... I believe with this implementation consumers won't be able to override breakpoints used in `.js` files because they are always imported from Paragon's `.scss` file. I don't think there is a way to fix it 🤔 Sadly we can't use scss file as a single source of truth it seems
> 
> Another idea I have is to create a context for breakpoints and export if from Paragon so consumers could override those values in all component they use in their apps, this would require them to override both css value and values used in component but I don't think there is another way What do you think?

@viktorrusakov Missed this comment before I merged, my bad! 🙃 

As we discussed in our meeting though, we agreed this is likely an issue but are going to defer on it for the time being as the current state is that consumers already cannot override the `breakpoints.js` import from Paragon in a custom SCSS theme. I filed a new issue that we can prioritize down the road to investigate a solution to allow custom themes to override the values in the `breakpoints.js` import, or a different solution altogether.

PR #1353, opened 2022-06-03
build: use shared browserslist config

Jawayria:
Removes custom browserslist configuration in favor of using a [shared configuration](https://github.com/edx/browserslist-config).
This change reduces the resultant asset bundle size.
Created this PR due to some issues in rebasing the [older PR](https://github.com/openedx/paragon/pull/868)

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "@babel/plugin-proposal-object-rest-spread": "^7.16.7",
     "@babel/preset-env": "^7.16.8",
     "@babel/preset-react": "^7.16.7",
+    "@edx/browserslist-config": "1.1.0",
     "@edx/eslint-config": "^3.0.0",
     "@formatjs/cli": "^4.8.4",
     "@semantic-release/changelog": "^5.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -46,8 +46,10 @@
     "@fortawesome/react-fontawesome": "^0.1.18",
     "@popperjs/core": "^2.11.4",
     "airbnb-prop-types": "^2.16.0",
+    "axios": "^0.27.2",
     "bootstrap": "^4.6.1",
     "classnames": "^2.3.1",
+    "dotenv": "^16.0.1",
```
adamstankiewicz, line 52: [curious] Why were these packages added to the root package.json file?

adamstankiewicz: @Jawayria Bump.

Jawayria: I guess there was a mistake, thanks for pointing out.

adamstankiewicz (approved):
LGTM, though the latest version of `@edx/browserslist-config` is now 1.1.0, so we could use that instead of 1.0.0 before we merge.

Issue #1308, opened 2022-05-20
Add analytics for non-component exports (e.g., hooks) to Paragon Usage Insights


adamstankiewicz:
Original ticket: https://openedx.atlassian.net/browse/PAR-822

Currently, the [Usage Insights](https://paragon-openedx.netlify.app/insights) page only shows components that are imported by consumers. It does not display any insights into the usage of the hooks that Paragon provides (e.g., `useToggle`, etc.) or the `breakpoints` export, so we have limited visibility into whether/how consumers are using these exports.

**AC**

- “Summary” tab
  - Change “…with an average of X component instances per project.” to “…with an average of X imports per project.” since the aggregate number here should be about all Paragon usage, not just components.
  - Change “Overall component usage” to “Overall usage”
  - In the summary table, include all imports (e.g., from hooks and breakpoints), not just components.
    - Add new column “Type” that represents whether an import is a component, hook, or other utility function/variable (e.g., breakpoints)
    - Add ability to filter by “Type” column with checkboxes
    - Sort expandable table row contents by project name alphabetically (ascending)
- “Projects” tab
  - Change “Instance Count” in table header to “Import Count”
  - Sort expandable table row contents by component name alphabetically (ascending)
  - Add `Tabs` inside expandable table row to differentiate usage between “Components” and “Hooks” and “Utility” (e.g., breakpoints; open to other name suggestions!)
- “Components” tab
  - Don’t believe any changes are needed.
- Add new “Hooks” tab
  - Similar to “Components” tab, but showing hooks usage instead (e.g., `useToggle`, `useWindowSize`).
  - This should include any pass-thru hooks, as well (e.g., `useMediaQuery`). Refer to index.js for all exports.
- Add new “Utility” tab (open to other name suggestions!)
  - Similar to “Components” tab, showing utility imports (e.g., `breakpoints`, `Variant`)

PR #1365, opened 2022-06-09
[BD-46] docs: change behaviour of insights

Inferato:
BREAKING CHANGE: the behavior of the usage analyzer was changed(The dependent-usage.json have a modified structure) and it reflects on the insight page

## Description

**AC**
- Summary” tab
    - Change “…with an average of X component instances per project.” to “…with an average of X imports per project.” since the aggregate number here should be about all Paragon usage, not just components.
    - Change “Overall component usage” to “Overall usage”
In the summary table, include all imports (e.g., from hooks and breakpoints), not just components.
    - Add new column “Type” that represents whether an import is a component, hook, or other utility function/variable (e.g., breakpoints)
    - Add ability to filter by “Type” column with checkboxes (example [here](https://paragon-openedx.netlify.app/components/datatable/#sidebar-filter))
    - Sort expandable table row contents by project name alphabetically (ascending)

- “Projects” tab
    - Change “Instance Count” in table header to “Import Count”
    - Sort expandable table row contents by component name alphabetically (ascending)
    - [TO DO] Add `Tabs` inside expandable table row to differentiate usage between “Components” and “Hooks” and “Utility” (e.g., breakpoints; open to other name suggestions!)
- Components” tab
    - Don’t believe any changes are needed.
- Add new “Hooks” tab
    - Similar to “Components” tab, but showing hooks usage instead (e.g., `useToggle`, `useWindowSize`).
    - This should include any pass-thru hooks, as well (e.g., `useMediaQuery`). Refer to `index.js` for all exports.
- Add new “Utility” tab (open to other name suggestions!)
    - Similar to “Components” tab, showing utility imports (e.g., `breakpoints`, `Variant`)

https://github.com/openedx/paragon/issues/1308

### Deploy Preview

https://deploy-preview-1365--paragon-openedx.netlify.app/insights

## Merge Checklist

## Post-merge Checklist


natabene:
@Inferato Thank you for your contribution. Is that part of the Blended work?

viktorrusakov:
@natabene Yes, it is, sorry for the confusion, I've added `BD-46` label.

www/src/pages/insights.jsx
```diff
@@ -88,6 +120,13 @@ const SummaryUsage = () => {
             accessor: 'name',
           },
           { Header: 'Instance Count', accessor: 'count' },
```
adamstankiewicz, line 122: Let's add `disableFilters` to this column so there isn't a text filter input field for this column. It doesn't really make sense to want to filter by a specific instance count here.

dependent-usage-analyzer/index.js
```diff
@@ -126,6 +126,30 @@ function getComponentUsagesInFiles(files, rootDir) {
           addComponentUsage(fullComponentName, node.loc.start);
         }
       },
+      AssignmentExpression(node) {
```
adamstankiewicz, line 129: It might be helpful to add comments for each of these different types to describe their purpose for future us and other developers. For example, what does an `AssignmentExpression` represent? What about a `CallExpression`? Adding a couple comments in code could help clarify.

www/src/components/insights/insights.scss
```diff
@@ -0,0 +1,3 @@
+.pgn__data-table-layout-wrapper .pgn__data-table-layout-sidebar {
+  padding: .75rem;
```
adamstankiewicz, line 2: Curious why we needed to change the default padding for `DataTable` here? We should try to minimize custom CSS styling when using Paragon components as a consumer, as we are here.

www/src/context/SettingsContext.jsx
```diff
@@ -1,7 +1,9 @@
 import React, { createContext, useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Helmet } from 'react-helmet';
+import * as Components from '~paragon-react'; // eslint-disable-line
 import { THEMES } from '../../theme-config';
+import getParagonComponentsTypes from '../utils/getParagonComponentsTypes';
```
adamstankiewicz, line 6: Hmm, these feel a bit strange to be adding to a `SettingsContext` which should be bounded to the context of what settings are available for the user in the "Settings" `Sheet` by clicking the gear icon in the header.

Also, `SettingsContext` wraps the entire Gatsby application, where presumably the `getParagonComponentsTypes` is only relevant to the "Insights" page. Should there be an `InsightsContext` specific to the "Insights" page if it needs to be placed in a context instead?

www/src/pages/insights.jsx
```diff
@@ -44,36 +49,63 @@ const componentsUsage = dependentProjectsUsages.reduce((accumulator, project) =>
...
+  const filterValues = Object.keys(paragonTypes)
+    .map((key) => paragonTypes[key])
+    .filter((v, i, a) => a.indexOf(v) === i)
+    .map(type => ({ name: type, number: typeCount[type], value: type }));
+
   const summaryTableData = summaryComponentsUsage.sort((a, b) => {
     const nameA = a.name.toUpperCase();
     const nameB = b.name.toUpperCase();
-    return (nameA < nameB) ? -1 : (nameA > nameB) ? 1 : 0;
+    return nameA < nameB ? -1 : 1;
```
adamstankiewicz, line 86: Is this fundamentally changing the sort logic? It's removing the case where `nameA` is the same as `nameB`, [e.g.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort):
```js
function compare(a, b) {
  if (a is less than b by some ordering criterion) {
    return -1;
  }
  if (a is greater than b by the ordering criterion) {
    return 1;
  }
  // a must be equal to b
  return 0;
}
```

www/src/pages/insights.jsx
```diff
@@ -216,3 +299,23 @@ InsightsPage.propTypes = {
...
+
+ComponentUsage.propTypes = {
+  name: PropTypes.string.isRequired,
+  componentUsageInProjects: PropTypes.arrayOf(PropTypes.shape({
+    name: PropTypes.string,
+    folderName: PropTypes.string,
+    version: PropTypes.string,
+    repositoryUrl: PropTypes.string,
+    componentUsageCount: PropTypes.number,
+    usages: PropTypes.arrayOf(PropTypes.shape({})),
```
adamstankiewicz, line 311: Can we expand on what this `PropTypes.shape({ ... })` should contain?

www/src/utils/getParagonComponentsTypes.js
```diff
@@ -0,0 +1,52 @@
+const MANUALLY_INCLUDED_COMPONENTS = {
```
adamstankiewicz, line 1: [curious] Why were these specific ones necessary to manually handle?

www/src/utils/getParagonComponentsTypes.js
```diff
@@ -0,0 +1,52 @@
...
+    if (componentType === 'Component') {
+      Object.keys(component).forEach(subcomponentName => {
+        const subcomponent = component[subcomponentName];
+
+        if (subcomponent && ((typeof subcomponent === 'function'
+        && subcomponentName[0] === subcomponentName[0].toUpperCase())
+        || (subcomponent.constructor.name === 'Object' && !!subcomponent.render)
+        || (subcomponentName === 'Consumer' || subcomponentName === 'Provider'))) {
+          componentsWithTypes[`${componentName}.${subcomponentName}`] = componentType;
+        }
```
adamstankiewicz, line 45: This is a pretty complex conditional. Could we break it up into its subparts with extra variables? It's hard to follow exactly why we're adding the conditionals we are here. For example:

```
(subcomponentName === 'Consumer' || subcomponentName === 'Provider')
```

Why is this necessary? Are components with these names not recognized as components for some reason?

adamstankiewicz (commented):
This will be great to have these extra insights into how the library is used by consumers! I left a few comments/questions/suggestions.

In addition, I noticed that when the "Showing X of Y" counts don't properly update when filtering. For example, in the screenshot below, I have "Hooks" applied with 16 results per the count in the sidebar but the table is saying there are 140 results. Along those lines, I'm actually confused what the 16 is referring to as well. Since there's 3 hooks returned in the table with that type filter applied, I would have expected the count in the filters sidebar for "Hooks" to be equal to 3, matching the number of total rows in the table:

image

Also, I don't know if it's just me, but the performance running this locally feels a bit sluggish to apply filters. Do you see any issues with performance on your end?

www/src/pages/insights.jsx
monteri: Added `disableFilters`

dependent-usage-analyzer/index.js
monteri: Yes, I added for all methods in `walk.simple`

www/src/components/insights/insights.scss
monteri: Actually, I found our standard paddings to small and decided to match padding for the table header.
image
I also went to the Figma and found that it has difference in paddings also.
https://www.figma.com/file/eGmDp94FlqEr4iSqy1Uc1K/Paragon-Design-System?node-id=10469%3A38059
Maybe it is good idea to improve styles for these filter `Card` in the `DataTable`?

www/src/context/SettingsContext.jsx
monteri: Yes, correct, I added `InsightsContext`. Is it good idea to wrap also whole application? I just think of `getParagonComponentsTypes` is loaded before user goes to `/insights` so that page will be rendered faster. Is it correct?

www/src/pages/insights.jsx
monteri: `nameA < nameB ? -1 : 1;` will be okay to omit `0` case. Only `-1` case results into values swap, `0` and `1` leave them as they are in the array. Correct me if I am wrong. Also in the table we can not have duplicates.
image

www/src/utils/getParagonComponentsTypes.js
monteri: I have just realized that in the `Card` after overriding we lost `Img`, `Title` and `Subtitle`. I add to the Card and remove `MANUALLY_INCLUDED_COMPONENTS `.
```
Card.Img = BaseCard.Img;
Card.Title = BaseCard.Title;
Card.Subtitle = BaseCard.Subtitle;
```

www/src/pages/insights.jsx
monteri: Sure, updated

www/src/pages/insights.jsx
```diff
@@ -138,7 +184,7 @@ const ComponentUsage = ({ name, componentUsageInProjects }) => (
     <DataTable
       isExpandable
       isSortable
-      itemCount={componentUsageInProjects.length} // eslint-disable-line
+        itemCount={componentUsageInProjects.length} // eslint-disable-line
```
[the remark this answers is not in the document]
monteri: Thank you, fixed

src/Card/index.jsx
```diff
@@ -52,6 +52,9 @@ Card.defaultProps = {
   isClickable: false,
 };
 
+Card.Img = BaseCard.Img;
+Card.Title = BaseCard.Title;
+Card.Subtitle = BaseCard.Subtitle;
```
adamstankiewicz, line 57: Hm, so I don't think we actually want to expose these subcomponents any more, since card titles and subtitles should come through `Card.Header` now and `Card.Img` was largely replaced by `Card.ImageCap`.

www/gatsby-browser.js
```diff
@@ -1,8 +1,16 @@
 const React = require('react');
 const { SettingsContextProvider } = require('./src/context/SettingsContext');
+const { InsightsContextProvider } = require('./src/context/InsightsContext');
+
 
 // wrap whole app in settings context
-exports.wrapRootElement = ({ element }) => <SettingsContextProvider>{element}</SettingsContextProvider>;
+exports.wrapRootElement = ({ element }) => (
+  <SettingsContextProvider>
+    <InsightsContextProvider>
```
adamstankiewicz, line 9: [curious] Why would these entire documentation site need access to the `InsightsContext`? Should this be isolated to just the "Usage Insights" page/route?

www/src/context/InsightsContext.jsx
```diff
@@ -0,0 +1,35 @@
...
+  useEffect(() => {
+    setParagonTypes(getParagonComponentsTypes(Components));
+  }, []);
+
+  const isParagonIcon = (name) => name in Icons;
+
+  const contextValue = {
+    paragonTypes,
+    isParagonIcon,
+  };
```
adamstankiewicz, line 21: Generally, context values should be memoized (e.g., with `useMemo`) to avoid re-renders. Each re-render of `InsightsContextProvider` would cause `contextValue` to get re-created, which will ultimately causes any children of `InsightsContext.Provider` to re-render as well.

adamstankiewicz (commented):
Left a few comments/questions; two other things worth asking:
1. Performance. Running this locally, it seems to be pretty laggy and occasionally freezes the Chrome tab. Are you seeing anything similar, or is this a "just on my machine" issue?
2. When applying filters in the sidebar, it looks like it's updating the table contents but the count in the status "Showing 20 of 207" doesn't seem to change. For example, when there's a filter of "hooks" with just 5 results, I think that message would update to "Showing 5 of 5"?

src/Card/index.jsx
monteri: Yes, I understand, removed them

www/gatsby-browser.js
monteri: I did some investigation and probably there is only way to wrap the whole app instead of just "Usage Insights"

www/src/context/InsightsContext.jsx
monteri: Wrapped in `useMemo`. It will be triggered once when `useEffect` changes value for `paragonTypes`.

www/src/pages/insights.jsx
```diff
@@ -203,7 +320,16 @@ export default function InsightsPage({ pageContext: { tab } }) {
             <ProjectsUsage />
           </Tab>
           <Tab eventKey={INSIGHTS_TABS.COMPONENTS} title="Components">
-            <ComponentsUsage />
+            <ComponentsUsage data={components} />
```
adamstankiewicz, line 323: So, running this locally, I'm seeing it have pretty degraded performance when interacting with the UI (e.g., clicking filters, changing tabs, etc.).

I think this is because we're trying to render the contents of all the tabs even for tabs that are not active, thus doing a lot of unnecessary extra work.

Could we conditionally render the tab content only if the tab is active (example below)? Doing so seems to drastically improve the performance of this page for me locally.

```jsx
<Tab eventKey={INSIGHTS_TABS.SUMMARY} title="Summary">
  {tab === INSIGHTS_TABS.SUMMARY && (
    <SummaryUsage />
  )}
</Tab>
```

www/src/pages/insights.jsx
```diff
@@ -43,40 +53,79 @@ const componentsUsage = dependentProjectsUsages.reduce((accumulator, project) =>
...
         <h2>Overview</h2>
         <p>
           Paragon is used by at least <strong>{dependentProjects.length} projects</strong>, with an average
-          of <strong>{round(averageComponentsUsedPerProject)} component instances</strong> per project.
+          of <strong>{round(averageComponentsUsedPerProject)} imports</strong> per project.
         </p>
       </div>
-      <h3>Overall component usage</h3>
+      <h3>Overall usage</h3>
       <DataTable
```
adamstankiewicz, line 118: From the previous review:

> When applying filters in the sidebar, it looks like it's updating the table contents but the count in the status "Showing 20 of 207" doesn't seem to change. For example, when there's a filter of "hooks" with just 5 results, I think that message would update to "Showing 5 of 5"?

I think this is still an outstanding issue? In the screenshot below, I have "Hooks" and "Objects" selected; the entire data set of results is only 5 results with those filters applied, but the table status is saying "Showing 5 **of 207**" instead of "Showing 5 **of 5**".

As is, the "Showing 5 of 207" hints there should be table pagination, but with those filters applied there isn't 😄 

image

monteri: Hm, I agree with this. I went to check how it works in examples in the `DataTable` and seems it is default behavior for filtering. As for me it is also correct to display `Showing 5 of 5.` instead of `of 207.`
image
image
I think meaning of `Showing 5 of 207.` is perfect when `isPaginated` is not used. But with pagination we can not see how much records per chosen filters. So maybe it is good idea to change behavior for this label when `isPaginated` is passed in `DataTable` component. What do you think?

www/src/pages/insights.jsx
monteri: I updated to the suggestion. But now probably the new problem is the loading time of the tab change.

monteri:
Updated

adamstankiewicz (approved):
Thanks for the work on this update! LGTM!

openedx-webhooks:
@Inferato 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1365&entry.1484655318=https://github.com/Inferato&entry.752974735=2022-06-09+14:25&entry.1917517419=2022-10-21+12:38&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1378, opened 2022-06-13
[BD-46] docs: improve design and usability of Icons page

monteri:
## Description

Features and fixes:
- according to big number of icons it required time to render all the icons. To avoid instant rendering of 2000+ icons `react-window` was used
- search against 2000+ icons is also a heavy operation. To avoid multiple searches `debounce` from `lodash` was used
- preview of the icon is added on the top of the table. If user scrolls to the bottom of the table and clicks on some icon then he is auto-scrolled to the preview.
- added styles
- fix for the SearchField: re-rendering of the `Wrapper` inside the `SearchField` makes the inner components to unmount and mount again that causes problems when writing a text in the input. Added `useCallback` to prevent re-rendering

### Deploy Preview

[Icons page](https://deploy-preview-1378--paragon-openedx.netlify.app)

## Merge Checklist

## Post-merge Checklist


openedx-webhooks:
Thanks for the pull request, @monteri! I've created [BLENDED-1314](https://openedx.atlassian.net/browse/BLENDED-1314) to keep track of it in Jira.   More details are on the [BD-46](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3134685276/BD-46+Paragon+Blended+Support+and+Maintenance) project page.


viktorrusakov (commented):
Left couple nits, otherwise LGTM!

Also in the deploy preview I noticed something weird in the deploy preview
image

But it's not caused by your changes, I see the same issue on the production site too and not just on the Icons page but on all non-component pages...
I'll file a separate ticket for that.

www/src/components/IconsTable.jsx
```diff
@@ -1,27 +1,177 @@
...
-          <h3 className="mb-0 mr-3 flex-grow-1 h5">{iconName}</h3>
-          <code className="bg-light px-2 py-1 rounded">
+          <code
+            role="presentation"
+            className="rounded"
+            onClick={() => copyToClipboard(currentIconImport)}
+          >
             <small>
-              {`import { ${iconName} } from '@edx/paragon/icons';`}
+              {currentIconImport}
```
adamstankiewicz, line 109: I'm thinking it might be a good idea to disclose that it is possible to copy this import string, without having to stumble on it. Maybe we show an inline/small `ContentCopy` icon to the right of the import statement that is the trigger to copy?

www/src/components/IconsTable.jsx
```diff
@@ -1,27 +1,177 @@
...
-            src={IconComponents[iconName]}
+            key={currentIcon}
+            src={IconComponents[currentIcon]}
           />
-          <h3 className="mb-0 mr-3 flex-grow-1 h5">{iconName}</h3>
-          <code className="bg-light px-2 py-1 rounded">
+          <code
+            role="presentation"
+            className="rounded"
+            onClick={() => copyToClipboard(currentIconImport)}
```
adamstankiewicz, line 106: Let's add a Segment event to track clicks, e.g. `openedx.paragon.icons-table.selected-icon.copied`, and maybe pass the name of the icon as a `name` property as well?

www/src/components/IconsTable.jsx
```diff
@@ -1,27 +1,177 @@
...
+              {currentIconImport}
             </small>
           </code>
         </div>
-      ))}
-    </div>
+      )}
+      {iconsList.length ? (
+        <div className="pgn-doc__icons-table">
+          <AutoSizer>
```
adamstankiewicz, line 116: It appears there's a double-scroll situation in which this table can be scrolled as well as the main page viewport. Is there a way we can eliminate the need for the table to be scrollable?

If we're trying to eliminate displaying icons that aren't visible in the viewport, I'd think additional rows of icons would start appearing as the main page scroll gets closer to the end of the page, if that makes sense.

adamstankiewicz, line 116: When there aren't many icons displayed, it appears there's a lot of excessive/unnecessary whitespace between the end of the icons table and the footer.

image

www/src/components/IconsTable.scss
```diff
@@ -0,0 +1,84 @@
...
+    margin: 0 auto;
+  }
+}
+
+.pgn-doc-icons-table__grid {
+  margin: 0 auto;
+}
+
+.pgn-doc__icons-table__cell {
+  @include pgn-box-shadow(1, "centered");
```
adamstankiewicz, line 22: With the cells flush together, the box-shadow with adjacent cells kind of conflicts. I wonder if the UI might be a bit cleaner without any shadows/borders, except on hover/focus:

image

www/src/components/IconsTable.jsx
monteri: Yes, I added `ContentCopy ` to the right and hover styles

www/src/components/IconsTable.jsx
monteri: I changed to own infinite scroll implementation

monteri: Updated to `min-height: 200px`

adamstankiewicz (approved):
Awesome! I think this came out really cool 😄 Great work!

openedx-webhooks:
@monteri 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1378&entry.1484655318=https://github.com/monteri&entry.752974735=2022-06-13+13:10&entry.1917517419=2022-06-24+03:24&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1388, opened 2022-06-16
docs: ADR to expose all SCSS variables that MFEs use as CSS variables [BB-6307]

xitij2000:
## Description

Adds an ADR for a plan to expose all Bootstrap variables used by Paragon and MFEs as CSS variables.

### Deploy Preview

NA

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
+++ b/docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
@@ -1,5 +1,5 @@
-0018 CSS Variables
-##################
+0018 Expose CSS Variables for Consumers of Paragon
+##################################################
 
 
 Status
@@ -10,46 +10,90 @@ Draft
 Context
 *******
 
-.. This section describes the forces at play, including technological, political, social, and project local. These forces are probably in tension, and should be called out as such. The language in this section is value-neutral. It is simply describing facts.
-
 All Open edX MFEs use Bootstrap via Paragon for styling. Each MFE builds its
 own styling which is a combination of common styles from Bootstrap + Paragon,
 on which it addds its own SCSS.
 
+Currently the SCSS code is structured as follows:
+
+- At the code we have Bootstrap
+- Then we have Paragon, which uses Bootstrap and react-bootstrap
+- Then we have the branding package, that has SCSS files that can override
+  Paragon defaults and add additional SCSS code.
+- Then we have MFEs that import from the branding package and Paragon, and
+  usually hav some of their own SCSS code.
+
+Paragon uses Bootstrap variables, and the branding theme uses Paragon/Bootstrap
+variables, any changes to any variables mean that the entire codebase needs to
+be rebuilt. However, this need not be the case, if the dependencies between
+these packages is expressed via CSS variables instead of SCSS variables, then
+any changes to values will be picked up at runtime. Such a setup would allow
+building each of these packages separately and only redeploying the bits that
+change.
+
+This ADR only concerns itself consumers of Paragon, and as such still expects
+Paragon to continue using SCSS variables itself.
+
 If we want to be able to support a common pre-built core stylesheet that can be
-deployed independently of MFEs, then MFEs cannot use any SCSS variables from
-Paragon + Bootstrap since in that case any changes to the core theme will
-require a rebuild of each MFE.
-
-Decision
-********
-
-We will export all theme variables that an MFE could need as css variables so
-that these can be applyed at runtime rather than build time.
+used by all MFEs and deployed independently of MFEs, then MFEs cannot use any
+SCSS variables from Paragon + Bootstrap since in that case any changes to the
+core theme will require a rebuild of each MFE, and as such these variables
+should instead be used as CSS variables so that they can be looked up at
+runtime rather then build time.
 
 Bootstrap already exports a number of CSS variables (these can be found in
 `this document for Bootstrap 4.x`_), however they don't cover all usecases. For
 instance, usage of colour variables like `primary-400`, or `light-300` are
-abundant. Often they can be used via existing Bootstrap classes, but sometimes
-they can't.
+abundant in MFEs, however these are not expored by Bootstrap even though they
+are available as SCSS variables.
 
-Paragon can create CSS variable exports like: `--pa-blue`, `--pa-primary`,
-`--pa-primary-200`` etc. The reason for prefixing with `--pa` is to isolate
+Decision
+********
+
+We will export all theme variables that an MFE could need as CSS variables so
+that these can be applyed at runtime rather than build time.
+
+Paragon can create CSS variable exports like: `--pgn-blue`, `--pgn-primary`,
+`--pgn-primary-200`` etc. The reason for prefixing with `--pgn` is to isolate
 changes to these names in Bootstrap. For example, in Bootstrap 4.x these are
-named as `--blue, --primary` etc. however in `Bootstrap 5`_, they are prefixed
-with `--bs-blue, --bs-primary`.
+named as `--blue, --primary` etc. However in `Bootstrap 5`_, they are prefixed
+with `--bs-blue, --bs-primary`, this incompatibility can be handled at the
+Paragon level whenever it switches from 4.x to 5.x without needing changes from
+MFEs.
+
+Paragon will pick up a standardised set of SCSS variables such as those
+relating to colours, colour shades, font styles, and responsible breakpoints
+and expose these as CSS variables with a ``--pgn`` prefix. These variabels will
+be exported by the Paragon package, however, they will be use variables defined
+in the branding package thus allowing them to be redefined easily.
+
+Any current usage of SCSS variables in MFEs can then be replaced with the
+equivalent CSS variable.
+
 
 Consequences
 ************
 
-Wherever we are currently using SASS variables, we can now use CSS variables.
+Once this is implemented, MFEs, and other consumers of Pargaon can start using
+CSS variables, wherever they are currently using SASS variables. The
+limitiation right now is that many of the variables needed by MFEs are only
+available as SCSS variables.
+
 Since these will be defined in the common theme, that means rebuilding it once
 and deploying it to all MFEs (or to a common place shared by all MFEs) will
 automatically apply all the latest variables.
 
-With this, the common branding theme becomes a runtime dependency rather than a
-build time dependency, allowing it to be replaced at run time without needing to
-rebuild a growing number of MFEs.
+After this ADR we can create a common stylesheet containing Bootstrap + Paragon
++ the Branding package that is loaded by all MFEs. The MFEs themselves should
+not import this as SCSS but as compiled CSS and only use CSS variables. This
+way each MFE will then have two stylesheets, a common stylesheet used by all
+MFEs which also includes variable definitions, and a MFE-specific stylesheet
+that uses only CSS variables. So the common stylesheet becomes a runtime
+dependency rather than a build time dependency, allowing it to be replaced at
+run time without needing to rebuild a growing number of MFEs.
+
+Implementing this ADR will not accomplish the above, however it paves the way
+for such a change. It will be the subject of a future ADR.
 
 `CSS Variables are well supported`_ in major browsers since many years, and
 aligns well with the currently supported browsers.
@@ -63,7 +107,9 @@ broad use or can be made into a component, however there are cases where this
 is less useful. For instance with one-off UI elements that only make sense in a
 particular MFE (`example
 <https://github.com/openedx/frontend-app-learning/blob/6257cb4b588fc4f9903113e22b318a63d1ddfe8e/src/course-home/progress-tab/course-completion/CompletionDonutChart.scss#L57-L74>`_)
-2. Keep a common core theme but also rebuild each MFEs when it changes. This removes some of the benefits of a single theme file, such as a single common theme deployment, and quicker theme compilation and deployment.
+2. Keep a common core theme but also rebuild each MFEs when it changes. This
+removes some of the benefits of a single theme file, such as a single common
+theme deployment, and quicker theme compilation and deployment.
 
 
 References
--- a/docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
+++ b/docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
@@ -1,0 +1,75 @@
+0018 CSS Variables
+##################
+
+
+Status
+******
+
+Draft
+
+Context
+*******
+
+.. This section describes the forces at play, including technological, political, social, and project local. These forces are probably in tension, and should be called out as such. The language in this section is value-neutral. It is simply describing facts.
+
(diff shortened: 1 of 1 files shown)


openedx-webhooks:
Thanks for the pull request, @xitij2000!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
```diff
@@ -0,0 +1,75 @@
...
+Bootstrap already exports a number of CSS variables (these can be found in
+`this document for Bootstrap 4.x`_), however they don't cover all usecases. For
+instance, usage of colour variables like `primary-400`, or `light-300` are
+abundant. Often they can be used via existing Bootstrap classes, but sometimes
+they can't.
+
+Paragon can create CSS variable exports like: `--pa-blue`, `--pa-primary`,
+`--pa-primary-200`` etc. The reason for prefixing with `--pa` is to isolate
+changes to these names in Bootstrap. For example, in Bootstrap 4.x these are
+named as `--blue, --primary` etc. however in `Bootstrap 5`_, they are prefixed
```
adamstankiewicz, line 39: nit: Capitalize "however"

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
```diff
@@ -0,0 +1,75 @@
...
+All Open edX MFEs use Bootstrap via Paragon for styling. Each MFE builds its
+own styling which is a combination of common styles from Bootstrap + Paragon,
+on which it addds its own SCSS.
+
+If we want to be able to support a common pre-built core stylesheet that can be
+deployed independently of MFEs, then MFEs cannot use any SCSS variables from
+Paragon + Bootstrap since in that case any changes to the core theme will
+require a rebuild of each MFE.
+
+Decision
```
adamstankiewicz, line 24: I'm in full support of this ADR and decision. As you mentioned, we can start moving towards CSS variables in Paragon since Open edX officially no longer supports older browsers like IE 11, per our shared [`@edx/browserslist-config`](https://github.com/edx/browserslist-config).

There are still some open questions around the Bootstrap 5 upgrade, which is why it hasn't necessarily been prioritized yet. I agree with the approach to namespace all the Paragon specific CSS variables with `--pgn` so they are distinct from the CSS variables provided by Bootstrap, if/when we do upgrade.

For what it's worth, adopting CSS variables is on the Paragon Working Group team's roadmap, likely in a few weeks, as part of our effort to adopt a more scalable solution to design tokens, using something like `style-dictionary` ([demo](https://www.style-dictionary-play.dev/)), where design tokens are drafted as JSON files in a standard format, and then `style-dictionary` transforms/exports those design tokens into various other formats accepted by various platforms (e.g., JS, CSS variables, SCSS, iOS, Android, Figma, Sketch, etc.). We are likely planning to build out a prototype of this concept for design tokens in the next month or two through the Blended Development project BD-39 that I'm leading with Raccoon Gang.

What is the timeline you're looking for here in an implementation? Were you planning on making the contribution to Paragon to create these CSS variables yourself? If so, can you expand on this decision in your ADR to help explain the plan for how we'll manage the creation of these CSS variables in more detail?

I'm planning to draft an ADR for Paragon about `style-dictionary` that refers to this PR, as a way to implement design tokens and automation around generation of CSS and SCSS variables (among other platforms as well).

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
```diff
@@ -0,0 +1,75 @@
...
+Consequences
+************
+
+Wherever we are currently using SASS variables, we can now use CSS variables.
+Since these will be defined in the common theme, that means rebuilding it once
+and deploying it to all MFEs (or to a common place shared by all MFEs) will
+automatically apply all the latest variables.
+
+With this, the common branding theme becomes a runtime dependency rather than a
+build time dependency, allowing it to be replaced at run time without needing to
```
adamstankiewicz, line 51: Can we expand on this explanation in more detail? While I understand the concept, I think it would be useful to expand on the _how_ it can replaced at runtime without needing a rebuild in this ADR. Where will the CSS variables be created in the current structure, for example?

adamstankiewicz, line 51: [clarification] How does this approach to run time theming differ to run time theming with CSS-in-JS, with a `theme` object that you pass to a `ThemeProvider` type component? Are there advantages/disadvantages?

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
```diff
@@ -0,0 +1,75 @@
...
+Paragon can create CSS variable exports like: `--pa-blue`, `--pa-primary`,
+`--pa-primary-200`` etc. The reason for prefixing with `--pa` is to isolate
+changes to these names in Bootstrap. For example, in Bootstrap 4.x these are
+named as `--blue, --primary` etc. however in `Bootstrap 5`_, they are prefixed
+with `--bs-blue, --bs-primary`.
+
+Consequences
+************
+
+Wherever we are currently using SASS variables, we can now use CSS variables.
```
adamstankiewicz, line 45: [clarification] Are these consequences for the consumers of Paragon (e.g., MFEs), or for the component library itself? Or both? It might be worth separating out how each stakeholder might be impacted by CSS variables.

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
xitij2000: Sure, I will expand on this in the ADR.

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
xitij2000: Sure, I will do that. 

But roughly, my idea was to leave Paragon's usage of SCSS intact, since it is a shared library, so it can be built and deployed once. I think all of Paragon's SCSS is included with each MFE anyway so right now it doesn't matter as much. 

However, in the future if all of Paragon also uses CSS variables then it too can be built once, and ideally we boil down the compilation to just the variable definitions. The benefit of that will be much quicker compilation of the "theme", and potential near-real-time theme previewing.

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
```diff
@@ -0,0 +1,75 @@
...
+Context
+*******
+
+.. This section describes the forces at play, including technological, political, social, and project local. These forces are probably in tension, and should be called out as such. The language in this section is value-neutral. It is simply describing facts.
+
+All Open edX MFEs use Bootstrap via Paragon for styling. Each MFE builds its
+own styling which is a combination of common styles from Bootstrap + Paragon,
+on which it addds its own SCSS.
+
+If we want to be able to support a common pre-built core stylesheet that can be
```
farhaanbukhsh, line 19:
```suggestion
If there is a need to support a common pre-built core stylesheet that can be
```

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
farhaanbukhsh: I read up on [style-dictionary](https://amzn.github.io/style-dictionary/#/) the use case that they are trying to solve is maintaining the constant style across platforms while our need is to make the style changes independent of the build process, right?

The other point I have is the need to use CSS variable will help us to reduce/eliminate the need to use SCSS preprocessor. While if we try to migrate/use style-dictionary we will again fall into the parser trap. Where our build time is dependent on how fast the parser is.

Do we really need that?

farhaanbukhsh (changes requested):

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
xitij2000: I don't think eliminating SCSS is a goal here. Even if SCSS takes time to build it does provide a lot of benefits. Additionally, there will still need to be a pre-processor involved to deduplicate and minifiy the css code. 

This ADR is not currently too concerned about the build time for paragon or common theme. The idea is to make it possible to inject variables after the theme is built so that a lot of common theming usecases can be accomplished without recompiling dozens of MFEs. 

This is a first step I see. Going from needing to rebuild 12 MFEs to rebuilding one theme. The next step can then involve reducing how much of that theme needs to be rebuilt till we reach a point where a single css file having all the necessary variables is all that's needed, and can be even generated dynamically.

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
xitij2000: I'm assuming in this ADR that the need is there, so I think I'm happy with the original phrasing.

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
```diff
@@ -0,0 +1,143 @@
...
+Decision
+********
+
+We will export all theme variables that an MFE could need as CSS variables so
+that these can be applyed at runtime rather than build time.
+
+Paragon can create CSS variable exports like: `--pgn-blue`, `--pgn-primary`,
+`--pgn-primary-200`` etc. The reason for prefixing with `--pgn` is to isolate
+changes to these names in Bootstrap. For example, in Bootstrap 4.x these are
+named as `--blue, --primary` etc. However in `Bootstrap 5`_, they are prefixed
```
adamstankiewicz, line 78: Another reason I'd love for Paragon to move away from Bootstrap nomenclature is because consumers often rely on the internals of Bootstrap (classes, mixins, variables, etc), which in theory _should_ be just an implementation detail of the Paragon library, not really exposed to the consumer.

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
```diff
@@ -0,0 +1,143 @@
...
+with `--bs-blue, --bs-primary`, this incompatibility can be handled at the
+Paragon level whenever it switches from 4.x to 5.x without needing changes from
+MFEs.
+
+Paragon will pick up a standardised set of SCSS variables such as those
+relating to colours, colour shades, font styles, and responsible breakpoints
+and expose these as CSS variables with a ``--pgn`` prefix. These variabels will
```
adamstankiewicz, line 85: nit: variables

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
```diff
@@ -0,0 +1,143 @@
...
+in the branding package thus allowing them to be redefined easily.
+
+Any current usage of SCSS variables in MFEs can then be replaced with the
+equivalent CSS variable.
+
+
+Consequences
+************
+
+Once this is implemented, MFEs, and other consumers of Pargaon can start using
```
adamstankiewicz, line 96: nit: Paragon

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
```diff
@@ -0,0 +1,143 @@
...
+CSS variables, wherever they are currently using SASS variables. The
```
adamstankiewicz, line 97: I think it'd be great to strive to theming capabilities with either CSS variables _or_ SCSS variables depending on the needs of the consumer/application. Is the intention here to no longer rely on SCSS variables all, and if so, it's worth calling out the impacts of that on consumers in terms of necessary refactors, etc. We'd want to ensure we minimize breaking changes as much as possible.

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
dcoa, line 78: The link isn't redirect because the link reference is `Bootstrap 5.x`

dcoa: I think is a good idea find a runtime theming option and this ADR is a good start point, and could show the capabilities of CSS variables for the open edx ecosystem.

xitij2000:
I've created a starting implementation draft PR for this.

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
```diff
@@ -0,0 +1,143 @@
...
+
+- they can be changed at runtime by changing the stylesheet, and can be
+  mainpulated using JavaScript.
+- they can be scoped to a particular CSS class or media query.
+
+SCSS/SASS still offers a number of other advantages such as nested classes,
+mixins, functions, and utilities that are currently not available with CSS.
+
+All Open edX MFEs use Bootstrap via Paragon for styling. Each MFE builds its
+own stylesheet which is a combination Bootstrap, Paragon, and some MFE-specific
```
tecoholic, line 32:
```suggestion
own stylesheet which is a combination of Bootstrap, Paragon, and some MFE-specific
```

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
```diff
@@ -0,0 +1,143 @@
...
+
+This ADR only concerns itself consumers of Paragon, and as such still expects
+Paragon to continue using SCSS variables itself.
+
+If we want to be able to support a common pre-built core stylesheet that can be
+used by all MFEs and deployed independently of MFEs, then MFEs cannot use any
+SCSS variables from Paragon + Bootstrap since in that case any changes to the
+core theme will require a rebuild of each MFE, and as such these variables
+should instead be used as CSS variables so that they can be looked up at
+runtime rather then build time.
```
tecoholic, line 61:
```suggestion
runtime rather than build time.
```

docs/decisions/0018-expose-all-sass-vars-as-css-vars.rst
```diff
@@ -0,0 +1,143 @@
...
+
+Bootstrap already exports a number of CSS variables (these can be found in
+`this document for Bootstrap 4.x`_), however they don't cover all usecases. For
+instance, usage of colour variables like `primary-400`, or `light-300` are
+abundant in MFEs, however these are not expored by Bootstrap even though they
```
tecoholic, line 66:
```suggestion
abundant in MFEs, however these are not exported by Bootstrap even though they
```

openedx-webhooks:
@xitij2000 Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1388&entry.1484655318=https://github.com/xitij2000&entry.752974735=2022-06-16+12:32&entry.1917517419=2022-12-26+07:34&entry.2133058324=No) so we can improve your experience in the future.

PR #1400, opened 2022-06-20
[BD-46] feat: show truncation examples for Card component

PKulkoRaccoonGang:
## Description

Show truncation examples for Card component.

### Deploy Preview

[Card component](https://deploy-preview-1400--paragon-openedx.netlify.app/components/card/)

## Merge Checklist

## Post-merge Checklist


openedx-webhooks:


package.json
```diff
@@ -61,6 +61,7 @@
     "react-popper": "^2.2.5",
     "react-proptype-conditional-require": "^1.0.4",
     "react-responsive": "^8.2.0",
+    "react-string-clamp": "^0.3.1",
```
adamstankiewicz, line 64: Hmm, I have some reservations around adopting this NPM package. It doesn't appear broadly adopted in the JS ecosystem (only ~500 weekly downloads) and its last publish was 3 years ago, so I'd potentially worried around maintenance issues.

`react-truncate` also has a similar maintenance concern in that there hasn't been a release in several years but there have been more recent unreleased merge commits 👀 

At a higher level, though, I think it'd be helpful to export a standalone `Truncate` function from Paragon for consumers to use in their own applications and to be composed by other Paragon components, if needed. In theory, then consumers could pass in their own `Truncate` to the existing props API for `Card`, e.g.:

```jsx
<Card>
  <Card.Header
    title={(
      <Truncate lines={2}>
        edX Demonstration Course with A Really Long Title That Never Seems To End
      </Truncate>
    )}
  />
</Card>
```

The rationale for this design would be that it takes advantage of composability in our component design.

I guess my question to the team is what would it take to build our own `Truncate` component for Paragon, potentially influenced by `react-truncate` and/or `react-string-clamp`? FWIW, the props API for `react-truncate` is more in line with the component design patterns we have throughout Paragon, in terms of preferring the use of `children` over passing things in like `text` via props.

The base requirements for a standalone component would be as follows:
* Customize the number of lines of text should be preserved.
* Customize the ellipsis, but it should default to `…` (accepting string or another component)
* Provide an option to trim unnecessary whitespace.. That is, any extra spaces between a truncated string and the `…, e.g. "courses ..." would become "courses..."
* Standalone component page on the Paragon documentation site with example usage.

Maybe we repurpose this PR to make progress towards this? What do you think? 🤔 😃

PKulkoRaccoonGang: @adamstankiewicz yes, it`s an exciting proposal. I have decided to use Clamp because I observe problems with a responsive view and it does not correctly work when dynamically changing the container's width and the elements' padding. I think we should think about how we can make our tool the most convenient. Thanks

adamstankiewicz: Got it, yeah. I'll be curious to see what the team comes up with for a custom component :)

openedx-webhooks:
@PKulkoRaccoonGang Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1400&entry.1484655318=https://github.com/PKulkoRaccoonGang&entry.752974735=2022-06-20+12:31&entry.1917517419=2022-07-15+14:48&entry.2133058324=No) so we can improve your experience in the future.

PR #1439, opened 2022-07-04
[BD-46] feat: add Card Status

monteri:
## Description

- add `Card.Status` component
- add tests
- add example to see both `horizontal` and `vertical` variants

### Deploy Preview

[Card Page](https://deploy-preview-1439--paragon-openedx.netlify.app/components/card/#card-status)

## Merge Checklist

## Post-merge Checklist


openedx-webhooks:


src/Card/CardStatus.jsx
```diff
@@ -0,0 +1,49 @@
...
+import classNames from 'classnames';
+import Alert from '../Alert';
+
+const CardStatus = React.forwardRef(({
+  className,
+  children,
+  variant,
+  icon,
+}, ref) => (
+  <Alert
```
adamstankiewicz, line 12: We decided we will _not_ use the existing `Alert` for the following reasons:
* a11y requirements around screen readers (e.g., if there are many card with a Card.Status rendered at once).
* We would not like to create a dark variant for `Alert` to support this.

Instead, we'll replicate the `Alert` behavior without the a11y needs of it and all color variants of `Card.Status` as speced.

monteri: Updated. Added css classes for `CardStatus` instead of using the `Alert` component

src/Card/README.md
```diff
@@ -670,3 +671,53 @@ it is meant to be used as a single horizontal row of Cards, not as a grid. See C
   </Card>
 </CardDeck>
 ```
+
+### Card status
```
adamstankiewicz, line 675: I think the heading level should be up a level for the heading hierarchy on the page? As is, it looks like it's nested under `CardDeck`. Also, it might make sense to put "Card status" section further up on the page, maybe after "Horizontal variant".

src/Card/README.md
```diff
@@ -670,3 +671,53 @@ it is meant to be used as a single horizontal row of Cards, not as a grid. See C
...
+        <Form.RadioSet
+          name="orientation"
+          onChange={handleChangeOrientation}
+          value={orientation}
+        >
+          <Form.Radio value="vertical">Vertical</Form.Radio>
+          <Form.Radio value="horizontal">Horizontal</Form.Radio>
+        </Form.RadioSet>
+      </Form.Group>
+      <Card orientation={orientation} className="flex-column">
```
adamstankiewicz, line 698: If the `Card` is in a vertical orientation, should it have a `maxWidth` to limit its width so the horizontal vs. vertical variants are more distinguishable?

src/Card/README.md
```diff
@@ -670,3 +671,53 @@ it is meant to be used as a single horizontal row of Cards, not as a grid. See C
...
+        <Card.Body className={orientation === 'horizontal' ? 'd-flex' : ''}>
```
adamstankiewicz, line 699: image

I'm unsure about the "d-flex" class on the `Card.Body` here. I think the intention is that the header title would appear above the section title, like so:

image

src/Card/README.md
adamstankiewicz, line 698: Is `flex-column` a style property we would be able to programmatically add to `Card` when `orientation == "horizontal"` so the consumer doesn't have to remember to add it?

src/Card/README.md
```diff
@@ -670,3 +671,53 @@ it is meant to be used as a single horizontal row of Cards, not as a grid. See C
...
+            title="Card title"
+          />
+          <Card.Section
+            title="Section title"
+          >
+            This is a wider card with supporting text below as a natural lead-in to
+            additional content. This card has even longer content than the first to
+            show that equal height action.
+          </Card.Section>
+          <Card.Footer className="justify-content-end">
```
adamstankiewicz, line 710: I believe the buttons in the `Card.Footer` are already justified for `flex-end` in the vertical orientation; should we only apply this class name for the horizontal orientation? That said, this also raises the (design) question of whether the content inside of `Card.Footer` should have `justify-content: flex-end` for the horizontal orientation already 👀

src/Card/README.md
```diff
@@ -670,3 +671,53 @@ it is meant to be used as a single horizontal row of Cards, not as a grid. See C
...
+            <Button>Save</Button>
+            <Button variant="danger">Remove</Button>
+          </Card.Footer>
+        </Card.Body>
+        
+        <Card.Status icon={Warning}>
```
adamstankiewicz, line 716: I'm wondering if we should include visual examples of all the variants of `Card.Status` (maybe another form field to configure the example by, e.g. `Form.Control as="select"`?).

src/Card/CardStatus.jsx
```diff
@@ -0,0 +1,50 @@
...
+  /** Specifies variant to use. */
+  variant: PropTypes.oneOf([
+    'primary',
+    'secondary',
+    'success',
+    'danger',
+    'warning',
+    'info',
+    'light',
+    'dark',
```
adamstankiewicz, line 39: I'm wondering if we want to expose all these variants as options in the prop types if the design spec only calls for warning, success, danger, dark.

src/Card/CardStatus.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+const CardStatus = React.forwardRef(({
+  className,
+  children,
+  variant,
+  icon,
+}, ref) => (
+  <div
+    className={classNames('pgn__card-status',
+      `alert-${variant}`, 'alert', 'alert-content',
```
adamstankiewicz, line 14: [curious] Which color variant should a consumer use for the darker `Card.Status` variant (primary-500 with white text)? If we're simply using the `alert` classes here, we're still limited by how it's styled, which is using the `*-100` color level for each variant (e.g., primary-100) so as far as I can tell, there's no way to get this color variant? 

image

adamstankiewicz: Also related to the above screenshot, how should a consumer implement a heading and content inside of the `children` for this component?

If the API were similar to `Alert`, there might be a `Card.Status.Heading` component and allow something like:

```jsx
<Card.Status icon={Check} variant="success">
  <Card.Status.Heading>Locked lorem ipsum dolor sit amet</Card.Status.Heading>
  <p>
    Review your grades below for more details.
  </p>
</Card.Status>
```

However, it seems this would result in the following behavior currently:

image

src/Card/README.md
monteri: Placed under `horizontal` variant and made it as `h2` heading

src/Card/README.md
monteri: Right, added `w-50` class for vertical

src/Card/README.md
monteri: Put the `Card.Header` out of the `Card.Body`

src/Card/README.md
monteri: I guess `Card.Status` is a unique case for the `orientation` property because we want to have him always at the bottom. Because of that there are such workarounds with `className="flex-column"` and `d-flex` on the `Card.Body`. Actually passing `orientation={orientation}` is only useful for inner components that depend on the `Card.Context`.

src/Card/README.md
monteri: Until we have an answer I added `className={!isVertical ? 'justify-content-end' : ''}`

src/Card/README.md
monteri: Added `Form.Control as="select"` for variants. Added comment that this `jsx` snippet is for example only.

src/Card/CardStatus.jsx
monteri: For the first comment: I overridden `primary` variant using css only for `Card.Status` (`primary-500` background and white text)
For the second comment: I added `title` prop that is similar to `Card.Section` `title` prop

adamstankiewicz: This works as well :)

src/Card/Card.scss
```diff
@@ -252,6 +252,31 @@
   }
 }
 
+.pgn__card-status {
+  box-shadow: none;
+  margin-bottom: 0;
+
+  p {
+    margin-bottom: 0;
+  }
```
adamstankiewicz, line 261: Related to the other comment, if the consumer is passing any number of `<p>` tags, we'd only want the last `p` element to have no bottom margin; otherwise, there won't be space between any of the paragraphs.

src/Card/CardStatus.jsx
```diff
@@ -0,0 +1,53 @@
...
+const CardStatus = React.forwardRef(({
+  className,
+  children,
+  variant,
+  icon,
+  title,
+}, ref) => (
+  <div
+    className={classNames('pgn__card-status',
+      `alert-${variant}`, 'alert', 'alert-content',
```
adamstankiewicz, line 15: I'm still a bit hesitant to rely on the `alert`-related class names, even though the styles are similar/identical as it creates a coupling between `Card.Status` and `Alert` from a styles perspective that we might not actually want. For example, if we change `Alert` styling, it'll also impact `Card.Status`.

I think I'd still be in favor of not using any `alert-*` related class names from Bootstrap, in favor of replicating those styles specifically with `pgn` prefixed class names.

src/Card/CardStatus.jsx
```diff
@@ -0,0 +1,53 @@
...
+      className)}
+    ref={ref}
+  >
+    {icon && <Icon src={icon} className="alert-icon" />}
+    <div className="alert-message-content">
+      {title && <div className="alert-heading h4">{title}</div>}
+      {children && <p>{children}</p>}
```
adamstankiewicz, line 22: I think we can make `children` a required prop and always render it here.

src/Card/Card.scss
monteri: I also took an example from alert and set `margin-bottom: 0;` for `&> :last-child`

src/Card/CardStatus.jsx
monteri: Now all `alert-*` are removed and it is `pgn__card-status__{variant}`, `pgn__card-status__heading` etc. All `alert` styles are transferred to `pgn__card-status`

src/Card/CardStatus.jsx
```diff
@@ -0,0 +1,53 @@
...
+}, ref) => (
+  <div
+    className={classNames('pgn__card-status',
+      `alert-${variant}`, 'alert', 'alert-content',
+      className)}
+    ref={ref}
+  >
+    {icon && <Icon src={icon} className="alert-icon" />}
+    <div className="alert-message-content">
+      {title && <div className="alert-heading h4">{title}</div>}
```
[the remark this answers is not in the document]
monteri: Instead of `h4` for heading, I added $h4-font-size for `pgn__card-status__heading`

src/Card/CardStatus.jsx
monteri: Yes, agree, added

src/Card/CardStatus.jsx
```diff
@@ -0,0 +1,53 @@
...
+  icon: PropTypes.func,
+  /** Specifies variant to use. */
+  variant: PropTypes.oneOf([
+    'primary',
+    'success',
+    'danger',
+    'warning',
+  ]),
+  /** Specifies title for the `Card.Status`. */
+  title: PropTypes.element,
```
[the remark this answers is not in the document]
monteri: Yeah, added `string`

adamstankiewicz (approved):

openedx-webhooks:
@monteri 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1439&entry.1484655318=https://github.com/monteri&entry.752974735=2022-07-04+13:04&entry.1917517419=2022-08-12+11:00&entry.2133058324=Yes) so we can improve your experience in the future.

Issue #1385, opened 2022-06-15
Add SCSS variables for each permutation of elevation direction


adamstankiewicz:
https://paragon-openedx.netlify.app/foundations/elevation

There are CSS utility classes and a SCSS mixin to implement shadows in all levels and directions, however, the SCSS variables only account for the level.

**AC**
* [ ] Ensure there are SCSS variables for each permutation of level and heading.

PR #1443, opened 2022-07-05
[BD-46]: vars for elevation

Inferato:
## Description

There are CSS utility classes and a SCSS mixin to implement shadows in all levels and directions, however, the SCSS variables only account for the level.

This PR ensure there are SCSS variables for each permutation of level and heading.

Github issue: https://github.com/openedx/paragon/issues/1385

### Deploy Preview

https://deploy-preview-1443--paragon-openedx.netlify.app/foundations/elevation

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/scss/core/_variables.scss
+++ b/scss/core/_variables.scss
@@ -478,18 +478,48 @@ $level-3-box-shadow:          0 0 .625rem rgba(0, 0, 0, .15), 0 0 1rem rgba(0, 0
 $level-4-box-shadow:          0 .625rem 1.25rem rgba(0, 0, 0, .15), 0 .5rem 1.25rem rgba(0, 0, 0, .15) !default;
 $level-5-box-shadow:          0 1.25px 2.5rem rgba(0, 0, 0, .15), 0 .5rem 2.5rem rgba(0, 0, 0, .15) !default;
 
+$box-shadow-down-1:           0 .0625rem .125rem rgba(0, 0, 0, .15), 0 .0625rem .25rem rgba(0, 0, 0, .15) !default;
+$box-shadow-down-2:           0 .125rem .25rem rgba(0, 0, 0, .15), 0 .125rem .5rem rgba(0, 0, 0, .15) !default;
+$box-shadow-down-3:           0 .5rem 1rem rgba(0, 0, 0, .15), 0 .25rem .625rem rgba(0, 0, 0, .15) !default;
+$box-shadow-down-4:           0 .625rem 1.25rem rgba(0, 0, 0, .15), 0 .5rem 1.25rem rgba(0, 0, 0, .15) !default;
+$box-shadow-down-5:           0 1.25px 2.5rem rgba(0, 0, 0, .15), 0 .5rem 2.5rem rgba(0, 0, 0, .15) !default;
+
+$box-shadow-left-1:           -.0625rem 0 .125rem rgba(0, 0, 0, .15), -.0625rem 0 .25rem rgba(0, 0, 0, .15) !default;
+$box-shadow-left-2:           -.125rem 0 .25rem rgba(0, 0, 0, .15), -.125rem 0 .5rem rgba(0, 0, 0, .15) !default;
+$box-shadow-left-3:           -.5rem 0 1rem rgba(0, 0, 0, .15), -.25rem 0 .625rem rgba(0, 0, 0, .15) !default;
+$box-shadow-left-4:           -.625rem 0 1.25rem rgba(0, 0, 0, .15), -.5rem 0 1.25rem rgba(0, 0, 0, .15) !default;
+$box-shadow-left-5:           -1.25rem 0 2.5rem rgba(0, 0, 0, .15), -.5rem 0 3rem rgba(0, 0, 0, .15) !default;
+
+$box-shadow-up-1:             0 -.0625rem .125rem rgba(0, 0, 0, .15), 0 -.0625rem .25rem rgba(0, 0, 0, .15) !default;
+$box-shadow-up-2:             0 -.125rem .25rem rgba(0, 0, 0, .15), 0 -.125rem .5rem rgba(0, 0, 0, .15) !default;
+$box-shadow-up-3:             0 -.5rem 1rem rgba(0, 0, 0, .15), 0 -.25rem .625rem rgba(0, 0, 0, .15) !default;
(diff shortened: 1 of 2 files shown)


openedx-webhooks:
Thanks for the pull request, @Inferato!


viktorrusakov:
@adamstankiewicz this one should be merged together with https://github.com/edx/brand-edx.org/pull/43 I think, so that it does not affect edX.org theme

scss/core/_variables.scss
```diff
@@ -472,11 +472,35 @@ $border-radius-sm:            .25rem !default;
 
 $rounded-pill:                50rem !default;
 
-$level-1-box-shadow:          0 .0625rem .125rem rgba(0, 0, 0, .15), 0 .0625rem .25rem rgba(0, 0, 0, .15) !default;
-$level-2-box-shadow:          0 .125rem .25rem rgba(0, 0, 0, .15), 0 .125rem .5rem rgba(0, 0, 0, .15) !default;
-$level-3-box-shadow:          0 0 .625rem rgba(0, 0, 0, .15), 0 0 1rem rgba(0, 0, 0, .15) !default;
-$level-4-box-shadow:          0 .625rem 1.25rem rgba(0, 0, 0, .15), 0 .5rem 1.25rem rgba(0, 0, 0, .15) !default;
-$level-5-box-shadow:          0 1.25px 2.5rem rgba(0, 0, 0, .15), 0 .5rem 2.5rem rgba(0, 0, 0, .15) !default;
```
adamstankiewicz, line 479: [curious] If we rename/remove these variables, this will have to be treated as a breaking change, and consumers will need to migrate to the new `box-shadow-*` variables instead as part of the upgrade. Would it make sense to have this change be additive, rather than removing these existing variables to avoid a breaking change?

scss/core/_variables.scss
```diff
@@ -472,11 +472,35 @@ $border-radius-sm:            .25rem !default;
...
+$box-shadow-right-2:          .125rem 0 .25rem rgba(0, 0, 0, .15), .125rem 0 .5rem rgba(0, 0, 0, .15) !default;
+$box-shadow-right-3:          .5rem 0 1rem rgba(0, 0, 0, .15), .25rem 0 .625rem rgba(0, 0, 0, .15) !default;
+$box-shadow-right-4:          .625rem 0 1.25rem rgba(0, 0, 0, .15), .5rem 0 1.25rem rgba(0, 0, 0, .15) !default;
+$box-shadow-right-5:          1.25rem 0 2.5rem rgba(0, 0, 0, .15), .5rem 0 3rem rgba(0, 0, 0, .15) !default;
+
+$box-shadow-centered-1:       0 0 .125rem rgba(0, 0, 0, .15), 0 0 .25rem rgba(0, 0, 0, .15) !default;
+$box-shadow-centered-2:       0 0 .25rem rgba(0, 0, 0, .15), 0 0 .5rem rgba(0, 0, 0, .15) !default;
+$box-shadow-centered-3:       0 0 .625rem rgba(0, 0, 0, .15), 0 0 1rem rgba(0, 0, 0, .15) !default;
+$box-shadow-centered-4:       0 0 1.25rem rgba(0, 0, 0, .15), 0 0 1.25rem rgba(0, 0, 0, .15) !default;
+$box-shadow-centered-5:       0 0 2.5rem rgba(0, 0, 0, .15), 0 0 3rem rgba(0, 0, 0, .15) !default;
```
adamstankiewicz, line 503: [curious] There is the `pgn-box-shadow` mixin below that computes these same values. Could these variables be created with the mixin so that the mixin is the source of truth for each level/side?  That is, if we change the mixin, it'll change the variables, too?

www/src/pages/foundations/elevation.jsx
```diff
@@ -238,32 +238,46 @@ export default function ElevationPage() {
...
+            <code className="d-block mb-2 bg-gray-100 p-3">
+              {'.box-shadow-{direction}-{level}'}
+            </code>
+            <p>For example:</p>
+            <code key={'right'} className="d-block mb-2 bg-gray-100 p-3">
+              .box-shadow-right-2
+            </code>
+            <code key={'right'} className="d-block mb-2 bg-gray-100 p-3">
+               .box-shadow-up-3
             </code>
```
adamstankiewicz, line 251: nit: is this extra indented?

www/src/pages/foundations/elevation.jsx
```diff
@@ -238,32 +238,46 @@ export default function ElevationPage() {
...
+            <code className="d-block mb-2 bg-gray-100 p-3">
+              {'@include pgn-box-shadow({level}, {side})'}
+            </code>
+            <p>For example:</p>
+            <code key={'right'} className="d-block mb-2 bg-gray-100 p-3">
+              @include <strong>pgn-box-shadow(1, "down")</strong>
+            </code>
+            <code className="d-block mb-2 bg-gray-100 p-3">
+              @include <strong>pgn-box-shadow(3, "left")</strong>;
             </code>
```
adamstankiewicz, line 265: nit: is this extra indented?

www/src/pages/foundations/elevation.jsx
```diff
@@ -238,32 +238,46 @@ export default function ElevationPage() {
...
+            <code className="d-block mb-2 bg-gray-100 p-3">
+              {'$box-shadow-{direction}-{level}'}
+            </code>
+            <p>For example:</p>
+            <code key={'right'} className="d-block mb-2 bg-gray-100 p-3">
+              $box-shadow-right-2
+            </code>
+            <code key={'right'} className="d-block mb-2 bg-gray-100 p-3">
+              $box-shadow-up-3
+            </code>
```
adamstankiewicz, line 279: nit: is this extra indented?

scss/core/_variables.scss
viktorrusakov: You're right, we should leave these variable as is and just add new ones, thanks!

scss/core/_variables.scss
viktorrusakov: Yes, I agree, we should have one source of truth, but I think it makes more sense for variables to be the source of truth. We define variables first and then reuse them in mixin. That's how we do it with other mixins also

viktorrusakov:
@adamstankiewicz since @Inferato is out this week, I've updated the PR myself, please take a look.

scss/core/_variables.scss
adamstankiewicz: Makes sense!

adamstankiewicz (approved):

PR #1478, opened 2022-07-15
[BD-46] docs: added support for making `Card` a hyperlink

### LINKED ISSUE #1368: Add support for making `Card` a hyperlink
In the example below, the consumer wrapped the `Card` with a hyperlink (e.g., `a` tag), in order to make it such that the full card will navigate to a new destination or route. However, in doing so, the styles end up getting kind of funky with the default link styling in this situation:

image

**AC**

* Ensure the styles for `Card` don't show default link styling when the component is wrapped with an `a`.
* Similarly, some consumers may be using something like `react-router-dom`, in which case the `Card` might be wrapped with a `Link` component instead.
* Consider whether the props API for `Card` should automatically wrap the `Card` with `a`/`Link` tags, or whether this should be left up to the consumer.

PKulkoRaccoonGang:
## Description

Added support for making `Card` a hyperlink

### Deploy Preview

[Card component](https://deploy-preview-1478--paragon-openedx.netlify.app/components/card/)

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Card/Card.scss
+++ b/src/Card/Card.scss
@@ -6,6 +6,10 @@
   width: 100%;
 }
 
+a:hover {
+  text-decoration: none;
+}
+
 .pgn__card-grid {
   .row > div[class*="col-"] {
     margin-bottom: $card-grid-margin;
@@ -129,6 +133,10 @@
   display: flex;
   align-items: center;
   justify-content: center;
+
+  a:hover {
+    text-decoration: underline;
+  }
 
   &.vertical {
     justify-content: flex-end;
--- a/src/Card/Card.scss
+++ b/src/Card/Card.scss
@@ -9,6 +9,7 @@
 a .pgn__card {
   color: $gray-700;
   display: inline-block;
+  margin-right: 100%;
 }
 
 .pgn__card-grid {
--- a/src/Card/Card.scss
+++ b/src/Card/Card.scss
@@ -6,13 +6,9 @@
   width: 100%;
 }
 
-.link {
-  color: $black;
-
-  &:hover {
-    text-decoration: none;
-    color: $black;
-  }
+a .pgn__card {
+  color: $gray-700;
+  display: inline-block;
 }
 
 .pgn__card-grid {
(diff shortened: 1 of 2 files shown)


openedx-webhooks:


ihor-romaniuk (approved):

src/Card/Card.scss
```diff
@@ -6,6 +6,12 @@
   width: 100%;
 }
 
+a .pgn__card {
+  color: $gray-700;
+  display: inline-block;
```
adamstankiewicz, line 11: nit : should this be an inline-flex vs. inline-block?

adamstankiewicz (approved):

src/Card/Card.scss
PKulkoRaccoonGang: I think in this case we can leave the block element inside the link. The obsolete HTML 4 specification forbade such nesting, but in [HTML 5 the rules](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) have changed. We can wrap a link, if necessary, in elements that do not contain interactive elements (other links) inside.
Also, in this case, it is important for us to use exactly the inline-block because text decorations do not work on "floating" elements. Inline flex unfortunately will not work.

adamstankiewicz (approved):

openedx-webhooks:
@PKulkoRaccoonGang 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1478&entry.1484655318=https://github.com/PKulkoRaccoonGang&entry.752974735=2022-07-15+11:53&entry.1917517419=2022-08-12+11:12&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1498, opened 2022-07-25
[BD-46] feat: add Multiselect component

monteri:
## Description

- modify React Select according to design to get Multiselect component
- tests
- update `component-generator` to place proper quotes

### Deploy Preview

## Merge Checklist

## Post-merge Checklist


openedx-webhooks:


src/Multiselect/Multiselect.scss
```diff
@@ -0,0 +1,126 @@
+// Put styles related to your component here
```
adamstankiewicz, line 1: nit: unneeded :)

src/Multiselect/Multiselect.scss
```diff
@@ -0,0 +1,126 @@
...
+    line-height: $multiselect-placeholder-height;
+    grid-area: auto;
+    max-width: 100%;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+
+    &.show {
+      z-index: $multiselect-floating-label-zindex;
+      padding: 0 .5em;
```
adamstankiewicz, line 26: nit: should this be `rem`?

src/Multiselect/MultiselectComponents.jsx
```diff
@@ -0,0 +1,136 @@
+import React from 'react';
+import classNames from 'classnames';
+import { components } from 'react-select';
+import PropTypes from 'prop-types';
+import Icon from '../Icon';
+import { Close, KeyboardArrowDown } from '../../icons';
+
+const ClearIndicator = (props) => (
+  <components.ClearIndicator {...props}>
+    <Icon src={Close} />
```
adamstankiewicz, line 10: I'm curious if we should be using an `IconButton` here instead of a regular `Icon` to get its focus/hover styles, though I imagine it might mess with the spacing inside of the input. There might also be a11y implications for rendering a clickable icon outside of an element that indicates its an interactive element (i.e., its not currently inside of a `button` element or an element with a `role="button"` for example).

src/Multiselect/MultiselectComponents.jsx
```diff
@@ -0,0 +1,136 @@
...
+  </components.ClearIndicator>
+);
+
+const DropdownIndicator = (props) => (
+  <components.DropdownIndicator {...props}>
+    <Icon src={KeyboardArrowDown} />
```
adamstankiewicz, line 16: Similar question regarding `IconButton` vs `Icon`

src/Multiselect/MultiselectComponents.jsx
```diff
@@ -0,0 +1,136 @@
+import React from 'react';
+import classNames from 'classnames';
+import { components } from 'react-select';
```
adamstankiewicz, line 3: nit: `import { components as reactSelectComponents } from 'react-select';` just to be a tad more explicit

src/Multiselect/MultiselectComponents.jsx
```diff
@@ -0,0 +1,136 @@
...
+  </components.DropdownIndicator>
+);
+
+const ValueContainer = ({ children, ...props }) => (
+  <components.ValueContainer {...props}>
+    {React.Children.map(children, (child) => (child && child.type !== components.Placeholder ? child : null))}
+  </components.ValueContainer>
+);
+
+const Control = ({
```
adamstankiewicz, line 26: [curious] Did you attempt to make use of `Form.Control` in any way here to avoid having to replicate CSS styles? Might not be possible or overly complex, but just wondering if we explored that avenue 😄

src/Multiselect/MultiselectComponents.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+const Menu = (props) => (
+  <components.Menu className="pgn__multiselect__menu" {...props} />
+);
+
+const Option = (props) => (
+  <components.Option className={classNames({ 'is-focus': props.isFocused })} {...props} />
+);
+
+const MultiValueContainer = ({ innerProps, ...props }) => {
```
adamstankiewicz, line 60: [curious] Did we try to explore whether we could make use of the existing `Chip` here? The `Chip`-like elements in this multiselect implementation are different from the behavior of the Paragon `Chip` component in that the entire element is clickable to dismiss it, not just the icon.

Related, is there any way to focus with keyboard only navigation on these chips for a11y support?

src/Multiselect/README.md
```diff
@@ -0,0 +1,140 @@
...
+      label: 'Silver',
+      color: '#666666'
+    },
+  ];
+  
+  const handleChange = (e) => console.log(e);
+
+  return (
+    <Stack direction="vertical">
+      <div className="w-100 p-4">
```
adamstankiewicz, line 79: I don't think the horizontal padding is necessary for the examples without a background as it creates unnecessary whitespace.

src/Multiselect/Multiselect.scss
monteri: Yes, correct, mistake

src/Multiselect/MultiselectComponents.jsx
monteri: Used `IconButton` instead of default component. Styled to fit the design. Added some changes according to focus/hover.

src/Multiselect/MultiselectComponents.jsx
monteri: The same here

src/Multiselect/MultiselectComponents.jsx
monteri: I didn't try because if I place `Form.Control` instead of `Control` it will appear as `Form.Control` inside another `Form.Control` and there a lot of change will be needed vs make some style and logic updates. The only problem that took a lot of time was to set up floating label.

src/Multiselect/MultiselectComponents.jsx
monteri: Yes, `Chip` in the `Multiselect` is different from the Paragon's `Chip` and actually it was easier to change some styles to fit design vs replacing whole component. In last commit I added a11y support (focus for `Chip` icon and `onKeyPress` activates `onClick`)

src/Multiselect/README.md
monteri: Changed to `py-4`

PR #1664, opened 2022-10-06
Automate Browserlist DB Update

abdullahwaheed:
### Ticket
[Automate browserslist DB update ](https://github.com/openedx/frontend-wg/issues/101)

### What has changed
Added a Github workflow that would run periodically(weekly), update browserslist DB, [create a PR](https://github.com/marketplace/actions/create-pull-request) and [merge](https://github.com/peter-evans/enable-pull-request-automerge) it automatically.
Used a shared [workflow](https://github.com/openedx/.github/blob/master/.github/workflows/update-browserslist-db.yml) for this


.github/workflows/update-browserlist.yml
```diff
@@ -0,0 +1,41 @@
+name: Update Browserlist DB
```
adamstankiewicz, line 1: Is this Github Action workflow essentially the same for all repos that use browserslist? If so, should we define the workflow once in a shared place that can be used by many repos?

Like what we do with the shared commit linting Github Action workflow file across all the repos: https://github.com/openedx/paragon/blob/master/.github/workflows/lockfileversion-check.yml#L11-L13

The shared Github Action workflows are defined here: https://github.com/openedx/.github/tree/master/workflow-templates, and should be accessible across repos without having to duplicate workflow files, resulting in more maintainable code.

Also, nit: let's change "browserlist" -> "browser**s**list" to match the actual name of browserslist.

.github/workflows/update-browserlist.yml
```diff
@@ -0,0 +1,41 @@
...
+on:
+  schedule:
+    - cron:  '0 0 * * 1'
+  workflow_dispatch:
```
adamstankiewicz, line 5: Great to provide a manual way to trigger the workflow! 🙌

.github/workflows/update-browserlist.yml
abdullahwaheed: yeah it seems a right way to do this, let me create a PR there

abdullahwaheed: created [this](https://github.com/openedx/.github/pull/29) PR for shared workflow

adamstankiewicz: @abdullahwaheed Left some comments on the PR for the shared workflow!

abdullahwaheed: yes, updating that PR on the basis of your review

abdullahwaheed:
@adamstankiewicz its ready for review

adamstankiewicz (approved):

adamstankiewicz:
Thanks @abdullahwaheed! :) I'll go ahead and merge this when the CI passes.

PR #1688, opened 2022-10-14
feat: added accentB background for page banner

mehaknasir:
## Description

Feature Ticket: https://2u-internal.atlassian.net/browse/INF-599
Figma designs supporting required change: https://www.figma.com/file/LVFnO9oRHYiamoicZuXDAC/Discussions?node-id=6884%3A261359

Page banners do not accept background color classes hence we need to add a new variant that can help achieve a design requirement in discussion MFE. 

- accentB color background is added with variant name accentB.
- documentation is updated to state available variants
- test cases added to ensure the required color class is added in the background.

### Deploy Preview

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/PageBanner/README.md
+++ b/src/PageBanner/README.md
@@ -13,60 +13,64 @@ notes: |
 
 A ``Page Banner`` displays an important, succinct message, and provides actions for users to address (or dismiss the banner). It requires a user action to be dismissed.
 
-## Basic Usage
+## Basic Usage Variants
 
 ```jsx live
 () => {
-  const [show, setShow] = useState(true);
-  return (
-    <>
-      <PageBanner
-        show={show}
-        dismissible
-        onDismiss={() => {setShow(false)}}
-      >
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore.
-      </PageBanner>
-      {!show && (<Button onClick={() => setShow(true)}>Show Page Banner</Button>)}
-    </>
-  )
+    const [showPageBanner, setShowPageBanner] = useState(true);
+    const [variant, setVariant] = useState('light');
+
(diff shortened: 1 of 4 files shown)


src/PageBanner/README.md
```diff
@@ -13,6 +13,14 @@ notes: |
 
 A ``Page Banner`` displays an important, succinct message, and provides actions for users to address (or dismiss the banner). It requires a user action to be dismissed.
 
+## Available Variants
+``dark``
+``light``
+``warning``
+``accentA``
+``accentB``
```
adamstankiewicz, line 21: Hmm, I think it'd make sense to instead of adding this new "Available Variants" section, add an example of using `accentB` in the existing "Variants" section below.

Alternatively, there is a `ExamplePropsForm` component that could be used here as well, which will allow you to configure a yellow box listing out all supported variants where the user can switch between them. See [this example](https://paragon-openedx.netlify.app/components/annotation/#referring-to-other-elements) or the below screenshot.

image

mehaknasir: added prop box

adamstankiewicz: @mehaknasir

image


If we're using the prop box here, I think it'd make sense to include the rendered `PageBanner` controlled by the prop box below the prop box instead of above. There should also be some vertical spacing between the example and the prop box.

If we're keeping the existing pre-rendered variants (current below the prop box), we should add the `accentB` variant there, too. That said, I think if we're using the prop box we should get rid of the other pre-existing variant examples since they are now largely redundant with the example driven by the prop box.

mehaknasir: Thanks, Adam, I was thinking the same but was unsure if I should remove the current examples. I have removed those please review it again and let me know if that's okay or need some further improvements,

PR #1696, opened 2022-10-19
docs: update fallback image src example for Card component

adamstankiewicz:
## Description

Small cleanup to force the fallback image src `Card` example to use the fallback src instead of the default image src by providing invalid image URLs.

### Deploy Preview

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Card/README.md
+++ b/src/Card/README.md
@@ -15,9 +15,9 @@ categories:
 - Content
 status: 'Stable'
 designStatus: 'Done'
-devStatus: 'In Progress'
+devStatus: 'Done'
 notes: |
-  A pass through from react-bootstrap
+  Partially a pass-thru from react-bootstrap, with custom subcomponents.
 ---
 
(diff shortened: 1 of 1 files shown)


kiram15 (approved):
LGTM! Didn't know if you wanted to add the fallback image decision in this PR, but if not, its gtg

adamstankiewicz:
> LGTM! Didn't know if you wanted to add the fallback image decision in this PR, but if not, its gtg

Let's tackle the fallback image decision as a separate chunk of work. Ideally, I think we may actually want to write a brief ADR in this repo about the decision and how it impacts the `@edx/brand-edx.org` and `@edx/brand-openedx` repos since I think based on the discussion in PWG, I think this would be the most reasonable place to store these sorts of images on a theme-by-theme basis.

I think for now we can use the same fallback image across both brand-openedx and brand-edx.org (file duplicated between those 2 repos), but that'll give us the flexibility to make them theme-specific later on.

We would want to update the docs in Paragon to refer and/or show an example of importing the fallback from `@edx/brand-openedx`. I will write up some of these items as a Github Issue for now.

PR #1712, opened 2022-10-25
[BD-46] feat: Clickable steps in Stepper component

PKulkoRaccoonGang:
## Description

Make steps clickable after the user visits them, clicking should navigate the user to clicked step.

### Deploy Preview

[Stepper component](https://deploy-preview-1712--paragon-openedx.netlify.app/components/stepper/)

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Stepper/Stepper.scss
+++ b/src/Stepper/Stepper.scss
@@ -1,13 +1,3 @@
-.pgn__stepper-header-step-list {
-  list-style: none;
-  padding: .25rem 0;
-  display: flex;
-  align-items: center;
-  margin: 0;
-  flex-grow: 1;
-  justify-content: center;
-}
-
 .pgn__stepper-header {
   display: flex;
   justify-content: center;
@@ -15,59 +5,83 @@
   background: $white;
   padding: .75rem 1rem;
   min-height: 5.13rem;
-}
 
-.pgn__stepper-header-step {
-  display: flex;
-  align-items: center;
-  color: $primary;
-  flex-shrink: 1;
-  min-width: 0;
+  .pgn__stepper-header-step-list {
+    list-style: none;
+    padding: .25rem 0;
+    display: flex;
+    align-items: center;
+    margin: 0;
+    flex-grow: 1;
+    justify-content: center;
 
-  .pgn__bubble {
-    margin-inline-end: .5rem;
-    flex-shrink: 0;
+    .pgn__stepper-header-line {
+      display: block;
+      height: 1px;
+      background: $light;
+      flex-basis: 80px;
+      margin: 0 .5rem;
+    }
   }
 
-  .pgn__stepper-header-step-title-description {
+  .pgn__stepper-header-step {
+    display: flex;
+    align-items: center;
+    color: $primary;
+    flex-shrink: 1;
     min-width: 0;
-  }
+    padding: .75rem 1rem;
 
-  .pgn__stepper-header-step-title {
-    white-space: nowrap;
-    overflow: hidden;
-    min-width: 0;
-    text-overflow: ellipsis;
-  }
-
-  .pgn__stepper-header-step-description {
-    font-size: $x-small-font-size;
-  }
-
-  &.pgn__stepper-header-step-active ~ .pgn__stepper-header-step {
-    color: $gray-500;
-  }
-
-  &.pgn__stepper-header-step-has-error {
     .pgn__bubble {
-      background: transparent;
-      box-shadow: inset 0 0 0 3px $danger;
+      margin-inline-end: .5rem;
+      flex-shrink: 0;
+    }
 
-      * {
+    .pgn__stepper-header-step-title-description {
+      min-width: 0;
+    }
+
+    .pgn__stepper-header-step-title {
+      white-space: nowrap;
+      overflow: hidden;
+      min-width: 0;
+      text-overflow: ellipsis;
+    }
+
+    .pgn__stepper-header-step-description {
+      font-size: $x-small-font-size;
+    }
+
+    &.pgn__stepper-header-step-active ~ .pgn__stepper-header-step {
+      color: $gray-500;
+    }
+
+    &.pgn__stepper-header-step-has-error {
+      .pgn__bubble {
+        background: transparent;
+        box-shadow: inset 0 0 0 3px $danger;
+
+        * {
+          color: $danger;
+        }
+      }
+
+      .pgn__stepper-header-step-description {
         color: $danger;
       }
     }
 
-    .pgn__stepper-header-step-description {
-      color: $danger;
+    &.pgn__stepper-header-step-clickable {
+      cursor: pointer;
+
+      &:hover,
+      &:focus {
+        color: $primary;
+
+        .pgn__bubble {
+          background-color: $primary;
+        }
+      }
     }
   }
-}
-
-.pgn__stepper-header-line {
-  display: block;
-  height: 1px;
-  background: $light;
-  flex-basis: 80px;
-  margin: 0 .5rem;
 }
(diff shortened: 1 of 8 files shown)


openedx-webhooks:


src/Stepper/Stepper.scss
```diff
@@ -1,73 +1,87 @@
...
+      cursor: pointer;
+
+      &:hover,
+      &:focus {
+        color: $primary;
+
+        .pgn__bubble {
+          background-color: $primary;
+        }
+      }
```
PKulkoRaccoonGang, line 84: These changes have been added for clickable titles. The rest of the styles in the file have been formatted to respect SCSS nesting. :)

src/Stepper/Stepper.scss
```diff
@@ -1,73 +1,87 @@
...
 
-  .pgn__stepper-header-step-title-description {
+  .pgn__stepper-header-step {
+    display: flex;
+    align-items: center;
+    color: $primary;
+    flex-shrink: 1;
     min-width: 0;
-  }
+    padding: .75rem 1rem;
```
PKulkoRaccoonGang, line 33: These changes have been added to clickable titles.

src/Stepper/StepperHeader.jsx
```diff
@@ -9,29 +9,34 @@ function StepListSeparator() {
   return <li aria-hidden="true" className="pgn__stepper-header-line" />;
 }
 
-function StepList({ steps, activeKey }) {
+function StepList({ steps, activeKey, isClickable }) {
```
ihor-romaniuk, line 12: I'm not sure about the naming. Because the `isClickable` prop seems to be a bool value.
Maybe it's better to change to `clickHandler` or something similar?

PKulkoRaccoonGang: Yes, you are right
Thanks

src/Stepper/StepperHeader.jsx
```diff
@@ -9,29 +9,34 @@ function StepListSeparator() {
...
-      ))}
+        return (
+          <React.Fragment key={stepProps.eventKey}>
+
+            {index !== 0 && <StepListSeparator />}
+            <StepperHeaderStep
+              {...stepProps}
+              index={index}
+              isActive={activeKey === stepProps.eventKey}
+              onClick={isClickable && handleClick}
```
viktorrusakov, line 26: maybe do `onClick={isClickable ? () => isClickable(stepProps.eventKey} : undefined`?

PKulkoRaccoonGang: Corrected, thanks

src/Stepper/StepperHeader.jsx
```diff
@@ -71,11 +72,15 @@ StepperHeader.propTypes = {
   className: PropTypes.string,
   /** A component that receives `activeStepIndex` and `totalSteps` props to display them. */
   PageCountComponent: PropTypes.elementType,
+  /** Specifies whether the `Stepper` headers is clickable, if `true` appropriate `hover` and `focus` styling
+   * will be added and opportunities toggle content. */
```
adamstankiewicz, line 76: nit: what is meant by "opportunities toggle content"?

adamstankiewicz (commented):
Overall, this looks good! I like the simplicity of opting into to clickable header steps with a single prop. I did notice a possible bug, though. Steps to reproduce:

1. Go to the [error state example](https://deploy-preview-1712--paragon-openedx.netlify.app/components/stepper/#error-state), and add the `handleStepClick` prop.
2. Check the box and go to the second step.
3. Go back to the first step and uncheck the box.
4. Now, click the second header step. You can get to the second step without the validation of ensuring the checkbox is checked.
5. Go back to first step.
6. Click "Next" button with the checkbox still unchecked and note the error validation, but the second header step is still clickable.

src/Stepper/StepperHeader.jsx
```diff
@@ -71,11 +72,15 @@ StepperHeader.propTypes = {
   className: PropTypes.string,
   /** A component that receives `activeStepIndex` and `totalSteps` props to display them. */
   PageCountComponent: PropTypes.elementType,
+  /** Specifies whether the `Stepper` headers is clickable, if `true` appropriate `hover` and `focus` styling
```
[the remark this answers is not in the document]
PKulkoRaccoonGang: Corrected, thanks

src/Stepper/StepperHeader.jsx
PKulkoRaccoonGang: This is redundant and incomprehensible, deleted

src/Stepper/StepperHeaderStep.jsx
```diff
@@ -7,28 +7,36 @@ import { StepperContext } from './StepperContext';
...
+  const isViewed = getIsViewed(index);
   const stepIcon = isComplete ? <Icon src={Check} /> : <span>{index + 1}</span>;
   const errorIcon = <Icon src={Error} />;
+  const isClickable = onClick && isViewed && !isActive;
 
   return (
     <li
+      /* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */
+      tabIndex={isClickable ? 0 : -1}
+      role="presentation"
```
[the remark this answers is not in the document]
PKulkoRaccoonGang: Yes, you are right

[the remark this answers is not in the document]
viktorrusakov: I think eslint just looks at the `li` tag (and doesn't pay much attention to role) and basically says that `li` tag is non-interactive by definition no matter the role. Actually if you just plainly add `role="button"` eslint starts complaining with `Non-interactive elements should not be assigned interactive roles.(jsx-a11y/no-noninteractive-element-to-interactive-role)`.

I don't see a workaround here, I think if step is clickable we need to just use `button` tag instead of `li` to comply with `a11y` rules

src/Stepper/StepperContext.jsx
```diff
@@ -51,7 +56,7 @@ export function StepperContextProvider({ children, activeKey }) {
         registerStep,
         steps,
         removeStep,
-        getIsComplete,
```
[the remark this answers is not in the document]
PKulkoRaccoonGang: Yes, this function was only internal, consumers couldn’t access it since we don’t export context components (also checked usage insights to make sure)

src/Stepper/StepperHeaderStep.jsx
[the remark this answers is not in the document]
adamstankiewicz: @PKulkoRaccoonGang @viktorrusakov Thanks for the clarifications! I think it's worth trying to make use of `button` when it's clickable in this case to avoid needing to ignore a11y rules, given a11y is core to Paragon.

PKulkoRaccoonGang:
@adamstankiewicz please review this pull request. Thank you :)

viktorrusakov:
@adamstankiewicz I've refactored out solution to add `onClick` handler for `StepperStep` component instead of `StepperHeader` component. I don't really see any other way to adequately handle error state 😞 . The downside is you have to provide onClick handler for every step separately, but I don't think it's too bad of a requirement. And you also get more control over click behaviour this way.

adamstankiewicz (approved):

openedx-webhooks:
@PKulkoRaccoonGang 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1712&entry.1484655318=https://github.com/PKulkoRaccoonGang&entry.752974735=2022-10-25+07:52&entry.1917517419=2023-05-05+15:08&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1789, opened 2022-11-22
[BD-46] feat: add variants for Card

monteri:
## Description

- remove muted from `Card` and from `Card.Section`
- added variants `light`, `dark`, `muted`

### Deploy Preview

https://deploy-preview-1789--paragon-openedx.netlify.app/components/card/

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Card/Card.scss
+++ b/src/Card/Card.scss
@@ -5,7 +5,30 @@
 .pgn__card-body {
   width: 100%;
 
-  &.is-muted {
+  &.pgn__card-dark {
+    background-color: $primary-500;
+    color: $white;
+
+    .pgn__card-header {
+      .pgn__card-header-title,
+      .pgn__card-header-title-sm,
+      .pgn__card-header-title-md {
+        color: $white;
+      }
+
+      .pgn__card-header-subtitle,
+      .pgn__card-header-subtitle-sm,
+      .pgn__card-header-subtitle-md {
+        color: $white;
+      }
+    }
+
+    .pgn__card-section .pgn__card-section-title {
+      color: $white;
+    }
+  }
+
+  &.pgn__card-muted {
     background-color: $light-200;
   }
 }
@@ -101,10 +124,6 @@ a .pgn__card {
   padding: $card-spacer-x;
   flex-grow: 1;
   text-align: start;
-
-  &.is-muted {
-    background-color: $light-200;
-  }
 
   .pgn__card-section-title {
     color: $black;
--- a/src/Card/Card.scss
+++ b/src/Card/Card.scss
@@ -125,6 +125,10 @@ a .pgn__card {
   flex-grow: 1;
   text-align: start;
 
+  &.is-muted {
+    background-color: $light-200;
+  }
+
   .pgn__card-section-title {
     color: $black;
     font-weight: $font-weight-bold;
--- a/src/Card/Card.scss
+++ b/src/Card/Card.scss
@@ -4,33 +4,32 @@
 .pgn__card,
 .pgn__card-body {
   width: 100%;
+  border: 0;
 
-  &.pgn__card-dark {
-    border: 0;
-    background-color: $primary-500;
+  %dark-variant-content {
     color: $white;
 
-    .pgn__card-header {
-      .pgn__card-header-title,
-      .pgn__card-header-title-sm,
-      .pgn__card-header-title-md {
-        color: $white;
-      }
-
-      .pgn__card-header-subtitle,
-      .pgn__card-header-subtitle-sm,
-      .pgn__card-header-subtitle-md {
-        color: $white;
-      }
-    }
-
-    .pgn__card-section .pgn__card-section-title {
+    a {
       color: $white;
     }
   }
 
+  &.pgn__card-dark {
+    background-color: $card-bg-dark;
+
+    @extend %dark-variant-content;
+
+    .pgn__card-header {
+      @extend %dark-variant-content;
+    }
+
+    .pgn__card-section {
+      @extend %dark-variant-content;
+    }
+  }
+
   &.pgn__card-muted {
-    background-color: $light-200;
+    background-color: $card-bg-muted;
   }
 }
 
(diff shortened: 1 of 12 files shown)


openedx-webhooks:


src/Card/CardSection.jsx
```diff
@@ -11,15 +11,14 @@ const CardSection = React.forwardRef(({
   children,
   title,
   actions,
-  muted,
```
adamstankiewicz, line 14: I think we may want to leave the `muted` prop on `Card.Section` to support use cases where consumers want to implement a light variant card with one muted section inside of it. Removing this prop would also technically be a breaking change.

src/Card/index.jsx
```diff
@@ -11,27 +11,42 @@ import CardImageCap from './CardImageCap';
...
-      })}
-      ref={ref}
-      tabIndex={isClickable ? '0' : '-1'}
-    />
-  </CardContextProvider>
-));
+}, ref) => {
+  if (muted) {
+    // eslint-disable-next-line no-console
+    console.error('Prop muted is deprecated, please, use variant="muted" instead');
```
adamstankiewicz, line 27: [curious] Could/should we use `withDeprecatedProps` to mark this as a deprecated prop (see other usages of it in the repo)?

src/Card/README.md
```diff
@@ -51,34 +51,45 @@ This component uses a `Card` from react-bootstrap as a base component and extend
...
-    <Card style={{ width: isExtraSmall ? "100%" : "18rem" }} muted>
-      <Card.ImageCap 
-        src="https://picsum.photos/360/200/"
-        srcAlt="Card image"
-      />
-      <Card.Header
-        title="Card Title"
+    <>
+      {/* start example form block */}
+      <ExamplePropsForm
```
adamstankiewicz, line 66: Note: it looks like the checkboxes in the `ExamplePropsForm` are functional (they switch between the variants), but they don't show when they are checked.

image

src/Card/Card.scss
```diff
@@ -5,7 +5,30 @@
 .pgn__card-body {
   width: 100%;
 
-  &.is-muted {
+  &.pgn__card-dark {
```
adamstankiewicz, line 8: It appears the background color is coming through as if its a border applied to the `Card`: 

image

This makes it appear as if it has active/focus styling when it shouldn't, and makes the actual active/focus styles largely indistinguishable, e.g. focus styles on the light variant:

image

src/Card/CardContext.jsx
```diff
@@ -20,12 +23,15 @@ CardContextProvider.propTypes = {
   isLoading: PropTypes.bool,
   /** Specifies content of the component. */
   children: PropTypes.node,
+  /** Specifies `Card` style variant */
+  variant: PropTypes.oneOf(['light', 'dark', 'muted']),
```
adamstankiewicz, line 27: nit: Should this re-use the `CARD_VARIANTS` constant?

src/Card/CardSection.jsx
viktorrusakov: True, we returned it, thanks!

src/Card/index.jsx
viktorrusakov: yeah, we probably should, updated

src/Card/README.md
viktorrusakov: fixed, thanks! (it was because of another `ExamplePropsForm` we have on the page which conflicted with this one)

src/Card/CardContext.jsx
viktorrusakov: I don't think we should, because it looks weird on docs site then (if constant is defined outside of the file it's used in it can't be properly parsed by a package we use to render props data)
image

src/Card/Card.scss
viktorrusakov: yeah, it looks like this border is the issue, I removed it for now with `border: 0` but then I'm not sure what the focus styles should be for dark `Card`.. (if we add border only for focus state then it gets jumpy when you focus). Might need to think more about this

adamstankiewicz: @viktorrusakov We might consider having the focus state border for `Card` be offset from the card element similar to the focus state on `Button` for all its variants, e.g.:

image

image

adamstankiewicz: @viktorrusakov I also believe we'll want to add `border: 0;` for all `Card` variants, not just dark. otherwise the card heights may be slightly inconsistent if rendering different `Card` variants side by side (e.g., light variant next to dark variant).

src/Card/CardContext.jsx
adamstankiewicz: Ah, right. I always forget about this haha. Thanks for clarifying.

src/Card/tests/CardSection.test.jsx
```diff
@@ -54,18 +54,4 @@ describe('<CardSection />', () => {
     expect(wrapper.exists('.pgn__card-section-loader')).toBe(true);
     expect(wrapper.props().isLoading).toBe(true);
   });
-  it('renders muted variant', () => {
```
adamstankiewicz, line 57: [curious] Should we leave this test in considering we're leaving the `muted` prop for `Card.Section`?

src/Card/Card.scss
```diff
@@ -5,7 +5,31 @@
...
+    border: 0;
+    background-color: $primary-500;
+    color: $white;
+
+    .pgn__card-header {
+      .pgn__card-header-title,
+      .pgn__card-header-title-sm,
+      .pgn__card-header-title-md {
+        color: $white;
```
adamstankiewicz, line 17: @viktorrusakov @monteri We will need to handle if/when a `title` and/or `subtitle` have a nested hyperlink in the `variant="dark"` use case, e.g.:

```jsx
<Card variant="dark">
  <Card.Header
    title={(
      <Hyperlink
        destination="https://google.com"
        target="_blank"
      >
        Card Title
       </Hyperlink>
    )}
  />
</Card>
```

image

openedx-webhooks:
@monteri 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1789&entry.1484655318=https://github.com/monteri&entry.752974735=2022-11-22+10:40&entry.1917517419=2022-12-21+12:46&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1804, opened 2022-11-29
feat: DataTable selections clean up and improvements

adamstankiewicz:
## Description

* Refactored to be more explicit about how often `DataTable` calls `fetchData` as well as what data it passes in its arguments.
* Make `children` optional for `Card.Section` and `Card.Footer`.
* Implements a loading state for `CardView`; consumer may override the skeleton card component to use with the `SkeletonCardComponent` prop to `CardView`.
* Exposes a new `onSelectedRowsChanged` prop to `DataTable` as a way to inform parent/consuming components when `DataTable` selections has changed in some way (e.g., a row was selected/unselected).
  * This approach is similar to [this official code example](https://codesandbox.io/embed/tannerlinsleyreact-table-row-selection-controlled-euyku?fontsize=14&hidenavigation=1&theme=dark) for "controlled" selections.
* Ensures `TableFooter` adheres to whether `RowStatus` is overridden by the `RowStatusComponent` prop on `DataTable`.

### Deploy Preview

https://deploy-preview-1804--paragon-openedx.netlify.app/components/card/
https://deploy-preview-1804--paragon-openedx.netlify.app/components/datatable/

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Card/CardFooter.jsx
+++ b/src/Card/CardFooter.jsx
@@ -42,7 +42,7 @@ const CardFooter = React.forwardRef(({
 
 CardFooter.propTypes = {
   /** Specifies contents of the component. */
-  children: PropTypes.node.isRequired,
+  children: PropTypes.node,
   /** Specifies class name to append to the base element. */
   className: PropTypes.string,
   /** Optional node to display near actions. Should be either a plain text or an element containing text (e.g. link). */
@@ -58,6 +58,7 @@ CardFooter.propTypes = {
 };
 
 CardFooter.defaultProps = {
+  children: null,
   className: undefined,
   textElement: undefined,
   isStacked: false,
--- a/src/Card/CardSection.jsx
+++ b/src/Card/CardSection.jsx
@@ -45,7 +45,7 @@ CardSection.propTypes = {
   /** Specifies class name to append to the base element. */
   className: PropTypes.string,
   /** Specifies contents of the component. */
-  children: PropTypes.node.isRequired,
+  children: PropTypes.node,
   /** Specifies title of the `Section`. */
   title: PropTypes.node,
   /** Specifies node to render on the bottom right of the `Section` (i.e. `ActionRow`). */
@@ -59,6 +59,7 @@ CardSection.propTypes = {
 };
 
 CardSection.defaultProps = {
+  children: null,
   className: undefined,
   title: undefined,
   actions: undefined,
--- a/src/DataTable/CardView.jsx
+++ b/src/DataTable/CardView.jsx
@@ -40,7 +40,7 @@ function CardItem({
 
 function DefaultSkeletonCardComponent() {
   return (
-    <Card isLoading>
+    <Card isLoading data-testid="default-skeleton-card-component">
       <Card.ImageCap logoSkeleton />
       <Card.Section className="pgn__data-table-card-view-default-skeleton-card-section" />
       <Card.Footer />
@@ -48,12 +48,15 @@ function DefaultSkeletonCardComponent() {
   );
 }
 
+export const DEFAULT_SKELETON_CARD_COUNT = 8;
+
 function CardView({
   columnSizes,
   CardComponent,
   className,
   selectionPlacement,
   SkeletonCardComponent = DefaultSkeletonCardComponent,
+  skeletonCardCount,
 }) {
   const {
     getTableProps, prepareRow, displayRows,
@@ -78,7 +81,7 @@ function CardView({
         className={classNames('pgn__data-table-card-view', className)}
         columnSizes={columnSizes}
       >
-        {[...new Array(8)].map(() => <SkeletonCardComponent key={uuidv4()} />)}
+        {[...new Array(skeletonCardCount)].map(() => <SkeletonCardComponent key={uuidv4()} />)}
       </CardGrid>
     );
   }
@@ -128,6 +131,7 @@ CardView.defaultProps = {
   className: undefined,
   selectionPlacement: 'right',
   SkeletonCardComponent: undefined,
+  skeletonCardCount: 8,
 };
 
 CardView.propTypes = {
@@ -151,6 +155,8 @@ CardView.propTypes = {
   selectionPlacement: PropTypes.oneOf(['left', 'right']),
   /** Overrides default skeleton card component for loading state in CardView */
   SkeletonCardComponent: PropTypes.func,
+  /** Customize the number of loading skeleton cards to display in CardView */
+  skeletonCardCount: PropTypes.number,
 };
 
 export default CardView;
--- a/src/DataTable/CardView.jsx
+++ b/src/DataTable/CardView.jsx
@@ -1,7 +1,7 @@
 import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import { nanoid } from 'nanoid';
+import { v4 as uuidv4 } from 'uuid';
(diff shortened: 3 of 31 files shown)


src/Card/CardFooter.jsx
```diff
@@ -42,7 +42,7 @@ const CardFooter = React.forwardRef(({
 
 CardFooter.propTypes = {
   /** Specifies contents of the component. */
-  children: PropTypes.node.isRequired,
+  children: PropTypes.node,
```
adamstankiewicz, line 45: turns out it is a valid use case to use `Card.Footer` without `children`

src/Card/CardSection.jsx
```diff
@@ -45,7 +45,7 @@ CardSection.propTypes = {
   /** Specifies class name to append to the base element. */
   className: PropTypes.string,
   /** Specifies contents of the component. */
-  children: PropTypes.node.isRequired,
+  children: PropTypes.node,
```
adamstankiewicz, line 48: turns out it is a valid use case to use `Card.Section` without `children`

src/DataTable/CardView.jsx
```diff
@@ -37,21 +38,51 @@ function CardItem({
...
     return null;
   }
 
+  if (isLoading) {
+    return (
+      <CardGrid
+        className={classNames('pgn__data-table-card-view', className)}
+        columnSizes={columnSizes}
+      >
+        {[...new Array(8)].map(() => <SkeletonCardComponent key={nanoid()} />)}
```
adamstankiewicz, line 81: note: might consider allowing consumers to override the number of loading skeleton cards to display

src/DataTable/DataTable.scss
```diff
@@ -299,12 +303,11 @@ $data-table-pagination-dropdown-min-width: 6rem !default;
...
   &.selection-right {
     > :first-child {
+      margin-right: map_get($spacers, 2);
+
       [dir="rtl"] & {
         margin-left: map_get($spacers, 2);
-      }
-
-      [dir="ltr"] & {
-        margin-right: map_get($spacers, 2);
```
adamstankiewicz, line 307: [clarification] This style wasn't applying in MFEs since the dev build doesn't compile SCSS/CSS with the same PostCSS RTL plugin as production builds.

src/DataTable/TableFooter.jsx
```diff
@@ -9,7 +9,13 @@ import TablePaginationMinimal from './TablePaginationMinimal';
 function TableFooter({ className, children }) {
   return (
     <div className={classNames(className, 'pgn__data-table-footer')}>
-      {children}
+      {children || (
+        <>
+          <RowStatus />
+          <TablePagination />
+          <TablePaginationMinimal />
+        </>
```
adamstankiewicz, line 17: moved from default props to be explicit in the component

src/DataTable/hooks.jsx
```diff
@@ -27,14 +26,15 @@ export const useSelectionActions = (
 ) => {
   const [{ selectedRows, isEntireTableSelected }, dispatch] = controlledTableSelections;
 
-  const clearSelection = () => {
+  const clearSelection = useCallback(() => {
```
adamstankiewicz, line 29: This exported function should be memoized so it's not responsible for any erroneous component re-renders in consuming components.

src/DataTable/index.jsx
```diff
@@ -107,17 +108,30 @@ function DataTable({
...
-  // any state changes to current row selections as we don't want to re-fetch data whenever row(s) are selected.
-  const tableStateWithoutSelections = { ...instance.state };
-  delete tableStateWithoutSelections.selectedRowIds;
+  const {
+    pageSize: tableStatePageSize,
+    pageIndex: tableStatePageIndex,
+    sortBy: tableStateSortBy,
+    filters: tableStateFilters,
+    selectedRowIds: tableStateSelectedRowIds,
+  } = instance.state;
```
adamstankiewicz, line 117: These data are the same fields as were previously passed as arguments to `fetchData` below, so it is not a breaking change from that perspective. This change was necessary to prevent the `useEffect` that calls `fetchData` below from calling when selections were being made and allows us to move away from the more risky/hacky `delete` approach in the previous implementation.

src/DataTable/index.jsx
```diff
@@ -107,17 +108,30 @@ function DataTable({
...
+        filters: tableStateFilters,
+      });
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [fetchData, JSON.stringify(tableStateWithoutSelections)]);
+  }, [fetchData, tableStatePageSize, tableStatePageIndex, tableStateSortBy, tableStateFilters]);
+
+  useEffect(() => {
+    if (onSelectedRowsChanged) {
+      onSelectedRowsChanged(tableStateSelectedRowIds);
```
adamstankiewicz, line 132: Inform the parent/consuming component that row selections were changed in some way.

src/DataTable/index.jsx
```diff
@@ -134,6 +148,7 @@ function DataTable({
     disableElevation,
     isLoading,
     isSelectable,
+    manualSelectColumn,
```
adamstankiewicz, line 151: This needed to be exposed in `DataTableContext` for it to be accessible by the `CardView`.

src/Skeleton/Skeleton.scss
```diff
@@ -0,0 +1,3 @@
+.react-loading-skeleton {
+    z-index: unset;
```
adamstankiewicz, line 2: `react-loading-skeleton` puts `z-index: 1` on their skeletons. This negatively impacted the shadow layering within a `FullscreenModal` and `Stepper`. Unsetting removes the styling from `react-loading-skeleton`.

src/DataTable/index.jsx
adamstankiewicz: Also note, the original "controlled" selections implementation co-exists with this new approach. Need to sanity check that other uses of `DataTable` selections relying on the original selections implementations won't break as a result of this change.

Long term, we'll probably want to move away from or otherwise deprecate the current selections strategy, where "controlled" selections are not really controlled.

src/DataTable/CardView.jsx
```diff
@@ -37,21 +38,54 @@ function CardItem({
...
+  const {
+    isSelectable,
+    manualSelectColumn,
+    isLoading,
+  } = useContext(DataTableContext);
+
   // use the same component for card selection that is used for row selection
   // otherwise view switching might break if row selection uses component that supports backend filtering / sorting
-  const selectionComponent = manualSelectColumn?.Cell || selectColumn.Cell;
+  const SelectionComponent = manualSelectColumn?.Cell || selectColumn.Cell;
```
brobro10000, line 72: [curious]
Would rewriting the logic as `const SelectionComponent = manualSelectColumn?.Cell || selectColumn?.Cell;` prevent non-loading page instances if `selectColumn` happens to be undefined?

src/Card/CardSection.jsx
brobro10000: Love the built in spacers of the `Card.Section` :)

brobro10000 (approved):

src/DataTable/CardView.jsx
adamstankiewicz: @brobro10000 The `selectColumn` variable (defined in `getVisibleColumns`) is a constant, so it'll always exist :)

PR #1870, opened 2022-12-23
fix: Takes into account pagination pageCount <=1 page

brobro10000:
## Description

https://2u-internal.atlassian.net/browse/ENT-6567

Fixes issue with pagination where if pageCount === (0 || 1) , allows for infinite pagination. This fix would disable the pagination buttons if the pageCount <= 1.  

Takes into account all variants, including the default. 

Testing WIP

### Deploy Preview

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Pagination/index.jsx
+++ b/src/Pagination/index.jsx
@@ -106,8 +106,7 @@ class Pagination extends React.Component {
 
     if (page === 1) {
       this.nextButtonRef.focus();
-    } else
-    if (page === pageCount) {
+    } else if (page === pageCount) {
       this.previousButtonRef.focus();
     }
     this.setState({ currentPage: page });
@@ -266,7 +265,8 @@ class Pagination extends React.Component {
       buttonLabels, pageCount, icons, variant, size,
     } = this.props;
     const { currentPage } = this.state;
-    const isLastPage = (currentPage === pageCount) || (pageCount <= 1);
+    const isLastPage = (currentPage === pageCount);
+    const isDisabled = isLastPage || (pageCount <= 1);
     const nextPage = isLastPage ? null : currentPage + 1;
     const iconSize = (variant !== VARIANTS.reduced && size !== 'small') || variant === VARIANTS.minimal;
 
@@ -280,7 +280,7 @@ class Pagination extends React.Component {
         className={classNames(
           'page-item',
           {
-            disabled: isLastPage,
+            disabled: isDisabled,
           },
         )}
       >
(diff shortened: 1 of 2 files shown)


src/Pagination/index.jsx
```diff
@@ -105,7 +106,8 @@ class Pagination extends React.Component {
 
     if (page === 1) {
       this.nextButtonRef.focus();
-    } else if (page === pageCount) {
+    } else
+    if (page === pageCount) {
```
adamstankiewicz, line 110: nit: unnecessary line break in the `} else if (page === pageCount) {`

src/Pagination/index.jsx
```diff
@@ -264,7 +266,7 @@ class Pagination extends React.Component {
       buttonLabels, pageCount, icons, variant, size,
     } = this.props;
     const { currentPage } = this.state;
-    const isLastPage = currentPage === pageCount;
+    const isLastPage = (currentPage === pageCount) || (pageCount <= 1);
```
adamstankiewicz, line 269: Is it necessary to have `isLastPage` be dependent on `pageCount` when both cases are handled independently already in `disabled` below?

I also realized we'll probably want to ensure the `tabIndex={-1}` is kept in sync with whenever `isDisabled` is true.

Perhaps something along the lines of:

```jsx
const isLastPage = currentPage === pageCount;
const isDisabled = isLastPage || pageCount <= 1;

// ...

tabIndex={isDisabled ? -1 : undefined}
disabled={isDisabled}
```

Similar feedback for the previous chevron `IconButton` as well.

src/Pagination/index.jsx
```diff
@@ -343,6 +345,9 @@ class Pagination extends React.Component {
     });
 
     return pages.map((pageIndex) => {
+      if (pageCount <= 1) {
+        return null;
+      }
```
adamstankiewicz, line 350: [curious] should we even need to iterate through `pages` array if `pageCount <= 1`?

src/Pagination/index.jsx
brobro10000: Ideally no, but not taking it into account allows the next button to be clickable on the default variant state when `pageCount = 0`, because the default `currentPage` is set to 1. So by setting the last page boolean based on strictly the value of `pageCount` as opposed to checking if the `currentPage === pageCount`. 
Screen Shot 2022-12-23 at 1 01 20 PM

brobro10000: Ahh but looking at your fix, that does fix the issue as well

src/Pagination/index.jsx
```diff
@@ -264,7 +265,8 @@ class Pagination extends React.Component {
       buttonLabels, pageCount, icons, variant, size,
     } = this.props;
     const { currentPage } = this.state;
-    const isLastPage = currentPage === pageCount;
+    const isLastPage = (currentPage === pageCount);
+    const isDisabled = isLastPage || (pageCount <= 1);
```
adamstankiewicz, line 269: Did we want to make a similar change to introduce a `isDisabled` variable for the "previous" button as well?

adamstankiewicz (approved):

PR #1910, opened 2023-01-12
[BD-46] feat: fixed buttons inconsistencies with design

PKulkoRaccoonGang:
## Description

- for openedx theme fix obvious issues (e.g. tertiary button disappearing in disabled state)
1) Screenshot 2023-01-12 at 23 32 53
2) Screenshot 2023-01-12 at 23 33 27

### Deploy Preview

[Button component](https://deploy-preview-1910--paragon-openedx.netlify.app/components/button/)

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Button/_variables.scss
+++ b/src/Button/_variables.scss
@@ -43,3 +43,8 @@ $btn-border-radius-lg:        $border-radius-lg !default;
 $btn-border-radius-sm:        $border-radius-sm !default;
 
 $btn-transition: null;
+
+$distance: $btn-focus-width + $btn-focus-gap !default;
+$distance-include-btn-border: $distance + $btn-border-width !default;
+
+$btn-border-radius-value: calc(#{$btn-border-radius} + #{$distance}) !default;
--- a/src/Button/_variables.scss
+++ b/src/Button/_variables.scss
@@ -44,7 +44,7 @@ $btn-border-radius-sm:        $border-radius-sm !default;
 
 $btn-transition: null;
 
-$distance: $btn-focus-width + $btn-focus-gap !default;
-$distance-include-btn-border: $distance + $btn-border-width !default;
+$btn-focus-border-gap: $btn-focus-width + $btn-focus-gap !default;
+$btn-focus-distance-to-border: $btn-focus-border-gap + $btn-border-width !default;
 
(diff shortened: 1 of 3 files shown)


openedx-webhooks:


viktorrusakov:
@PKulkoRaccoonGang can we move `$distance` variable from [here](https://github.com/openedx/paragon/blob/master/src/Button/Button.scss#L138) to `_variables.scss` file and override it in brand-edx repo once instead of writing `border-radius: 0` for every `Button` class

src/Button/_variables.scss
```diff
@@ -43,3 +43,8 @@ $btn-border-radius-lg:        $border-radius-lg !default;
 $btn-border-radius-sm:        $border-radius-sm !default;
 
 $btn-transition: null;
+
+$distance: $btn-focus-width + $btn-focus-gap !default;
+$distance-include-btn-border: $distance + $btn-border-width !default;
+
+$btn-border-radius-value: calc(#{$btn-border-radius} + #{$distance}) !default;
```
viktorrusakov, line 50: can we rename this to something more obvious? right now it's not clear what this is used for, especially since we already have `$btn-border-radius` variable. I think this one is used only during focus state, maybe name it something like `btn-focus-border-radius`?

Also, same comment to `distance` variables above, it's not clear what distance they represent, maybe rename them to something like `btn-focus-distance-to-border`? (I think they represent distance to broder during focus state, but maybe I'm wrong)

PKulkoRaccoonGang: Corrected, thanks

adamstankiewicz (commented):
This is looking pretty good. I did notice that the `inverse-brand` and `inverse-primary` lose their background color when `disabled=true`. 

I think the Open edX theme should roughly match what's depicted for the edX.org theme here.

**Current state**

The disabled `inverse-brand` and `inverse-primary` buttons get a transparent background even though their non-disabled backgrounds are solid white.

image

**Expected state**

The disabled `inverse-brand` and `inverse-primary` buttons keep their white background color even when `disabled=true`, but with 0.65 opacity like how the `brand` variant's disabled background color is derived.

image

adamstankiewicz (approved):

PR #1911, opened 2023-01-13
[BD-46] refactor Pagination component

viktorrusakov:
## Description

Refactors `Pagination` component to fix its style and API issues (and to improve it overall to ease maintenance).

**Note**: should be merged after #1190 and https://github.com/edx/brand-edx.org/pull/49, as they would fix ongoing issues with styles.

### Deploy Preview

https://deploy-preview-1911--paragon-openedx.netlify.app/components/pagination/

## Merge Checklist

## Post-merge Checklist


openedx-webhooks:
Thanks for the pull request, @viktorrusakov!


src/Pagination/constants.js
```diff
@@ -1,2 +1,16 @@
...
 export const ELLIPSIS = '...';
+
+export const PAGINATION_VARIANTS = {
+  default: 'default',
+  secondary: 'secondary',
+  reduced: 'reduced',
+  minimal: 'minimal',
+};
+
+export const PAGINATION_BUTTON_LABEL_PREV = 'Previous';
```
PKulkoRaccoonGang, line 10: A good option is to move these constants related to PAGINATION_BUTTON to an object.
For example: `export const PAGINATION_BUTTON = { label: { prev: 'Previous' ... }, iconButton: { nextAlt: 'Go to next page', ... } }`

src/Pagination/Pagination.test.jsx
```diff
@@ -161,18 +204,16 @@ describe('<Pagination />', () => {
...
-        const wrapper = mount((
+        render((
           <ResponsiveContext.Provider value={{ width: breakpoints.large.maxWidth }}>
             <Pagination {...props} />
           </ResponsiveContext.Provider>
         ));
 
-        wrapper.find('.btn').at(2).simulate('click');
-        expect(wrapper.state('currentPage')).toEqual(2);
+        fireEvent.click(screen.getByLabelText('Page 2'));
```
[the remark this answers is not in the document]
viktorrusakov: Yes! Good catch, thanks, replaced everything with `userEvent` without any problems 🙂

src/Pagination/PaginationContext.jsx
```diff
@@ -0,0 +1,191 @@
...
+    currentPage: PropTypes.string,
+    pageOfCount: PropTypes.string,
+  }).isRequired,
+  currentPage: PropTypes.number,
+  maxPagesDisplayed: PropTypes.number.isRequired,
+  icons: PropTypes.shape({
+    leftIcon: PropTypes.oneOfType([PropTypes.element, PropTypes.func]),
+    rightIcon: PropTypes.oneOfType([PropTypes.element, PropTypes.func]),
+  }).isRequired,
+  variant: PropTypes.oneOf(['default', 'secondary', 'reduced', 'minimal']).isRequired,
```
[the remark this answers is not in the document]
viktorrusakov: Good idea, updated!

src/Pagination/Pagination.test.jsx
```diff
@@ -30,46 +38,64 @@ describe('<Pagination />', () => {
...
-      const wrapper = mount(<Pagination {...props} />);
-      expect(wrapper.state('currentPage')).toEqual(initialPage);
-      wrapper.setProps({
-        currentPage: newPage,
-      });
-      wrapper.update();
-      expect(wrapper.state('currentPage')).toEqual(newPage);
+  it('correctly handles initial page prop', () => {
+    render(<Pagination {...baseProps} currentPage={undefined} initialPage={3} />);
+    expect(screen.getByLabelText(/current page/i)).toHaveTextContent('3');
```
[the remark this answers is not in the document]
viktorrusakov: Yeah, it definitely should be better, thanks. I've updated the tests, no more hardcoding 🙂

src/Pagination/constants.js
viktorrusakov: I think in this case it's slightly better to separate these constants, just like they are right now. To me it seems that they are easier to use this way - instead of dragging the whole object across multiple files I can import just a single constant which is needed for the specific component.

I'd like to keep them as they are right now.

PKulkoRaccoonGang: Ok, thanks 👍

src/DataTable/TablePaginationMinimal.jsx
```diff
@@ -1,6 +1,7 @@
 import React, { useContext } from 'react';
 import DataTableContext from './DataTableContext';
 import Pagination from '../Pagination';
+import { ArrowBackIos, ArrowForwardIos } from '../../icons';
 
 function TablePaginationMinimal() {
```
monteri, line 6: Can we add `propTypes` for this component?

viktorrusakov: This component does not receive any props, so I don't think `propTypes` is needed here

src/Pagination/index.jsx
```diff
@@ -483,40 +116,34 @@ Pagination.propTypes = {
    * string, symbol, etc. Default is chevrons rendered using fa-css.
    */
   icons: PropTypes.shape({
-    leftIcon: PropTypes.node,
-    rightIcon: PropTypes.node,
+    leftIcon: PropTypes.oneOfType([PropTypes.element, PropTypes.func]),
+    rightIcon: PropTypes.oneOfType([PropTypes.element, PropTypes.func]),
```
adamstankiewicz, line 120: [clarification] Is this intended to be a breaking change from what's supported today? In the current state, consumers can technically pass any node rather than just an element. 

That said, it only looks like this impacts 1 usage of `Pagination` though as a breaking change according to usage insights; I think it's probably fine to go this route. That 1 usage should be using the `buttonLabels` instead of passing a node with `<Icon>` and `<span className="sr-only">...</span>` elements. Everything in the `sr-only` span really should probably be in the `buttonLabels` prop.

src/Pagination/constants.js
```diff
@@ -1,2 +1,16 @@
...
+export const PAGINATION_BUTTON_LABEL_NEXT = 'Next';
+export const PAGINATION_BUTTON_LABEL_PAGE = 'Page';
+export const PAGINATION_BUTTON_LABEL_CURRENT_PAGE = 'Current Page';
+export const PAGINATION_BUTTON_LABEL_PAGE_OF_COUNT = 'of';
+export const PAGINATION_BUTTON_ICON_BUTTON_NEXT_ALT = 'Go to next page';
+export const PAGINATION_BUTTON_ICON_BUTTON_PREV_ALT = 'Go to previous page';
```
adamstankiewicz, line 16: Also asked in another comment related to Pagination's default props, but wondering if these default English strings should be marked for i18n translation?

src/Pagination/subcomponents/index.jsx
```diff
@@ -0,0 +1,7 @@
+export { default as Ellipsis } from './Ellipsis';
```
adamstankiewicz, line 1: tiniest of nits: this could be a vanilla `.js` file as it doesn't render JSX

src/index.js
```diff
@@ -97,16 +97,16 @@ export {
...
-  default as Pagination,
   PAGINATION_BUTTON_LABEL_PREV,
   PAGINATION_BUTTON_ICON_BUTTON_NEXT_ALT,
   PAGINATION_BUTTON_ICON_BUTTON_PREV_ALT,
   PAGINATION_BUTTON_LABEL_PAGE_OF_COUNT,
   PAGINATION_BUTTON_LABEL_CURRENT_PAGE,
   PAGINATION_BUTTON_LABEL_NEXT,
   PAGINATION_BUTTON_LABEL_PAGE,
-} from './Pagination';
+} from './Pagination/constants';
```
adamstankiewicz, line 109: nit: i wonder if it's worth sticking to convention in this file of exporting constants from the top-level component file? Could consider adding the following to `./src/Pagination/index.jsx`:

```jsx
export * from './constants';
```

src/Pagination/index.jsx
viktorrusakov: Good catch! That was a while ago, took me some time to remember why I did this haha, I was definitely trying to avoid making breaking changes, but here it's almost unavoidable I think.

The reason is that in original implementation these icon props worked only for the `default` variant of `Pagination`, i.e. they were ignored for all other variants and you could not override icons for `secondary` variant for example. I felt like this definitely should be fixed during this refactoring and I don't really see any other way, because we use `IconButton`s for other variants and they expect our icon instances, not any node.

We could probably add some magic with our `withDeprecatedProps` HOC to support both behaviours, but I think that would be pretty ugly... not sure if it's worth doing that, I feel like this breaking change is pretty minor and we can go with it to keep this component clean.

src/Pagination/index.jsx
```diff
@@ -483,40 +116,34 @@ Pagination.propTypes = {
...
-    previous: PAGINATION_BUTTON_LABEL_PREV,
-    next: PAGINATION_BUTTON_LABEL_NEXT,
-    page: PAGINATION_BUTTON_LABEL_PAGE,
-    currentPage: PAGINATION_BUTTON_LABEL_CURRENT_PAGE,
-    pageOfCount: PAGINATION_BUTTON_LABEL_PAGE_OF_COUNT,
+    previous: 'Previous',
+    next: 'Next',
+    page: 'Page',
+    currentPage: 'Current Page',
+    pageOfCount: 'of',
```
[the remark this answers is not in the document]
viktorrusakov: Yeah, it crossed my mind, but I think in this case it's harder than it seems because we use these strings in a way that does not support translation I think, e.g. the screen reader text for the component is generated as follows:
```jsx
const getScreenReaderText = () => `${buttonLabels.page} ${currentPage}, ${buttonLabels.currentPage}, ${buttonLabels.pageOfCount} ${pageCount}`;
```
(this would generate following text `Page 3, Current Page, of 20`)

I'm pretty sure that for it to be translated correctly, the whole string has to be marked for translation, not just parts of it (in this case it would be `page`, `currentPage` and `pageOfCount` strings)

So yeah, I'm not sure that just marking these props for i18n would be useful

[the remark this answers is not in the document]
viktorrusakov: I think we might need to make another breaking change here to correctly support i18n, i.e. make consumers pass the whole text instead of parts of it, for example add `screenReaderText` prop and allow adding placeholders there, e.g. default value would be `Page {currentPage}, Current Page, of {totalPagesCount}` where we have predefined placeholders that will be replaced by the component with correct values. What do you think?

src/index.js
viktorrusakov: Yeah, makes sense, updated, thanks!

PKulkoRaccoonGang:
New PR: https://github.com/openedx/paragon/pull/2837

openedx-webhooks:
@viktorrusakov Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1911&entry.1484655318=https://github.com/viktorrusakov&entry.752974735=2023-01-13+07:13&entry.1917517419=2023-11-21+17:41&entry.2133058324=No) so we can improve your experience in the future.

PR #1962, opened 2023-01-27
[BD-46] chore: add type-check step to CI and update typescript configs

viktorrusakov:
## Description

Adds type check step to CI, simplifies typescript configs and fixes `target` to be `es6`.

### Deploy Preview

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,8 @@ jobs:
         npm ci
         npm run generate-component-install
         npm run doc-site-install
+    - name: Check Types
+      run: npm run type-check
     - name: Lint
       run: npm run lint
     - name: Test
--- a/www/src/pages/insights.tsx
+++ b/www/src/pages/insights.tsx
@@ -17,6 +17,7 @@ import SummaryUsageExamples, { ISummaryUsageExamples } from '../components/insig
 import ProjectUsageExamples, { IProjectUsageExamples } from '../components/insights/ProjectUsageExamples';
 import ComponentUsageExamples, { IComponentUsageExamples } from '../components/insights/ComponentUsageExamples';
 import getGithubProjectUrl from '../utils/getGithubProjectUrl';
+// @ts-ignore
 import dependentProjectsAnalysis from '../../../dependent-usage.json'; // eslint-disable-line
 import { INSIGHTS_TABS, INSIGHTS_PAGES } from '../config';
 import InsightsContext from '../context/InsightsContext';
@@ -84,7 +85,7 @@ export interface IComponentUsage {
 
 const dependentProjects: IDependentUsage[] = [];
 
-const componentsUsage: Record<string, IComponentUsage[]> = dependentProjectsUsages
+const componentsUsage: Record<string, IComponentUsageData[]> = dependentProjectsUsages
   .reduce((accumulator: any, project: any) => {
     dependentProjects.push({
       ...project,
--- a/www/tsconfig.json
+++ b/www/tsconfig.json
@@ -1,35 +1,14 @@
 {
+  "extends": "../tsconfig.json",
   "compilerOptions": {
-    "disableSourceOfProjectReferenceRedirect": true,
-    "disableSolutionSearching": true,
-    "target": "es6",
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "es6"
-    ],
-    "jsx": "react",
-    "module": "es6",
-    "moduleResolution": "node",
-    "preserveSymlinks": true,
-    "allowJs": true,
-    "skipLibCheck": true,
-    "esModuleInterop": true,
-    "allowSyntheticDefaultImports": true,
-    "strict": true,
-    "forceConsistentCasingInFileNames": true,
-    "noFallthroughCasesInSwitch": true,
+    "rootDir": "src",
     "resolveJsonModule": true,
-    "isolatedModules": true,
-    "noEmit": false,
-    "baseUrl": "src",
+    "noImplicitAny": false,
     "paths": {
-      "~paragon-react": ["../../src"],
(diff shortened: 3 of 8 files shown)


openedx-webhooks:
Thanks for the pull request, @viktorrusakov!


.github/workflows/ci.yml
```diff
@@ -25,6 +25,8 @@ jobs:
         npm ci
         npm run generate-component-install
         npm run doc-site-install
+    - name: Check Types
```
viktorrusakov, line 28: this is embarrassing, but we never checked our types... (not sure if anyone did it locally too 😅)

www/src/pages/insights.tsx
```diff
@@ -17,6 +17,7 @@ import SummaryUsageExamples, { ISummaryUsageExamples } from '../components/insig
 import ProjectUsageExamples, { IProjectUsageExamples } from '../components/insights/ProjectUsageExamples';
 import ComponentUsageExamples, { IComponentUsageExamples } from '../components/insights/ComponentUsageExamples';
 import getGithubProjectUrl from '../utils/getGithubProjectUrl';
+// @ts-ignore
 import dependentProjectsAnalysis from '../../../dependent-usage.json'; // eslint-disable-line
```
viktorrusakov, line 21: typescript complains that we export a file outside of `rootDIr` specified in `tsconfig.json`, I didn't find an easy fix for this except for moving the file to `www` project, but I'm not sure that we should do that

www/tsconfig.json
```diff
@@ -1,35 +1,14 @@
...
-    "allowSyntheticDefaultImports": true,
-    "strict": true,
-    "forceConsistentCasingInFileNames": true,
-    "noFallthroughCasesInSwitch": true,
+    "rootDir": "src",
     "resolveJsonModule": true,
-    "isolatedModules": true,
-    "noEmit": false,
-    "baseUrl": "src",
+    "noImplicitAny": false,
```
viktorrusakov, line 6: Sooo this is not good and should be definitely set to `true` but we cannot do it now. Reason is that core Paragon `index.js` file (the file that exports all component) does not have any typings which results in an error when we try to import something from it into TypeScript file, e.g. if we do `import { Hyperlink } from '~paragon-react';`) typescript does not find any types for `Hyperlink` and thinks it has an implicit `any` type.

We probably need to provide types for `index.js` file, maybe in the follow-up PR? also, we might want to eventually publish our types as a separate package under `@types/paragon`, I think a lot of projects do this when they have typescript, (e.g. [react-bootstrap's types package](https://www.npmjs.com/package/@types/react-bootstrap))

adamstankiewicz: @viktorrusakov This types issue came up recently in conversation with @marlonkeating who is trying to bring TypeScript into `frontend-app-admin-portal` where TypeScript was complaining that Paragon's `Button` component (and others) didn't have types defined and would need to define them locally in the application to get around it.

It looks like publishing `@types/paragon` might even be possible to do self-service, according to their docs: 

* https://github.com/DefinitelyTyped/DefinitelyTyped#make-a-pull-request
* https://devblogs.microsoft.com/typescript/changes-to-how-we-manage-definitelytyped/#what-pull-requests-can-be-merged-by-definition-owners
* https://github.com/DefinitelyTyped/dt-mergebot/

Based on these readings, it looks like an initial PR would need to be reviewed by the DT maintainers but subsequent edits might not (self-serve), which is cool.

Given this, yeah, I agree with your proposal here... get types defined for `index.js` as a follow-up PR and then try to get `@types/paragon` published. 

I filed two follow-up issues related to this:
* https://github.com/openedx/paragon/issues/1964
* https://github.com/openedx/paragon/issues/1965

www/src/pages/insights.tsx
adamstankiewicz: Hmm, yeah I don't think we'll be able to do that, at least not trivially. Good call to ignore for now.

www/tsconfig.json
adamstankiewicz: I suppose the other short term fix for consumers' type issues would be change `noImplicitAny` to `false` in the root `tsconfig.json` temporarily as well.

I'd probably lean towards introducing types vs. changing `noImplicitAny=false` temporarily, though curious to hear what you think!

adamstankiewicz: Follow-up from that last comment, it might actually be worth changing `noImplicitAny=false` for now as the quickest path forward. 

This `tsconfig.json` is largely acting as if its a consuming application (e.g., an MFE). When we introduce `tsconfig.json` into `@edx/frontend-build` ([PR](https://github.com/openedx/frontend-build/pull/294)), it's likely we'll want to keep `noImplicitAny=false` in that repo as we don't want the default for MFEs to be strictly-typed since that would require a large effort to refactor everything to TS, so leaving it as optionally-typed supports a more incremental, opt-in migration for MFEs.

That said, we could also decide for the Paragon documentation website that we _do_ want to be strictly-typed as you're proposing here; though, yes, in that case, we would need to define types for all the exports in `index.js`.

adamstankiewicz (approved):

openedx-webhooks:
@viktorrusakov 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1962&entry.1484655318=https://github.com/viktorrusakov&entry.752974735=2023-01-27+09:20&entry.1917517419=2023-01-27+21:38&entry.2133058324=Yes) so we can improve your experience in the future.

PR #1984, opened 2023-01-30
[BD-46] docs: renamed Paragon Segment events

PKulkoRaccoonGang:
## Description

Rename Segment events to follow a common convention: `openedx.paragon.{module}.{event descriptor}.{action} # action in past tense`

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/www/src/context/SettingsContext.tsx
+++ b/www/src/context/SettingsContext.tsx
@@ -49,9 +49,7 @@ function SettingsContextProvider({ children }) {
 
   const toggleSettings = (value: boolean) => {
     setShowSettings(value);
-    global.analytics.track(`openedx.paragon.docs.settings.${value ? 'opened' : 'closed'}`, {
-      value: value ? 'show' : 'hide',
-    });
+    global.analytics.track(`openedx.paragon.docs.settings.${value ? 'opened' : 'closed'}`);
   };
 
   // this hook will be called after the first render, so we can safely access localStorage
--- a/www/src/context/SettingsContext.tsx
+++ b/www/src/context/SettingsContext.tsx
@@ -44,12 +44,14 @@ function SettingsContextProvider({ children }) {
     }
     setSettings(prevState => ({ ...prevState, [key]: value }));
     global.localStorage.setItem('pgn__settings', JSON.stringify({ ...settings, [key]: value }));
-    global.analytics.track(`${key[0].toUpperCase() + key.slice(1)} change`, { [key]: value });
+    global.analytics.track(`openedx.paragon.docs.settings.${key}.changed`, { [key]: value });
   };
 
   const toggleSettings = (value: boolean) => {
     setShowSettings(value);
-    global.analytics.track('Toggle Settings', { value: value ? 'show' : 'hide' });
+    global.analytics.track(`openedx.paragon.docs.settings.${value ? 'opened' : 'closed'}`, {
+      value: value ? 'show' : 'hide',
+    });
   };
 
   // this hook will be called after the first render, so we can safely access localStorage
--- a/www/src/pages/insights.tsx
+++ b/www/src/pages/insights.tsx
@@ -365,7 +365,7 @@ export default function InsightsPage({ pageContext: { tab } }: { pageContext: { 
 
   const handleOnSelect = (value: string) => {
     if (value !== tab) {
-      global.analytics.track(`openedx.paragon.docs.insights.tabs.${value.toLowerCase().trim()}.clicked`, { tab: value });
+      global.analytics.track(`openedx.paragon.docs.insights.tabs.${value.toLowerCase().trim()}.clicked`);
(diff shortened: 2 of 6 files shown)


openedx-webhooks:


www/src/pages/insights.tsx
```diff
@@ -365,7 +365,7 @@ export default function InsightsPage({ pageContext: { tab } }: { pageContext: {
 
   const handleOnSelect = (value: string) => {
     if (value !== tab) {
-      global.analytics.track('Usage Insights', { tab: value });
+      global.analytics.track(`openedx.paragon.docs.insights.tabs.${value.toLowerCase().trim()}.clicked`, { tab: value });
```
viktorrusakov, line 368: Same here, we probably don't need to send additional data (`{ tab: value }`) here because event name already indicates which tab was clicked (e.g., sending `openedx.paragon.docs.insights.tabs.summary.clicked` event is enough to know that `Summary` tab was clicked, no need to also send `{ tab: 'summary' }`)

www/src/context/SettingsContext.tsx
```diff
@@ -44,12 +44,14 @@ function SettingsContextProvider({ children }) {
...
     global.localStorage.setItem('pgn__settings', JSON.stringify({ ...settings, [key]: value }));
-    global.analytics.track(`${key[0].toUpperCase() + key.slice(1)} change`, { [key]: value });
+    global.analytics.track(`openedx.paragon.docs.settings.${key}.changed`, { [key]: value });
   };
 
   const toggleSettings = (value: boolean) => {
     setShowSettings(value);
-    global.analytics.track('Toggle Settings', { value: value ? 'show' : 'hide' });
+    global.analytics.track(`openedx.paragon.docs.settings.${value ? 'opened' : 'closed'}`, {
+      value: value ? 'show' : 'hide',
```
viktorrusakov, line 53: No need to send additional data here because event name already describes whether settings were opened or closed. (right now when settings are opened you send  `'openedx.paragon.docs.settings.opened', { value: 'show' }` event, but it is enough to just send `openedx.paragon.docs.settings.opened` event without additional data)

www/src/context/SettingsContext.tsx
```diff
@@ -44,12 +44,14 @@ function SettingsContextProvider({ children }) {
     }
     setSettings(prevState => ({ ...prevState, [key]: value }));
     global.localStorage.setItem('pgn__settings', JSON.stringify({ ...settings, [key]: value }));
-    global.analytics.track(`${key[0].toUpperCase() + key.slice(1)} change`, { [key]: value });
+    global.analytics.track(`openedx.paragon.docs.settings.${key}.changed`, { [key]: value });
```
viktorrusakov, line 47: I think we should try to send additional data in the form `{ value: 'some_value' }` to be consistent, e.g. here when language is changed you currently send `'openedx.paragon.docs.settings.language.changed', { language: 'Spanish' }`, instead we could send `'openedx.paragon.docs.settings.language.changed', { value: 'Spanish' }`, where event name describes which value changed and additional data provides new value. Same could be done for other events.

@adamstankiewicz what do you think? I actually don't know how it affects Snowflake dashboard, if it does at all 😅

www/src/pages/insights.tsx
PKulkoRaccoonGang: Corrected, thanks

www/src/context/SettingsContext.tsx
adamstankiewicz: Hmm, good call out @viktorrusakov. Given that the `*.language.changed` is in the event name, I don't believe we'd ever need to do a `group by` on the `language`/`value` attribute in the event properties here.

If we had this event named something like `openedx.paragon.docs.settings.changed` and passed `{ settingsAttribute: 'language', value: 'Spanish' }`, then we might want the `value` attribute because that SQL query would definitely require a `group by` on a attribute that's consistent across all setting fields.

When looking at each explicit event per setting field (e.g., `openedx.paragon.docs.settings.language.changed`), the SQL query can just `group by` on the known  event-specific property (e.g., `language`).

adamstankiewicz (approved):

openedx-webhooks:
@PKulkoRaccoonGang 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1984&entry.1484655318=https://github.com/PKulkoRaccoonGang&entry.752974735=2023-01-30+23:05&entry.1917517419=2023-02-01+13:36&entry.2133058324=Yes) so we can improve your experience in the future.

Issue #1967, opened 2023-01-27
undefined/null fallback image logo source on Card working incorrectly


brobro10000:
The `fallbackLogoSrc` is incorrectly calling image URL when the `logoSrc` prop is not correctly fetching the image.

If the `fallbackLogoSrc` is not declared in the prop and the `logoSrc` fails, the component will attempt to render the fallbackLogoSrc infinitely 
<img width="752" alt="Screenshot 2023-01-27 at 9 37 04 AM" src="https://user-images.githubusercontent.com/82611798/215112638-d89c627b-2c59-48ca-abff-2c213a0df151.png">

If the `fallbackLogoSrc` retrieves either a null or undefined value, the same behavior will occur as above: the component will attempt to infinitely call the `fallbackLogoSrc`
<img width="756" alt="Screenshot 2023-01-27 at 9 39 39 AM" src="https://user-images.githubusercontent.com/82611798/215113061-9b418bcf-ecaf-4c07-aacb-7ca86b193877.png">
<img width="729" alt="Screenshot 2023-01-27 at 9 41 20 AM" src="https://user-images.githubusercontent.com/82611798/215113424-e07d7c19-edc8-486c-b59e-42dbf20c4140.png">

The resulting behavior on the card component is the image will render a blank white background in the logoSrc or show a blinking broken image link logo (blinking occurs due to attempted image src retrieval)
<img width="313" alt="Screenshot 2023-01-27 at 9 43 48 AM" src="https://user-images.githubusercontent.com/82611798/215113983-dc32284f-637b-4c67-8aea-26badad08f4c.png">

## **Acceptance Criteria**:
- [ ] Ensure when `logoSrc` image fails to load and no valid `fallbackLogoSrc` image is specified, it doesn't attempt to re-render the image infinitely.
- [ ] Ensure when `logoSrc` fails to load and no valid `fallbackLogoSrc` image is provided, the logo in `Card.ImageCap` should not appear as a white background with no image. Instead, we should hide the logo's wrapper altogether.

PKulkoRaccoonGang:
PR: https://github.com/openedx/paragon/pull/1994

PR #1994, opened 2023-02-03
[BD-46] fix: fixed undefined/null image logo src

PKulkoRaccoonGang:
## Description

The fallbackLogoSrc is incorrectly calling the image URL when the logoSrc prop is not correctly fetching the image.
Issue: https://github.com/openedx/paragon/issues/1967

### Deploy Preview

[Card Fallback example](https://deploy-preview-1994--paragon-openedx.netlify.app/components/card/#fallback-image)

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Card/CardImageCap.jsx
+++ b/src/Card/CardImageCap.jsx
@@ -46,14 +46,20 @@ const CardImageCap = React.forwardRef(({
     );
   }
 
-  const handleSrcFallback = (event, altSrc) => {
+  const handleSrcFallback = (event, altSrc, imageKey) => {
     const { currentTarget } = event;
 
     if (!altSrc || currentTarget.src.endsWith(altSrc)) {
-      currentTarget.style.display = 'none';
-    } else if (currentTarget.src !== altSrc) {
-      currentTarget.src = altSrc;
+      if (imageKey === 'imageCap') {
+        currentTarget.src = cardSrcFallbackImg;
+      } else {
+        setShow(s => ({ ...s, logoCap: false }));
+      }
+
+      return;
     }
+
+    currentTarget.src = altSrc;
   };
 
   return (
@@ -61,16 +67,16 @@ const CardImageCap = React.forwardRef(({
       <img
         className={classNames('pgn__card-image-cap', { show: show.imageCap })}
         src={src}
-        onError={(event) => handleSrcFallback(event, fallbackSrc)}
-        onLoad={() => setShow({ ...show, imageCap: src })}
+        onError={(event) => handleSrcFallback(event, fallbackSrc, 'imageCap')}
+        onLoad={() => setShow(s => ({ ...s, imageCap: !!src }))}
         alt={srcAlt}
       />
       {!!logoSrc && (
         <img
           className={classNames('pgn__card-logo-cap', { show: show.logoCap })}
           src={logoSrc}
-          onError={(event) => handleSrcFallback(event, fallbackLogoSrc)}
-          onLoad={() => setShow({ ...show, logoCap: logoSrc })}
+          onError={(event) => handleSrcFallback(event, fallbackLogoSrc, 'logoCap')}
+          onLoad={() => setShow(s => ({ ...s, logoCap: !!logoSrc }))}
           alt={logoAlt}
         />
       )}
--- a/src/Card/CardImageCap.jsx
+++ b/src/Card/CardImageCap.jsx
@@ -23,7 +23,8 @@ const CardImageCap = React.forwardRef(({
   className,
 }, ref) => {
   const { orientation, isLoading } = useContext(CardContext);
-  const [show, setShow] = useState({ imageCap: false, logoCap: false });
+  const [showImageCap, setShowImageCap] = useState(false);
+  const [showLogoCap, setShowLogoCap] = useState(false);
   const wrapperClassName = `pgn__card-wrapper-image-cap ${orientation}`;
 
   if (isLoading) {
@@ -53,7 +54,7 @@ const CardImageCap = React.forwardRef(({
       if (imageKey === 'imageCap') {
         currentTarget.src = cardSrcFallbackImg;
       } else {
-        setShow(s => ({ ...s, logoCap: false }));
+        setShowLogoCap(false);
       }
 
       return;
@@ -66,19 +67,19 @@ const CardImageCap = React.forwardRef(({
     <div className={classNames(className, wrapperClassName)} ref={ref}>
       {!!src && (
         <img
-          className={classNames('pgn__card-image-cap', { show: show.imageCap })}
+          className={classNames('pgn__card-image-cap', { show: showImageCap })}
           src={src}
           onError={(event) => handleSrcFallback(event, fallbackSrc, 'imageCap')}
-          onLoad={() => setShow(s => ({ ...s, imageCap: true }))}
+          onLoad={() => setShowImageCap(true)}
           alt={srcAlt}
         />
       )}
       {!!logoSrc && (
         <img
-          className={classNames('pgn__card-logo-cap', { show: show.logoCap })}
+          className={classNames('pgn__card-logo-cap', { show: showLogoCap })}
           src={logoSrc}
           onError={(event) => handleSrcFallback(event, fallbackLogoSrc, 'logoCap')}
-          onLoad={() => setShow(s => ({ ...s, logoCap: true }))}
+          onLoad={() => setShowLogoCap(true)}
           alt={logoAlt}
         />
       )}
--- a/src/Card/CardImageCap.jsx
+++ b/src/Card/CardImageCap.jsx
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react';
+import React, { useContext, useState } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import Skeleton from 'react-loading-skeleton';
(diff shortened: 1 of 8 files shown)


openedx-webhooks:


brobro10000 (approved):
LGTM, thanks for the fix. 🥇

src/Card/CardImageCap.jsx
```diff
@@ -47,7 +47,10 @@ const CardImageCap = React.forwardRef(({
 
   const handleSrcFallback = (event, altSrc) => {
     const { currentTarget } = event;
-    if (currentTarget.src !== altSrc) {
+
+    if (!altSrc || currentTarget.src.endsWith(altSrc)) {
+      currentTarget.style.display = 'none';
+    } else if (currentTarget.src !== altSrc) {
```
adamstankiewicz, line 53: It looks like this approach ends up rendering an empty logo div in the `Card` briefly before this logic hides it, leading to a brief flicker of this empty, white logo div.

I wonder if we could flip this logic around a bit? Keep the logo image hidden until we know we loaded an image properly, and then toggle the logo div to be visible; otherwise, keep it hidden.

Curious to hear your two cents @PKulkoRaccoonGang :)

brobro10000 (changes requested):
Looking at a re-rendering issue with the logo

src/Card/CardImageCap.jsx
adamstankiewicz: Example: 

[pgn-card-fallback-img-review.webm](https://user-images.githubusercontent.com/2828721/216644321-2d6e2ff3-6e11-4976-889e-fc906701d44b.webm)

brobro10000 (approved):
Great job fixing the render bug. LGTM

src/Card/CardImageCap.jsx
```diff
@@ -47,24 +48,29 @@ const CardImageCap = React.forwardRef(({
 
   const handleSrcFallback = (event, altSrc) => {
     const { currentTarget } = event;
-    if (currentTarget.src !== altSrc) {
+
+    if (!altSrc || currentTarget.src.endsWith(altSrc)) {
+      currentTarget.style.display = 'none';
```
adamstankiewicz, line 53: If we're now using the `show` class to control the image visibility, is setting the `display` style to "none" here redundant? Should we instead rely solely on the `show` state variable rather than having 2 potentially competing approaches to hiding the element?

adamstankiewicz: I wonder if this function could possible get simplified to something similar to this:

```jsx
const handleSrcFallback = (event, altSrc, imageKey) => {
  const { currentTarget } = event;
  if (!altSrc || currentTarget.src === altSrc) {
    setShow(s => ({ ...s, [imageKey]: false }));
    return;
  }
  currentTarget.src = altSrc;
};
```

where `imageKey` is `imageCap` or `logoCap` depending on the caller of `handleSrcFallback`. 

[aside] I wonder if having separate state variables (one per image) would simplify this suggestion at all, though.

src/Card/CardImageCap.jsx
```diff
@@ -47,24 +48,29 @@ const CardImageCap = React.forwardRef(({
...
   };
 
   return (
     <div className={classNames(className, wrapperClassName)} ref={ref}>
       <img
-        className="pgn__card-image-cap"
+        className={classNames('pgn__card-image-cap', { show: show.imageCap })}
         src={src}
         onError={(event) => handleSrcFallback(event, fallbackSrc)}
+        onLoad={() => setShow({ ...show, imageCap: src })}
```
adamstankiewicz, line 65: Using the `show` state variable in the `setShow` state setter function is an anti-pattern. Instead, to update state based on previous state, you can use the callback function approach to set state instead, i.e:

```jsx
setShow(s => ({ ...s, imageCap: true }))
```

adamstankiewicz, line 65: [suggestion] If this new state variable is initialized with boolean values, we should keep the data types consistent. In this line, it looks like the string `src` is being passed instead of a true boolean data type.

src/Card/CardImageCap.jsx
PKulkoRaccoonGang: Yes, this is misbehavior. Thanks 👍

src/Card/CardImageCap.jsx
```diff
@@ -45,26 +46,37 @@ const CardImageCap = React.forwardRef(({
     );
   }
 
-  const handleSrcFallback = (event, altSrc) => {
+  const handleSrcFallback = (event, altSrc, imageKey) => {
     const { currentTarget } = event;
-    if (currentTarget.src !== altSrc) {
-      currentTarget.src = altSrc;
+
+    if (!altSrc || currentTarget.src.endsWith(altSrc)) {
```
adamstankiewicz, line 52: nit/curious: could this be simplified to `if (!altSrc || currentTarget.src === altSrc) {`?

src/Card/CardImageCap.jsx
```diff
@@ -45,26 +46,37 @@ const CardImageCap = React.forwardRef(({
...
+      if (imageKey === 'imageCap') {
+        currentTarget.src = cardSrcFallbackImg;
```
adamstankiewicz, line 54: Good idea here to fall back to the default `cardSrcFallbackImg` in this case! Should we ensure this line gets covered in the tests?

src/Card/CardImageCap.jsx
```diff
@@ -45,26 +46,37 @@ const CardImageCap = React.forwardRef(({
...
 
   return (
     <div className={classNames(className, wrapperClassName)} ref={ref}>
       <img
-        className="pgn__card-image-cap"
+        className={classNames('pgn__card-image-cap', { show: show.imageCap })}
         src={src}
-        onError={(event) => handleSrcFallback(event, fallbackSrc)}
+        onError={(event) => handleSrcFallback(event, fallbackSrc, 'imageCap')}
+        onLoad={() => setShow(s => ({ ...s, imageCap: !!src }))}
```
adamstankiewicz, line 71: [curious/question] If `src` is an optional prop for `Card.ImageCap`, should we render the `` element conditionally similar to `!!logoSrc && (...)` below?

src/Card/CardImageCap.jsx
```diff
@@ -45,26 +46,37 @@ const CardImageCap = React.forwardRef(({
...
         alt={srcAlt}
       />
       {!!logoSrc && (
         <img
-          className="pgn__card-logo-cap"
+          className={classNames('pgn__card-logo-cap', { show: show.logoCap })}
           src={logoSrc}
-          onError={(event) => handleSrcFallback(event, fallbackLogoSrc)}
+          onError={(event) => handleSrcFallback(event, fallbackLogoSrc, 'logoCap')}
+          onLoad={() => setShow(s => ({ ...s, logoCap: !!logoSrc }))}
```
adamstankiewicz, line 79: nit: since we're already in the `!!logoSrc` conditional here, you might just pass `true` here.

src/Card/CardImageCap.jsx
PKulkoRaccoonGang: Yes, it`s true
Thanks

src/Card/CardImageCap.jsx
PKulkoRaccoonGang: I agree, two different approaches are not what we need

src/Card/CardImageCap.jsx
PKulkoRaccoonGang: Using the option you suggested, we run the risk of getting an infinite image `src` call. This comparison `!altSrc || currentTarget.src === altSrc` successfully processes the passed link, but if we try to pass a string value to `fallbackLogoSrc` (for example, `fallbackLogoSrc="undefined "`), we will get an endless request along the path passed to the `src`.

image

src/Card/CardImageCap.jsx
PKulkoRaccoonGang: Good, thanks

src/Card/CardImageCap.jsx
PKulkoRaccoonGang: Yes, let's do this

src/Card/CardImageCap.jsx
PKulkoRaccoonGang: Done ✅

src/Card/CardImageCap.jsx
viktorrusakov: yeah, doing just `currentTarget.src === altSrc` fails if you provide relative url to the image instead of full url

src/Card/CardImageCap.jsx
viktorrusakov: > [aside] I wonder if having separate state variables (one per image) would simplify this suggestion at all, though.

I'm pretty sure it would, in this case using composite state does not give any benefit I think, just makes code a bit harder to understand

@PKulkoRaccoonGang let's introduce 2 states (`showImageCap`, `showLogoCap`) instead of one, this should simplify code

src/Card/CardImageCap.jsx
adamstankiewicz: Thanks for the clarification 👍

adamstankiewicz (approved):

openedx-webhooks:
@PKulkoRaccoonGang 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/1994&entry.1484655318=https://github.com/PKulkoRaccoonGang&entry.752974735=2023-02-03+00:52&entry.1917517419=2023-02-09+15:48&entry.2133058324=Yes) so we can improve your experience in the future.

PR #2112, opened 2023-03-03
[BD-46] feat: add description for tokens

### LINKED ISSUE #1948: Ensure core design tokens are documented with a `description` attribute

monteri:
## Description

Add description for tokens.

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/tokens/src/core/alias/size.json
+++ b/tokens/src/core/alias/size.json
@@ -1,11 +1,11 @@
 {
   "size": {
     "border": {
-      "width": { "value": "1px", "type": "dimension", "source": "$border-width" },
+      "width": { "value": "1px", "type": "dimension", "source": "$border-width", "description": "Border width value." },
       "radius": {
-        "base": { "value": "6px", "type": "dimension", "source": "$border-radius" },
-        "lg": { "value": "7px", "type": "dimension", "source": "$border-radius-lg" },
-        "sm": { "value": "4px", "type": "dimension", "source": "$border-radius-sm" }
+        "base": { "value": "6px", "type": "dimension", "source": "$border-radius", "description": "Border radius basic." },
+        "lg": { "value": "7px", "type": "dimension", "source": "$border-radius-lg", "description": "Border radius large." },
+        "sm": { "value": "4px", "type": "dimension", "source": "$border-radius-sm", "description": "Border radius small." }
       }
     }
   }
--- a/tokens/src/core/alias/size.json
+++ b/tokens/src/core/alias/size.json
@@ -1,11 +1,11 @@
 {
   "size": {
     "border": {
-      "width": { "value": "1px", "type": "dimension", "source": "$border-width", "description": "Border width value." },
+      "width": { "value": "1px", "type": "dimension", "source": "$border-width", "description": "Default border width." },
       "radius": {
-        "base": { "value": "6px", "type": "dimension", "source": "$border-radius", "description": "Border radius basic." },
-        "lg": { "value": "7px", "type": "dimension", "source": "$border-radius-lg", "description": "Border radius large." },
-        "sm": { "value": "4px", "type": "dimension", "source": "$border-radius-sm", "description": "Border radius small." }
+        "base": { "value": "6px", "type": "dimension", "source": "$border-radius", "description": "Default border radius." },
+        "lg": { "value": "7px", "type": "dimension", "source": "$border-radius-lg", "description": "Large border radius." },
+        "sm": { "value": "4px", "type": "dimension", "source": "$border-radius-sm", "description": "Small border radius." }
       }
     }
   }
--- a/tokens/src/core/global/breakpoints.json
+++ b/tokens/src/core/global/breakpoints.json
@@ -1,12 +1,12 @@
 {
   "size": {
     "breakpoint": {
-      "xs": { "value": "0", "type": "dimension" },
-      "sm": { "value": "576px", "type": "dimension" },
-      "md": { "value": "768px", "type": "dimension" },
-      "lg": { "value": "992px", "type": "dimension" },
-      "xl": { "value": "1200px", "type": "dimension" },
-      "xxl": { "value": "1400px", "type": "dimension" }
+      "xs": { "value": "0", "type": "dimension", "description": "Starting value for extra-small breakpoint." },
+      "sm": { "value": "576px", "type": "dimension", "description": "Starting value for small breakpoint." },
+      "md": { "value": "768px", "type": "dimension", "description": "Starting value for medium breakpoint." },
+      "lg": { "value": "992px", "type": "dimension", "description": "Starting value for large breakpoint." },
+      "xl": { "value": "1200px", "type": "dimension", "description": "Starting value for extra-large breakpoint." },
+      "xxl": { "value": "1400px", "type": "dimension", "description": "Starting value for extra-extra-large breakpoint." }
     }
   }
 }
--- a/tokens/src/core/global/breakpoints.json
+++ b/tokens/src/core/global/breakpoints.json
@@ -1,12 +1,12 @@
 {
   "size": {
     "breakpoint": {
-      "xs": { "value": "0", "type": "dimension", "description": "Starting value for extra-small breakpoint." },
-      "sm": { "value": "576px", "type": "dimension", "description": "Starting value for small breakpoint." },
-      "md": { "value": "768px", "type": "dimension", "description": "Starting value for medium breakpoint." },
-      "lg": { "value": "992px", "type": "dimension", "description": "Starting value for large breakpoint." },
-      "xl": { "value": "1200px", "type": "dimension", "description": "Starting value for extra-large breakpoint." },
-      "xxl": { "value": "1400px", "type": "dimension", "description": "Starting value for extra-extra-large breakpoint." }
+      "xs": { "value": "0", "type": "dimension", "description": "Starting breakpoint for portrait phones." },
+      "sm": { "value": "576px", "type": "dimension", "description": "Starting breakpoint for landscape phones." },
+      "md": { "value": "768px", "type": "dimension", "description": "Starting breakpoint for tablets." },
+      "lg": { "value": "992px", "type": "dimension", "description": "Starting breakpoint for desktops." },
+      "xl": { "value": "1200px", "type": "dimension", "description": "Starting breakpoint for large desktops." },
+      "xxl": { "value": "1400px", "type": "dimension", "description": "Starting breakpoint for large desktops, more than 1400px." }
     }
   }
 }
--- a/tokens/src/core/global/display.json
+++ b/tokens/src/core/global/display.json
@@ -2,13 +2,13 @@
   "typography": {
     "display": {
       "line-height": {
-        "base": { "value": "1", "type": "dimension", "source": "$display-line-height", "description": "Line height basic value." },
-        "mobile": { "value": "3.5rem", "type": "dimension", "source": "$display-mobile-line-height", "description": "Line height for mobile devices." }
+        "base": { "value": "1", "type": "dimension", "source": "$display-line-height", "description": "Standard line height." },
+        "mobile": { "value": "3.5rem", "type": "dimension", "source": "$display-mobile-line-height", "description": "Mobile line height." }
       },
-      "1": { "value": "3.75rem", "type": "dimension", "source": "$display1-size", "description": "Font size of level 1." },
-      "2": { "value": "4.875rem", "type": "dimension", "source": "$display2-size", "description": "Font size of level 2." },
-      "3": { "value": "5.625rem", "type": "dimension", "source": "$display3-size", "description": "Font size of level 3." },
-      "4": { "value": "7.5rem", "type": "dimension", "source": "$display4-size", "description": "Font size of level 4." },
+      "1": { "value": "3.75rem", "type": "dimension", "source": "$display1-size", "description": "Font size for heading of level 1." },
+      "2": { "value": "4.875rem", "type": "dimension", "source": "$display2-size", "description": "Font size for heading of level 2." },
+      "3": { "value": "5.625rem", "type": "dimension", "source": "$display3-size", "description": "Font size for heading of level 3." },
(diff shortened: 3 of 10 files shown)


tokens/src/core/alias/size.json
```diff
@@ -1,11 +1,11 @@
 {
   "size": {
     "border": {
-      "width": { "value": "1px", "type": "dimension", "source": "$border-width" },
+      "width": { "value": "1px", "type": "dimension", "source": "$border-width", "description": "Border width value." },
```
brian-smith-tcril, line 4: Is the plan to have *every* token include a description? I don't think "Border width value" provides any any additional context over `border-width` or `side.border.width`.

If the idea is to be able to rely exclusively on descriptions then I understand having this, but I would suggest simply "Border width" without "value"

tokens/src/core/alias/size.json
```diff
@@ -1,11 +1,11 @@
...
       "radius": {
-        "base": { "value": "6px", "type": "dimension", "source": "$border-radius" },
-        "lg": { "value": "7px", "type": "dimension", "source": "$border-radius-lg" },
-        "sm": { "value": "4px", "type": "dimension", "source": "$border-radius-sm" }
+        "base": { "value": "6px", "type": "dimension", "source": "$border-radius", "description": "Border radius basic." },
+        "lg": { "value": "7px", "type": "dimension", "source": "$border-radius-lg", "description": "Border radius large." },
+        "sm": { "value": "4px", "type": "dimension", "source": "$border-radius-sm", "description": "Border radius small." }
```
brian-smith-tcril, line 8: Another place to decide on naming conventions. My gut feeling is "Large border radius" feels better than "Border radius large.", but I'd like to hear more thoughts.

tokens/src/core/global/breakpoints.json
```diff
@@ -1,12 +1,12 @@
...
-      "md": { "value": "768px", "type": "dimension" },
-      "lg": { "value": "992px", "type": "dimension" },
-      "xl": { "value": "1200px", "type": "dimension" },
-      "xxl": { "value": "1400px", "type": "dimension" }
+      "xs": { "value": "0", "type": "dimension", "description": "Starting value for extra-small breakpoint." },
+      "sm": { "value": "576px", "type": "dimension", "description": "Starting value for small breakpoint." },
+      "md": { "value": "768px", "type": "dimension", "description": "Starting value for medium breakpoint." },
+      "lg": { "value": "992px", "type": "dimension", "description": "Starting value for large breakpoint." },
+      "xl": { "value": "1200px", "type": "dimension", "description": "Starting value for extra-large breakpoint." },
+      "xxl": { "value": "1400px", "type": "dimension", "description": "Starting value for extra-extra-large breakpoint." }
```
brian-smith-tcril, line 9: Is there something special about these that warrants adding "Starting value" to the description? I could also see adding more context via these descriptions such as *why* each breakpoint was picked.

tokens/src/core/global/display.json
```diff
@@ -2,19 +2,19 @@
   "typography": {
     "display": {
       "line-height": {
-        "base": { "value": "1", "type": "dimension", "source": "$display-line-height" },
-        "mobile": { "value": "3.5rem", "type": "dimension", "source": "$display-mobile-line-height" }
+        "base": { "value": "1", "type": "dimension", "source": "$display-line-height", "description": "Line height basic value." },
+        "mobile": { "value": "3.5rem", "type": "dimension", "source": "$display-mobile-line-height", "description": "Line height for mobile devices." }
```
brian-smith-tcril, line 6: Maybe something like "Standard line height" and "Mobile line height" instead here? To reiterate, don't take these comments as prescriptive, I'm just trying to spark conversation around what our description conventions should be.

tokens/src/core/global/display.json
```diff
@@ -2,19 +2,19 @@
...
       },
-      "1": { "value": "3.75rem", "type": "dimension", "source": "$display1-size" },
-      "2": { "value": "4.875rem", "type": "dimension", "source": "$display2-size" },
-      "3": { "value": "5.625rem", "type": "dimension", "source": "$display3-size" },
-      "4": { "value": "7.5rem", "type": "dimension", "source": "$display4-size" },
-      "mobile": { "value": "3.25rem", "type": "dimension", "source": "$display-mobile-size" },
+      "1": { "value": "3.75rem", "type": "dimension", "source": "$display1-size", "description": "Font size of level 1." },
+      "2": { "value": "4.875rem", "type": "dimension", "source": "$display2-size", "description": "Font size of level 2." },
+      "3": { "value": "5.625rem", "type": "dimension", "source": "$display3-size", "description": "Font size of level 3." },
+      "4": { "value": "7.5rem", "type": "dimension", "source": "$display4-size", "description": "Font size of level 4." },
```
brian-smith-tcril, line 11: "level 1" etc. doesn't seem to add any context here. Not sure what we should have here instead though.

tokens/src/core/global/elevation.json
```diff
@@ -1,19 +1,19 @@
...
+      "200": { "value": 200, "type": "ratio", "description": "z-index of level 200." },
+      "400": { "value": 400, "type": "ratio", "description": "z-index of level 400." },
+      "600": { "value": 600, "type": "ratio", "description": "z-index of level 600." },
+      "800": { "value": 800, "type": "ratio", "description": "z-index of level 800." },
+      "1000": { "value": 1000, "type": "ratio", "description": "z-index of level 1000." },
+      "1200": { "value": 1200, "type": "ratio", "description": "z-index of level 1200." },
+      "1400": { "value": 1400, "type": "ratio", "description": "z-index of level 1400." },
+      "1600": { "value": 1600, "type": "ratio", "description": "z-index of level 1600." },
+      "1800": { "value": 1800, "type": "ratio", "description": "z-index of level 1800." },
+      "2000": { "value": 2000, "type": "ratio", "description": "z-index of level 2000." },
```
brian-smith-tcril, line 14: These feel like another example of the "do we expect to only see the descriptions?" question.

tokens/src/core/global/transition.json
```diff
@@ -1,7 +1,7 @@
 {
   "transition": {
-    "base": { "value": "all .2s ease-in-out", "type": "transition", "source": "$transition-base" },
-    "fade": { "value": "opacity .15s linear", "type": "transition", "source": "$transition-fade" },
-    "collapse": { "value": "height .35s ease", "type": "transition", "source": "$transition-collapse" }
+    "base": { "value": "all .2s ease-in-out", "type": "transition", "source": "$transition-base", "description": "Generic transition for any property change" },
+    "fade": { "value": "opacity .15s linear", "type": "transition", "source": "$transition-fade", "description": "Opacity transition that takes 150ms" },
+    "collapse": { "value": "height .35s ease", "type": "transition", "source": "$transition-collapse", "description": "Collapse transition that takes 350ms" }
```
brian-smith-tcril, line 5: These raise the question of what is being described. Is the point of the description to describe the *value* or the *variable*?

"Generic transition for any property changed" seems to describe the *variable*, since it has no mention of "ease-in-out" or 200ms.

The other two seem to describe the *value*, directly mentioning the type of transition and amount of time.

I'd think the goal would be to describe the variable, so when someone sets a new value for it they'd know where it would be used. This would prevent the situation where someone updates the fade transition to be slower and the description still says "150ms"

tokens/src/core/global/typography.json
```diff
@@ -5,72 +5,77 @@
...
           "serif": {
             "value": "-apple-system, BlinkMacSystemFont, \"Segoe UI\",  Roboto, \"Helvetica Neue\", Arial, \"Noto Sans\", sans-serif, \"Apple Color Emoji\",  \"Segoe UI Emoji\", \"Segoe UI Symbol\", \"Noto Color Emoji\"",
             "type": "fontFamily",
-            "source": "$font-family-sans-serif"
+            "source": "$font-family-sans-serif",
+            "description": "Sans-serif font family."
           }
         },
-        "serif": { "value": "serif", "source": "$font-family-serif" },
+        "serif": { "value": "serif", "source": "$font-family-serif", "description": "Serif font style" },
```
brian-smith-tcril, line 19: the other ones say "font family" but this one stays "font style"

brian-smith-tcril (commented):
Overall I really like the idea of having descriptions for design tokens. I think there's a lot of potential here.

This PR raises quite a few questions around how we want to use those descriptions. I left some comments with the ones that came to mind.

Hopefully these comments can spark some discussion!

tokens/src/core/alias/size.json
viktorrusakov: I *think* eventually we want to have description for every token, but not in this PR. At some point in the future we will have a UI to update token values without explicitly opening a PR (e.g., designers could update token values without engineers' help), and ideally we'd also want to have some kind of a description for tokens in that UI, our idea is to store and take such descriptions from the JSON files.

I agree though that most of the token names are self-explanatory and maybe we don't need descriptions for every token, on the other hand we could add it just to be consistent.

Also, yeah, `value` is redundant in the description, I'd suggest `Default border with` instead

tokens/src/core/global/breakpoints.json
viktorrusakov: Agree, we should also describe intention of the breakpoints, e.g. for which device they are used.

tokens/src/core/global/transition.json
viktorrusakov: Totally agree with that, we should only describe the variable

tokens/src/core/global/display.json
viktorrusakov: I think you variant is more human-friendly, so I would use it instead 🙂 Also, I think we definitely should avoid using word `value` in the description, it just doesn't contain any useful information for the reader.

tokens/src/core/global/display.json
viktorrusakov: I think here would be a good idea to look where we use these variables and come up with more detailed description. I think these are font sizes for headers? Maybe the description would be something like `Base font size for h1 element.` (not ideal but probably better than saying `level 1`)

viktorrusakov:
@brian-smith-tcril thanks for the great feedback! That's much appreciated 🙂

tokens/src/core/global/elevation.json
viktorrusakov: Yeah, that's a good example of redundant description if we are able to see variable name, which we most likely will.
I can't come up with any meaningful description here, maybe we should just leave it empty.

@adamstankiewicz curious what you think

tokens/src/core/alias/size.json
```diff
@@ -1,11 +1,11 @@
 {
   "size": {
     "border": {
-      "width": { "value": "1px", "type": "dimension", "source": "$border-width" },
+      "width": { "value": "1px", "type": "dimension", "source": "$border-width", "description": "Border width value." },
       "radius": {
-        "base": { "value": "6px", "type": "dimension", "source": "$border-radius" },
-        "lg": { "value": "7px", "type": "dimension", "source": "$border-radius-lg" },
-        "sm": { "value": "4px", "type": "dimension", "source": "$border-radius-sm" }
+        "base": { "value": "6px", "type": "dimension", "source": "$border-radius", "description": "Border radius basic." },
```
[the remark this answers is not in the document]
monteri: Changed to `default` here

tokens/src/core/alias/size.json
monteri: Changed to `Default border width`

tokens/src/core/alias/size.json
monteri: Changed to `Large border radius` and for the second one also

tokens/src/core/global/breakpoints.json
monteri: Changed to `Starting breakpoint for portrait phones` and so on for other breakpoints

tokens/src/core/global/display.json
monteri: Changed to "Standard line height" and "Mobile line height"

tokens/src/core/global/display.json
monteri: Changed to "Font size for heading of level 1." and so on

tokens/src/core/global/transition.json
monteri: Left here because there can be more transitions and description should specify the difference between them so I assume that timings will be user-friendly here. Yes, they shouldn't be forgotten when those values will be changed in future.

tokens/src/core/global/typography.json
monteri: Changed to `Serif font family.`

tokens/src/core/global/elevation.json
adamstankiewicz: Let's leave as is for the time being, but plan on revisiting once we have more clarity how these descriptions will ultimately get used. Goal of this PR is largely to get some reasonable descriptions in place to solidify the decision to do so and showcase a few examples.

tokens/src/core/global/transition.json
adamstankiewicz: I think it's reasonable for now to include the value in this case. Can always revisit later. Generally, though, I agree the descriptions should be describing the intent of the variable, not so much the value itself.

The descriptions will be mostly useful to theme authors to get a better idea of what each design token's purpose is.

adamstankiewicz (approved):
LGTM as a first pass of sample descriptions. We can revisit descriptions again once there's more clarity around how they will be used/written.

openedx-webhooks:
@monteri 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/2112&entry.1484655318=https://github.com/monteri&entry.752974735=2023-03-03+10:45&entry.1917517419=2023-03-24+13:57&entry.2133058324=Yes) so we can improve your experience in the future.

Issue #2107, opened 2023-03-01
`Form.Autosuggest` intended behavior of `onSelected` vs `onChange` attributes


MaxFrank13:
### Description

1. While developing with the `Form.Autosuggest` component, I noticed that the function passed to the `onSelected` attribute is being called during any `onChange` event — meaning it was calling this function after each character was entered into the input box. The `value` being passed to the `onSelected` function was `null`. See [Paragon docs](https://paragon-openedx.netlify.app/components/form/form-autosuggest/) for more details on how `Form.Autosuggest` and its attributes work.

2. In addition, the `onSelected` function will be called if the input is an exact match for one of the autosuggested options (i.e. if "Teacher" is an option, entering the word "teacher" will then trigger the `onSelected` function). Technically, the user hasn't actually made the selection at this point, they have only entered input into the search box.

After a discussion in Paragon Working Group, we came to the conclusion that this is most likely NOT the intended behavior. The `onSelected` function is intended to only be called when a selection is made, whereas the `onChange` is intended for tracking the `onChange` event of the root `input` element.

### Observations

I believe the below `setDisplayValue` function is actually responsible for both of the above issues.

https://github.com/openedx/paragon/blob/81f746f104b873802ae7655747ac6f4b3c1a7c5f/src/Form/FormAutosuggest.jsx#L147-L170

#### `null` value being passed to `onSelected`

On line 164, we are calling `setValue(null)`. The `null` value is what then gets passed to the `onSelected` function.

#### Selecting when an exact match is found

On line 155, we are filtering through the list of options for a match. If one is found, we end up feeding that value to the `setValue` function (linked below), which then calls the `onSelected` function.

https://github.com/openedx/paragon/blob/81f746f104b873802ae7655747ac6f4b3c1a7c5f/src/Form/FormAutosuggest.jsx#L34-L45

### Video

https://user-images.githubusercontent.com/92897870/223162632-dc43b4d4-6e3b-47e4-8e4b-cf7ea06fd0ec.mov

PR #2127, opened 2023-03-10
fix: onChange behavior for `Form.Autosuggest`

MaxFrank13:
## Description

This PR is to update the `Form.Autosuggest` component.

The `onSelected` function is being called during any `onChange` event. See below:

`onChange` => `setDisplayValue` => `setValue` => `onSelected` — we only want to call `onSelected` when a selection is made, not during every onChange.

Unless the user's input is an exact match for an option, `onSelected` will be passed a `null` value.  Meaning after each character is entered in the input, `onSelected(null)` is called. 

In addition, if "Teacher" is an option, entering the word "teacher" will then call `onSelected("Teacher")` — auto-selecting "Teacher". Technically, the user hasn't actually made the selection at this point. This can lead to unintended results — imagine if "Teacher Assistant" was another option the user wanted to select.

This can be resolved by deleting the lines that call the `setValue` function inside `setDisplayValue`.

If the intent is to have "auto-selection" as an option for the `Autosuggest` component, then perhaps it should be added as an attribute. Developers would have the ability to turn the option on when needed.

#### JIRA
[APER-2298](https://2u-internal.atlassian.net/browse/APER-2298)

#### Github Issue
[`Form.Autosuggest` intended behavior](https://github.com/openedx/paragon/issues/2107)

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Form/tests/FormAutosuggest.test.jsx
+++ b/src/Form/tests/FormAutosuggest.test.jsx
@@ -26,9 +26,11 @@ function FormAutosuggestWrapper(props) {
 }
 
 describe('FormAutosuggest', () => {
-  it('renders component without error', () => {
-    mount(<FormAutosuggestWrapper />);
+  afterEach(() => {
+    jest.clearAllMocks();
   });
+
+  const onSelected = jest.fn();
 
   const container = mount(
     <FormAutosuggestWrapper
@@ -36,6 +38,7 @@ describe('FormAutosuggest', () => {
       floatingLabel="floatingLabel text"
       helpMessage="Example help message"
       errorMessageText="Example error message"
+      onSelected={onSelected}
     >
       <FormAutosuggestOption>Option 1</FormAutosuggestOption>
       <FormAutosuggestOption>Option 2</FormAutosuggestOption>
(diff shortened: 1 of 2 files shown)


src/Form/tests/FormAutosuggest.test.jsx
MaxFrank13, line 1: Thanks for the feedback, @viktorrusakov! I noticed there were a couple tests for selecting the option so I added a few assertions there that check specifically for the `onSelected` function to be called. I also took the opportunity to reformat the file as it was getting pretty large. Let me know if you think there is something we're still not testing for — happy to address additional changes 👍

viktorrusakov: Awesome, @MaxFrank13, thank you! LGTM, I can't think of anything else 😄. We'll have to wait for @adamstankiewicz to take a look and merge though. 🙂

viktorrusakov (approved):

adamstankiewicz (approved):
LGTM as well. Thanks for taking a look @viktorrusakov 😄 Feel free to merge away @MaxFrank13!

PR #2272, opened 2023-05-06
feat: add `actions` prop to `Card.Status` and resolve few `Card`-related style issues

adamstankiewicz:
## Description

* Adds an `actions` prop to `Card.Status` to allow consumers to pass use CTA buttons in its children.
* Updates `Card` component page in the docs site to include example of new `actions` prop.
* Displays a `Skeleton` for `Card.Status` when the card is in a loading state.
* Forces the image cap skeleton height to be 100% when the card is in a horizontal orientation.
* Fixes rounded corners on the image cap such that the image cap itself doesn't specify the roundedness; it now relies on the parent card element's border radius with an added `overflow: hidden` to hide the image that goes beyond the rounded corners.

image

image

image

image

image

### Deploy Preview

https://deploy-preview-2272--paragon-openedx.netlify.app/components/card

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/scss/core/core.scss
+++ b/scss/core/core.scss
@@ -5,7 +5,7 @@
 @import "~bootstrap/scss/reboot";
 @import "typography";
 @import "grid";
-@import "~react-loading-skeleton/dist/skeleton";
+@import "~react-loading-skeleton/dist/skeleton.css";
 @import "~bootstrap/scss/transitions";
 @import "utilities";
 @import "~bootstrap/scss/media";
--- a/scss/core/core.scss
+++ b/scss/core/core.scss
@@ -5,7 +5,7 @@
 @import "~bootstrap/scss/reboot";
 @import "typography";
 @import "grid";
-@import "~react-loading-skeleton/dist/skeleton.css";
+@import "~react-loading-skeleton/dist/skeleton";
 @import "~bootstrap/scss/transitions";
 @import "utilities";
 @import "~bootstrap/scss/media";
--- a/src/Card/CardImageCap.jsx
+++ b/src/Card/CardImageCap.jsx
@@ -25,24 +25,27 @@ const CardImageCap = React.forwardRef(({
   const { orientation, isLoading } = useContext(CardContext);
   const [showImageCap, setShowImageCap] = useState(false);
   const [showLogoCap, setShowLogoCap] = useState(false);
+
   const wrapperClassName = `pgn__card-wrapper-image-cap ${orientation}`;
 
   if (isLoading) {
     return (
-      <div className={classNames(className, wrapperClassName)} data-testid="image-loader-wrapper">
+      <div
+        className={classNames(wrapperClassName, className)}
+        data-testid="image-loader-wrapper"
+      >
         <Skeleton
           containerClassName="pgn__card-image-cap-loader"
-          height={skeletonHeight}
+          height={orientation === 'horizontal' ? '100%' : skeletonHeight}
           width={skeletonWidth}
         />
-        {logoSkeleton
-          && (
+        {logoSkeleton && (
           <Skeleton
             containerClassName="pgn__card-logo-cap"
             height={logoSkeletonHeight}
             width={logoSkeletonWidth}
           />
-          )}
+        )}
       </div>
     );
   }
--- a/src/Card/tests/CardStatus.test.jsx
+++ b/src/Card/tests/CardStatus.test.jsx
@@ -4,7 +4,7 @@ import renderer from 'react-test-renderer';
 import { render, screen } from '@testing-library/react';
 
 import Button from '../../Button';
-import Card from '../index';
+import Card from '..';
 import CardStatus from '../CardStatus';
 import WarningFilled from '../../../icons/jsx/WarningFilled';
 
--- a/src/Card/tests/CardStatus.test.jsx
+++ b/src/Card/tests/CardStatus.test.jsx
@@ -1,7 +1,10 @@
 import React from 'react';
 import { mount } from 'enzyme';
 import renderer from 'react-test-renderer';
+import { render, screen } from '@testing-library/react';
 
+import Button from '../../Button';
+import Card from '../index';
 import CardStatus from '../CardStatus';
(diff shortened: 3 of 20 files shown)


scss/core/core.scss
```diff
@@ -5,7 +5,7 @@
 @import "~bootstrap/scss/reboot";
 @import "typography";
 @import "grid";
-@import "~react-loading-skeleton/dist/skeleton";
+@import "~react-loading-skeleton/dist/skeleton.css";
```
adamstankiewicz, line 8: For some reason, the dev server won't start for me without this `.css` file extensions after upgrading `react-loading-skeleton`.

src/Card/CardImageCap.jsx
```diff
@@ -25,24 +25,27 @@ const CardImageCap = React.forwardRef(({
...
   const [showImageCap, setShowImageCap] = useState(false);
   const [showLogoCap, setShowLogoCap] = useState(false);
+
   const wrapperClassName = `pgn__card-wrapper-image-cap ${orientation}`;
 
   if (isLoading) {
     return (
-      <div className={classNames(className, wrapperClassName)} data-testid="image-loader-wrapper">
+      <div
+        className={classNames(wrapperClassName, className)}
```
adamstankiewicz, line 34: Reverses the `wrapperClassName` and `className` such that consumers' `className` (if any) takes precedence over the component class names. Otherwise, they may be effectively ignored.

iloveagent57 (approved):

src/Card/CardImageCap.jsx
brian-smith-tcril: Is this a pattern across any other components? If so, would it make sense to make an issue to standardize these?

adamstankiewicz: @brian-smith-tcril A large portion (majority) of components allows consumers to pass in a `className` prop to add custom class names or use the CSS utility classes (which rely on `!important`) to the actual class names on the component.

I'll plan to file a new GitHub issue for:
* Writing a small ADR to describe/document this standard to make it "official".
* Find other instances of this issue throughout the design system and implement a fix.

adamstankiewicz: @brian-smith-tcril I may have been wrong about this being an issue after all. I just did some test cases more explicitly about this. It appears it's not so much the order of the passed `className` needing to coming after any `.pgn__` namespaced classes.

Rather, it seems more important where the custom `className` is defined in relation to the Paragon class names. If the custom `className` is defined prior to the Paragon class names, it appears that's when consumers may run into this issue.

For example, defining a `.custom-styles-class` above all the other styles in `Card.scss` and passing `className="custom-styles-class"` to the `ActionRow` component makes no difference whether the custom `className` is ordered first or after the `.pgn__` namespaced classes in the `classNames` function.

adamstankiewicz:
> I'm not sure if needing to add this to each preview example is a good long term solution, but I do think it'd be good to add something similar to this component preview so it looks nice on mobile!

@brian-smith-tcril Ah, good catch! I've updated the component preview to account for mobile screen widths:

image

The reason we do this on many of the component previews is that `Card`, by default, takes up 100% width of its container. Without constraining its width on larger screens, the `Card` component previews can get excessively wide.

FWIW, I opted to use `CardGrid` and conditionally update the `columnSizes` prop based on the whether the component is being rendered in a vertical orientation (6 columns) or horizontal orientation (12 columns, full width), which feels like it could be a bit more approachable/understandable than relying on the `useMediaQuery` hook in the code blocks. Thoughts?

brian-smith-tcril:
> FWIW, I opted to use CardGrid and conditionally update the columnSizes prop based on the whether the component is being rendered in a vertical orientation (6 columns) or horizontal orientation (12 columns, full width), which feels like it could be a bit more approachable/understandable than relying on the useMediaQuery hook in the code blocks. Thoughts?

I completely agree! It even has the added benefit of being able to quickly copy the card within the grid in the live editor and see changes side by side!

src/Card/CardImageCap.jsx
brian-smith-tcril: Glad to hear it's not an issue! Thanks for digging into that!

brian-smith-tcril (approved):

adamstankiewicz:
[inform] Thanks for the approvals! I will hold off on merging until the contribution is reviewed at the Paragon Working Group meeting tomorrow to get broader design input beyond the enterprise designers at 2U/edX.

viktorrusakov (approved):
Left one tiny nit, otherwise LGTM! The `Card` keeps growing. 🙂 

I do think the example code is a little hard to follow if you want to copy just one variant (e.g., `primary` variant with `horizontal` orientation and no actions), but I don't really see right now how this can be simplified. (Also, maybe it's just me 🤷 )

adamstankiewicz:
> I do think the example code is a little hard to follow if you want to copy just one variant (e.g., primary variant with horizontal orientation and no actions), but I don't really see right now how this can be simplified. (Also, maybe it's just me 🤷 )

@viktorrusakov Yeah, I think this is a broader issue than just `Card.Status`, though; when we try to put a lot of different examples in one component preview block, the tradeoff is it makes it less copy/pasteable for consumers for the added benefit of not having so many separate sections/code blocks to demonstrate all the permutations of the component.

I also don't quite have any recommendations yet on how to mitigate this, but I agree it is a bit of a concern 😃

src/Card/tests/CardStatus.test.jsx
```diff
@@ -1,7 +1,10 @@
 import React from 'react';
 import { mount } from 'enzyme';
 import renderer from 'react-test-renderer';
+import { render, screen } from '@testing-library/react';
 
+import Button from '../../Button';
+import Card from '../index';
```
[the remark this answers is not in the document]
adamstankiewicz: @viktorrusakov Looks like there's a few more instances than just here with the `../index` (this was copy/pasted from another test file). I've updated all these other instances, too, as you're correct that normally we just use `..` :) 

image

PR #2277, opened 2023-05-08
feat: provide functionality for initial value in autosuggest component

syedsajjadkazmii:
## Description

This PR adds functionality in the `Form.Autosuggest` component to have an already selected value initially. Currently, it always initializes the field to an empty string.

**WHY DO WE NEED THIS?**

We are trying to use `Form.Autosuggest` for the country field in `frontend-app-authn`. We auto-populate the country field by getting the user country from the IP. Therefore we need the functionality to have already selected value/initial value in the autoSuggest component.

Please let me know if this approach has any side-effects or if the approach should be changed.

### Deploy Preview

## Merge Checklist

## Post-merge Checklist


zainab-amir (approved):

src/Form/FormAutosuggest.jsx
```diff
@@ -33,7 +33,7 @@ function FormAutosuggest({
   });
   const [isMenuClosed, setIsMenuClosed] = useState(true);
   const [state, setState] = useState({
-    displayValue: '',
+    displayValue: value || '',
```
adamstankiewicz, line 36: [curious] Is there a test that should be added to account for this change in logic so we don't introduce future regressions? There is a `FormAutosuggest.test.jsx` test suite.

syedsajjadkazmii: @adamstankiewicz yes, you are right. i have added the test

viktorrusakov: @syedsajjadkazmii Thanks for adding the test, LGTM!

viktorrusakov (approved):

PR #2300, opened 2023-05-15
feat: improvement to initial value in autosuggest component

syedsajjadkazmii:
## Description

Improvement to this [PR](https://github.com/openedx/paragon/pull/2277)
 
### Deploy Preview

## Merge Checklist

## Post-merge Checklist


mubbsharanwar (approved):

src/Form/FormAutosuggest.jsx
```diff
@@ -157,6 +157,15 @@ function FormAutosuggest({
     };
   });
 
+  useEffect(() => {
+    if (value) {
```
viktorrusakov, line 161: Should we also check that `value !== state.displayValue` to avoid unnecessary re-renders? I think without it if I open a dropdown and manually select an item it will first update `displayValue` internally and then call `onSelected` function that will update `value` state of a parent component that will trigger `setState` in this `useEffect` which is not needed because `displayValue` is already up-to-date.

Also, should we consider a case when `value` becomes an empty string for some reason? I think we might want to update `displayValue` in such case too, just in case.

syedsajjadkazmii: @viktorrusakov  thanks for the suggestion. 

Also regarding your second point, if `value` changes to `''` from lets say some value `test value`, then i guess this useEffect will run and will update the value to `''`. So this will also work.

viktorrusakov: Yeah, that's true. Thanks for the update, LGTM!

src/Form/FormAutosuggest.jsx
```diff
@@ -157,6 +157,15 @@ function FormAutosuggest({
...
   });
 
+  useEffect(() => {
+    if (value !== state.displayValue) {
+      setState(prevState => ({
+        ...prevState,
+        displayValue: value,
+      }));
+    }
+  }, [value]);
```
adamstankiewicz, line 167: I'm not sure why the CI linter is passing here, but locally ESLint complains that `state.displayValue` is missing from the `useEffect` dependency array. Without it, if/when `state.displayValue` changes, this effect won't trigger.

image

PR #2330, opened 2023-05-25
[SEO-478] feat: allows users to set loading type to images on CardImageCap comp…

ochavarria2u:
## Description

[SEO-478](https://2u-internal.atlassian.net/browse/SEO-478)

In the interest of LCP we want to lazy load images that are not in the view port. Browser level lazy loading with suffice for a first pass.

### Deploy Preview

N/A

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Card/CardImageCap.jsx
+++ b/src/Card/CardImageCap.jsx
@@ -135,7 +135,7 @@ CardImageCap.defaultProps = {
   logoSkeletonHeight: LOGO_SKELETON_HEIGHT_VALUE,
   skeletonWidth: undefined,
   logoSkeletonWidth: undefined,
-  imageLoadingType: undefined,
+  imageLoadingType: 'eager',
 };
 
 export default CardImageCap;
--- a/src/Card/CardImageCap.jsx
+++ b/src/Card/CardImageCap.jsx
@@ -21,6 +21,7 @@ const CardImageCap = React.forwardRef(({
   logoSkeletonHeight,
   logoSkeletonWidth,
   className,
+  imageLoadingType,
 }, ref) => {
   const { orientation, isLoading } = useContext(CardContext);
   const [showImageCap, setShowImageCap] = useState(false);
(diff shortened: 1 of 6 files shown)


src/Card/CardImageCap.jsx
```diff
@@ -130,6 +135,7 @@ CardImageCap.defaultProps = {
   logoSkeletonHeight: LOGO_SKELETON_HEIGHT_VALUE,
   skeletonWidth: undefined,
   logoSkeletonWidth: undefined,
+  imageLoadingType: undefined,
```
viktorrusakov, line 138: Should the default value be `eager` here? This is what the browser will default to anyway, but still might be helpful to explicitly specify it here.

viktorrusakov (commented):
LGTM! Left one tiny comment regarding props definition.

src/Card/CardImageCap.jsx
ochavarria2u: Tried to add it but tests start failling due it will render with <CardImageCap> ..... loading={'eager'} </> I made the fix for the test snapshots  but that means users will also have the same test issue with the snapshots. You can check it [here](https://github.com/openedx/paragon/pull/2328/commits/e4687a49dd7829804e9cf19a3d26f69a9fdfdaa4). Let me know your thoughts.

adamstankiewicz: @ochavarria2u I agree about defaulting to `eager` in the `CardImageCap.defaultProps`. If the internals of a Paragon component changes, that shouldn't impact consumers' tests. If it does, I feel they should be testing in a different way that doesn't rely on Paragon's internals. All that to say, adding `loading="eager"` by default on `CardImageCap` here is an implementation detail of the component so should be fine to change imo.

ochavarria2u: Got it, I just added a commit with the changes.

adamstankiewicz (approved):

PR #2378, opened 2023-06-16
[BD-46] feat: create chip carousel

monteri:
## Description

Create chip carousel component

### Deploy Preview

https://deploy-preview-2378--paragon-openedx.netlify.app

## Merge Checklist

## Post-merge Checklist


openedx-webhooks:


src/ChipCarousel/ChipCarousel.scss
```diff
@@ -0,0 +1,24 @@
...
+
+.pgn__chip-carousel {
+  position: relative;
+
+  .pgn__overflow-scroll-overflow-container {
+    --pgn-overflow-scroll-opacity-mask-transparent: rgba(0, 0, 0, 0);
+  }
+}
+
+.pgn__chip-carousel__right-control,
```
PKulkoRaccoonGang, line 11: These selectors are children `.pgn__chip-carousel`?

src/ChipCarousel/ChipCarousel.test.jsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import { IntlProvider } from 'react-intl';
+import ChipCarousel from './index';
```
PKulkoRaccoonGang, line 4:
```suggestion
import ChipCarousel from '.';
```

src/ChipCarousel/README.md
```diff
@@ -0,0 +1,92 @@
...
+- ChipCarousel
+categories:
+- Content
+status: 'New'
+designStatus: 'Done'
+devStatus: 'Done'
+notes: |
+---
+
+``ChipCarousel`` component creates a scrollable horizontal block of chips with buttons for navigating to the previous and next set of chips.
```
PKulkoRaccoonGang, line 14:
```suggestion
The ``ChipCarousel`` component creates a scrollable horizontal block of chips with buttons for navigating to the previous and next set of chips.
```

src/ChipCarousel/README.md
```diff
@@ -0,0 +1,92 @@
...
+          onClick={() => console.log(`Chip #${index + 1} clicked`)}
+        >
+          Chip #{index + 1}
+        </Chip>
+      )
+    )}
+  />
+```
+
+## `Chip` Carousel
```
PKulkoRaccoonGang, line 34: Interactive Сhip Сarousel maybe

src/ChipCarousel/README.md
```diff
@@ -0,0 +1,92 @@
...
+            onClick={scrollToNext}
+            size="sm"
+            disabled={isScrolledToEnd}
+          >
+            Next
+          </Button>
+        </div>
+        <div ref={setOverflowRef} className="d-flex">
+          <OverflowScroll.Items>
+            <Chip iconAfter={Close}>New</Chip>
```
PKulkoRaccoonGang, line 66: I propose to simulate an array in this way. This will help us keep the example compact.

```
{[...Array(12).keys()].map(item => (
  <Chip key={item + 1} iconAfter={Close}>New</Chip>
))}
```

src/ChipCarousel/index.tsx
```diff
@@ -0,0 +1,140 @@
+import React, { ForwardedRef } from 'react';
+import PropTypes from 'prop-types';
+import { useIntl } from 'react-intl';
+import classNames from 'classnames';
+// @ts-ignore
+import { OverflowScroll, OverflowScrollContext } from '../OverflowScroll';
```
PKulkoRaccoonGang, line 6: Let's try to ignore the necessary import rule for the whole file, this will help to avoid numerous `@ts-ignore`

src/ChipCarousel/messages.js
```diff
@@ -0,0 +1,16 @@
+import { defineMessages } from 'react-intl';
+
+const messages = defineMessages({
+  scrollToPrevious: {
+    id: 'pgn.ChipCarousel.scrollToPrevious',
```
PKulkoRaccoonGang, line 5: We do not use camelСase in the declaration of transfer identifiers, let's use exclusively lowercase identifiers

src/ChipCarousel/messages.js
```diff
@@ -0,0 +1,16 @@
...
+    defaultMessage: 'Scroll to previous',
+    description: 'Accessibility text describing the action for navigating a scrollable carousel to the previous element',
+  },
+  scrollToNext: {
```
PKulkoRaccoonGang, line 9: Also in other translations, I noticed that we use a slightly different format for recording the key for translations. I suggest sticking to uniformity.

`  'subscription.details.program.type.microMasters': {
    id: 'subscription.details.program.type.microMasters',
    defaultMessage: 'MicroMasters',
    description: 'Label to render the MicroMasters Certificate program type.',
  },`

src/index.scss
```diff
@@ -11,6 +11,7 @@
 @import "./Collapsible/Collapsible";
 @import "./CloseButton/CloseButton";
 @import "./Chip/Chip";
+@import "./ChipCarousel/ChipCarousel";
```
PKulkoRaccoonGang, line 14: Curious moment. It seems to me that if the style files are called index, we can remove the duplication of the name from the path here

For example: `@import "./ChipCarousel"`

src/ChipCarousel/_variables.scss
```diff
@@ -0,0 +1 @@
+$chip-carousel-controls-top-offset:   -3px !default;
```
PKulkoRaccoonGang, line 1: It will be necessary to add a new token :)

src/ChipCarousel/ChipCarousel.scss
monteri: Yes, but if we don't need nesting, it is appropriate to avoid it to have more readable styles.

src/ChipCarousel/README.md
monteri: I just forgot to delete this example

src/ChipCarousel/index.tsx
monteri: I would do it as a separate PR

src/ChipCarousel/messages.js
monteri: I just stick to the other `id` in the Paragon, do you think we need to change to lowercase all of them?

src/index.scss
monteri: Yeah, I think it can be considered as separate issue

src/ChipCarousel/messages.js
PKulkoRaccoonGang: `id: 'pgn.chip-carousel.scrollToPrevious',` maybe

src/ChipCarousel/README.md
```diff
@@ -0,0 +1,73 @@
...
+      {/* start example form block */}
+      <ExamplePropsForm
+        inputs={[
+          {
+            value: offset,
+            setValue: setOffset,
+            range: {
+              min: 0,
+              max: offsetType === 'percentage' ? 100 : 1000,
+              step: offsetType === 'percentage' ? 1 : 50,
```
adamstankiewicz, line 35: On page load, `offsetType` is defaulted to "fixed" at a value of 120. However, if the user switches to "percentage", 120 > the `max` of 100, so the percentage offset changes to 100%, which doesn't quite demonstrate the percentage increment option effectively.

I wonder if we should reset the `offset` value when switching between `offsetType`s so the offset values make sense for each type upon switching?

src/ChipCarousel/index.tsx
adamstankiewicz: [clarification] Once the Paragon types PR merges in, I'm assuming these `@ts-ignore` comment would become unnecessary?

src/ChipCarousel/README.md
```diff
@@ -0,0 +1,73 @@
...
+            value: offsetType,
+            setValue: setOffsetType,
+            options: ['percentage', 'fixed'],
+            name: 'offsetType'
+          },
+          {
+            value: gap,
+            setValue: setGap,
+            range: { min: 0, max: 6, step: 0.5 },
+            name: 'offset'
```
adamstankiewicz, line 49: Should this be named `gap` instead?

adamstankiewicz, line 49: [curious] It looks like there's no difference in the gap between `0` and `0.5`; is this expected?

image

image

src/ChipCarousel/messages.js
adamstankiewicz: > We do not use camelСase in the declaration of transfer identifiers, let's use exclusively lowercase identifiers

@PKulkoRaccoonGang [clarification] Can you expand on this comment a bit more?

src/OverflowScroll/OverflowScroll.jsx
```diff
@@ -84,7 +84,9 @@ OverflowScroll.propTypes = {
   onScrollPrevious: PropTypes.func,
   /** Callback function for when the user scrolls to the next element. */
   onScrollNext: PropTypes.func,
-  offset: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  /** A value specifying the distance the scroll should move. */
```
adamstankiewicz, line 87: 🎉

src/ChipCarousel/ChipCarousel.test.jsx
```diff
@@ -0,0 +1,68 @@
...
+      <ChipCarousel ariaLabel={ariaLabel} items={items} {...props} />
+    </IntlProvider>
+  );
+}
+
+describe('<ChipCarousel />', () => {
+  it('should render the carousel correctly', () => {
+    render(<TestingChipCarousel />);
+
+    const carousel = document.getElementsByClassName('pgn__chip-carousel');
```
adamstankiewicz, line 49: Is there a way around needing to rely on the implementation details of the component in these tests by relying more on the `@testing-library/react` (RTL) API?

This is kind of using RTL almost as if it were Enzyme by relying on implementation details like class names. Perhaps we rely more on `data-testid` attributes here for selections?

```jsx
function TestingChipCarousel(props) {
  return (
    <IntlProvider locale="en">
      <ChipCarousel data-testid="chip-carousel" ariaLabel={ariaLabel} items={items} {...props} />
    </IntlProvider>
  );
}

render(<TestingChipCarousel />);

const carousel = screen.getByTestId('chip-carousel');
```

Similarly for the `chipItems` below, we could do something like add a `data-testid="chip-component"` on each child object passed to `ChipCarousel` such that we could query using RTL like: `screen.queryByTestId('chip-component')`

Reference: https://kentcdodds.com/blog/testing-implementation-details

src/ChipCarousel/ChipCarousel.test.jsx
```diff
@@ -0,0 +1,68 @@
...
+      expect(chipItems[i].textContent).toBe(items[i].props.children);
+    }
+  });
+
+  it('should call onClick when a chip item is clicked', () => {
+    render(<TestingChipCarousel />);
+
+    const chipItems = document.getElementsByClassName('pgn__chip');
+    for (let i = 0; i < chipItems.length; i++) {
+      fireEvent.click(chipItems[i]);
```
adamstankiewicz, line 64: Typically, `userEvent.click` is preferred over using `fireEvent` directly. See https://stackoverflow.com/a/71031849.

src/ChipCarousel/messages.js
adamstankiewicz: @PKulkoRaccoonGang @monteri After discussing, we decided to stick with the camelCase message ID format (it's used through frontend-app-learning and some other MFEs. Up to the owning team of repo to decide the standard).

src/ChipCarousel/ChipCarousel.test.jsx
monteri: Agree on this point. Refactored tests to use `data-testid`

src/ChipCarousel/ChipCarousel.test.jsx
monteri: updated to use `userEvent.click`

src/ChipCarousel/README.md
monteri: Added function to be more user-friendly when `offsetType` is switched. It calculates ration and sets relatively equal value between `fixed` and `percentage`.

src/ChipCarousel/README.md
monteri: updated to `gap`

monteri: Yes, probably I have talked about this case that there is no `.5` `spacer` value.
```
$spacers: map-merge(
  (
    0: 0,
    1: ($spacer * .25),
    1\.5: ($spacer * .375),
    2: ($spacer * .5),
    2\.5: ($spacer * .75),
    3: $spacer,
    3\.5: ($spacer * 1.25),
    4: ($spacer * 1.5),
    4\.5: ($spacer * 2),
    5: ($spacer * 3),
    5\.5: ($spacer * 4),
    6: ($spacer * 5),
  ),
  $spacers
);
```

adamstankiewicz: Ah, yes. Thank you for the reminder. Noted :)

src/ChipCarousel/ChipCarousel.scss
adamstankiewicz: Yeah, I'd say the nesting isn't a hard requirement given we have the convention of BEM class names (mostly) prefixed with `pgn__{componentname}`. Unlikely for collisions to happen, though still possible.

adamstankiewicz (approved):
LGTM! Just looks like it might need a rebase with `master`, though.

monteri:
> Yeah, I'd say the nesting isn't a hard requirement given we have the convention of BEM class names (mostly) prefixed with pgn__{componentname}. Unlikely for collisions to happen, though still possible.

Updated to use nesting in this case. Also noticed that when we changed `componentName.scss` to `index.scss` we forgot to update `component-generator`. Made quick fix as part of this PR

adamstankiewicz:
> Updated to use nesting in this case. Also noticed that when we changed `componentName.scss` to `index.scss` we forgot to update `component-generator`. Made a quick fix as part of this PR

Good catch!

adamstankiewicz (approved):

openedx-webhooks:
@monteri 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/2378&entry.1484655318=https://github.com/monteri&entry.752974735=2023-06-16+11:47&entry.1917517419=2023-08-04+12:16&entry.2133058324=Yes) so we can improve your experience in the future.

Issue #2373, opened 2023-06-12
[paragon-openedx.netlify.app] Feedback (on /components/alert/)


gabew1984:
## Improvements

Please add a "Leave feedback" button to the top of the component page template, and style the button as an "**small, outline primary button**".
![Screenshot 2023-06-12 at 11 20 10 AM](https://github.com/openedx/paragon/assets/14075470/8ae7faed-9f92-418f-a12f-89d86c4849af)
![Screenshot 2023-06-12 at 11 18 14 AM](https://github.com/openedx/paragon/assets/14075470/55af11a2-ef7e-4b67-89c3-7054ab7332fe)

Migrate existing "Leave feedback" hyperlink to a similar button as at the top of component page.

adamstankiewicz:
Related issue: https://github.com/openedx/paragon/issues/2366

PKulkoRaccoonGang:
> Related issue: #2366

@adamstankiewicz This is for the alpha branch.

adamstankiewicz:
Closed via https://github.com/openedx/paragon/pull/2395

PR #2395, opened 2023-06-21
[BD-46] docs: added 'Leave feedback' button to the top of the component page

PKulkoRaccoonGang:
## Description

Issue: https://github.com/openedx/paragon/issues/2373

### Deploy Preview

## Merge Checklist

## Post-merge Checklist


openedx-webhooks:


www/src/components/LeaveFeedback.tsx
```diff
@@ -3,7 +3,11 @@ import PropTypes from 'prop-types';
 import { Nav } from '~paragon-react';
 import { useLocation } from '@reach/router';
 
-function LeaveFeedback(props) {
+export interface LeaveFeedbackProps {
+  as?: keyof JSX.IntrinsicElements;
```
monteri, line 7: Is it better to make like this `as?: keyof JSX.IntrinsicElements | React.ElementType`?

PKulkoRaccoonGang: Done

www/src/templates/component-page-template.tsx
```diff
@@ -136,7 +138,10 @@ export default function PageTemplate({
             <p className="small mb-0">{mdx.frontmatter.notes}</p>
           </Alert>
         )}
-        <h1 className="mb-4">{mdx.frontmatter.title}</h1>
+        <div className="d-flex justify-content-between">
+          <h1 className="mb-4">{mdx.frontmatter.title}</h1>
+          <LeaveFeedback as={Hyperlink} />
```
adamstankiewicz, line 143: If we're using `Hyperlink` here to get it to have the external launch icon, we may also want to include it on the other in-page "Leave feedback" link towards the bottom.

Likewise, we could also use `as={Hyperlink}` in the docs site `footer` such that we get the correct `rel="noopener noreferrer"` attribute for security. That said, given none of the other external links in the footer include an external launch icon, we can use `Hyperlink`'s `showLaunchIcon={false}` to hide the icon for the footer only.

adamstankiewicz, line 143: Per the [issue](https://github.com/openedx/paragon/issues/2373), I believe the intent for "Leave feedback" now is to be a small `Button` (`variant="outline-primary"`) that acts as a `Hyperlink` here in `component-page-template` and below the page content (though not the footer).

adamstankiewicz, line 143: It looks like this defaults to using `Nav.Link` styles by default, even though most uses of `LeaveFeedback` don't actually intend to use those styles. I wonder if the use of `LeaveFeedback` in the footer should be the exception that gets the `Nav.Link` styles rather than having to override that in 2 of the 3 places the `LeaveFeedback` component is used.

Is there a way to perhaps make this component a bit more flexible so the differing styles in the various use cases don't conflict?

PKulkoRaccoonGang: Thanks for the review! I replaced the top link and the bottom link with buttons, I also changed the implementation of links in the site footer. It seems to me that all links that lead to third-party resources should have rel="noopener noreferrer" and target="_blank ".

www/src/components/LeaveFeedback.tsx
```diff
@@ -1,29 +1,57 @@
...
+function LeaveFeedback({ isLink, ...props }: LeaveFeedbackProps) {
   const location = useLocation();
   const FEEDBACK_URL = `https://github.com/openedx/paragon/issues/new?title=%5Bparagon-openedx.netlify.app%5D%20Feedback%20(on%20${location.pathname})&amp;labels=docs&template=feedback_template.md`;
+  const leaveFeedbackLinkTitle = 'Leave feedback';
 
   const handleLinkFeedbackClick = () => {
     global.analytics.track('openedx.paragon.docs.leave_feedback.clicked');
   };
 
+  if (isLink) {
```
viktorrusakov, line 19: I think it also make more sense to leave `Nav.Link` implementation here since we only use it inside `Nav` component, e.g.
```javascript
if (isNavLink) {
  return (
    <Nav.Link
      onClick={handleLinkFeedbackClick}
      href={FEEDBACK_URL}
      target="_blank"
      rel="noopener noreferrer"
      {...props}
    >
      {leaveFeedbackLinkTitle}
    </Nav.Link>
  )
}
```

www/src/components/PageLayout.tsx
```diff
@@ -114,28 +115,40 @@ function Layout({
...
+              showLaunchIcon={false}
+              target="_blank"
             >
               Code of Conduct
-            </Nav.Link>
+            </Hyperlink>
           </Nav.Item>
           <Nav.Item>
-            <Nav.Link className="muted-link" href="https://open.edx.org/">
+            <Hyperlink
```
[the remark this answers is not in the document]
PKulkoRaccoonGang: @viktorrusakov thanks for the review 👍 

The change from `Nav.Link` to `HyperLink` was due to my assumptions about the semantic meaning of the `Nav.Link` component. After researching the [Bootstrap page](https://react-bootstrap-v4.netlify.app/components/navs/) for the `Nav.Link` component, it seemed to me that the correct solution is to use the `HyperLink` component for navigation links that refer to internal pages of the application. It is for this reason that the `Nav.Link` component does not have corresponding attributes.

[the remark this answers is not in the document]
viktorrusakov: It still just renders an `a` tag whether you use `Nav.Link` or `HyperLink`, but `Nav.Link` ensures that the link looks correctly inside a `nav` while for `HyperLink` you need to add additional class (which is internal to `Nav.Link`) to make it work.

I agree with your semantics reasoning (that nav probably shouldn't have external links in it), but your solution does not solve the underlying problem - in the DOM it still renders a `nav` element with bunch of external links inside it. If we really care about this, then we need to remove `nav` altogether and render `HyperLink`'s inside a `ul` I think.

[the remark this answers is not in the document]
PKulkoRaccoonGang: Ok, I agree with you 
Thanks 👍

www/src/components/LeaveFeedback.tsx
PKulkoRaccoonGang: Done

adamstankiewicz (approved):
LGTM!

openedx-webhooks:
@PKulkoRaccoonGang 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/2395&entry.1484655318=https://github.com/PKulkoRaccoonGang&entry.752974735=2023-06-21+20:18&entry.1917517419=2023-07-14+12:46&entry.2133058324=Yes) so we can improve your experience in the future.

PR #2431, opened 2023-07-11
feat: POC for isHidden prop on ModalDialog

adamstankiewicz:
## Description

Demos a proof-of-concept of updating the `ModalDialog` component to introduce an `isHidden` prop for the following types of use cases:

> By default, ``ModalDialog`` only renders its modal contents to the DOM when the modal is explicitly marked as open via the `isOpen` prop. However, in some cases, needing to re-render the contents of the modal body each and every time the modal closes and re-opens can be expensive and/or a poor user experience (UX), e.g. if the modal contents are dynamic and loaded from an external source like some iframed content or a slow loading API request.
>
> To mitigate these performance concerns, ``ModalDialog`` supports an optional ``isHidden`` prop that when used in conjunction with ``isOpen``, would result in the modal contents and the backdrop still being kept in the React Portal DOM, but visually hidden in the UI.

### Deploy Preview

TBD

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Modal/ModalLayer.jsx
+++ b/src/Modal/ModalLayer.jsx
@@ -39,7 +39,7 @@ ModalContentContainer.defaultProps = {
  * component is that if a modal object is visible then it is "enabled"
  */
 function ModalLayer({
-  children, onClose, isOpen, isBlocking, zIndex,
+  children, onClose, isOpen, isHidden, isBlocking, zIndex,
 }) {
   useEffect(() => {
     if (isOpen) {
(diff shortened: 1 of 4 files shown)


src/Modal/ModalLayer.jsx
```diff
@@ -52,29 +52,31 @@ function ModalLayer({
...
+          className={classNames({
+            'pgn__modal-layer': isOpen && !isHidden,
+            [`zindex-${zIndex}`]: !!zIndex,
+          })}
         >
           <ModalContentContainer>
-            <ModalBackdrop onClick={onClickOutside} />
-            {children}
+            {!isHidden && <ModalBackdrop onClick={onClickOutside} />}
+            <div className={classNames({ 'd-none': isHidden })}>
```
adamstankiewicz, line 77: Note: we should avoid using CSS utility classes with `!important` in Paragon components; define a custom `.pgn__` namespaced class name to use instead with `display: none;` to be more themable.

src/Modal/ModalLayer.jsx
```diff
@@ -52,29 +52,31 @@ function ModalLayer({
...
-          enabled={isOpen}
+          enabled={isOpen && !isHidden}
           onEscapeKey={onClose}
           onClickOutside={onClickOutside}
-          className={classNames(
-            'pgn__modal-layer',
-            zIndex ? `zindex-${zIndex}` : '',
-          )}
+          className={classNames({
+            'pgn__modal-layer': isOpen && !isHidden,
```
adamstankiewicz, line 71: [inform] Conditionally include the `pgn__modal-layer` class name only when the modal contents and the backdrop should actually be visible. Without making this class name conditional, the resulting `div` from `FocusOn` acts a non-clickable overlay.

src/Modal/ModalLayer.jsx
```diff
@@ -52,29 +52,31 @@ function ModalLayer({
     };
   }, [isOpen]);
 
-  if (!isOpen) {
+  if (!isOpen && !isHidden) {
```
adamstankiewicz, line 55: [consideration] It's possible this conditional could remain unchanged given `isOpen` would need to be truthy for `isHidden` to work anyways.

src/Modal/ModalLayer.jsx
```diff
@@ -52,29 +52,31 @@ function ModalLayer({
...
 
   return (
-    <ModalContextProvider onClose={onClose} isOpen={isOpen} isBlocking={isBlocking}>
+    <ModalContextProvider onClose={onClose} isOpen={isOpen} isHidden={isHidden} isBlocking={isBlocking}>
       <Portal>
         <FocusOn
           allowPinchZoom
           scrollLock
-          enabled={isOpen}
+          enabled={isOpen && !isHidden}
```
adamstankiewicz, line 67: `FocusOn` is used in `ModalDialog` to handle the accessibility of auto-focus on the first interactive element in the modal and handling the `Esc` keypress or backdrop click.

PR #2452, opened 2023-07-18
feat: add ref support in container

ahtesham-quraish:
Description:
Add react ref support in container component

related to : [VAN-1537](https://2u-internal.atlassian.net/browse/VAN-1537)


viktorrusakov (approved):
LGTM, thanks for the fix!

PR #2461, opened 2023-07-20
fix: selectable box set props forwarding

MaxFrank13:
## Description

This sets up the ability to forward props directly to the root element of the `SelectableBoxSet` component. This came up originally to enable to ability to provide an `aria-labelledby` attribute to the `SelectableBoxSet`. This will support adding other attributes should the need come up in the future.

### Deploy Preview

[Preview of changes](https://deploy-preview-2461--paragon-openedx.netlify.app/components/selectablebox/)

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/SelectableBox/SelectableBoxSet.jsx
+++ b/src/SelectableBox/SelectableBoxSet.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { getInputType } from './utils';
+import { requiredWhenNot } from '../utils/propTypes';
 
 const INPUT_TYPES = [
   'radio',
@@ -19,6 +20,9 @@ const SelectableBoxSet = React.forwardRef(({
   type,
   columns,
   className,
+  ariaLabel,
+  ariaLabelledby,
+  ...props
 }, ref) => {
   const inputType = getInputType('SelectableBoxSet', type);
 
@@ -35,6 +39,9 @@ const SelectableBoxSet = React.forwardRef(({
         `pgn__selectable_box-set--${columns || DEFAULT_COLUMNS_NUMBER}`,
         className,
       ),
+      'aria-label': ariaLabel,
+      'aria-labelledby': ariaLabelledby,
+      ...props,
     },
     children,
   );
@@ -62,6 +69,17 @@ SelectableBoxSet.propTypes = {
   columns: PropTypes.number,
   /** A class that is be appended to the base element. */
   className: PropTypes.string,
+  /**
+   * The ID of the label for the `SelectableBoxSet`.
+   *
+   * An accessible label must be provided to the `SelectableBoxSet`.
+   */
+  ariaLabelledby: PropTypes.string,
+  /**
+   * A label for the `SelectableBoxSet`.
+   *
+   * If not using `ariaLabelledby`, then `ariaLabel` must be provided */
+  ariaLabel: requiredWhenNot(PropTypes.string, 'ariaLabelledby'),
 };
 
 SelectableBoxSet.defaultProps = {
@@ -72,6 +90,8 @@ SelectableBoxSet.defaultProps = {
   type: 'radio',
   columns: DEFAULT_COLUMNS_NUMBER,
   className: undefined,
+  ariaLabelledby: undefined,
+  ariaLabel: undefined,
 };
 
 export default SelectableBoxSet;
--- a/src/SelectableBox/README.md
+++ b/src/SelectableBox/README.md
@@ -42,6 +42,7 @@ As ``Checkbox``
         onChange={handleChange}
         name="cheeses"
         columns={isExtraSmall ? 1 : 2}
+        ariaLabel="cheese selection"
       >
         <SelectableBox value="swiss" type={type} aria-label="swiss checkbox">
           <div>
@@ -83,6 +84,7 @@ As ``Checkbox``
       onChange={handleChange}
       name="colors"
       columns={isExtraSmall ? 1 : 3}
+      ariaLabel="color selection"
     >
       <SelectableBox value="red" type={type} aria-label="red checkbox">
         <div>
@@ -144,6 +146,7 @@ As ``Checkbox`` with ``isIndeterminate``
         onChange={handleChange}
         name="cheeses"
         columns={isExtraSmall ? 1 : 3}
+        ariaLabel="cheese selection"
       >
         <SelectableBox value="swiss" type={type} aria-label="swiss checkbox">
           <div>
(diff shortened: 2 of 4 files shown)


src/SelectableBox/SelectableBoxSet.jsx
```diff
@@ -19,6 +19,7 @@ const SelectableBoxSet = React.forwardRef(({
   type,
   columns,
   className,
+  ...props
```
adamstankiewicz, line 22: [non-blocking suggestion] I wonder if it's worth adding a test case to assert that prop spreading is supported (e.g., in case the line were to get removed erroneously somehow)?

src/SelectableBox/SelectableBoxSet.jsx
```diff
@@ -35,6 +36,7 @@ const SelectableBoxSet = React.forwardRef(({
         `pgn__selectable_box-set--${columns || DEFAULT_COLUMNS_NUMBER}`,
         className,
       ),
+      ...props,
```
adamstankiewicz, line 39: Actually, perhaps a quick clarifying question to verify with @wittjeff: Are the `aria-label` or `aria-labelledby` attributes _required_ from an a11y perspective, or optional but recommended?

If **required** in order to be generally a11y compliant as a standalone component, then I suppose we may want to consider enforcing the specific required a11y props (e.g., I wonder if there's a way to have both `aria-label` and `aria-labelledby` optional in the prop types but still conditionally required in some way). If were to enforce such an explicit prop as required, it would technically be a breaking for consumers, though there's not that many consumers of `SelectableBox` yet anyways, i.e. may need to release as a new major version if we were to make either of these props required.

All that said, I think we should go forward how you have it here for now; but would recommend getting that clarification from @wittjeff in terms of whether we should actually make it truly _required_ / enforced. We could also consider a middle ground where if either an `aria-label` or `aria-labelledby` is not provided, a `console.warn` is called, which consumers may see during local development as a hint to include these attributes.

Non-blocking feedback for this immediate PR, though!

adamstankiewicz, line 39: IIRC, I believe there was also some discussion around ensuring the Paragon documentation website page(s) pertaining to `SelectableBox.Set` included either the `aria-label` or `aria-labelledby` attribute(s) in the docs site examples themselves so its more evident to other consumers that either of these attributes is recommended / required(?).

src/SelectableBox/SelectableBoxSet.jsx
MaxFrank13: Good callout! Quick question on this, though. I notice the tests for this component are using `enzyme`. Is Paragon also planning to migrate to React 18 (I would think, yes)? And if so, will we also be following this [proposed OEP](https://github.com/openedx/open-edx-proposals/blob/de17de9aee40a2baa29cd1a87c1640fca2586489/oeps/best-practices/oep-0011/decisions/0001-enzyme-deprecation.rst)?

adamstankiewicz: We've decided we will not be using `react-unit-test-utils` in the Paragon repo for shallow/snapshot testing. As you noted, Raccoon Gang is actively migrating away from Enzyme tests in favor of RTL / `react-test-renderer` (where appropriate). Generally, my two cents is that explicit assertions are generally preferable than snapshots.

I would recommend using RTL to render the component with an `data-testid` prop and then assert that `expect(screen.getByTestId(...)).toBeInTheDocument()`.

src/SelectableBox/SelectableBoxSet.jsx
adamstankiewicz: > That being said, I don't know if there is a way to require one OR the other within our components? 

@MaxFrank13 would likely need to write a custom prop type definition, e.g.: https://stackoverflow.com/a/46337223

[This file](https://github.com/openedx/paragon/blob/5dff07e45f779021e2a4ade0738b8b9fe48e3953/src/utils/propTypes/utils.js) in Paragon is a collection of a handful of re-usable prop types. Consider looking into whether `customPropTypeRequirement` might be applicable here?

wittjeff: For div role="group" or div role="radiogroup" (the role should match whether we're using checkboxes or radio inputs), an accessible Name is a WCAG requirement (which in these cases would come from aria-label or aria-labelledby), so please make it a requirement here.

The only exception I can think of is if someone took our current control and wrapped it with a <fieldset> and <legend> (which are semantically equivalent to the above), but that seems exceedingly unlikely.

adamstankiewicz: > so please make it a requirement here.

Thanks @wittjeff!

@MaxFrank13 Given this, let's try to make explicit props for `aria-label` and `aria-labelledby`, where `aria-label` becomes required if `aria-labelledby` is not already set, possibly utilizing `customPropTypeRequirement` mentioned above. And, then keep the additional prop spreading in place since that adds support for attrs like `data-testid`.

adamstankiewicz: @MaxFrank13 If we do conditionally require one of these attributes, I would consider whether it needs to be released a breaking change (e.g., if a consuming MFE using `SelectableBox.Set` upgrades to this version, will any of their tests/CI begin failing?). My hunch is that it should be fine to release the required prop without a breaking change but worth verifying.

wittjeff: I'm not clear on your logic with the last comment @adamstankiewicz . This would seem to be a breaking change, yes? I mean I suspect this error is probably out there. When I get back later I'll file issues with the pages that use this control without labels already.

adamstankiewicz: @wittjeff I guess I just wanted to do a quick sanity check that the prop type warnings do indeed fail CI (linting?), as opposed to simply logging a prop type warning in the console devtools that may be effectively ignored / not seen by engineers. You're right that it should and, if so, should be released as a breaking change.

MaxFrank13: Thanks for the feedback, @wittjeff ! 

@adamstankiewicz that is a very useful file! Thanks for the tip. I didn't know we could do this with our prop types. 

From what I gather, this needs to be released as a breaking change. When it comes time to merge this, I will ask for a sanity check before I hit the big, scary green button. From the README, I gather that it is pretty straightforward but I would like to know if there is anything else that is maybe not included in there. I assume we need to alert the communication channels in some way?

src/SelectableBox/SelectableBoxSet.jsx
MaxFrank13: I was not able to get the tests working with `toBeInTheDocument()`. I was seeing this error:
`TypeError: expect(...).toBeInTheDocument is not a function`

However, `toBeTruthy()` was able to run the assertion just fine. I'm happy to change this but may need some guidance if we have to use `toBeInTheDcoument`.

src/SelectableBox/SelectableBoxSet.jsx
MaxFrank13: I've added details in the PropTypes definition. Do we think this information should be more visible? Such as getting it's own section on the doc site page?

src/SelectableBox/SelectableBoxSet.jsx
adamstankiewicz: Ah, @MaxFrank13: I believe that error is likely due to not including an import for the following in the test file:

```js
import '@testing-library/jest-dom/extend-expect';
```

adamstankiewicz (approved):

MaxFrank13:
Minor changes in latest push were:

- changed `toBeTruthy()` to `toBeInTheDocument()`
- added `ariaLabel` to test components and updated snapshot accordingly

PR #2472, opened 2023-07-24
fix: color picker error message fix

kiram15:
## Description

Quick style bug fix of error message of the ColorPicker component. 
Screenshot 2023-07-24 at 4 13 41 PM
Screenshot 2023-07-24 at 4 11 03 PM

### Deploy Preview

[Include a direct link to your changes in this PR's deploy preview here (e.g., a specific component page).
](http://localhost:8000/components/colorpicker/)

## Merge Checklist

## Post-merge Checklist


src/ColorPicker/index.jsx
```diff
@@ -66,18 +66,20 @@ function ColorPicker({
...
-          <Form.Group className="pgn__hex-form" size="sm">
-            <h5>Hex</h5>
-            <Form.Control
-              className="form-field"
-              isInvalid={!hexValid}
-              value={color}
-              onChange={(e) => validateHex(e.target.value)}
-              data-testid="hex-input"
-            />
+          <Form.Group className="pgn__hex-form flex-wrap" size="sm">
```
adamstankiewicz, line 69: Paragon components cannot use CSS utility classes on them because they include `!important`. It makes it challenges to theme the component for brand authors. Would recommend putting `flex-wrap` equivalent in the CSS class name definition for `.pgn__hex-form`. 

Similar comments for all other CSS utility classes below, e.g. `d-inline-flex`, `flex-fill`, and `ml-2`.

src/ColorPicker/index.jsx
```diff
@@ -66,18 +66,20 @@ function ColorPicker({
...
+            <div className="d-inline-flex">
+              <h5>Hex</h5>
```
adamstankiewicz, line 71: The `Form.Control` below does not have a label for a11y. From a screen reader's POV, there is no connection between this "Hex" heading and the `Form.Control` input field.

If you can't make use of `Form.Label` here, might recommend implementing `aria-labelledby` to associate the `h5` with the `Form.Control`.

See this example from the W3: https://www.w3.org/WAI/tutorials/forms/labels/#using-aria-labelledby

src/ColorPicker/index.jsx
```diff
@@ -49,7 +49,7 @@ function ColorPicker({
               `pgn__color-picker-${size}`,
             )}
             style={{
-              ...(color && hexValid ? { background: `${color}` } : {}),
+              ...(color && hexValid ? { background: `${color} !important` } : {}),
```
adamstankiewicz, line 52: [curious] Why is `!important` needed here? My understanding is that inline styles through the `style` attribute should already have preference. Needing to use `!important` is generally a code smell (though not _always_).

brian-smith-tcril (approved):
:shipit:

adamstankiewicz (approved):

Issue #2437, opened 2023-07-12
Form.Autosuggest Accessibility - Tab Behavior


brian-smith-tcril:
When tabbing out of the `<input>` field, the dropdown of options remains displayed. The list should be hidden when focus is removed from the `<input>`.

https://github.com/openedx/paragon/assets/92897870/e25a4ff9-48df-4304-8995-c199f15fde97

Issue #2435, opened 2023-07-12
Form.Autosuggest Accessibility - Trivial fixes


brian-smith-tcril:
- `<input>` element should have:
  - `role=”combobox”`
  - `aria-autocomplete=”list”`
  - `autocomplete=”off”`
    - This disables the native autocomplete https://github.com/openedx/paragon/issues/2434
- The `<div>` that houses the dropdown items should be a `<ul>` with:
  - `role=”listbox”`
- The `<button>` elements that render in the list should be `<li>` with:
  - `role=”option”`

PR #2513, opened 2023-08-09
fix: disable browser autocomplete and edit dropdown items elements

### LINKED ISSUE #2557: Form.Autosuggestion Accessibility - Keyboard Navigation Behavior in Dropdown
While working on #2435, the changes made broke keyboard navigation which prompted us to ask if the preferred behavior would be to be able to tab through dropdown item options and/or use arrow keys to flip through options. 

Per Jeff W's advice on [Slack](https://openedx.slack.com/archives/C02NR285KV4/p1692212283039559):

> Re Autocomplete / autosuggest implementation:  I believe the
> AlphaGov (UK govt design system) reference — see
> https://alphagov.github.io/accessible-autocomplete/examples/ --
>  is consistent with the latest draft of ARIA 1.3 spec at
> https://w3c.github.io/aria/#combobox
> and it works will all of the screen readers we support.  I suggest we use that as our reference.
> It uses up and down arrow key to change selection, so I recommend going with that.
> 
> The AlphaGov example puts tabindex=-1 on each of the list items in suggestions list. The ARIA spec doesn’t seem to include that but I don’t think it matters??  Adding tabindex=0 to each option (as with the edx.org/search box) puts all items in the focus order, which is probably not a problem but I suggest we go with the tabindex=-1 method for Paragon.

**TL;DR:** Properly handle keyboard navigation behavior in dropdown by taking these steps
- use up/down arrow keys for navigation
- don't allow tab for navigation to happen throughout dropdown
- ensure all items in the dropdown have a tabindex=-1

httpsmenahassan:
Add attributes to `input` element to disable browser autocomplete on autosuggest component and change the div that houses the dropdown items to be a `ul`, change the button elements that render the list to be `li`. add new `role` to each element

## Description

Resolves [#2435](https://github.com/openedx/paragon/issues/2435) and resolves #2557.

We had to make more changes than anticipated here given that we previously relied on the `button` value attribute. After changing button elements to `li`, we realized that `li`'s value attribute is an integer value, which posed some problems.

Note that this doesn't fix the issue where if you try to tab out while in the dropdown, the menu is still displayed. (#2437)

### Deploy Preview

https://deploy-preview-2513--paragon-openedx.netlify.app/components/form/form-autosuggest/

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Form/FormAutosuggest.jsx
+++ b/src/Form/FormAutosuggest.jsx
@@ -39,7 +39,7 @@ function FormAutosuggest({
   });
 
   const handleItemClick = (e, onClick) => {
-    const clickedValue = e.currentTarget.value;
+    const clickedValue = e.currentTarget.getAttribute('data-value');
 
     if (onSelected && clickedValue !== value) {
       onSelected(clickedValue);
@@ -66,7 +66,7 @@ function FormAutosuggest({
       return React.cloneElement(child, {
         ...rest,
         children,
-        value: children,
+        'data-value': children,
         onClick: (e) => handleItemClick(e, onClick),
       });
     });
@@ -219,6 +219,9 @@ function FormAutosuggest({
         <FormControl
           aria-expanded={(state.dropDownItems.length > 0).toString()}
           aria-owns="pgn__form-autosuggest__dropdown-box"
+          role="combobox"
+          aria-autocomplete="list"
+          autoComplete="off"
           value={state.displayValue}
           aria-invalid={state.errorMessage}
           onChange={handleOnChange}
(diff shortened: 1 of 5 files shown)


src/Menu/index.scss
```diff
@@ -92,3 +92,4 @@
     position: relative;
   }
 }
+
```
adamstankiewicz, line 95: nit: is this extra blank line necessary?

adamstankiewicz (approved):
LGTM!

src/Form/FormAutosuggest.jsx
```diff
@@ -283,8 +286,6 @@ FormAutosuggest.propTypes = {
   ignoredArrowKeysNames: PropTypes.arrayOf(PropTypes.string),
   /** Specifies loading state. */
   isLoading: PropTypes.bool,
-  /** An ARIA role describing the form autosuggest. */
-  role: PropTypes.string,
```
httpsmenahassan, line 287: We discussed this during the Paragon Working Group. We made the decision to remove consumers' ability to change the role to something that would break WCAG.

src/Menu/index.scss
httpsmenahassan: Thanks for catching that! It's now gone.

src/Form/FormAutosuggest.jsx
```diff
@@ -240,25 +243,25 @@ function FormAutosuggest({
...
         ) : state.dropDownItems.length > 0 && state.dropDownItems}
-      </div>
+      </ul>
     </div>
   );
 }
 
 FormAutosuggest.defaultProps = {
-  arrowKeyNavigationSelector: 'a:not(:disabled),button:not(:disabled, .btn-icon),input:not(:disabled)',
+  arrowKeyNavigationSelector: 'a:not(:disabled),li:not(:disabled, .btn-icon),input:not(:disabled)',
```
httpsmenahassan, line 262: We realized that after disabling the browser autocomplete that changing `button` to `li` broke the keyboard navigation through the dropdown items but this was easily solved by changing this line to include `li`.

brian-smith-tcril:
@httpsmenahassan I know you've made changes (that, from what I can tell, look great!) since this PR was approved. This is good to go now right?

httpsmenahassan:
@brian-smith-tcril yes! this PR is all set on my end.

brian-smith-tcril:
@adamstankiewicz it sounds like this one should be 100% good to go!

adamstankiewicz (approved):

Issue #2510, opened 2023-08-08
Autosuggest: use something other than `state.dropDownItems.length > 0` to filter document clicks


brian-smith-tcril:
In https://github.com/openedx/paragon/issues/1732 I noticed
> there is also no error message if text is entered that doesn't match anything in the dropdown (for example, typing "foo" in any of the sample boxes)

it seems this is happening because we have `state.dropDownItems.length > 0` in
https://github.com/openedx/paragon/blob/13b7488f495dfe260811fe3a8d0565929713b10b/src/Form/FormAutosuggest.jsx#L115
where we are handling all clicks on the document

if we remove `state.dropDownItems.length > 0` and instead just have
```jsx
if (parentRef.current && !parentRef.current.contains(e.target)) {
```
interacting with any autosuggest component on the docs page causes error states in the others

I believe the intention here is to not set the error state on components other than the one the user is interacting with. This implies that `state.dropDownItems.length > 0` is being used to mean "this is the autosuggest component the user is interacting with right now."

In the case where we have entered text that filters the dropdownitems down to a length of zero, the component behaves as if it is **not** the one the user is interacting with, and doesn't show an error.

We should not rely on `state.dropDownItems.length > 0` to mean "this is the autosuggest component the user is interacting with right now," as it is possible to filter the `dropDownItems` down to a length of zero while interacting with the component.

Issue #2516, opened 2023-08-09
Autosuggest: set focus back to text box when pressing esc in dropdown


brian-smith-tcril:
Observed: When pressing esc in the dropdown in the Form.Autosuggest component, the dropdown closes but the focus seems to still be inside the dropdown. The user must hit shift-tab twice after hitting escape to get back to typing in the textbox.

Expected: When pressing esc in the dropdown in the Form.Autosuggest compoent, focus returns to the text box

PR #2517, opened 2023-08-09
fix: add isActive to FormAutosuggest state

cintnguyen:
## Description

state.dropDownItems.length > 0 used to determine if an autosuggest component was active or not which was inaccurate. Found an alternative way to specify which component is being interacted with through setting a new bool isActive on the state.

Resolves : #2510 

### Deploy Preview

https://deploy-preview-2517--paragon-openedx.netlify.app/components/form/form-autosuggest/

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Form/FormAutosuggest.jsx
+++ b/src/Form/FormAutosuggest.jsx
@@ -32,6 +32,7 @@ function FormAutosuggest({
     ignoredKeys: ignoredArrowKeysNames,
   });
   const [isMenuClosed, setIsMenuClosed] = useState(true);
+  const [isActive, setIsActive] = useState(false);
   const [state, setState] = useState({
     displayValue: value || '',
     errorMessage: '',
@@ -87,6 +88,7 @@ function FormAutosuggest({
     };
 
     if (isMenuClosed) {
+      setIsActive(true);
       newState.dropDownItems = getItems(state.displayValue);
       newState.errorMessage = '';
     }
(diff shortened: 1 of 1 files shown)


brian-smith-tcril:
@adamstankiewicz just a heads up on this one that the only failing part is codecov (and the overall coverage is actually higher than it was before this PR)

src/Form/FormAutosuggest.jsx
```diff
@@ -124,13 +127,12 @@ function FormAutosuggest({
...
 
   const keyDownHandler = e => {
-    if (e.key === 'Escape') {
+    if (e.key === 'Escape' && isActive) {
       e.preventDefault();
 
       setState(prevState => ({
         ...prevState,
         dropDownItems: [],
-        errorMessage: !state.displayValue ? errorMessageText : '',
```
adamstankiewicz, line 133: [clarification] If removing the `errorMessage` state here and below, is it correct that this component will no longer be using the `errorMessageText` prop anymore?

brian-smith-tcril: it seems to still be working on https://deploy-preview-2517--paragon-openedx.netlify.app/components/form/form-autosuggest/

image

cintnguyen: We're not changing the errorMessage state in the keyDownHandler since the user will still be focused on the component even when they type esc (because this only closes the drop down) #2516. Therefore, we wouldn't want there to be any reported errors until they have left the component. 

We still are keeping errorMessageText prop since we use it in handleDocumentClick function above.

brian-smith-tcril (approved):

adamstankiewicz (approved):

adamstankiewicz:
@cintnguyen LGTM. Looks like there's just a conflict in the `FormAutosuggest` component that will need to be addressed now.

cintnguyen:
@adamstankiewicz Merge conflict resolved!

Issue #1964, opened 2023-01-27
Define types for Paragon exports


adamstankiewicz:

PKulkoRaccoonGang:
Pull Request: https://github.com/openedx/paragon/pull/2177

monteri:
Reacreated PR: https://github.com/openedx/paragon/pull/2566

bradenmacdonald:
Paragon has working type definitions now, although they're incomplete. Closing this.

PR #2566, opened 2023-08-21
[BD-46] feat: defined types for Paragon export

monteri:
## Description

Define types for Paragon exports
Issue: https://github.com/openedx/paragon/issues/1964

### Deploy Preview

## Merge Checklist

## Post-merge Checklist


PKulkoRaccoonGang (approved):

bradenmacdonald:
I'm really happy to see this PR! I can help test/review if you like - let me know how I can help.

I have a couple questions:
1. How were these `.d.ts` files generated? Are they manually authored or generated from propTypes, etc.? How do we ensure they stay in sync with propTypes and the actual implementation?
2. Don't we need to add `exports.*.types:` and `typesVersions` etc. to `package.json` for this to work? Something like [this](https://github.com/openedx/frontend-platform/pull/619/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519).

PKulkoRaccoonGang:
@bradenmacdonald Thank you for paying attention to this PR.
The initial idea was to rewrite the Paragon components into Typescript. After discussions with @adamstankiewicz, we decided this would be a fairly breaking change and wrote the types separately without changing the component file extensions.

1. The `.d.ts` files were written by hand. Now we have type declarations for exported constants and `propTypes`. We did not consider the option of automatically generating types because the Paragon components use Bootstrap and we do not have access to the full props API. If you have any ideas on how to solve this problem it would be great 💯 
2. We tested type checking in MFE and it worked. Your proposal may solve a problem we were unaware of at the time of development.

bradenmacdonald:
@PKulkoRaccoonGang Thanks for the details!

@PKulkoRaccoonGang @adamstankiewicz What do you think about splitting this up into smaller PRs and trying to get them merged separately? I think we should really try to make at least the tests and the documentation site consume the TypeScript types, so that we have some way of checking that they're used and in sync.

Here's an example of what I mean: https://github.com/openedx/paragon/pull/3016 (I noticed that this PR doesn't include types for the `<Icon>` component, and the existing types were incorrect, so I fixed the types and made changes to the tests and the documentation site so that they're using the types. That way if the types change and introduce a bug, we should see errors when checking the tests or the docs site.)

bradenmacdonald:
This PR is a bit old now, and I think we've decided to focus on [converting components to TypeScript](https://github.com/openedx/paragon/issues/3739) rather than just adding separate type information, so I'm going to close it for now. It could be a useful reference for people as they convert components to TypeScript though.

PR #2570, opened 2023-08-22
feat: add attributes to input and wrapper component

### LINKED ISSUE #2438: Form.Autosuggest Accessibility - Attributes
### aria-activedescendent

This attribute tells assistive technologies which item in the dropdown is currently focused. The attribute should be attached to the `<input>` element and its value should be the ID of the currently focused item in the list. Currently, we do not require `<Form.AutosuggestOption>` to have an ID, so we would first need to update the component to require an ID or at least create ID's dynamically within `<Form.Autosuggest>`. Then, we'll need to provide functionality in `<Form.Autosuggest>` that utilizes the `onFocus` event for each item in the list. The callback for the `onFocus` event should update `aria-activedescendent`.

We could configure the `onFocus` event similar to how we are already providing an `onClick` event to the child components:
https://github.com/openedx/paragon/blob/d383b57b5db6c1096e38e991d0eb5cbcac5d83a0/src/Form/FormAutosuggest.jsx#L66-L71

### aria-live=”assertive”

This attribute is used to announce something to the user. The “assertive” value tells screen readers that it should take priority over anything else being read. This attribute can be attached to a `<div>` with the text to announce inside. The `<div>` can be inserted above the component as seen in the example. This task involves working with screen readers to ensure the correct number of options are announced to the user. 

```
<ComponentWrapper>
     <div aria-live="assertive" className="sr-only">
	 {dropDownItems.length} options found
     </div>
     {...Component}
<ComponentWrapper>
```

httpsmenahassan:
## Description

Resolves #2438. 

- Create IDs for items in dropdown
- Add `aria-activedescendant` attribute to `input` to identify to assistive technologies which item in the dropdown is being focused
- Add `aria-live=”assertive”` to wrapper component to allow screen readers to read out the number of options in the dropdown

### Deploy Preview

https://deploy-preview-2570--paragon-openedx.netlify.app/components/form/form-autosuggest/

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Form/FormAutosuggest.jsx
+++ b/src/Form/FormAutosuggest.jsx
@@ -2,6 +2,7 @@ import React, {
   useEffect, useState,
 } from 'react';
 import PropTypes from 'prop-types';
+import { v4 as uuidv4 } from 'uuid';
 import { useIntl } from 'react-intl';
 import { KeyboardArrowUp, KeyboardArrowDown } from '../../icons';
 import Icon from '../Icon';
@@ -38,6 +39,11 @@ function FormAutosuggest({
     errorMessage: '',
     dropDownItems: [],
   });
+  const [activeMenuItemId, setActiveMenuItemId] = useState(null);
+
+  const handleMenuItemFocus = (menuItemId) => {
+    setActiveMenuItemId(menuItemId);
+  };
 
   const handleItemClick = (e, onClick) => {
     const clickedValue = e.currentTarget.getAttribute('data-value');
@@ -63,12 +69,15 @@ function FormAutosuggest({
     let childrenOpt = React.Children.map(children, (child) => {
       // eslint-disable-next-line no-shadow
       const { children, onClick, ...rest } = child.props;
+      const menuItemId = uuidv4();
 
       return React.cloneElement(child, {
         ...rest,
         children,
         'data-value': children,
         onClick: (e) => handleItemClick(e, onClick),
+        id: menuItemId,
+        onFocus: () => handleMenuItemFocus(menuItemId),
       });
     });
 
@@ -219,6 +228,9 @@ function FormAutosuggest({
 
   return (
     <div className="pgn__form-autosuggest__wrapper" ref={parentRef}>
+      <div aria-live="assertive" className="sr-only" data-testid="autosuggest-screen-reader-options-count">
+        {`${state.dropDownItems.length} options found`}
+      </div>
       <FormGroup isInvalid={!!state.errorMessage}>
         <FormControl
           aria-expanded={(state.dropDownItems.length > 0).toString()}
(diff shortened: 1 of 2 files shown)


src/Form/FormAutosuggest.jsx
```diff
@@ -59,21 +62,26 @@ function FormAutosuggest({
   };
 
   function getItems(strToFind = '') {
-    let childrenOpt = React.Children.map(children, (child) => {
+    let childrenOpt = React.Children.map(children, (child, index) => {
       // eslint-disable-next-line no-shadow
       const { children, onClick, ...rest } = child.props;
+      // Generate a unique ID for each menu item
+      const menuItemId = `pgn__form-autosuggest__menuItem-${index}`;
```
adamstankiewicz, line 69: Two questions here:

1. Duplicate IDs are possible here (e.g., if there are multiple `FormAutosuggest` components rendered at the same time in the UI with the same number of children. There might be an opportunity to be use `name` to differentiate between multiple `FormAutosuggest` components?
3. Is there a concern if/when the children change order between re-rendered? In theory, the `index` would

brian-smith-tcril:
@adamstankiewicz just to verify/document decisions from Paragon WG today:
* make name a required prop for autosuggest
* ensure we generate unique ids (something like `autoSuggestName-uuid`) to set `aria-activedescendant`

src/Form/FormAutosuggest.jsx
brian-smith-tcril: it seems on http://react-autosuggest.js.org/ the ids will update as the items in the dropdown change

image

image

adamstankiewicz: As long as the `id` changes in all the appropriate places at the same time (which I'm assuming is the same; that is, it doesn't update the ID in one place and leave it unchanged somewhere else), I imagine the re-rendering/re-ordering of `React.Children` wouldn't be an issue. Just worth a sanity check :)

adamstankiewicz:
@brian-smith-tcril @httpsmenahassan 

> make name a required prop for autosuggest

I'd like to better understand how and why `name` is used today. It looks like it is only passed to a `feedback-for` attribute, but on a quick search, I'm struggling to find any documentation on if this is a standard attribute related to a11y. I haven't seen `feedback-for` used on the `FormControlFeedback` component before...

If we truly need it, then we'll need to consider making it a required prop. Ideally, though, the solution for generating unique IDs wouldn't rely on the prop as if it did, we'd need to make a breaking change to make the `name` (or otherwise) prop required.

> ensure we generate unique ids (something like autoSuggestName-uuid) to set aria-activedescendant

If the ID is guaranteed to be unique based on a uuid, I don't think the `name` bit above is relevant? FWIW, we generate a UUID for use in React `key` attributes in some places to ensure uniqueness:

```
import { v4 as uuidv4 } from 'uuid';
```

We also use this `newId` function in Paragon ([source](https://github.com/openedx/paragon/blob/b6a889571a4c02076508dd17ea2947b09208ca54/src/utils/newId.js#L3)) to generate unique IDs in some places.

brian-smith-tcril:
@adamstankiewicz if we can avoid making name a required prop and instead just do the uuid thing that works perfectly imo

brian-smith-tcril (approved):
Awesome! :shipit:

PR #2611, opened 2023-09-08
chore: move autosuggest tests from Enzyme to RTL

cintnguyen:
## Description

Move autosuggest tests from Enzyme to RTL

### Deploy Preview

N/A

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Form/FormAutosuggest.jsx
+++ b/src/Form/FormAutosuggest.jsx
@@ -102,6 +102,7 @@ function FormAutosuggest({
   const iconToggle = (
     <IconButton
       className="pgn__form-autosuggest__icon-button"
+      data-testid="autosuggest_iconbutton"
       src={isMenuClosed ? KeyboardArrowDown : KeyboardArrowUp}
       iconAs={Icon}
       size="sm"
@@ -230,6 +231,7 @@ function FormAutosuggest({
           onChange={handleOnChange}
           onClick={handleClick}
           trailingElement={iconToggle}
+          data-testid="autosuggest_textbox_input"
           {...props}
         />
 
@@ -253,7 +255,7 @@ function FormAutosuggest({
       >
         {isLoading ? (
           <div className="pgn__form-autosuggest__dropdown-loading">
-            <Spinner animation="border" variant="dark" screenReaderText={screenReaderText} />
+            <Spinner animation="border" variant="dark" screenReaderText={screenReaderText} data-testid="autosuggest_loading_spinner" />
           </div>
         ) : state.dropDownItems.length > 0 && state.dropDownItems}
       </ul>
--- a/src/Form/FormAutosuggestOption.jsx
+++ b/src/Form/FormAutosuggestOption.jsx
@@ -12,8 +12,9 @@ function FormAutosuggestOption({
   return (
     <MenuItem
       as="li"
+      data-testid="autosuggest_optionitem"
       role="option"
-      tabindex="-1"
+      tabIndex="-1"
       onClick={onClick}
       className={classNames(className, 'dropdown-item')}
       {...props}
--- a/src/Form/tests/FormAutosuggest.test.jsx
+++ b/src/Form/tests/FormAutosuggest.test.jsx
@@ -1,21 +1,11 @@
 import React from 'react';
-import { mount } from 'enzyme';
-import { act } from 'react-dom/test-utils';
+import PropTypes from 'prop-types';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom';
 import { IntlProvider } from 'react-intl';
 import FormAutosuggest from '../FormAutosuggest';
 import FormAutosuggestOption from '../FormAutosuggestOption';
-
-const createDocumentListenersMock = () => {
-  const listeners = {};
-  const handler = (domEl, e) => listeners?.[e]?.({ target: domEl });
-
-  document.addEventListener = jest.fn((e, fn) => { listeners[e] = fn; });
-  document.removeEventListener = jest.fn(e => { delete listeners[e]; });
-
-  return {
-    click: domEl => handler(domEl, 'click'),
-  };
-};
 
 function FormAutosuggestWrapper(props) {
   return (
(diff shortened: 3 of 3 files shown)


src/Form/FormAutosuggestOption.jsx
```diff
@@ -13,7 +13,7 @@ function FormAutosuggestOption({
     <MenuItem
       as="li"
       role="option"
-      tabindex="-1"
```
brian-smith-tcril, line 16: good catch!

brian-smith-tcril (approved):

src/Form/tests/FormAutosuggest.test.jsx
```diff
@@ -25,141 +14,179 @@ function FormAutosuggestWrapper(props) {
...
+    expect(screen.getByDisplayValue('Test Value')).toBeInTheDocument();
+  });
 
-    it('renders component with options', () => {
-      container.find('input').simulate('click');
-      const optionsList = container.find('.pgn__form-autosuggest__dropdown').find('li');
+  it('renders component with options', () => {
+    const { getByTestId, container } = render(<FormAutosuggestTestComponent />);
+    const input = getByTestId('autosuggest_textbox_input');
+    fireEvent.click(input);
```
adamstankiewicz, line 68: [curious] Is there a reason to use `fireEvent` here (and elsewhere throughout the tests) versus using `userEvent`? 

If it's possible to use `userEvent.click` instead, it's generally preferable since it simulates more of the true user interactions in the browser, not just the low-level click itself. 

For example, `userEvent.click` also simulates a previous `hover` event that a real user would trigger prior to clicking.

Resources:
* https://testing-library.com/docs/guide-events/
* https://blog.mimacom.com/react-testing-library-fireevent-vs-userevent/

src/Form/tests/FormAutosuggest.test.jsx
```diff
@@ -25,141 +14,179 @@ function FormAutosuggestWrapper(props) {
...
+  it('renders component without error', () => {
+    render(<FormAutosuggestWrapper />);
+  });
 
-      expect(wrapper.exists('.pgn__form-autosuggest__dropdown-loading')).toBe(true);
-      expect(wrapper.props().isLoading).toBe(true);
-    });
+  it('renders without loading state', () => {
+    const { container } = render(<FormAutosuggestTestComponent />);
+    expect(container.querySelector('.pgn__form-autosuggest__dropdown-loading')).toBeNull();
```
adamstankiewicz, line 52: Is there a way to avoid using the class name implementation detail in this test?

For example, if the component renders a `Spinner` with the supplied `screenReaderText` prop, could this test assert that the `Spinner` is not rendered based on a `data-testid` attribute versus relying on the class name that could change at any time. If the class name did change down the road, this test would still pass.

src/Form/tests/FormAutosuggest.test.jsx
```diff
@@ -25,141 +14,179 @@ function FormAutosuggestWrapper(props) {
...
+  });
 
-    it('renders the auto-populated value if it exists', () => {
-      const wrapper = mount(<FormAutosuggestWrapper value="Test Value" />);
+  it('render with loading state', () => {
+    const { container } = render(<FormAutosuggestWrapper isLoading />);
+    expect(container.querySelector('.pgn__form-autosuggest__dropdown-loading')).toBeTruthy();
```
adamstankiewicz, line 57: Similar to above comment, could we either assert that the `screenReaderText` passed to the `Spinner` is rendered, or alternatively, rely on a more stable `data-testid` instead of the class name?

adamstankiewicz (approved):
LGTM!

src/Form/tests/FormAutosuggest.test.jsx
cintnguyen: Fixed in:
https://github.com/openedx/paragon/compare/299441986865ab87377629aa4d99f13ed8fc47d7..6c9a96ea3e2ba2ef4ab37d5f266b4d7205f5c718

src/Form/tests/FormAutosuggest.test.jsx
cintnguyen: Fixed in:
https://github.com/openedx/paragon/compare/299441986865ab87377629aa4d99f13ed8fc47d7..6c9a96ea3e2ba2ef4ab37d5f266b4d7205f5c718

brian-smith-tcril:
So the current failing test is the error state test. It was written by having the simulated user enter the textbox and press escape. This worked before https://github.com/openedx/paragon/commit/4816baac16644eabb6362405627555d71ce77cd8#diff-96e13c1a773ec8d218ea5667d3dec6c8f6eeadcb00c81584f50d6c3d4d52135fL133, but now that doesn't trigger the error state.

It should probably be updated to simulate a click outside, as that will still trigger the error state (I verified on https://paragon-openedx.netlify.app/components/form/form-autosuggest/)

cintnguyen:
> So the current failing test is the error state test. It was written by having the simulated user enter the textbox and press escape. This worked before [4816baa#diff-96e13c1a773ec8d218ea5667d3dec6c8f6eeadcb00c81584f50d6c3d4d52135fL133](https://github.com/openedx/paragon/commit/4816baac16644eabb6362405627555d71ce77cd8#diff-96e13c1a773ec8d218ea5667d3dec6c8f6eeadcb00c81584f50d6c3d4d52135fL133), but now that doesn't trigger the error state.
> 
> It should probably be updated to simulate a click outside, as that will still trigger the error state (I verified on https://paragon-openedx.netlify.app/components/form/form-autosuggest/)

Fix here: https://github.com/openedx/paragon/compare/ba5085a9a58e9a35f545070541decaba69c8130f..33fb35b955f6e7de8b872791fe5794616300cff9

brian-smith-tcril (approved):

src/Form/FormAutosuggest.jsx
```diff
@@ -102,6 +102,7 @@ function FormAutosuggest({
   const iconToggle = (
     <IconButton
       className="pgn__form-autosuggest__icon-button"
+      data-testid="autosuggest_iconbutton"
```
PKulkoRaccoonGang, line 105: We recently [rewrote all tests](https://github.com/openedx/paragon/pull/2581) (except `FormAutosuggest`) to RTL. For uniformity of test identifiers, we agreed to use a kebab case. Can you convert the names of the identifiers to a uniform format? `autosuggest-iconbutton`

src/Form/FormAutosuggest.jsx
```diff
@@ -253,7 +255,7 @@ function FormAutosuggest({
       >
         {isLoading ? (
           <div className="pgn__form-autosuggest__dropdown-loading">
-            <Spinner animation="border" variant="dark" screenReaderText={screenReaderText} />
+            <Spinner animation="border" variant="dark" screenReaderText={screenReaderText} data-testid="autosuggest_loading_spinner" />
```
PKulkoRaccoonGang, line 258: Maybe more readable
```suggestion
            <Spinner 
                animation="border" 
                variant="dark" 
                screenReaderText={screenReaderText} 
                data-testid="autosuggest-loading-spinner" 
            />
```

src/Form/FormAutosuggest.jsx
```diff
@@ -230,6 +231,7 @@ function FormAutosuggest({
           onChange={handleOnChange}
           onClick={handleClick}
           trailingElement={iconToggle}
+          data-testid="autosuggest_textbox_input"
```
PKulkoRaccoonGang, line 234:
```suggestion
          data-testid="autosuggest-textbox-input"
```

src/Form/FormAutosuggestOption.jsx
```diff
@@ -12,8 +12,9 @@ function FormAutosuggestOption({
   return (
     <MenuItem
       as="li"
+      data-testid="autosuggest_optionitem"
```
PKulkoRaccoonGang, line 15:
```suggestion
      data-testid="autosuggest-optionitem"
```

src/Form/tests/FormAutosuggest.test.jsx
```diff
@@ -1,22 +1,12 @@
 import React from 'react';
-import { mount } from 'enzyme';
-import { act } from 'react-dom/test-utils';
+import PropTypes from 'prop-types';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom';
```
PKulkoRaccoonGang, line 5: Is it better to move this line to `setupTests.js`? Did you encounter some problems with this?

brian-smith-tcril:
@PKulkoRaccoonGang I'm sorry I was a bit hasty and didn't see these comments before hitting merge. We should be able to address them in a quick follow-up PR.

Issue #2458, opened 2023-07-20
Extend Form.Control with input mask


monteri:
Input mask can nr useful in cases with phone number, credit card, date.
There are existing solutions for React that can be either integrated with `Form.Control` or partially taken to create own input mask functionality.
Example:
[react-input-mask by sanniassin](http://sanniassin.github.io/react-input-mask/demo.html) 

Considered enough of a PITA for users where it’s a WCAG concern. (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/tel)

PKulkoRaccoonGang:
PR: (DRAFT) https://github.com/openedx/paragon/pull/2626

PR #2626, opened 2023-09-15
[BD-46] feat: added FormControl input mask

PKulkoRaccoonGang:
## Description

- added FormControl input mask.

**Issue:** https://github.com/openedx/paragon/issues/2458

### Deploy Preview

[Form.Control component](https://deploy-preview-2626--paragon-openedx.netlify.app/components/form/form-control/)

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Form/FormControl.jsx
+++ b/src/Form/FormControl.jsx
@@ -2,6 +2,7 @@ import React, { useCallback, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import RBFormControl from 'react-bootstrap/FormControl';
+import { IMaskInput } from 'react-imask';
 import { useFormGroupContext } from './FormGroupContext';
 import FormControlFeedback from './FormControlFeedback';
 import FormControlDecoratorGroup from './FormControlDecoratorGroup';
@@ -17,6 +18,7 @@ const FormControl = React.forwardRef(({
   floatingLabel,
   autoResize,
   onChange,
+  withMask,
   ...props
 }, ref) => {
   const {
@@ -71,7 +73,7 @@ const FormControl = React.forwardRef(({
       className={className}
     >
       <RBFormControl
-        as={as}
+        as={withMask ? IMaskInput : as}
         ref={resolvedRef}
         size={size}
         isInvalid={isInvalid}
--- a/src/Form/FormControl.jsx
+++ b/src/Form/FormControl.jsx
@@ -9,8 +9,6 @@ import FormControlDecoratorGroup from './FormControlDecoratorGroup';
 
 import { callAllHandlers, useHasValue } from './fieldUtils';
 
-const DEFAULT_MASK_VALUE = '00-00-000';
-
 const FormControl = React.forwardRef(({
   as,
   className,
@@ -21,7 +19,6 @@ const FormControl = React.forwardRef(({
   autoResize,
   onChange,
   hasInputMask,
-  mask,
   ...props
 }, ref) => {
   const {
@@ -76,7 +73,7 @@ const FormControl = React.forwardRef(({
       className={className}
     >
       <RBFormControl
-        as={hasInputMask ? IMaskInput : as}
+        as={hasInputMask?.length ? IMaskInput : as}
         ref={resolvedRef}
         size={size}
         isInvalid={isInvalid}
@@ -85,7 +82,7 @@ const FormControl = React.forwardRef(({
           'has-value': hasValue,
         })}
         onChange={handleOnChange}
-        mask={mask}
+        mask={hasInputMask}
         {...controlProps}
       />
     </FormControlDecoratorGroup>
@@ -128,10 +125,8 @@ FormControl.propTypes = {
   isInvalid: PropTypes.bool,
   /** Only for `as="textarea"`. Specifies whether the input can be resized according to the height of content. */
   autoResize: PropTypes.bool,
-  /** Specifies whether to use an input mask for the input. */
-  hasInputMask: PropTypes.bool,
-  /** Specifies the input mask to be used if `hasInputMask` is set to `true`. */
-  mask: PropTypes.string,
+  /** Specifies what format to use for the input mask. */
+  hasInputMask: PropTypes.string,
 };
 
 FormControl.defaultProps = {
@@ -150,8 +145,7 @@ FormControl.defaultProps = {
   isValid: undefined,
   isInvalid: undefined,
   autoResize: false,
-  hasInputMask: false,
-  mask: DEFAULT_MASK_VALUE,
+  hasInputMask: undefined,
 };
 
 export default FormControl;
--- a/src/Form/FormControl.jsx
+++ b/src/Form/FormControl.jsx
@@ -18,7 +18,7 @@ const FormControl = React.forwardRef(({
   floatingLabel,
   autoResize,
   onChange,
-  hasInputMask,
+  inputMask,
   ...props
 }, ref) => {
   const {
@@ -73,7 +73,7 @@ const FormControl = React.forwardRef(({
       className={className}
     >
       <RBFormControl
-        as={hasInputMask?.length ? IMaskInput : as}
+        as={inputMask ? IMaskInput : as}
         ref={resolvedRef}
         size={size}
         isInvalid={isInvalid}
@@ -82,7 +82,7 @@ const FormControl = React.forwardRef(({
           'has-value': hasValue,
         })}
         onChange={handleOnChange}
-        mask={hasInputMask}
+        mask={inputMask}
         {...controlProps}
       />
     </FormControlDecoratorGroup>
@@ -126,7 +126,7 @@ FormControl.propTypes = {
   /** Only for `as="textarea"`. Specifies whether the input can be resized according to the height of content. */
   autoResize: PropTypes.bool,
   /** Specifies what format to use for the input mask. */
-  hasInputMask: PropTypes.string,
+  inputMask: PropTypes.string,
 };
 
 FormControl.defaultProps = {
@@ -145,7 +145,7 @@ FormControl.defaultProps = {
   isValid: undefined,
   isInvalid: undefined,
   autoResize: false,
-  hasInputMask: undefined,
+  inputMask: undefined,
 };
 
 export default FormControl;
--- a/src/Form/FormControl.jsx
+++ b/src/Form/FormControl.jsx
@@ -9,6 +9,8 @@ import FormControlDecoratorGroup from './FormControlDecoratorGroup';
 
 import { callAllHandlers, useHasValue } from './fieldUtils';
 
+const DEFAULT_MASK_VALUE = '00-00-000';
+
 const FormControl = React.forwardRef(({
   as,
   className,
@@ -18,7 +20,8 @@ const FormControl = React.forwardRef(({
   floatingLabel,
   autoResize,
   onChange,
-  withMask,
+  hasInputMask,
+  mask,
   ...props
 }, ref) => {
   const {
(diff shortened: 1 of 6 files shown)


openedx-webhooks:


src/Form/FormControl.jsx
```diff
@@ -17,6 +20,8 @@ const FormControl = React.forwardRef(({
   floatingLabel,
   autoResize,
   onChange,
+  hasInputMask,
+  mask,
```
monteri, line 24: Can we combine these two params in single one `mask`?

src/Form/FormControl.jsx
```diff
@@ -140,6 +150,8 @@ FormControl.defaultProps = {
   isValid: undefined,
   isInvalid: undefined,
   autoResize: false,
+  hasInputMask: false,
+  mask: DEFAULT_MASK_VALUE,
```
monteri, line 154: In case of combining maybe it is better to leave it `undefined` by default?

src/Form/tests/FormControl.test.jsx
```diff
@@ -1,26 +1,63 @@
...
+
     expect(onChangeFunc).toHaveBeenCalledTimes(1);
-    expect(ref.current.style.height).toEqual(`${ref.current.scrollHeight + ref.current.offsets}px`);
+    expect(ref.current.style.height).toEqual(`${ref.current.scrollHeight + ref.current.offsetHeight}px`);
+  });
+  it('should apply and accept input mask for phone numbers', () => {
+    render(<Component />);
+
+    const input = screen.getByTestId('form-control-with-mask');
+    fireEvent.change(input, { target: { value: '1234567890' } });
```
monteri, line 53: Please, use `userEvent` instead

PKulkoRaccoonGang: This is not possible for the pure value example. To pass a value and receive a cleared one, the [react-imask](https://github.com/uNmAnNeR/imaskjs) library provides an `onAccept` property, which does not work when using `userEvent`. Inside `react-imask`, processing of the value passed to `onAccept` occurs through manual transmission and processing (by library tools), in contrast to the familiar input to the `onChange` property.
Let's leave `fireEvent` here since this is the only tool that can recognize the cleared value at the moment.

src/Form/FormControl.jsx
```diff
@@ -122,6 +125,8 @@ FormControl.propTypes = {
   isInvalid: PropTypes.bool,
   /** Only for `as="textarea"`. Specifies whether the input can be resized according to the height of content. */
   autoResize: PropTypes.bool,
+  /** Specifies what format to use for the input mask. */
+  hasInputMask: PropTypes.string,
```
viktorrusakov, line 129: Usually, if a variable's name starts with `has` it indicates that it is of boolean type, in your case it's a string, so it would be better to name it just `inputMask`.

src/Form/FormControl.jsx
```diff
@@ -71,7 +73,7 @@ const FormControl = React.forwardRef(({
       className={className}
     >
       <RBFormControl
-        as={as}
+        as={hasInputMask?.length ? IMaskInput : as}
```
viktorrusakov, line 76: I think you can simplify this to `as={inputMask ? IMaskInput : as}` (because if `inputMask` is an empty string it will be considered falsy)

PKulkoRaccoonGang: Yeah, you are right. Thanks 💯

src/Form/FormControl.jsx
PKulkoRaccoonGang: Corrected, thanks

src/Form/form-control.mdx
```diff
@@ -163,6 +162,160 @@ or [select attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element
 }
 ```
 
+## Input masks
+Paragon uses the [react-imask](https://www.npmjs.com/package/react-imask) library,
+which allows you to add masks of different types for inputs.
+To create your own mask, you need to pass the required mask pattern (`+{1}(000)000-00-00`) to the `inputMask` property. <br />
```
adamstankiewicz, line 168: The US-based example phone number mask should likely be `+{1} (000) 000-0000` so that phone numbers are formatted like `+1 (555) 555-5555`, which is typical for a US-based phone number.

src/Form/form-control.mdx
```diff
@@ -163,6 +162,160 @@ or [select attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element
...
+  const [maskType, setMaskType] = useState('phone');
+  {/* end example state */}
+
+  const inputsWithMask = {
+    phone: (
+      <>
+        <h3>Phone</h3>
+        <Form.Group>
+          <Form.Control
+            inputMask="+{1}(000)000-00-00"
```
adamstankiewicz, line 183: The US-based example phone number mask should likely be `+{1} (000) 000-0000` so that phone numbers are formatted like `+1 (555) 555-5555`, which is typical for a US-based phone number.

src/Form/form-control.mdx
```diff
@@ -163,6 +162,160 @@ or [select attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element
...
+          leadingElement={<Icon src={FavoriteBorder} />}
+          floatingLabel="What is your credit card number?"
+        />
+      </Form.Group>
+    </>),
+    date: (<>
+      <h3>Date</h3>
+      <Form.Group>
+        <Form.Control
+          inputMask={Date}
```
adamstankiewicz, line 208: A bit curious about this one; it appears it doesn't let users input months without leading zeros? For example, when trying to type in June 10, I type `6` for the month, but it doesn't accept any input until I add a leading zero first.

Beyond this, I'm also wondering if we should keep the date input mask example of these examples given the `Form.Control` can be used specifically with a date picker input type.

src/Form/form-control.mdx
```diff
@@ -163,6 +162,160 @@ or [select attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element
...
+          value={value}
+          onChange={handleChange}
+          leadingElement={<Icon src={FavoriteBorder} />}
+          floatingLabel="What is your OPT password?"
+        />
+      </Form.Group>
+    </>),
+    price: (
+        <>
+          <h3>Course prise</h3>
```
adamstankiewicz, line 249: nit: "Price"

src/Form/form-control.mdx
```diff
@@ -163,6 +162,160 @@ or [select attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element
...
+
+```jsx live
+() => {
+  const [value, setValue] = useState('');
+
+  return (
+      <>
+        <Form.Group>
+          <Form.Control
+            inputMask="+{1}(000)000-00-00"
```
adamstankiewicz, line 301: The US-based example phone number mask should likely be `+{1} (000) 000-0000` so that phone numbers are formatted like `+1 (555) 555-5555`, which is typical for a US-based phone number.

src/Form/tests/FormControl.test.jsx
```diff
@@ -31,4 +49,21 @@ describe('FormControl', () => {
     expect(onChangeFunc).toHaveBeenCalledTimes(inputText.length);
     expect(ref.current.style.height).toEqual(`${ref.current.scrollHeight + ref.current.offsetHeight}px`);
   });
+
+  it('should apply and accept input mask for phone numbers', () => {
+    render(<Component />);
+
+    act(() => {
+      const input = screen.getByTestId('form-control-with-mask');
+      userEvent.type(input, '1234567890');
```
adamstankiewicz, line 58: [curious] generally, `userEvent.*` functions are called outside of `act` since the `userEvent.*` functions are already wrapped with `act` ([example source](https://github.com/testing-library/user-event/issues/497#issuecomment-729904582)).

src/Form/tests/FormControl.test.jsx
```diff
@@ -31,4 +49,21 @@ describe('FormControl', () => {
...
+      expect(input.value).toBe('+1(234)567-89-0');
+    });
+  });
+  it('should be cleared from the mask elements value', () => {
+    render(<Component isClearValue />);
+
+    const input = screen.getByTestId('form-control-with-mask');
+    fireEvent.change(input, { target: { value: '1234567890' } });
```
adamstankiewicz, line 66: [curious] Why use `userEvent.type` but `fireEvent.change` here? Ideally, we could rely on `userEvent` to more closely mimic real user behavior.

src/Form/tests/FormControl.test.jsx
PKulkoRaccoonGang: @adamstankiewicz thank you for bringing this issue to our attention 👍
At the moment we have a [lot of similar examples](https://github.com/search?q=repo%3Aopenedx%2Fparagon%20act(&type=code)) of wrapping in `act` `userEvent` in Paragon. I've looked into several interesting sources (including the [one](https://github.com/testing-library/user-event/issues/497#issuecomment-729904582) you shared), this is really an unnecessary solution, here are some confirmations:
1. [When should I use act() in react-testing-library?](https://flyingsquirrel.medium.com/when-should-i-use-act-in-react-testing-library-d7dd22a3340e)
2. [Internal implementation](https://github.com/testing-library/react-testing-library/blob/main/src/pure.js#L21-L23)
3. [Is wrapping with act now required?](https://github.com/testing-library/user-event/discussions/906#discussioncomment-2522468)

My solution is based on [Is wrapping with act now required?](https://github.com/testing-library/user-event/discussions/906#discussioncomment-2522468). As it turned out, we have differences in the `@testing-library/dom` subdependency in the versions of the `@testing-library/react@12.1.5` and `@testing-library/user-event@13.5.0` libraries. A dependency conflict negatively affects the performance of the `userEvent` and we see a warning after running tests.

image

**We have several options for further action:**
1. Apply now the solution with an override of the `@testing-library/dom` package (with version `9.3.3`). This way we will get rid of the problem of displaying warnings in the console and our tests will be cleaner. This solution can be developed in a separate task, where we remove the wrapping in `act` `userEvent`. I have verified this in the Tabs component and in my current case (FormControl).
2. Make these changes and “synchronize” the `@testing-library/dom` subdependency in `@testing-library/react` and `@testing-library/user-event` in the pull request for [updating to React 18](https://github.com/openedx/paragon/pull/2774). So It is quite likely that we will not have to do this since we will update the versions of `@testing-library/react` and `@testing-library/user-event`. But we cannot predict whether `@testing-library/dom` versions will change in the future. Therefore, the override option seems workable.

An [article about dependency overriding](https://github.com/npm/rfcs/blob/main/accepted/0036-overrides.md#motivation), in which we can make sure that our intentions are correct:

> There should only be a single copy of a given dependency in our tree, but npm's dependency resolution will result in a duplicate due to version conflicts.

src/Form/form-control.mdx
PKulkoRaccoonGang: Corrected

src/Form/form-control.mdx
PKulkoRaccoonGang: Corrected, thanks

src/Form/form-control.mdx
PKulkoRaccoonGang: It seems to me that this is a good solution if the mask requires entering a zero before the month.
But this example seems redundant since we have a better use case.
Removed this example

src/Form/tests/FormControl.test.jsx
adamstankiewicz: Thanks for digging into, this @PKulkoRaccoonGang! TIL that `@testing-library/dom` has to be synchronized :) I agree with your synopsis/options; interesting that I'm not sure this has come up before in MFEs that use RTL. But I agree the `overrides` is sufficient to hold over until the React 18 upgrade.

Issue #2594, opened 2023-09-04
ProgressBar.Annotated dir="rtl" problem


monteri:
![image](https://github.com/openedx/paragon/assets/36768631/b02b08b2-d9de-4170-b668-f5d49fe63783)

PKulkoRaccoonGang:
DRAFT: https://github.com/openedx/paragon/pull/2641

PR #2641, opened 2023-09-21
[BD-46] fix: fixed rtl direction in ProgressBar

PKulkoRaccoonGang:
## Description

- fixed `rtl` direction in ProgressBar component

**Issue:** https://github.com/openedx/paragon/issues/2594

### Deploy Preview

[ProgressBar component](https://deploy-preview-2641--paragon-openedx.netlify.app/components/progressbar/#annotated-variant)

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/ProgressBar/index.jsx
+++ b/src/ProgressBar/index.jsx
@@ -31,18 +31,27 @@ function ProgressBarAnnotated({
   thresholdHint,
   ...props
 }) {
+  const [direction, setDirection] = React.useState('ltr');
   const progressInfoRef = React.useRef();
   const thresholdInfoRef = React.useRef();
+  const progressAnnotatedRef = React.useRef();
   const thresholdPercent = (threshold || 0) - (now || 0);
   const isProgressHintAfter = now < HINT_SWAP_PERCENT;
   const isThresholdHintAfter = threshold < HINT_SWAP_PERCENT;
   const progressColor = VARIANTS.includes(variant) ? variant : PROGRESS_DEFAULT_VARIANT;
   const thresholdColor = VARIANTS.includes(thresholdVariant) ? thresholdVariant : THRESHOLD_DEFAULT_VARIANT;
 
+  useEffect(() => {
+    if (progressAnnotatedRef.current) {
+      const pageDirection = window.getComputedStyle(progressAnnotatedRef.current).getPropertyValue('direction');
+      setDirection(pageDirection);
+    }
+  }, []);
+
   const positionAnnotations = useCallback(() => {
-    placeInfoAtZero(progressInfoRef, isProgressHintAfter, ANNOTATION_CLASS);
-    placeInfoAtZero(thresholdInfoRef, isThresholdHintAfter, ANNOTATION_CLASS);
-  }, [isProgressHintAfter, isThresholdHintAfter]);
+    placeInfoAtZero(progressInfoRef, direction, isProgressHintAfter, ANNOTATION_CLASS);
+    placeInfoAtZero(thresholdInfoRef, direction, isThresholdHintAfter, ANNOTATION_CLASS);
+  }, [direction, isProgressHintAfter, isThresholdHintAfter]);
 
   useEffect(() => {
     positionAnnotations();
@@ -61,11 +70,11 @@ function ProgressBarAnnotated({
   );
 
   return (
-    <div className="pgn__progress-annotated">
+    <div ref={progressAnnotatedRef} className="pgn__progress-annotated">
       {!!label && (
         <div
           className="pgn__progress-info"
-          style={{ left: `${now}%` }}
+          style={direction === 'rtl' ? { right: `${now}%` } : { left: `${now}%` }}
           ref={progressInfoRef}
         >
           {!isProgressHintAfter && getHint(progressHint)}
@@ -96,7 +105,7 @@ function ProgressBarAnnotated({
       {(!!threshold && !!thresholdLabel) && (
         <div
           className="pgn__progress-info"
-          style={{ left: `${threshold}%` }}
+          style={direction === 'rtl' ? { right: `${threshold}%` } : { left: `${threshold}%` }}
           ref={thresholdInfoRef}
         >
           {!isThresholdHintAfter && getHint(thresholdHint)}
--- a/src/ProgressBar/index.jsx
+++ b/src/ProgressBar/index.jsx
@@ -20,7 +20,7 @@ function ProgressBar(props) {
   return <ProgressBarBase {...props} />;
 }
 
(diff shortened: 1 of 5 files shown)


openedx-webhooks:


src/ProgressBar/index.jsx
```diff
@@ -31,18 +31,27 @@ function ProgressBarAnnotated({
   thresholdHint,
   ...props
 }) {
+  const [direction, setDirection] = React.useState('ltr');
   const progressInfoRef = React.useRef();
   const thresholdInfoRef = React.useRef();
+  const progressAnnotatedRef = React.useRef();
```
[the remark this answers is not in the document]
PKulkoRaccoonGang: Corrected

src/ProgressBar/index.jsx
```diff
@@ -30,19 +30,29 @@ function ProgressBarAnnotated({
...
+  const resolvedRef = ref || progressAnnotatedRef;
   const thresholdPercent = (threshold || 0) - (now || 0);
   const isProgressHintAfter = now < HINT_SWAP_PERCENT;
   const isThresholdHintAfter = threshold < HINT_SWAP_PERCENT;
   const progressColor = VARIANTS.includes(variant) ? variant : PROGRESS_DEFAULT_VARIANT;
   const thresholdColor = VARIANTS.includes(thresholdVariant) ? thresholdVariant : THRESHOLD_DEFAULT_VARIANT;
 
+  useEffect(() => {
+    if (resolvedRef.current) {
+      const pageDirection = window.getComputedStyle(resolvedRef.current).getPropertyValue('direction');
```
monteri, line 47: @adamstankiewicz @viktorrusakov What do you think about creating such hook called `usePageDirection` where you can pass `ref` and retrieve `"rtl"/"ltr"` values as a result. Right now we have only one instance where it can be used. Maybe other components would also require additional changes in Javascript in order to achieve proper RTL display.

src/ProgressBar/index.jsx
```diff
@@ -61,11 +71,11 @@ function ProgressBarAnnotated({
   );
 
   return (
-    <div className="pgn__progress-annotated">
+    <div ref={progressAnnotatedRef} className="pgn__progress-annotated">
       {!!label && (
         <div
           className="pgn__progress-info"
-          style={{ left: `${now}%` }}
+          style={direction === 'rtl' ? { right: `${now}%` } : { left: `${now}%` }}
```
monteri, line 78:
```
const getOffestStyles = (value, direction) => direction === 'rtl' ? { right: `${value}%` } : { left: `${value}%` }
```
And let's move it to the `utils.js`

src/ProgressBar/utils.js
```diff
@@ -3,24 +3,30 @@
  * that the annotation pointer indicates on zero of the ProgressBar.
  *
  * @param {object} ref reference to the info block
+ * @param {string} direction directing elements to pages
```
monteri, line 6:
```suggestion
 * @param {string} direction of the page ("ltr" or "rtl")
```

src/ProgressBar/utils.js
```diff
@@ -3,24 +3,30 @@
...
+export const placeInfoAtZero = (
+  ref,
+  direction = 'ltr',
+  annotationOnly = true,
+  annotationClass = 'pgn__annotation',
+) => {
   if (!ref.current || !ref.current.style) { return false; }
   const { children } = ref.current;
-  let marginLeft = 0.0;
+  let horizontalMargin = 0.0;
```
monteri, line 19: 👍

src/ProgressBar/utils.js
PKulkoRaccoonGang: Done, thanks

src/ProgressBar/index.jsx
PKulkoRaccoonGang: Yeah, good catch! 💯

src/ProgressBar/index.jsx
viktorrusakov: Honestly, I'm not sure if it would be useful. Usually the page direction is set on the `body` or `html` tags so that the value is propagated to all elements on the page and there is no need to check specific element. Same here - why don't you just do
```javascript
const direction = window.getComputedStyle(document.body).getPropertyValue('direction');
```
this would allow you to remove this `useEffect` and state management.

I don't think there ever will be a case where `body`'s direction is different from this component's.

PKulkoRaccoonGang: @viktorrusakov In the DOM structure of the Paragon documentation site, we always have `dir` at the `body` tag. I checked some MFEs and noticed that the tag for indicating the direction might be different, for example in the [Auth MFE](https://app-studio-mfe-dev.raccoongang.com/authn/login?next=https%3A%2F%2Fapp-studio-mfe-dev.raccoongang.com%2Fcourse-authoring%2Fcourse%2Fcourse-v1%3AedX%2BDemoX%2BDemo_Course%2Fsettings%2Fadvanced) it is specified for `html` tag. It seems that if, in the case of the [ProgressBar component](https://paragon-openedx.netlify.app/components/progressbar/#annotated-variant), we get dir from the `body` tag, we run the risk of getting `undefined` in some of the MFEs.
After discussing with @adamstankiewicz the option of defining this functionality in the form of a custom React hook, we decided to implement it and provide css-in-JS users with an alternative to `[dir="rtl"]` from CSS.

viktorrusakov: Why do you think you would get `undefined`? If `dir` is specified in `html` tag that just means that all of the elements on the page will inherit this value, even `body`. I just tested it locally with our docs site and my solution works even with setting `dir` to `html` tag.

I think it's okay to provide a custom hook, but it looks like an overkill for this component. It just adds unnecessary state logic and complexity.

PKulkoRaccoonGang: @viktorrusakov Ok, I agree with you
Thanks 👍

adamstankiewicz (approved):
LGTM! Thanks for addressing this RTL issue 😄

openedx-webhooks:
@PKulkoRaccoonGang 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/2641&entry.1484655318=https://github.com/PKulkoRaccoonGang&entry.752974735=2023-09-21+13:23&entry.1917517419=2023-09-29+12:26&entry.2133058324=Yes) so we can improve your experience in the future.

PR #2686, opened 2023-10-04
feat: add twitter X icon to list of approved Bootstrap brandIconNames

kalebjdavenport:
## Description

This PR adds the new Twitter "X" icon which is the updated branding.

Edit (@brian-smith-tcril):
* Rebased on latest
* Updated `bootstrap-icons` package to `1.11.3` in the icons dir `package.json`
* Updated `copy-brand-icons` to look in `./node_modules` (the icons directory node modules) instead of `../`
* Ran build in icons dir (it seems `1.11.3` made some minor changes to the icons. this is why this PR now has 198 files changed)

https://deploy-preview-2686--paragon-openedx.netlify.app/foundations/brand-icons/

#### Before (`1.10.5`)
image

#### After (`1.11.3`)
image

## Merge Checklist


icons/copy-brand-icons.js
```diff
@@ -5,7 +5,7 @@ const fs = require('fs');
 */
 
 const brandIconNames = [
-  'github', 'google', 'linkedin', 'slack', 'instagram', 'facebook', 'twitter', 'twitch', 'youtube', 'discord',
+  'github', 'google', 'linkedin', 'slack', 'instagram', 'facebook', 'twitter', 'twitter-x', 'twitch', 'youtube', 'discord',
```
adamstankiewicz, line 8: In addition to adding `twitter-x` here, I believe you'll also need to run the icons build process locally to ensure the SVG is copied over (i.e., run `npm run build` in the `icons` directory) to generate the associated JSX/JS files as well. Without doing so, the icon won't exist on the docs site (e.g., it doesn't appear on the deploy preview for this PR) or exposed to consumers of the `@edx/paragon` library.

Docs: https://github.com/openedx/paragon/tree/master/icons#paragon-icons

kalebjdavenport: Will do!

kalebjdavenport: Okay I ran `npm install` and `npm run build` in the icons directory. I'm not seeing any JSX directly related to twitter-x so if I should do some more research and figure out any steps I missed let me know!

viktorrusakov: I think the problem here might be that `twitter-x` icon is new and we don't have the latest version of `bootstrap-icons` setup in the project, currently the latest version is `v1.11.1` but we have `v1.10.5`, maybe it was only added somewhere in between these releases? Can you try upgrading to `v1.11.1` and running `npm run build` again?

Also, I believe for this you have to run `npm install` in the root directory of the project, not in the icons one, because we [look for bootstrap icons in the root node modules](https://github.com/openedx/paragon/blob/master/icons/copy-brand-icons.js#L17) (that's just a by-product of using npm workspaces 🙂)

PKulkoRaccoonGang (approved):
Perhaps in the future, we should consider adding an icon for [Threads](https://seeklogo.com/vector-logo/490178/threads) 😀

icons/jsx/index.jsx
```diff
@@ -472,7 +472,6 @@ export { default as Check } from "./Check";
 export { default as CheckBoxIcon } from "./CheckBoxIcon";
 export { default as CheckBoxOutlineBlank } from "./CheckBoxOutlineBlank";
 export { default as CheckCircle } from "./CheckCircle";
-export { default as CheckCircleLightOutline } from "./CheckCircleLightOutline";
```
brian-smith-tcril, line 475: I think we should get https://github.com/openedx/paragon/pull/2869 merged in first to avoid losing this icon

brian-smith-tcril: @kalebjdavenport #2869 has been merged - would you mind rebasing this PR and ensuring `CheckCircleLightOutline` isn't removed?

adamstankiewicz (approved):

Issue #2645, opened 2023-09-22
`Chip` implementation updates


adamstankiewicz:
The `Chip` was previously updated to reflect the chip-like UI within the spec'd `Multiselect` component ([WIP](https://www.figma.com/file/eGmDp94FlqEr4iSqy1Uc1K/Paragon-Design-System?type=design&node-id=9076-2632&mode=design&t=Vt7bsBprO6xU1Pcb-4)). However, this decision had the unfortunate consequence of changing the `Chip` to something we didn't actually want it to be for majority of existing use cases.

Now, design has some capacity to go back to the drawing board for `Chip` to ensure its design spec reflects the intent we have for the `Chip` component moving forward. The `Chip` used in the `Multiselect` component's spec remains unchanged, but the intent is that `Multiselect` will adopt the pattern spec'd for `Chip` as is in the future.

The spec updates for `Chip` lives here: https://github.com/openedx/paragon/issues/2645

Dismissable `Chip` in this work remains largely deferred from a design perspective, but we will still need to support it for existing use cases by some product teams. The `Chip` component accepts `onIconBeforeClick` and `onIconAfterClick` props to allow consumers to interact with clicks on the icon.

Based on the WIP dismissable `Chip` specs in the above Figma spec, I would recommend updating to conditionally use an `IconButton` for the icon passed to `iconAfter` and/or `iconBefore` when the associated `onIconBeforeClick` and/or `onIconAfterClick` is passed.

```[tasklist]
### Tasks
- [ ] Implement `Chip` based on the Figma spec linked above.
- [ ] There should be some way for consumers to persist a `Chip` as "selected".
- [ ] Consider making the icons rendered within `IconButton` when `onIconBeforeClick` and/or `onIconAfterClick` are passed.
- [ ] On docs site today, it looks like `onIconBeforeClick` may not be doing anything. Verify its behavior is working as intended.
```

PKulkoRaccoonGang:
PR: https://github.com/openedx/paragon/pull/2836

brian-smith-tcril:
I think this has been addressed by #2836. If not we can reopen.

PR #2713, opened 2023-10-11
[OLD] feat: Chip component redesign

PKulkoRaccoonGang:
## Description

- Chip component redesign.

**Issue:** https://github.com/openedx/paragon/issues/2645

### Deploy Preview

[Chip component](https://deploy-preview-2713--paragon-openedx.netlify.app/components/chip/)

## Merge Checklist

## Post-merge Checklist


openedx-webhooks:


src/Chip/index.tsx
```diff
@@ -32,46 +36,73 @@ const Chip = React.forwardRef(({
...
-));
+      {iconAfter && (
+        <div className={classNames('pgn__chip__icon-after', { active: onIconAfterClick })}>
+          {onIconAfterClick ? (
+            <IconButton
+              onClick={onIconAfterClick}
+              onKeyPress={onIconAfterClick}
+              src={iconAfter}
+              iconAs={Icon}
+              alt="Chip icon after"
```
monteri, line 94: Should we wrap with i18n?

src/Chip/index.scss
```diff
@@ -1,91 +1,206 @@
...
+        border: 2px solid $chip-light-selected-focus-border-color;
+        border-radius: 50%;
+      }
+    }
+
+    &:focus {
+      @include chip-outline(
+        $chip-light-selected-focus-border-color,
+        calc($chip-light-focus-outline-distance * -1),
+        calc($chip-border-radius + .188rem)
```
monteri, line 128: Move to `px` and to variable

src/Chip/index.tsx
```diff
@@ -32,46 +36,73 @@ const Chip = React.forwardRef(({
...
-  >
-    {iconBefore && (
-      <div
-        className={classNames('pgn__chip__icon-before')}
-      >
-        <Icon src={iconBefore} />
-      </div>
-    )}
+}: IChip, ref: ForwardedRef<HTMLDivElement>) => {
+  const [isSelected, setIsSelected] = useState(false);
```
monteri, line 40: Let's add `isSelected` prop instead of internal state

monteri:
There is one issue: for the disabled Chip you can highlight the text.

src/Chip/index.scss
PKulkoRaccoonGang: Ok, thanks

src/Chip/index.tsx
PKulkoRaccoonGang: Yeah, it`s good idea

src/Chip/index.tsx
PKulkoRaccoonGang: Done

src/Chip/mixins.scss
```diff
@@ -0,0 +1,12 @@
+@mixin chip-outline($ring-color: $white, $distance-to-border: 0, $border-radius: 50%, $border-width: .125rem) {
```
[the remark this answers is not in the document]
PKulkoRaccoonGang: Corrected

src/Chip/README.md
```diff
@@ -16,34 +16,82 @@ notes: |
...
 
 ```jsx live
-<div>
-  <Chip>New</Chip>
-  <Chip disabled>New</Chip>
-  <Chip variant="dark">New</Chip>
-</div>
+<Stack
+    gap={2}
+    direction="horizontal"
```
viktorrusakov, line 21: We usually indent JS code by 2 spaces in this repository, can you please do that in this file as well for consistency?

src/Chip/index.tsx
```diff
@@ -3,51 +3,74 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 // @ts-ignore
 import Icon from '../Icon';
+// @ts-ignore
+import IconButton from '../IconButton';
+
+export const ICON_AFTER_ALT_TEXT = 'Chip icon after';
```
viktorrusakov, line 9: I don't think it's a good idea have default alt text here, it heavily depends on the context and I don't think there is a reasonable default. We should make alt text props required only when consumers user icons with `Chip`, otherwise default to `undefined` (if no icons are used, then we do not need alt text at all).

We can achieve this behaviour with [requiredWhen proptype](https://github.com/openedx/paragon/blob/master/src/utils/propTypes/utils.js#L34), e.g.
```js
Chip.propTypes = {
  ...
  iconBeforeAlt: requiredWhen(PropTypes.string, 'iconBefore')
}

Chip.defaultProps = {
  iconBeforeAlt: undefined,
};
```

src/SelectableBox/tests/SelectableBoxSet.test.jsx
```diff
@@ -68,10 +68,13 @@ describe('<SelectableBox.Set />', () => {
       expect(screen.getByTestId('checkbox-set')).toBeInTheDocument();
     });
     it('renders with radio type if neither checkbox nor radio is passed', () => {
+      // eslint-disable-next-line no-console
```
viktorrusakov, line 71: Maybe just disable this rule for the whole file by adding `// eslint-disable no-console` at the top of the file instead of repeating this line multiple times?

src/Chip/Chip.test.jsx
```diff
@@ -77,5 +77,37 @@ describe('<Chip />', () => {
...
       expect(func).toHaveBeenCalled();
     });
+    it('onIconBeforeClick is triggered', async () => {
+      const func = jest.fn();
+      render(
+        <TestChip iconBefore={Close} onIconBeforeClick={func} />,
+      );
+      const iconBefore = screen.getByTestId('icon-before');
+      await userEvent.click(iconBefore);
+      expect(func).toHaveBeenCalled();
```
viktorrusakov, line 87: Can we avoid using test ids here? You can find these buttons by alt text for example, which should be a required prop now that's missing in your tests 🙂

src/Chip/Chip.test.jsx
```diff
@@ -77,5 +77,37 @@ describe('<Chip />', () => {
...
+      render(
+        <TestChip iconBefore={Close} onIconBeforeClick={func} />,
+      );
+      const iconBefore = screen.getByTestId('icon-before');
+      await userEvent.type(iconBefore, '{enter}');
+      expect(func).toHaveBeenCalled();
+    });
+    it('checks the absence of the `selected` class in the chip', async () => {
+      render(
+        <TestChip data-testid="chip" />,
```
viktorrusakov, line 100: Here also can we avoid using `data-testid`? I think `Chip` has `role="button"` set, can you try finding it by its role maybe?

PKulkoRaccoonGang: Yes, good idea

src/Chip/index.tsx
PKulkoRaccoonGang: Corrected

src/Chip/README.md
PKulkoRaccoonGang: Good catch, thanks 👍

src/SelectableBox/tests/SelectableBoxSet.test.jsx
PKulkoRaccoonGang: Done

src/Chip/ChipIcon.tsx
```diff
@@ -0,0 +1,50 @@
...
+        onClick={onClick}
+        onKeyPress={onClick}
+        iconAs={Icon}
+        alt={alt}
+        invertColors={variant === STYLE_VARIANTS[1]}
+      />
+    );
+  }
+
+  return <Icon src={src} className={className} />;
```
adamstankiewicz, line 34: [curious] I'm wondering if we should be using the `size="sm"` attribute for the `Chip` icons so it more closely matches the small font size used inside the `Chip`? 

image

src/Chip/ChipIcon.tsx
```diff
@@ -0,0 +1,50 @@
...
+}
+
+ChipIcon.propTypes = {
+  className: PropTypes.string.isRequired,
+  src: PropTypes.oneOfType([PropTypes.element, PropTypes.func]).isRequired,
```
adamstankiewicz, line 39: [curious] Could this prop type be reduced to `PropTypes.elementType`?

> // A React element type (eg. MyComponent).
// a function, string, or "element-like" object (eg. React.Fragment, Suspense, etc.)
// see https://github.com/facebook/react/blob/master/packages/shared/isValidElementType.js

src/Chip/index.tsx
```diff
@@ -2,73 +2,79 @@ import React, { ForwardedRef, KeyboardEventHandler, MouseEventHandler } from 're
...
+  iconAfterAlt,
   onIconBeforeClick,
   onIconAfterClick,
   disabled,
+  isSelected,
   ...props
 }: IChip, ref: ForwardedRef<HTMLDivElement>) => (
   <div
+    tabIndex={0}
+    role="button"
```
adamstankiewicz, line 43: If the left/right icons included in the `Chip` are interactive, I believe the intent is to _not_ have the top-level `Chip` be interactive as a button, where _only_ the icon(s) are interactive. This prevents possible usability issues where user is trying to dismiss the `Chip` but accidentally clicks on the parent `Chip` container instead of the `IconButton`.

Similarly, if the icon(s) are interactive, the parent `Chip` should remain unchanged from a hover styles perspective. The only hover focus in this case should be on the `IconButton` itself.

src/SelectableBox/tests/SelectableBoxSet.test.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
adamstankiewicz, line 1: nit: is this eslint disable line still relevant?

adamstankiewicz (commented):
This is looking great! Left a couple comments/suggestions.

src/SelectableBox/tests/SelectableBoxSet.test.jsx
viktorrusakov: Not after rebasing on master, thanks 🙂

src/Chip/ChipIcon.tsx
viktorrusakov: I don't think so, `elementType` will also allow you to pass strings like `'div'` which we do not want here. We usually use `elementType` type for  the `as` prop that allows you to specify base element for the component, but here I believe we only want actual React components, e.g. our icons.

src/Chip/index.tsx
viktorrusakov: Great points! I've refactored some stuff here and updated docs and I believe I achieved the behaviour you described, so please take a look again and let me know what you think 🙂

src/Chip/README.md
```diff
@@ -16,34 +16,139 @@ notes: |
...
+    iconAfterAlt="icon-after"
+    iconBeforeAlt="icon-before"
+    disabled
+  >
+    Both
+  </Chip>
+</Stack>
+```
+
+**Note**, however, that both `Chip` and its icons cannot be made interactive at the same time, e.g. if you provide both `onClick` and `onIconAfterClick` props,
```
adamstankiewicz, line 113: super tiny nit: if we keep the "Note" bold, perhaps change the comma to a colon, e.g. "**Note:** both `Chip` and its icons cannot be made interactive at the same time."

src/Chip/README.md
```diff
@@ -16,34 +16,139 @@ notes: |
...
+    iconAfterAlt="icon-after"
+    iconBeforeAlt="icon-before"
   >
-    New
+    Both
+  </Chip>
+</Stack>
+```
+
+### Clickable variant
```
adamstankiewicz, line 66: nit: perhaps mention "Clickable icon variant" to further clarify that these clickable variant examples are about clickable _icons_ (versus the identically named heading "Clickable Variant" above above full `Chip` being clickable).

adamstankiewicz (commented):
This is looking great; I added one suggestion related to `alt` text props and a couple minor nits. Even though none of the existing `Chip` props seem to have changed, I'm thinking this may technically need to be released as a breaking change due to the change in some of the SCSS theme variables that theme authors _may_ have been relying on. 

Given this `Chip` breaking change and the other breaking change we're considering (`Pagination`), perhaps we consider creating a single major version release that contains the breaking changes of both components to avoid back-to-back major version releases.

src/Chip/README.md
```diff
@@ -16,34 +16,139 @@ notes: |
...
+
+```jsx live
+<Chip isSelected>New</Chip>
 ```
 
 ## With Icon Before and After
+### Basic Usage
+
+Use `iconBefore` and `iconAfter` props to provide icons for the `Chip`, note that you also have to provide
+accessible names for these icons for screen reader support via `iconBeforeAlt` and `iconAfterAlt` respectively.
```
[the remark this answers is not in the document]
adamstankiewicz: @PKulkoRaccoonGang As is, I believe the alt text for the icons is always required whenever an icon is passed. I would recommend modifying the conditionally required alt text based not only on whenever an icon is passed, but also if its associated click handler is passed as well. E.g., perhaps something along the lines of:

```jsx
Chip.propTypes = {
  iconBeforeAlt: requiredWhen(PropTypes.string, ['iconBefore', `onIconBeforeClick`]),
};
```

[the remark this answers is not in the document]
viktorrusakov: @PKulkoRaccoonGang I think `requiredWhen` accepts only strings as a second argument and it does not know how to work with lists, so right now this prop validation does not work, can you please check this? (and modify `requiredWhen` if necessary)

Also, we probably need to change `have to` to `can` in the description of this example, because it's not always required to pass alt text anymore.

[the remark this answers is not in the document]
PKulkoRaccoonGang: Thanks, corrected

openedx-webhooks:
@PKulkoRaccoonGang Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/2713&entry.1484655318=https://github.com/PKulkoRaccoonGang&entry.752974735=2023-10-11+16:27&entry.1917517419=2023-11-21+17:36&entry.2133058324=No) so we can improve your experience in the future.

PR #2744, opened 2023-10-23
fix: Update docs for a final publish before we move this package.

feanil:
This is a part of https://github.com/openedx/axim-engineering/issues/23

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,12 @@
 # Paragon
 
+## Warning: Package Moved
+
+While the Paragon design system will continue to receive updates, these updates
+will no longer be published at the @edx/paragon NPM package. Instead, the
+latest versions can be found with
+[@openedx/paragon](https://www.npmjs.com/package/@openedx/paragon) instead.
+
 [![Build Status](https://github.com/openedx/paragon/actions/workflows/release.yml/badge.svg)](https://github.com/openedx/paragon/actions/workflows/release.yml)
 [![npm_version](https://img.shields.io/npm/v/@edx/paragon.svg)](@edx/paragon)
 ![status](https://img.shields.io/badge/status-Maintained-brightgreen)


README.md
```diff
@@ -1,5 +1,10 @@
 # Paragon
 
+## Warning: Package Moved
+
+This package will no longer receive updates and the latest version can be found
```
adamstankiewicz, line 5: [suggestion] At first glance, the way I read this was that Paragon itself will no longer be receiving updates, not just that older `@edx/paragon` NPM package won't be receiving further updates. 

Perhaps making "This package" more explicit by mentioning the old package name `@edx/paragon` instead as well as explicitly mentioning `@openedx/paragon` instead of just the URL to the package (i.e., non-zero effort to parse the package name from the URL):

> While the Paragon design system will continue to receive updates, these updates will no longer be published at the `@edx/paragon` NPM package. Instead, the latest versions can be found with [`@openedx/paragon`](https://www.npmjs.com/package/@openedx/paragon) instead.

README.md
```diff
@@ -1,5 +1,10 @@
...
+at https://www.npmjs.com/package/@openedx/paragon
```
adamstankiewicz, line 6: [clarification] Visiting this link brings me to a login screen. I don't think we should merge this until the NPM url for `@openedx/paragon` is broadly available to unauthenticated users.

feanil: @adamstankiewicz the URL breaks because we haven't published it to the new scope yet, that will be in a follow up PR that will land right after this one so the time when this URL is unavailable will be a few minutes at most.

README.md
brian-smith-tcril: I like that suggested rewording!

feanil: Yea, me too, I'll update it!

README.md
adamstankiewicz: Thanks for the added context!

adamstankiewicz (approved):

PR #2792, opened 2023-11-09
feat: Better support for multiple themes

xitij2000:
This change adds support for two CLI options to the build-tokens command.

The first, --all-themes makes the build-tokens command process all themes in the source directory as opposed to specifying all the names via --theme.

Secondly, --base-theme allows you to have multiple themes derived from the same common base. If specified, you can have a derived theme that still loads tokens from the light theme. For example, you can have a site theme called 'theme-one' and set the base theme to light so it will reuse the core light theme tokens and layer on changes from 'theme-one'.

**Testing instructions:**

- Create a base tokens directory structure from which tokens will be loaded, this will involve creating an empty directory with a folder called 'themes' inside which there will be multiple folders, with names other than 'light' i.e. test1, test2. So you have folders called 'test1' and 'test2' inside '/tmp/pr-2792/themes/' for example. 
- Just for test purposes create some tokens in both directories. You can just create a file called `colors.json' with the following contents: 
  ```json
  {
      "color": {
          "test-color": {
              "$value": "#112233"
          }
      }
  }
  ```
  You can use different color values or variable names to see if things are really working as expected. 
- You can now build tokens using `./bin/paragon-scripts.js build-tokens --source /tmp/pr-2792 --build-dir /tmp/pr-2792/build --themes test1 --themes test2`. You will need to specify both themes since currently, if you specify only one it will fail. 
- You should see `build` directory inside `/tmp/pr-2792` that a `core` an `themes` directory where the themes directory in turn will have a `variables.css` file with the variable defined above, and `utility-classes.css` will be empty. Delete this directory. 
- Check out this branch and try the following:
  * `./bin/paragon-scripts.js build-tokens --source /tmp/pr-2792 --build-dir /tmp/pr-2792/build --themes test1`: This would error before but will work now. 
  * `./bin/paragon-scripts.js build-tokens --source /tmp/pr-2792 --build-dir /tmp/pr-2792/build --all-themes`: This will build all themes without specifying a name.
  * `./bin/paragon-scripts.js build-tokens --source /tmp/pr-2792 --build-dir /tmp/pr-2792/build --themes test1 --base-theme light`: This will compile the test1 theme using the light theme as a base. So `utility-classes.css` will not longer be empty but contain styles from the light theme.

### DIFF
--- a/bin/paragon-scripts.js
+++ b/bin/paragon-scripts.js
@@ -107,6 +107,16 @@ const COMMANDS = {
         description: 'Enable verbose logging.',
         defaultValue: false,
       },
+      {
+        name: '--base-theme',
+        description: 'Specify the base theme to use in the token build. For example, to build the "high-contrast" theme on top of the light theme use "--theme high-contrast --base-theme light".',
+        defaultValue: 'Same as theme',
+      },
+      {
+        name: '--all-themes',
+        description: 'Build tokens for all themes in the source directory.',
+        defaultValue: false,
+      },
     ],
   },
   'replace-variables': {
--- a/bin/paragon-scripts.js
+++ b/bin/paragon-scripts.js
@@ -94,7 +94,8 @@ const COMMANDS = {
       {
         name: '-t, --themes',
         description: `Specify themes to include in the token build.
-          Can be provided as a comma-separated list (e.g., "light,dark") or multiple arguments (e.g., "-t light -t dark").`,
+          Can be provided as a comma-separated list (e.g., "light,dark") or multiple arguments (e.g., "-t light -t dark").
+          Cannot be used with --all-themes`,
         defaultValue: 'light',
       },
       {
@@ -114,7 +115,7 @@ const COMMANDS = {
       },
       {
         name: '--all-themes',
-        description: 'Build tokens for all themes in the source directory.',
+        description: 'Build tokens for all themes in the source directory. Cannot be used with --themes.',
         defaultValue: false,
       },
     ],
--- a/bin/paragon-scripts.js
+++ b/bin/paragon-scripts.js
@@ -108,8 +108,8 @@ const COMMANDS = {
         defaultValue: false,
       },
       {
-        name: '--base-theme',
-        description: 'Specify the base theme to use in the token build. For example, to build the "high-contrast" theme on top of the light theme use "--theme high-contrast --base-theme light".',
+        name: '--base-paragon-theme',
+        description: 'Specify the base theme to use in the token build. For example, to build the "high-contrast" theme on top of the light theme use "--theme high-contrast --base-paragon-theme light".',
         defaultValue: 'Same as theme',
       },
       {
--- a/lib/build-tokens.js
+++ b/lib/build-tokens.js
@@ -14,7 +14,8 @@ const { createIndexCssFile } = require('../tokens/utils');
  * @param {string[]} commandArgs - Command line arguments for building tokens.
  * @param {string} [commandArgs.build-dir='./build/'] - The directory where the build output will be placed.
  * @param {string} [commandArgs.source] - The source directory containing JSON token files.
- * @param {string} [commandArgs.base-theme] - The base theme to use from Paragon if named differently than the theme.
+ * @param {string} [commandArgs.base-paragon-theme] - The base theme to use from Paragon if named differently than
+ *                                                    the theme.
  * @param {boolean} [commandArgs.source-tokens-only=false] - Indicates whether to include only source tokens.
  * @param {string|string[]} [commandArgs.themes=['light']] - The themes (variants) for which to build tokens.
  * @param {boolean} [commandArgs.all-themes] - Indicated whether to process all themes.
@@ -22,7 +23,7 @@ const { createIndexCssFile } = require('../tokens/utils');
 async function buildTokensCommand(commandArgs) {
   const defaultParams = {
     themes: 'light',
-    'base-theme': null,
+    'base-paragon-theme': null,
     'build-dir': './build/',
     'source-tokens-only': false,
     'output-references': true,
@@ -44,7 +45,7 @@ async function buildTokensCommand(commandArgs) {
     'output-references': outputReferences,
     themes,
     verbose,
-    'base-theme': baseTheme,
+    'base-paragon-theme': baseParagonTheme,
     'all-themes': allThemes,
     'exclude-core': excludeCore,
   } = minimist(
@@ -177,10 +178,10 @@ async function buildTokensCommand(commandArgs) {
   }
 
   //  Add theme variants
-  for (const themeVariant of parsedThemes) {
-    const config = getStyleDictionaryConfig(themeVariant, baseTheme || themeVariant);
+  themesToProcess.forEach(themeVariant => {
+    const config = getStyleDictionaryConfig(themeVariant, baseParagonTheme || themeVariant);
     configs.push({ config, themeVariant });
-  }
+  });
 
   // Build tokens for each configuration
   await Promise.all(configs.map(async ({ config, themeVariant }) => {
--- a/lib/build-tokens.js
+++ b/lib/build-tokens.js
@@ -1,3 +1,4 @@
+const fs = require('fs');
 const path = require('path');
 const minimist = require('minimist');
 const {
@@ -13,17 +14,21 @@ const { createIndexCssFile } = require('../tokens/utils');
  * @param {string[]} commandArgs - Command line arguments for building tokens.
  * @param {string} [commandArgs.build-dir='./build/'] - The directory where the build output will be placed.
  * @param {string} [commandArgs.source] - The source directory containing JSON token files.
+ * @param {string} [commandArgs.base-theme] - The base theme to use from Paragon if named differently than the theme.
  * @param {boolean} [commandArgs.source-tokens-only=false] - Indicates whether to include only source tokens.
  * @param {string|string[]} [commandArgs.themes=['light']] - The themes (variants) for which to build tokens.
+ * @param {boolean} [commandArgs.all-themes] - Indicated whether to process all themes.
  */
 async function buildTokensCommand(commandArgs) {
   const defaultParams = {
-    themes: ['light'],
+    themes: 'light',
+    'base-theme': null,
     'build-dir': './build/',
     'source-tokens-only': false,
     'output-references': true,
     'exclude-core': false,
     verbose: false,
+    'all-themes': false,
   };
 
   const alias = {
@@ -39,19 +44,32 @@ async function buildTokensCommand(commandArgs) {
     'output-references': outputReferences,
     themes,
     verbose,
+    'base-theme': baseTheme,
+    'all-themes': allThemes,
     'exclude-core': excludeCore,
   } = minimist(
     commandArgs,
     {
       alias,
       default: defaultParams,
-      boolean: ['source-tokens-only', 'output-references', 'exclude-core', 'verbose'],
+      boolean: ['source-tokens-only', 'output-references', 'exclude-core', 'verbose', 'all-themes'],
     },
   );
+  let themesToProcess = null;
 
-  const parsedThemes = Array.isArray(themes) ? themes : themes.split(',').map(t => t.trim());
+  if (allThemes) {
+    const tokensPath = tokensSource || path.resolve(__dirname, '../tokens/src');
+    themesToProcess = fs
+      .readdirSync(`${tokensPath}/themes/`, { withFileTypes: true })
+      .filter(entry => entry.isDirectory())
+      .map(entry => entry.name);
+  } else if (Array.isArray(themes)) {
+    themesToProcess = themes;
+  } else {
+    themesToProcess = themes.split(',').map(t => t.trim());
+  }
 
-  const StyleDictionary = await initializeStyleDictionary({ themes: parsedThemes });
+  const StyleDictionary = await initializeStyleDictionary({ themes: themesToProcess });
 
   const coreConfig = {
     include: [
@@ -100,12 +118,12 @@ async function buildTokensCommand(commandArgs) {
     },
   };
 
-  const getStyleDictionaryConfig = (themeVariant) => ({
+  const getStyleDictionaryConfig = (themeVariant, baseThemeVariant) => ({
     ...coreConfig,
     include: [
       ...coreConfig.include,
-      path.resolve(__dirname, `../tokens/src/themes/${themeVariant}/**/*.json`),
-      path.resolve(__dirname, `../tokens/src/themes/${themeVariant}/**/*.toml`),
+      path.resolve(__dirname, `../tokens/src/themes/${baseThemeVariant}/**/*.json`),
+      path.resolve(__dirname, `../tokens/src/themes/${baseThemeVariant}/**/*.toml`),
     ],
     source: tokensSource
       ? [
@@ -160,7 +178,7 @@ async function buildTokensCommand(commandArgs) {
 
   //  Add theme variants
   for (const themeVariant of parsedThemes) {
-    const config = getStyleDictionaryConfig(themeVariant);
+    const config = getStyleDictionaryConfig(themeVariant, baseTheme || themeVariant);
     configs.push({ config, themeVariant });
   }
 
--- a/lib/build-tokens.js
+++ b/lib/build-tokens.js
@@ -22,7 +22,7 @@ const { createIndexCssFile } = require('../tokens/utils');
  */
 async function buildTokensCommand(commandArgs) {
   const defaultParams = {
-    themes: 'light',
+    themes: null,
     'base-paragon-theme': null,
     'build-dir': './build/',
     'source-tokens-only': false,
@@ -56,6 +56,10 @@ async function buildTokensCommand(commandArgs) {
       boolean: ['source-tokens-only', 'output-references', 'exclude-core', 'verbose', 'all-themes'],
     },
   );
+
+  if (themes !== null && allThemes) {
+    throw Error('Cannot specify themes with `--themes` when using `--all-themes`.');
+  }
   let themesToProcess = null;
 
   if (allThemes) {
@@ -67,7 +71,7 @@ async function buildTokensCommand(commandArgs) {
   } else if (Array.isArray(themes)) {
     themesToProcess = themes;
   } else {
-    themesToProcess = themes.split(',').map(t => t.trim());
+    themesToProcess = (themes || 'light').split(',').map(t => t.trim());
   }
 
   const StyleDictionary = await initializeStyleDictionary({ themes: themesToProcess });


openedx-webhooks:
Thanks for the pull request, @xitij2000!


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

mphilbrick211:
Hi @adamstankiewicz! Would you mind reviewing for us? Thanks!

lib/build-tokens.js
```diff
@@ -28,7 +33,20 @@ async function buildTokensCommand(commandArgs) {
...
-  } = minimist(commandArgs, { alias, default: defaultParams, boolean: 'source-tokens-only' });
+    'base-theme': baseTheme,
+    'all-themes': allThemes,
+  } = minimist(commandArgs, { alias, default: defaultParams, boolean: ['source-tokens-only', 'all-themes'] });
+
+  let themesToProcess = null;
+
+  if (allThemes) {
+    themesToProcess = fs
+      .readdirSync(`${tokensSource}/themes/`, { withFileTypes: true })
```
viktorrusakov, line 44: This line fails for me locally when I triy to build all themes for Paragon, because `tokensSource` is `undefined` by default. I think this setting is used to overwrite default tokens which we don't need to do in Paragon, so we don't use this parameter internally in repository.

Changing this line to ``.readdirSync(`${tokensSource || path.resolve(__dirname, '../tokens/src')}}/themes/`, { withFileTypes: true })`` seems to work for me (basically defaulting to the root folder of Paragon's tokens)

lib/build-tokens.js
```diff
@@ -9,13 +10,17 @@ const { createIndexCssFile } = require('../tokens/utils');
...
  * @param {string} [commandArgs.source] - The source directory containing JSON token files.
+ * @param {string} [commandArgs.base-theme] - The base theme to use from Paragon if named differently than the theme.
  * @param {boolean} [commandArgs.source-tokens-only=false] - Indicates whether to include only source tokens.
  * @param {string|string[]} [commandArgs.themes=['light']] - The themes (variants) for which to build tokens.
+ * @param {boolean} [commandArgs.all-themes] - Indicated whether to process all themes.
  */
 async function buildTokensCommand(commandArgs) {
   const defaultParams = {
-    themes: ['light'],
+    themes: 'light',
```
mavidser, line 20: Any specific reason to switch from `themes` argument as a list, to a string and splitting it later in `:48` ?

If so, please also update the `@param` doc in `:15` to reflect the same

xitij2000: The reason is that currently specifying a single theme via `--theme` is simply broken. You will get an error if you run the command. 

The reason for that is that if you specify a single theme as `--theme light` the value of themes above will be the string "light" and so the `themes.forEach` later on will fail. You'll need to either go with the default value, or specify themes twice as:

`paragon build-tokens --themes light --themes dark`

This change makes it possible to specify multiple themes as:

`paragon build-tokens --themes "light  dark"`

This is also a bit clumsy, but is more consistent. I'll update the docstring and also make this more robust to support the old syntax.

lib/build-tokens.js
xitij2000: I've applied your suggestion. Do tell if there are any other changed you'd like me to make to get this merged.

mphilbrick211:
Hi @brian-smith-tcril! Are you able to review / merge this for us?

brian-smith-tcril (commented):
I'm not fully sure how all these new options interact with each other:

| `--themes` | `--base-theme` | `--all-themes` | expected behavior |
| - | - | - | - |
| not set | not set | false | ? |
| not set | not set | true | ? |
| not set | set | false | ? |
| not set | set | true | ? |
| single theme | not set | false | ? |
| single theme | not set | true | ? |
| single theme | set | false | ? |
| single theme | set | true | ? |
| multiple themes | not set | false | ? |
| multiple themes | not set | true | ? |
| multiple themes | set | false | ? |
| multiple themes | set | true | ? |

xitij2000:
Sure, I can clarify the interactions. 

First, the `--base-theme` option:

This option is to allow creating theme variants. Currently, you can create your own `light` theme, Paragon will load the `light` theme tokens from the Paragon package, and overlay the branding version on top. However, if you use any other theme name, for example, "theme-a", then Paragon will try to load the corresponding `theme-a` tokens from Paragon and that won't work. We'd like the ability to have theme variants. For instance, if a site wants to support multiple light themes or have multiple themes for different microsites. 

With the `--base-theme` option, you can tell paragon that all the themes being specified on the command line are to use that theme as the base theme. For instance, if you wanted to create a high-contrast version of the light theme, you could do:

`paragon build-tokens --source ./tokens --build-dir ./paragon/css --source-tokens-only --themes high-contrast-light --base-theme light`

Here it would pull the tokens from the `high-contrast-light` directory of the tokens folder but combine it with the `light` tokens from the Paragon package. This will apply to all themes supplied via this command line, whether via `--theme` or `--all-themes`. 

Second, the interaction between `themes` and `all-themes`:

If `--all-themes` is supplied, then it will iterate over all the theme directories in the tokens path and ignore the `themes` option, since it overrides that. If the `all-themes` option is absent and `themes` is provided, then it will only process the themes specified via the `themes` option.

mphilbrick211:
> Sure, I can clarify the interactions.
> 
> First, the `--base-theme` option:
> 
> This option is to allow creating theme variants. Currently, you can create your own `light` theme, Paragon will load the `light` theme tokens from the Paragon package, and overlay the branding version on top. However, if you use any other theme name, for example, "theme-a", then Paragon will try to load the corresponding `theme-a` tokens from Paragon and that won't work. We'd like the ability to have theme variants. For instance, if a site wants to support multiple light themes or have multiple themes for different microsites.
> 
> With the `--base-theme` option, you can tell paragon that all the themes being specified on the command line are to use that theme as the base theme. For instance, if you wanted to create a high-contrast version of the light theme, you could do:
> 
> `paragon build-tokens --source ./tokens --build-dir ./paragon/css --source-tokens-only --themes high-contrast-light --base-theme light`
> 
> Here it would pull the tokens from the `high-contrast-light` directory of the tokens folder but combine it with the `light` tokens from the Paragon package. This will apply to all themes supplied via this command line, whether via `--theme` or `--all-themes`.
> 
> Second, the interaction between `themes` and `all-themes`:
> 
> If `--all-themes` is supplied, then it will iterate over all the theme directories in the tokens path and ignore the `themes` option, since it overrides that. If the `all-themes` option is absent and `themes` is provided, then it will only process the themes specified via the `themes` option.

Flagging this, @brian-smith-tcril, just in case you're still able to review.

lib/build-tokens.js
PKulkoRaccoonGang: @xitij2000 It seems we still have problems with the `tokensSource`. Could you check if everything is working correctly?

PKulkoRaccoonGang:
@xitij2000 Thanks for your improvements! We need to solve the problem of specifying specific themes. I like the shorter declaration format you proposed and the flexibility we can provide to consumers.

Unfortunately, I was not able to test this locally (maybe I missed something).

xitij2000:
> @xitij2000 Thanks for your improvements! We need to solve the problem of specifying specific themes. I like the shorter declaration format you proposed and the flexibility we can provide to consumers.
> 
> Unfortunately, I was not able to test this locally (maybe I missed something). We may still have a problem with `tokensSource`.

Could you tell me what didn't work when testing? Were you getting an error?

I've added test instructions in the PR description above. Do tell me if those steps do not work.

mphilbrick211:
@xitij2000 @PKulkoRaccoonGang just following up on this :)

mphilbrick211:
> @xitij2000 @PKulkoRaccoonGang just following up on this :)

Friendly ping, @PKulkoRaccoonGang @xitij2000!

brian-smith-tcril:
@PKulkoRaccoonGang I see you mentioned

> I was not able to test this locally

and @xitij2000 reponded

> I've added test instructions in the PR description above. Do tell me if those steps do not work.

@PKulkoRaccoonGang - were you able to test locally after the steps were added?
 
Either way, I know quite a few things have changed in the `alpha` branch since this PR was opened, and this sounds like a feature we'd like to have, so it'd be great to verify this still works.

I know @PKulkoRaccoonGang is working to update `alpha` to include new commits from `master` at the moment, so rebasing this PR **after** that happens would probably be ideal.

peterkulko:
> rebasing this PR after that happens would probably be ideal

Yes, I think after the rebase we will be able to test this functionality and make sure that everything works as we expect 💯

xitij2000:
@PKulkoRaccoonGang I've rebased this on paragon 23 of late. Do tell me how i can help move this along.

dependent-usage.json
```diff
@@ -1,5 +1,5 @@
 {
-  "lastModified": 1733926086658,
+  "lastModified": 1734012486681,
```
peterkulko: Should these changes be here?

peterkulko:
@xitij2000 Sorry for the long review wait for this PR. I'm trying to test this locally but I get stuck at this step:

> You should see build directory inside /tmp/pr-2792 that a core an themes directory where the themes directory in turn will have a variables.css file with the variable defined above, and utility-classes.css will be empty. Delete this directory.

After running the command `./bin/paragon-scripts.js build-tokens --source themes --build-dir build --themes test1 --themes test2` I get the following logs:

```shell
css
- build/core/variables.css
- build/core/custom-media-breakpoints.css

css
! build/themes/test1/variables.css, does not exist
- build/themes/test1/utility-classes.css

css
! build/themes/test2/variables.css, does not exist
- build/themes/test2/utility-classes.css
- build/themes/test2

css
✔︎ build/core/variables.css
✔︎ build/core/custom-media-breakpoints.css

css
No tokens for themes/test1/variables.css. File not created.
✔︎ build/themes/test1/utility-classes.css

css
No tokens for themes/test2/variables.css. File not created.
✔︎ build/themes/test2/utility-classes.css
An error occurred: TypeError: StyleDictionary.extend is not a function
```

I created a separate PR (https://github.com/openedx/paragon/pull/3613) where you can see my settings.
As you can see, there are no `variables.css` files in the `build/themes/<theme_name>` directory.

Could this be related to the error `An error occurred: TypeError: StyleDictionary.extend is not a function`? Or am I doing something wrong?

dependent-usage.json
xitij2000: Removed! Don't know how they ended up here.

xitij2000:
@PKulkoRaccoonGang 

Thank you so much for persisting with this! 

> Could this be related to the error `An error occurred: TypeError: StyleDictionary.extend is not a function`? Or am I doing something wrong?

I found a couple of issues with my code since it wasn't updated to take into account the latest changes. I've fixed it now.

Two small issues with the test environment from that PR:

- The build commands looks inside the `themes`  folder inside the source folder, so you'd need to either use `--source .` or move the theme tokens into `themes/themes`. 
- I think since the latest version of style dictionary, it should be "$value" instead of "value".

lib/build-tokens.js
peterkulko: I'm testing locally, this problem doesn't seem relevant to me anymore. Closed

peterkulko:
> Two small issues with the test environment from that 

@xitij2000 Thanks! Could you please update the "Testing instructions"?

xitij2000:
> > Two small issues with the test environment from that
> 
> @xitij2000 Thanks! Could you please update the "Testing instructions"?

I think other than the value -> $value change the rest of the instruction should be the same. I've made the change. Please do have a look when you are able.

lib/build-tokens.js
peterkulko: I tried to check how the build is currently happening on the [release-23.x](https://github.com/openedx/paragon/tree/release-23.x) branch using the following scenarios:

- build multiple themes
Command: `./bin/paragon-scripts.js build-tokens --build-dir test --themes light,dark,theme3`
Result: three themes (`light`, `dark`, `theme3`) have been successfully added to the `test` directory.

- build one theme
Command: `./bin/paragon-scripts.js build-tokens --build-dir test --themes theme3`
Result: one theme (`theme3`) has been successfully expanded in the `test` directory.

> The reason for that is that if you specify a single theme as --theme light the value of themes above will be the string "light" and so the themes.forEach later on will fail

We get this result because the `build-tokens` command supports adding **one** or **more** custom themes via a single `--themes` option (a list of themes separated by commas, no spaces). It seems to me that the fact that when specifying `--theme my-custom-theme` the user gets a build of the `light` theme is expected, because the user did not correctly pass the parameter to the `build-tokens` command.

> This change makes it possible to specify multiple themes as:
> 
> paragon build-tokens --themes "light dark"
> 
> This is also a bit clumsy, but is more consistent.

I don't really understand the benefits of changing the format for adding multiple themes from `--themes light,dark` to `--themes "light dark"`. Maybe I'm missing something, please correct me.

bin/paragon-scripts.js
```diff
@@ -106,6 +106,16 @@ const COMMANDS = {
         description: 'Enable verbose logging.',
         defaultValue: false,
       },
+      {
+        name: '--base-theme',
+        description: 'Specify the base theme to use in the token build. For example, to build the "high-contrast" theme on top of the ligth theme use "--theme high-contrast --base-theme light".',
```
peterkulko:
```suggestion
        description: 'Specify the base theme to use in the token build. For example, to build the "high-contrast" theme on top of the light theme use "--theme high-contrast --base-theme light".',
```

peterkulko:
I've prepared a some table to help test this:

№ | --themes | --base-theme | --all-themes | expected behavior | build result
-- |-- | -- | -- | -- | --
1 | not set | not set | false | `core` + `light` (with filled `utility-classes.css`) | [Source](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-1)
2 | not set | not set | true | `core` + `theme-1`, `theme-2 `and `theme-3` (all custom themes with empty `utility-classes.css`) | [Source](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-2)
3 | not set | set | false | `core` + `light` (with filled `utility-classes.css`) | [Source](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-3)
4 | not set | set | true | `core` + `theme-1`, `theme-2` and `theme-3` (all custom themes with filled `utility-classes.css`) | [Source](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-4)
5 | single theme | not set | false | `core` + `theme-1` (with empty `utility-classes.css`) | [Source](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-5)
6 | single theme | not set | true | `core` + `theme-1`, `theme-2` and `theme-3` (all custom themes with empty `utility-classes.css`) | [Source](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-6)
7 | single theme | set | false | `core` + `theme-1` (with filled `utility-classes.css`) | [Source](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-7)
8 | single theme | set | true | `core` + `theme-1`, `theme-2` and `theme-3` (all custom themes with filled `utility-classes.css`) | [Source](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-8)
9 | multiple themes | not set | false | `core` + `theme-1`, `theme-2` (all custom themes with empty `utility-classes.css`) | [Source](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-9)
10 | multiple themes | not set | true | `core` + `theme-1`, `theme-2` and `theme-3` (all custom themes with empty `utility-classes.css`) | [Source](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-10)
11 | multiple themes | set | false | `core` + `theme-1`, `theme-2` (all custom themes with filled `utility-classes.css`) | [Source](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-11)
12 | multiple themes | set | true | `core` + `theme-1`, `theme-2` and `theme-3` (all custom themes with filled `utility-classes.css`) | [Source](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-12)

peterkulko:
> ./bin/paragon-scripts.js build-tokens --source /tmp/pr-2792 --build-dir /tmp/pr-2792/build --themes test1 --base-theme light: This will compile the test1 theme using the light theme as a base. So utility-classes.css will not longer be empty but contain styles from the light theme.

@xitij2000 should `theme1` that is built based on `theme2` include in the CSS build results the variables generated from theme2's design tokens + new or redefined design tokens? Currently I see [this result in test case 4](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-4/themes)

bin/paragon-scripts.js
```diff
@@ -106,6 +106,16 @@ const COMMANDS = {
...
+        defaultValue: 'Same as theme',
+      },
+      {
+        name: '--all-themes',
```
peterkulko, line 117: I like your idea of ​​adding the `--all-themes` option. This is much better for building all themes, instead of manually entering the name of each theme 💯

xitij2000:
> @xitij2000 should `theme1` that is built based on `theme2` include in the CSS build results the variables generated from theme2's design tokens + new or redefined design tokens? Currently I see [this result in test case 4](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-4/themes)

Hmm, this makes me realise that base-theme is misnamed for what it's doing (and for what we want it to do for our purposes). 

The idea is to have a theme called `theme-1` still be based on the core "light" theme of paragon. If you need to have a dozen themes that are all a mild variation of the light theme you use the light theme as the base theme since by default it'll look for a theme with the same name in paragon and that won't exist. 

However, base theme does also seem to imply that you can create a theme called theme-1 and one called theme-2 and have theme-2 be based on theme-1 which this currently can't do since it's only looking for the core tokens inside the paragon dir. 

I see two options here: 
1. Change the name of this flag from --base-theme to --core-theme / --paragon-theme / --base-paragon-theme etc. 
2. Add support for basing one user theme on another. However, this leads into another issue, how do you then have a base theme from paragon? i.e. Base theme-2 on theme-1 on light because if you can't do that then you'll need a lot of heavy-lifting in theme-1. 

I considered having some kind of metadata where theme-2 can declare that it's based on theme-1 and then theme-1 can declare that's it's based on 'light' however it seems there is a much simple solution: symlinks. If you want theme-2 based on theme-1, simply symlink theme-1 inside theme-2's folder and it will build theme-2 including the theme-1 tokens as well. 

@PKulkoRaccoonGang What do you propose I do here? Should I rename the flag? Or is it better to instead have some official mechanism for haveing a chain of theme dependencies (other than symlinks which will have downsides).

peterkulko:
@xitij2000 

> What do you propose I do here? Should I rename the flag? Or is it better to instead have some official mechanism for haveing a chain of theme dependencies (other than symlinks which will have downsides).

> Add support for basing one user theme on another. However, this leads into another issue, how do you then have a base theme from paragon? i.e. Base theme-2 on theme-1 on light because if you can't do that then you'll need a lot of heavy-lifting in theme-1.

I think it will be enough for us for now if we focus on creating custom themes based on the `light` theme. The idea of ​​creating themes based on one of the custom themes (which is based on `light`) is interesting and can be useful. I am ready to create a separate issue for this, I have a case where it may be needed.

> Change the name of this flag from --base-theme to --core-theme / --paragon-theme / --base-paragon-theme etc.

Yes, I think the functionality of `--base-theme` would be more obvious if this parameter were called `--base-paragon-theme`. This way we emphasize that this theme is base and belongs to Paragon 👍 

Some comments on my question I asked earlier

> @xitij2000 should theme1 that is built based on theme2 include in the CSS build results the variables generated from theme2's design tokens + new or redefined design tokens? Currently I see [this result in test case 4](https://github.com/openedx/paragon/tree/Peter_Kulko/testing-CLI-fixes/build/test-case-4/themes)

Sorry, in this question, when I said "theme2", I meant the `light` theme that we use as a base for creating custom themes that override it. Now I'm sure that the build result of each custom theme should not include the `light` theme variables, since custom themes are overrides.

But I still don't understand why when we create custom themes (for example, theme-1, theme-2, theme-3) we have a filled `utility-classes.css` file after the build?

For example, I use the following configuration for each microsite, where the Paragon theme `light` is taken as the basis and only override variables (+ custom variables if any) are added as overrides. For example:

```json
{
  "MFE_CONFIG": {
    "PARAGON_THEME_URLS": {
      "core": {
        "url": "https://cdn.jsdelivr.net/npm/@openedx/paragon@<version>/dist/core.min.css"
      },
      "defaults": {
        "light": "light"
      },
      "variants": {
        "light": {
          "urls": {
            "default": "https://cdn.jsdelivr.net/npm/@openedx/paragon@<version>/dist/light.min.css",
            "brandOverride": "<PATH_TO_MY_CUSTOM_THEME>/light/variables.css"
          }
        }
      }
    }
  }
}
```

How do you work with microsites? Do you use `utility-classes.css`?

xitij2000:
@PKulkoRaccoonGang 
I've updated the param to --base-paragon-theme.

> How do you work with microsites? Do you use `utility-classes.css`?

This is the output of build-tokens. After running build-scss it should produce a combined CSS files for each theme.

brian-smith-tcril:
@xitij2000 I see that @PKulkoRaccoonGang filled out the table I made (https://github.com/openedx/paragon/pull/2792#pullrequestreview-1968849443) a while ago (https://github.com/openedx/paragon/pull/2792#issuecomment-2974331300)

Is that accurate to your understanding?

xitij2000:
> @xitij2000 I see that @PKulkoRaccoonGang filled out the table I made ([#2792 (review)](https://github.com/openedx/paragon/pull/2792#pullrequestreview-1968849443)) a while ago ([#2792 (comment)](https://github.com/openedx/paragon/pull/2792#issuecomment-2974331300))
> 
> Is that accurate to your understanding?

Sorry for the delay, I've been a bit under the weather. Here is my simplified chart for how this command should work:

| themes | all-themes | Result |
|--------|--------|--------|
| EMPTY | EMPTY | The light theme will be built |
| EMPTY | True | All themes in source directory will be built |
| theme-name | EMPTY | Theme with name `theme-name` will be built |
| theme-name | True | All themes in source directory will be built |

In other words, `all-themes` takes precedence over explicitly specified themes.

| themes | base-paragon-theme | Result |
|--------|--------|--------|
| EMPTY | EMPTY | Light theme tokens from the user theme are built based on Paragon light theme. |
| One or more specific themes | EMPTY | Specified themes are built based on a Paragon theme of the same name. This might fail if Paragon doesn't have a theme with that name.
| All themes | EMPTY | All themes in the source dir are built based on a Paragon theme of the same name. This might fail as above. |
| EMPTY | some-theme | Light theme tokens from the user theme are built based on the Paragon theme called `some-theme`. This might fail if there is no theme called `some-theme` in Paragon. |
| One or more themes | light | Specified themes from user theme are built based on the Paragon light theme. | 

In other words, when the base Paragon theme is not provided, all specified themes (by default only the `ligth` theme) are built against a theme of the same name in Paragon. When base Paragon theme **is** provided, then all specified themes are built again the provided base Paragon theme.

brian-smith-tcril:
Sorry for all the back and forth on this one. Overall I think this is a great improvement and we should aim to land it!

 I want to make sure I understand how the behavior with this PR compares to the current behavior (to ensure it's a non-breaking change)

This PR adds:
* `--base-paragon-theme`
* `--all-themes`

My understanding of the behavior before this PR is that it'd be the equivalent of `--base-paragon-theme` being set to match each entry in `--themes` (so if you had `light` and `dark`, `light` would use `light` as the base, and `dark` would use `dark` as the base). It appears this is handled properly by
https://github.com/openedx/paragon/blob/6f21b5355b5ef6260bbe45351cd3f03cb8d5c93b/lib/build-tokens.js#L182

:white_check_mark: Non-breaking!

As for `--all-themes`, the logic in the table looks sound, and it does appear to be a non-breaking change.

I do think it'd be good to document/add error messaging to handle the case where a user tries to use both `--themes` and `--all-themes`.

It looks like with this PR `--all-themes` takes precedence

https://github.com/openedx/paragon/blob/6f21b5355b5ef6260bbe45351cd3f03cb8d5c93b/lib/build-tokens.js#L61

https://github.com/openedx/paragon/blob/6f21b5355b5ef6260bbe45351cd3f03cb8d5c93b/lib/build-tokens.js#L67

so communicating that to users would be perfect.

Once some error messaging and or help text is added this should be good to merge!

xitij2000:
> Sorry for all the back and forth on this one. Overall I think this is a great improvement and we should aim to land it!

No worries about the back-and-forth, I wouldn't want to break anyone's flow so I can understand it's important to get the behaviour right, especially since this is now actually in use :-)
 
> I want to make sure I understand how the behavior with this PR compares to the current behavior (to ensure it's a non-breaking change)
> 
> My understanding of the behavior before this PR is that it'd be the equivalent of `--base-paragon-theme` being set to match each entry in `--themes` (so if you had `light` and `dark`, `light` would use `light` as the base, and `dark` would use `dark` as the base). It appears this is handled properly by
> 
> https://github.com/openedx/paragon/blob/6f21b5355b5ef6260bbe45351cd3f03cb8d5c93b/lib/build-tokens.js#L182
> 
> ✅ Non-breaking!

I'm not sure I follow. If one doesn't specify `--base-paragon-theme` then there should be no change in behaviour. However, if you do specify a theme for `--base-paragon-theme` then it will use the specified theme as the base insted of deriving it from the entry in themes.
> 
> As for `--all-themes`, the logic in the table looks sound, and it does appear to be a non-breaking change.
> 
> I do think it'd be good to document/add error messaging to handle the case where a user tries to use both `--themes` and `--all-themes`.
> 
> It looks like with this PR `--all-themes` takes precedence

Yes it does, I don't know how else to resolve the behaviour, perhaps by letting `--themes` take precidence?  Or I can just raise an error since they are incompatible options.

xitij2000:
I've pushed a commit that improves the help, and also now makes specifying both themes and base theme an error condition.

brian-smith-tcril:
> > I want to make sure I understand how the behavior with this PR compares to the current behavior (to ensure it's a non-breaking change)
> > 
> > My understanding of the behavior before this PR is that it'd be the equivalent of `--base-paragon-theme` being set to match each entry in `--themes` (so if you had `light` and `dark`, `light` would use `light` as the base, and `dark` would use `dark` as the base). It appears this is handled properly by
> > 
> > https://github.com/openedx/paragon/blob/6f21b5355b5ef6260bbe45351cd3f03cb8d5c93b/lib/build-tokens.js#L182
> > 
> > ✅ Non-breaking!
> 
> I'm not sure I follow. If one doesn't specify `--base-paragon-theme` then there should be no change in behaviour. However, if you do specify a theme for `--base-paragon-theme` then it will use the specified theme as the base insted of deriving it from the entry in themes.

I think we're on the same page here. I'm sorry I didn't communicate my thoughts in the most clear manner. Hopefully this table does a better job!

| `--base-paragon-theme` | Behavior before this PR | Behavior after this PR | Breaking? |
| - | - | - | - |
| Not set | Uses `light` as base for `light`, `dark` as base for `dark` etc. | Uses `light` as base for `light`, `dark` as base for `dark` etc. | Non-breaking! Behavior matches |
| Set | N/A | Uses theme set via `--base-paragon-theme` as base for all themes | Non-breaking! New functionality. |

_____

> Or I can just raise an error since they are incompatible options.

> now makes specifying both themes and base theme an error condition.

That's perfect! Thank you!

brian-smith-tcril:
Retrying on netlify didn't seem to fix the errors. I'm going to try closing and re-opening this to see if that triggers a successful deploy preview.

brian-smith-tcril:
It seems netlify is still trying to use an old ~~(Node 18)~~ (edit: Turns out it was an old version of Ubuntu that didn't like Node 24, I updated the netlify settings so it shouldn't be an issue anymore) base image for this despite it being rebased onto latest `release-23.x`. I think we are safe to ignore the netlify errors on this one.

> I've pushed a commit that improves the help

I don't see that change in here, maybe something got lost in the rebase?

xitij2000:
> I don't see that change in here, maybe something got lost in the rebase?

@brian-smith-tcril 
Sorry there is some issue on my side that's causing a lot of pushes to fail. I've pushed it now and can see it on github. 
It's in a separate commit to make it easier to review.

brian-smith-tcril (approved):

brian-smith-tcril:
Retrying on netlify worked, I'm going to try closing and reopening this PR one more time to see if the checks update.

xitij2000:
@brian-smith-tcril  Thanks!

PR #2899, opened 2023-12-07
feat!: update Autosuggest component to better support freeform and selected values

brian-smith-tcril:
## Description

This updates the Autosuggest component to provide the consumer with:
* The value entered in the textbox by the user
* The selected value (as it is displayed in the dropdown)
* The id of the selected option

This also allows the consumer to determine if the Autosuggest component should be in an error state when
* No text has been entered
* Text has been entered, but not selection has been made
* An external validation check has failed - "customError"

The consumer will provide error messages for the appropriate error states

BREAKING CHANGE: value of Autosuggest component is now an object instead of a string
BREAKING CHANGE: Autosuggest component now uses onChange instead of onSelected
BREAKING CHANGE: Autosuggest component now takes in different error messages for value/selection required, and custom errors

### Deploy Preview

https://deploy-preview-2899--paragon-openedx.netlify.app/components/form/form-autosuggest/

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Form/FormAutosuggest.jsx
+++ b/src/Form/FormAutosuggest.jsx
@@ -1,9 +1,10 @@
 import React, {
-  useEffect, useState, useRef,
+  useEffect, useState, useRef, forwardRef, useImperativeHandle,
 } from 'react';
 import PropTypes from 'prop-types';
 import { v4 as uuidv4 } from 'uuid';
 import { useIntl } from 'react-intl';
+import { requiredWhen } from '../utils/propTypes';
 import { KeyboardArrowUp, KeyboardArrowDown } from '../../icons';
 import Icon from '../Icon';
 import { FormGroupContextProvider, useFormGroupContext } from './FormGroupContext';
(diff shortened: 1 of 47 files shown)


src/Form/FormAutosuggest.jsx
```diff
@@ -334,9 +380,23 @@ FormAutosuggest.propTypes = {
...
-  errorMessageText: PropTypes.string,
+  value: PropTypes.shape({
+    userProvidedText: PropTypes.string,
+    selectionValue: PropTypes.string,
+    selectionId: PropTypes.string,
+  }),
+  /** Specifies if empty values trigger an error state */
+  valueRequired: PropTypes.bool,
+  /** Informs user they must input a value. */
+  valueRequiredErrorMessageText: PropTypes.string,
```
monteri, line 391: I assume that there should be `valueRequiredErrorMessageText` provided if `valueRequired` equals `true`. Please, consider:

```
import { requiredWhen } from '../utils/propTypes';

...

valueRequiredErrorMessageText: requiredWhen(PropTypes.string, 'valueRequired'),
```

And the same for `selectionRequired` and `selectionRequiredErrorMessageText`, `customError` and `customErrorMessageText`.

Another option, if no message is provided and there should only be a red box to indicate the error, I would remove the `x` under the input, otherwise it will appear incomplete.image

src/Form/form-autosuggest.mdx
```diff
@@ -69,6 +80,9 @@ Form auto-suggest enables users to manually select or type to find matching opti
 
 ```jsx live
 () => {
+    const [userProvidedText, setUserProvidedText] = useState('');
```
PKulkoRaccoonGang, line 83: It seems to me that this `mdx` file should have 2 indent spaces (currently 4). This can be seen by looking at other `mdx` files of the [Form component](https://github.dev/openedx/paragon/blob/716a6ddd932d61a5b5ab594e8e525c3dfffe20ab/src/Form/form-autosuggest.mdx).

src/Form/tests/FormAutosuggest.test.jsx
```diff
@@ -21,26 +23,53 @@ function FormAutosuggestTestComponent(props) {
...
-  onSelected: PropTypes.func,
+  /** Specifies onChange event handler. */
+  onChange: PropTypes.func,
   /** Specifies onClick event handler. */
   onClick: PropTypes.func,
+  /** Specifies if empty values trigger an error state */
+  valueRequired: PropTypes.bool,
+  /** Specifies if freeform values trigger an error state */
+  selectionRequired: PropTypes.bool,
+  /** Specifies the control is in a consumer provided error state */
```
PKulkoRaccoonGang, line 71: [question]: Does it make sense to add additional comments to the details in the test file? It seems to me that we already have these descriptions on the page of the component itself.

PKulkoRaccoonGang (commented):
@brian-smith-tcril leave some suggestions, please take a look

src/Form/FormAutosuggest.jsx
```diff
@@ -174,107 +212,112 @@ function FormAutosuggest({
...
-
-  const handleOnChange = (e) => {
-    const findStr = e.target.value;
 
-    if (onChange) { onChange(findStr); }
+    setHasValue(true);
+
+    const filteredItems = getItems(userProvidedText);
+    setDropdownItems(filteredItems);
+    // expandDropdown();
```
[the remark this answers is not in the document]
brian-smith-tcril: Good catch! I removed it.

src/Form/FormAutosuggest.jsx
```diff
@@ -174,107 +212,112 @@ function FormAutosuggest({
...
+
+    const matchingDropdownItem = filteredItems.find((o) => (
+      o.props.children.toLowerCase() === userProvidedText.toLowerCase()
+    ));
+    if (!matchingDropdownItem) {
```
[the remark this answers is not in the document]
brian-smith-tcril: I added some spaces and comments in here, hopefully that clears things up a bit.

src/Form/form-autosuggest.mdx
brian-smith-tcril: I fixed the indentation (by copying the jsx from the mdx file into a temp.jsx file and running lint fix)

src/Form/tests/FormAutosuggest.test.jsx
brian-smith-tcril: I was following the pattern of the existing ones, but I agree they felt redundant. I've removed them.

src/Form/FormAutosuggest.jsx
brian-smith-tcril: @monteri I have updated these to use `requiredWhen`

https://github.com/openedx/paragon/blob/f8a4e5ee118199936c3ff6653e3dfb389d12fffa/src/Form/FormAutosuggest.jsx#L421-L430

src/Form/form-autosuggest.mdx
```diff
@@ -19,52 +19,77 @@ Form auto-suggest enables users to manually select or type to find matching opti
...
+        User provided text
+        <Form.Control
+          onChange={(e) => setValue({
+            userProvidedText: e.target.value,
+            selectionValue: '',
+            selectionId: '',
+          })}
+          value={value.userProvidedText}
+        />
+        <Button onClick={forceUpdateErrorState}>Trigger validation</Button>
```
adamstankiewicz, line 90: nit: consider using `Stack` or CSS utility class to give some vertical whitespace between the `Form.Control` and the `Button`.

image

src/Form/form-autosuggest.mdx
```diff
@@ -19,52 +19,77 @@ Form auto-suggest enables users to manually select or type to find matching opti
 
 ```jsx live
 () => {
-  const [selected, setSelected] = useState('');
+  const [value, setValue] = useState({});
+  const [valueRequired, setValueRequired] = useState(false);
+  const [selectionRequired, setSelectionRequired] = useState(false);
+  const [customValidation, setCustomValidation] = useState(false);
```
adamstankiewicz, line 25: nit: perhaps prefix these boolean state variables with `isValueRequired`, `isSelectionRequired`, `hasCustomValidation`.

src/Form/FormAutosuggest.jsx
```diff
@@ -14,292 +15,358 @@ import Spinner from '../Spinner';
...
-        displayValue: value,
-      }));
-    }
-  }, [value]);
 
-  const setDisplayValue = (itemValue) => {
-    const optValue = [];
+    useImperativeHandle(ref, () => ({
+      updateErrorStateAndErrorMessage,
+    }));
```
adamstankiewicz, line 178: [nit/suggestion] May be a good idea to include a comment here or elsewhere documenting the use of `useImperativeHandle` to only include `updateErrorStateAndErrorMessage` on its exposed methods. That way, future contributions will have some guidance on how to keep this maintained such that methods intended to be exposed are intentionally included here? 🤔

src/Form/FormAutosuggest.jsx
```diff
@@ -14,292 +15,358 @@ import Spinner from '../Spinner';
...
+    },
+    ref,
+  ) => {
+    const intl = useIntl();
+    const formControlRef = useRef();
+    const parentRef = useArrowKeyNavigation({
+      selectors: arrowKeyNavigationSelector,
+      ignoredKeys: ignoredArrowKeysNames,
+    });
+    const [dropdownExpanded, setDropdownExpanded] = useState(false);
```
adamstankiewicz, line 45: nit: `isDowndownExpanded`

src/Form/FormAutosuggest.jsx
```diff
@@ -14,292 +15,358 @@ import Spinner from '../Spinner';
...
+        return '';
+      }
 
-    setState(prevState => ({
-      ...prevState,
-      dropDownItems: [],
-      displayValue: clickedValue,
-    }));
+      return value.userProvidedText;
+    });
```
adamstankiewicz, line 55: nit: could this be simplified to:

```jsx
const [displayValue, setDisplayValue] = useState(value?.userProvidedText || '');
```

src/Form/FormAutosuggest.jsx
```diff
@@ -14,292 +15,358 @@ import Spinner from '../Spinner';
...
-
-    setState(prevState => ({
-      ...prevState,
-      dropDownItems: [],
-      errorMessage: !state.displayValue ? errorMessageText : '',
-    }));
+    function getItems(strToFind = '') {
+      let childrenOpt = React.Children.map(children, (child) => {
+      // eslint-disable-next-line no-shadow
+        const { children, onClick, ...rest } = child.props;
```
adamstankiewicz, line 97: nit: to avoid `// eslint-disable-next-line no-shadow`, consider renaming `children` once deconstructed (i.e., `const { children: childChildren, ... } = child.props;` to avoid the naming conflict with the parent `children`.

src/Form/FormAutosuggest.jsx
```diff
@@ -14,292 +15,358 @@ import Spinner from '../Spinner';
...
-    setActiveMenuItemId(menuItemId);
-  };
-
-  const handleItemClick = (e, onClick) => {
-    const clickedValue = e.currentTarget.getAttribute('data-value');
-
-    if (onSelected && clickedValue !== value) {
-      onSelected(clickedValue);
-    }
+const FormAutosuggest = forwardRef(
```
adamstankiewicz, line 18: [nit/suggestion/consideration] I'm wondering if it might make sense to abstract some of this logic into custom React hook(s) to keep the component code itself more focused on the UI presentation. For example, a custom hook(s) could return the callback functions to the component for use, having been defined in the hook(s).

adamstankiewicz (approved):

src/Form/form-autosuggest.mdx
brian-smith-tcril: addressed here https://github.com/openedx/paragon/compare/f8a4e5ee118199936c3ff6653e3dfb389d12fffa..9310aa04d988eba1ee524cf920e5ad54dd39590f#diff-3ce694685ee1cc2b274e86b07ff39240bff96ddf3862d4c5366d0f24ad863d38R81-R91

src/Form/form-autosuggest.mdx
brian-smith-tcril: addressed throughout https://github.com/openedx/paragon/compare/f8a4e5ee118199936c3ff6653e3dfb389d12fffa..9310aa04d988eba1ee524cf920e5ad54dd39590f

src/Form/FormAutosuggest.jsx
brian-smith-tcril: added a comment here https://github.com/openedx/paragon/compare/f8a4e5ee118199936c3ff6653e3dfb389d12fffa..9310aa04d988eba1ee524cf920e5ad54dd39590f#diff-96e13c1a773ec8d218ea5667d3dec6c8f6eeadcb00c81584f50d6c3d4d52135fR170-R171

If you have another suggestion for a comment I'm more than happy to change it!

src/Form/FormAutosuggest.jsx
brian-smith-tcril: addressed https://github.com/openedx/paragon/compare/f8a4e5ee118199936c3ff6653e3dfb389d12fffa..9310aa04d988eba1ee524cf920e5ad54dd39590f#diff-96e13c1a773ec8d218ea5667d3dec6c8f6eeadcb00c81584f50d6c3d4d52135fR45

src/Form/FormAutosuggest.jsx
brian-smith-tcril: addressed https://github.com/openedx/paragon/compare/f8a4e5ee118199936c3ff6653e3dfb389d12fffa..9310aa04d988eba1ee524cf920e5ad54dd39590f#diff-96e13c1a773ec8d218ea5667d3dec6c8f6eeadcb00c81584f50d6c3d4d52135fR49

src/Form/FormAutosuggest.jsx
brian-smith-tcril: addressed https://github.com/openedx/paragon/compare/f8a4e5ee118199936c3ff6653e3dfb389d12fffa..9310aa04d988eba1ee524cf920e5ad54dd39590f#diff-96e13c1a773ec8d218ea5667d3dec6c8f6eeadcb00c81584f50d6c3d4d52135fR90

src/Form/FormAutosuggest.jsx
brian-smith-tcril: I'd like to think about this one a bit. I worry that *too* much abstraction might lead to consumers using the component in ways that feel broken, but that could likely be resolved by providing good examples.

Maybe it makes sense to break this out into a separate issue to do some discovery on it?

adamstankiewicz: The suggestion to abstract some of the logic shouldn't really affect consumers? The custom hooks wouldn't be exposed to consumers; just for use internally within the component.

Generally when a components implementation reaches several hundred plus lines that feels like the component might have a bit too much going on where it could be simplified through the use of hooks to improve readability, etc. 

Totally fine to defer and consider. Just a suggestion to make the component a bit more maintainable, where it makes sense.

brian-smith-tcril: > The custom hooks wouldn't be exposed to consumers; just for use internally within the component.

Gotcha, I misinterpreted the comment.

> Totally fine to defer and consider. Just a suggestion to make the component a bit more maintainable, where it makes sense.

That sounds good. It's definitely a good idea, but I think it makes sense to hold off until after getting this PR merged.

adamstankiewicz (approved):

Issue #2870, opened 2023-11-28
Paragon CLI version


PKulkoRaccoonGang:
In popular CLI interfaces, you will notice that it is possible to view the version of the CLI being used. It would be an interesting idea to add the ability to view the Paragon CLI version using the `paragon -v` (`paragon --version`) command.

For example:
<img width="509" alt="image" src="https://github.com/openedx/paragon/assets/93188219/ef2a96cd-6698-4da9-94ac-de752a63ad43">

```[tasklist]
### Tasks
- [ ] Ensure that after entering the `paragon -v` (`paragon --version`) command the Paragon CLI version is displayed
- [ ] Update supporting documentation (`paragon help` command)
- [ ] Ensure that the functionality works both locally in Paragon and in MFE consumers
```

PKulkoRaccoonGang:
PR: https://github.com/openedx/paragon/pull/2921

PR #2921, opened 2023-12-11
[BD-46] feat: added Paragon CLI version command

PKulkoRaccoonGang:
## Description

**Issue:** https://github.com/openedx/paragon/issues/2870

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/bin/paragon-scripts.js
+++ b/bin/paragon-scripts.js
@@ -2,8 +2,10 @@
 const chalk = require('chalk');
 const themeCommand = require('../lib/install-theme');
 const helpCommand = require('../lib/help');
+const versionCommand = require('../lib/version');
 
 const HELP_COMMAND = 'help';
+const VERSION_COMMANDS = ['-v', '--version'];
 
 const COMMANDS = {
   /**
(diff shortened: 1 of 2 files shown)


openedx-webhooks:


bin/paragon-scripts.js
```diff
@@ -47,12 +49,21 @@ const COMMANDS = {
...
 
 (async () => {
   const [command] = process.argv.slice(2);
   const executor = COMMANDS[command];
 
+  if (VERSION_COMMANDS.includes(command)) {
+    versionCommand();
+    return;
+  }
+
```
viktorrusakov, line 66: Let's not do this, this breaks the whole pattern we have for CLI commands - you should never call the command directly, but instead call `executor` method that you defined in `COMMANDS` object.

Instead, create an aliases map for command, e.g.
```js
// add this at the top of the file
commandAliases = {
  '-v': 'version',
  '--version': 'version',
}

// and extract executor this way
const [command] = process.argv.slice(2);
const resolvedCommand = commandAliases[command] || command;
const executor = COMMANDS[resolvedCommand];
```

then you don't need to add any `if` block and it will be very easy to add aliases for the future commands as well.

viktorrusakov:
@PKulkoRaccoonGang Also, I think it might be better to publish `paragon-cli` as a separate package, that's what other popular libraries do. Food for thought 🙂 :

bin/paragon-scripts.js
PKulkoRaccoonGang: Ok, thanks

viktorrusakov (approved):

PR #2940, opened 2023-12-15
fix: use `any` in `getDependentProjectUsages` to work around type-check errors

brian-smith-tcril:
## Description

As of https://github.com/openedx/paragon/commit/e6967cf63ed1b15dc0c0d5f236b867743afe73a7, the "Check Types" step of the CI has been failing. This is currently a problem for every open PR. I tried digging into the problem and seeing if I could fix the type definitions to get past the error, but didn't come to a good solution. I posted about what I learned in slack [here](https://openedx.slack.com/archives/C02NR285KV4/p1702662593338799) 

I would prefer to fix this in a way that doesn't rely on using `any`, but I figured it's better to have *some* way to get the error to disappear PR'd than nothing.

### Deploy Preview

Include a direct link to your changes in this PR's deploy preview here (e.g., a specific component page).

## Merge Checklist

* [ ] If your update includes visual changes, have they been reviewed by a designer? Send them a link to the Netlify deploy preview, if applicable.
* [ ] Does your change adhere to the documented [style conventions](https://github.com/openedx/paragon/blob/master/docs/decisions/0012-css-styling-conventions)?
* [ ] Do any prop types have missing descriptions in the Props API tables in the documentation site (check deploy preview)?
* [ ] Were your changes tested using all available themes (see theme switcher in the header of the deploy preview, under the "Settings" icon)?
* [ ] Were your changes tested in the `example` app?
* [ ] Is there adequate test coverage for your changes?
* [ ] Consider whether this change needs to reviewed/QA'ed for accessibility (a11y). If so, please add `wittjeff` and `adamstankiewicz` as reviewers on this PR.

## Post-merge Checklist

* [ ] Verify your changes were released to [NPM](https://www.npmjs.com/package/@edx/paragon) at the expected version.
* [ ] If you'd like, [share](https://github.com/openedx/paragon/discussions/new?category=show-and-tell) your contribution in [#show-and-tell](https://github.com/openedx/paragon/discussions/categories/show-and-tell).
* [ ] 🎉 🙌 Celebrate! Thanks for your contribution.

PR #2941, opened 2023-12-15
fix: use helper function to satisfy dependent project type checks

brian-smith-tcril:
## Description

This updates the type definitions for dependent project usage to match the json, and adds a helper function to get past typescript issues with dynamic keys.

An alternate solution would be to just use `any` as seen here: https://github.com/openedx/paragon/pull/2940

### Deploy Preview

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/www/src/types/types.ts
+++ b/www/src/types/types.ts
@@ -9,7 +9,7 @@ export interface IUsage {
   filePath: string,
   line: number,
   column: number,
-  version: string,
+  index: number,
 }
 
 export interface IComponentUsageData {
@@ -24,14 +24,14 @@ export interface IComponentUsageData {
 export interface IDependentProjectsUsages extends Omit<IDependentUsage, 'count'> {
   version: string,
   name: string,
-  repository: { type: string, url: string },
+  repository: { type: string, url: string } | string,
   folderName: string,
 }
 
 export interface IDependentUsage {
   version?: string,
   name?: string,
-  repository?: { type: string, url: string },
+  repository?: { type: string, url: string } | string,
   repositoryUrl?: string,
   count: number,
   folderName?: string,
--- a/www/src/utils/getDependentProjectsUsages.tsx
+++ b/www/src/utils/getDependentProjectsUsages.tsx
@@ -7,14 +7,27 @@ const {
   projectUsages: dependentProjectsUsages,
 } = dependentProjectsAnalysis;
 
+// Utility function to convert dynamic structure to expected structure
+function convertToExpectedStructure(project: any): IDependentProjectsUsages {
+  return {
+    ...project,
+    usages: Object.keys(project.usages).reduce((acc, key) => {
+      acc[key] = project.usages[key];
+      return acc;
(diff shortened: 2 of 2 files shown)


viktorrusakov:
I've been digging into it as well and I think I found what's causing all of the problems. We try to import a JSON file which has no types defined for it, so TypeScript tries to infer these types during import to help us and does it poorly. When it reads the file and specifically these lines
```js
...
"usages": {
  "Alert": [
    {
      "filePath": "openassessment/xblock/static/js/src/lms/components/WaitingStepContent.jsx",
      "line": 40,
      "column": 6,
      "index": 1251
    },
    {
      "filePath": "openassessment/xblock/static/js/src/lms/containers/WaitingStepDetailsContainer.jsx",
      "line": 68,
      "column": 12,
      "index": 2171
    }
  ],
  "Button": [
    {
      "filePath": "openassessment/xblock/static/js/src/lms/components/WaitingStepList.jsx",
      "line": 15,
      "column": 4,
      "index": 445
    }
...
```
it assumes that `usages` is an object with _literal_ keys `Alert`, `Button` and so on which does not correspond to our defined types, you can see this from this error message that is printed after running `npm run type-check` (see how the type definition includes literal keys here, instead of generic `[key: string]` type or something similar)
```
Type '{ Alert: { filePath: string; line: number; column: number; index: number; }[]; Button: { filePath: string; line: number; column: number; index: number; }[]; DataTable: { filePath: string; line: number; column: number; index: number; }[]; ... 318 more ...; Fieldset?: undefined; }' is not assignable to type '{ [key: string]: IUsage[]; }'.
```

My guess is that wasn't an issue before https://github.com/openedx/paragon/commit/e6967cf63ed1b15dc0c0d5f236b867743afe73a7 because the first `usages` object that TypeScript encounter was an empty one, so it probably inferred `any` type for keys or something like that...

Anyway, I think a better solution would be to somehow disable type inference for the JSON file or enforce our types here, but I haven't found a clean way to do that...

One thing that works for me and seems like an OK solution is updating first lines in `getDependentProjectsUsages.tsx` file as follows
```ts
// @ts-ignore
import dependentProjectsAnalysis from '../../../dependent-usage.json'; // eslint-disable-line
import getGithubProjectUrl from './getGithubProjectUrl';
import { IDependentProjectsUsages, IDependentUsage, IUsage } from '../types/types';

const { projectUsages } = dependentProjectsAnalysis;
const dependentProjectsUsages = projectUsages as unknown as IDependentProjectsUsages[];
```
which does the following:
1. `projectUsages as unknown` - clears the types inferred by TypeScript 
2. `as IDependentProjectsUsages[]` - casts `projectUsages` object to be of desired type

this works for me locally.

@brian-smith-tcril I'm going to approve this PR as I really don't have any strong preference for either of the solutions. I'm not a fan of TypeScript, even just simply switching type to `any` like you did in your other PR works for me 🙂.

viktorrusakov (approved):

brian-smith-tcril:
closing in favor of https://github.com/openedx/paragon/pull/2951

PR #3003, opened 2024-02-13
fix: unrecognized prop warning in form autosuggest

syedsajjadkazmii:
## Description

We encountered an issue where the `childChildren` prop passed to the `MenuItem` component was flagged as an unrecognized prop to DOM (locally), leading to a console warning. Simultaneously, this warning caused the occurrence of duplicate API calls within our MFE (frontend-app-authn).

**Resolution:** We renamed the `childChildren` to `childchildren` and the warning stopped appearing and the duplicate API calls issue was resolved.

We are still unsure why we were having duplicate API calls due to this warning. Maybe React was trying to re-render due to invalid props? But we made sure that there was no issue with our implementation and it was due to the unrecognized prop. But we would like to understand the actual reason behind this if possible.

Console Warning screenshot.
Screenshot 2024-02-13 at 12 15 43 PM

### Before
In this video, we can notice that we are getting single API calls until we click on the `FormAutoSuggest` field. On clicking the field, we get the console warning and all our API calls start firing twice. 

https://github.com/openedx/paragon/assets/52817156/5a5fc154-64d7-4177-a658-90f97c4f0bf3

### After
In this video, (after the fix) we can see that we are neither getting console warnings nor duplicate API calls. All the API calls are being fired once.

https://github.com/openedx/paragon/assets/52817156/bd6dd6f3-3c73-4570-b192-2138e20d3984

### Deploy Preview

https://deploy-preview-3003--paragon-openedx.netlify.app/

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Form/FormAutosuggest.jsx
+++ b/src/Form/FormAutosuggest.jsx
@@ -92,7 +92,7 @@ const FormAutosuggest = forwardRef(
 
         return React.cloneElement(child, {
           ...rest,
-          childChildren,
+          children: childChildren,
           'data-value': childChildren,
           onClick: (e) => handleItemSelect(e, onClick),
           id: menuItemId,


zainab-amir (approved):

src/Form/FormAutosuggest.jsx
```diff
@@ -87,13 +87,13 @@ const FormAutosuggest = forwardRef(
 
     function getItems(strToFind = '') {
       let childrenOpt = React.Children.map(children, (child) => {
-        const { children: childChildren, onClick, ...rest } = child.props;
+        const { children: childchildren, onClick, ...rest } = child.props;
```
brian-smith-tcril, line 90:
```suggestion
        const { children: childChildren, onClick, ...rest } = child.props;
```

src/Form/FormAutosuggest.jsx
```diff
@@ -87,13 +87,13 @@ const FormAutosuggest = forwardRef(
...
         const menuItemId = child.props.id ?? uuidv4();
 
         return React.cloneElement(child, {
           ...rest,
-          childChildren,
-          'data-value': childChildren,
+          childchildren,
+          'data-value': childchildren,
```
brian-smith-tcril, line 96:
```suggestion
          children: childChildren,
          'data-value': childChildren,
```

brian-smith-tcril (changes requested):
Thanks for pointing out this bug! I believe the intended behavior is to set `children` on the cloned element since the `childChildren` variable used to just have the name `children`

https://github.com/openedx/paragon/blob/d32e6d01dc9c9458f2456ace6ebb8da173a43e88/src/Form/FormAutosuggest.jsx#L75-L82

I've left suggestions that resolved the warning for me when testing locally.

syedsajjadkazmii:
> Thanks for pointing out this bug! I believe the intended behavior is to set `children` on the cloned element since the `childChildren` variable used to just have the name `children`
> 
> https://github.com/openedx/paragon/blob/d32e6d01dc9c9458f2456ace6ebb8da173a43e88/src/Form/FormAutosuggest.jsx#L75-L82
> 
> I've left suggestions that resolved the warning for me when testing locally.

Hi @brian-smith-tcril, thanks for the suggestions. I have updated the PR.

brian-smith-tcril (approved):

PR #3016, opened 2024-03-12
feat: working typings for Paragon, better types for <Icon> component

bradenmacdonald:
## Description

This represents my version of the **minimal changes to Paragon** to start getting (just a few!) useful type definitions shipping with it.

Before this PR, I wasn't seeing any types for Paragon imports, either in MFE code that uses it nor even in the `www` subproject of this repo, which nominally uses TypeScript.

After this PR, you still won't see many types, but TypeScript will at least know which component exist and which don't (it can check the imports). And a few components - `<Icon>`, `<Chip>`, `<ChipCarousel>`, and `<Bubble>` will have types because they are written in TypeScript (or in the case of `Icon` because there's a `.d.ts` file).

I deliberately didn't go any further than that because I want to keep the changeset small enough that we can confidently review and merge it. From that point on, it will be much easier to incrementally add types to individual components in separate PRs.

This PR:
1. Creates a `src/index.d.ts` file that makes all of the top-level exports visible to TypeScript. Without this fix, types were not being found in repos that import Paragon by path, such as the `www` subproject. It **will** be necessary to manually keep this file in sync with `src/index.js`. To make that as easy as possible, I made sure the files have a 1:1 correspondence line-by-line.
    
    Example: **`www/src/components/IconsTable.tsx`**
    | Before: `<Icon>` has no types | After: `<Icon>` is type checked |
    |-|-|
    | Before | After |

2. Adds `"types": "dist/index.d.ts",` to `package.json`. Without this change, types were not being found in MFEs that import Paragon via NPM.
2. Renames `/src/Icon/Icon.test.jsx` to be a `.tsx` file so that the `<Icon />` component tests will use/check TypeScript.
3. Fixes incorrect types in `src/Icon/index.d.ts`. Now that `Icon.test.tsx` is enabled for type checking, we see that it contained type errors, because the `.d.ts` file was incomplete:
    Screenshot 2024-03-12 at 2 14 49 PM
4. Adds type information to the `icons` sub-package. Now you get auto-completion and compile-time validation of the icon names. The new file, `icons/es5/index.ts` is completely auto-generated so there's no need to maintain the types manually.
    Screenshot 2024-03-12 at 2 16 21 PM
5. Adds some compile-time (TypeScript-only) tests to `Icon.test.tsx` to avoid regressions.
6. Renames a few `.jsx` files in the documentation site (`www` subfolder) to `.tsx` because otherwise the linter was complaining that `~paragon-react` couldn't be found. ~~This is because a `.jsx` file cannot import Paragon's `src/index.ts` anymore after the change in step 1.~~ This may not be an issue anymore after I changed the approach from step 1 above, but it doesn't hurt :)
7. Fixed the `tsconfig.json` (see inline comment) and separated out the build vs. dev tsconfigs.

### How to test in an MFE

1. Have the MFE and paragon in folders next to each other on your host. This example will assume that `frontend-app-course-authoring` and `paragon` are peer folders on your host.
2. Make sure the MFE's dependencies are up to date (e.g. `tutor dev run course-authoring npm install` to do this for course-authoring)
3. In Paragon, edit `package.json` and change the line `"@types/react": "17.0.0",` to `"@types/react": "../frontend-app-course-authoring/node_modules/@types/react",` (otherwise you'll get weird conflicts with this package). Substitute the appropriate folder name if you're using a different MFE.
4. Run `npm install` in the `paragon` folder.
5. Mount paragon into the MFE's container: `tutor mounts add course-authoring:/Users/yourname/path/to/paragon:/openedx/paragon` (adjust this command as needed)
6. In the MFE's `package.json` change `"@openedx/paragon": "^21.5.7",` to `"@openedx/paragon": "file:../paragon",`
7. Link in the local paragon copy using `tutor dev run course-authoring npm install --force` (warnings about "Conflicting peer dependency" are expected; ignore them)
8. In VS Code, open any `.jsx` file in the MFE that has `// @ts-check` on its first line (or add that line in!), and you'll see type checking. Note that `<Icon>` should have correct typing, and other Paragon components like `<Stack>` should not show any type errors.

Example, showing the types appearing in VS Code for Course Authoring MFE's [`ImportTagsWizard.jsx`](https://github.com/openedx/frontend-app-course-authoring/blob/master/src/taxonomy/import-tags/ImportTagsWizard.jsx):
Screenshot 2024-03-13 at 1 31 22 PM

### Next Steps

Once this PR is merged, I think we should follow up with additional PRs to add typing to individual components (e.g. one component per PR, or similar). That can be done either by adding a `.d.ts` file for the component, or by converting the component implementation to TypeScript. In either case, I would encourage people to follow the example in this PR and make sure that the tests for the component are also converted to TypeScript (or some other mechanism is used to check that the types are working).

Once _everything_ has a type (or close to it), the `index.js` file can be renamed to `index.ts` and the `index.d.ts` file can be deleted :)

### Deploy Preview

The code for this page was changed, but you should not be able to see any changes:
https://deploy-preview-3016--paragon-openedx.netlify.app/foundations/icons/

## Merge Checklist

## Post-merge Checklist


tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "noImplicitAny": true,
     "allowJs": false,
     "rootDir": "src",
+    "outDir": "dist"
```
bradenmacdonald, line 7: This fixed a major issue. Without this `outDir` setting, the `build` script was running `tsc` but tsc was putting its output `.d.ts` files into `./node_modules/@edx/typescript-config/dist/` !

www/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "rootDir": "src",
+    "rootDir": "../",
```
brian-smith-tcril, line 4: I'm curious as to what happened before this was changed.

brian-smith-tcril (commented):
Overall this looks great! Thank you so much for putting this together! I'm looking forward to pulling this down and testing it with an MFE!

www/tsconfig.json
bradenmacdonald: Without this change, I was getting errors like this when running `./node_modules/.bin/tsc --project www --noEmit` from the main `paragon` root:

```
www/src/components/AutoToc.tsx:5:24 - error TS6059: File '/Users/opencraft/Documents/tutor-devstack/paragon/src/index.ts' is not under 'rootDir' '/Users/opencraft/Documents/tutor-devstack/paragon/www/src'. 'rootDir' is expected to contain all source files.
  The file is in the program because:
    Imported via '~paragon-react' from file '/Users/opencraft/Documents/tutor-devstack/paragon/www/src/components/AutoToc.tsx'
```

However, it seems like the commit 688d212f49c3c2950b54c10e57e86ac0804eb2c6 resolved this in another way, by changing from `index.ts` to `index.d.ts` so this change to `www/tsconfig.json`'s `rootDir` is no longer needed. I'll just leave it as it was.

brian-smith-tcril:
I tested this out in `frontend-app-authn` by replacing `"@openedx/paragon": "^22.0.0",` in `package.json` with `"@openedx/paragon": "file:../paragon",`, and adding `// @ts-check` to the top of `EnterpriseSSO.jsx`. I then looked at this line https://github.com/brian-smith-tcril/frontend-app-authn/blob/2d37b8b0bfd8d289dc4da409d5d9c749b6c01f2b/src/common-components/EnterpriseSSO.jsx#L62 and saw
image

I'm not sure if this is a "task failed successfully" moment or if this means there's more that needs to be explored here.

bradenmacdonald:
@brian-smith-tcril That happens when you have two different versions of `@types/react` installed - Paragon has its own version that's different from what's in your MFE. It's a side effect of the way you added paragon as a local folder dependency. I ran into the same problem, and found a workaround, which is in the "How to test in an MFE" in the PR description.

Also, once this is merged and installed via NPM, its deduplication will make sure that doesn't happen. But here you're joining together two separate sets of `npm install`ed packages and two different `node_modules` folders, rather than one big one.

brian-smith-tcril:
@bradenmacdonald thanks for clarifying! I wasn't sure what you meant by "weird conflicts with this package" in that section. After following that step everything seems to work as expected!

brian-smith-tcril (approved):
Tested locally with `frontend-app-authn`, everything is working as expected!

adamstankiewicz (approved):

openedx-webhooks:
@bradenmacdonald 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/3016&entry.1484655318=https://github.com/bradenmacdonald&entry.752974735=2024-03-12+21:24&entry.1917517419=2024-03-25+15:57&entry.2133058324=Yes) so we can improve your experience in the future.

Issue #3050, opened 2024-05-13
[paragon-openedx.netlify.app] Feedback (on /components/hyperlink/)


adamstankiewicz:
## Improvements

`Hyperlink` is currently always a wrapper around the standard HTML anchor tag (i.e., `<a>`). As a result, it is not currently possible to use the `Hyperlink` component with a React Router `Link` for internal page routes within React applications. 

Given this, consumers implementing links to internal page routes typically rely on `Link` directly, accepting a tradeoff that the `Hyperlink` component's built-in color and link variants are not applicable to the internal route link.

To mitigate this, `Hyperlink` should be extended to support an `as` prop similar to other components throughout the Paragon component library. This addition would allow consumers to render `<Hyperlink as={Link} ... />` and still have consistently styled links even when the link used React routing for an internal page route.

```[tasklist]
### Tasks
- [ ] Ensure `Hyperlink` accepts an `as` prop such that consumers can rely on React Router's `Link` component, i.e. `<Hyperlink as={Link} destination="..." />`.
```

PR #3077, opened 2024-05-17
Add TypeScript types for <Hyperlink>

bradenmacdonald:
## Description

Working to incrementally expand typing in paragon. This adds types for `<Hyperlink>`

I removed the `content` prop which [has been deprecated for five years, since v4.0.2](https://github.com/openedx/paragon/commit/40227e164b5f6cc22f52ce760f05615169b72320).

I also clarified the description of props like `externalLinkAlternativeText` and `isInline` which were very unclear about what they were doing.

### Deploy Preview

Will be available later

## Merge Checklist

* [x] If your update includes visual changes, have they been reviewed by a designer? Send them a link to the Netlify deploy preview, if applicable. n/a

## Post-merge Checklist

### DIFF
--- a/src/Hyperlink/Hyperlink.test.tsx
+++ b/src/Hyperlink/Hyperlink.test.tsx
@@ -4,30 +4,34 @@ import userEvent from '@testing-library/user-event';
 
 import Hyperlink from '.';
 
-const content = 'content';
 const destination = 'destination';
+const content = 'content';
 const onClick = jest.fn();
 const props = {
-  content,
   destination,
   onClick,
 };
 const externalLinkAlternativeText = 'externalLinkAlternativeText';
 const externalLinkTitle = 'externalLinkTitle';
 const externalLinkProps = {
-  target: '_blank',
+  target: '_blank' as const,
   externalLinkAlternativeText,
   externalLinkTitle,
   ...props,
 };
 
 describe('correct rendering', () => {
+  beforeEach(() => {
+    onClick.mockClear();
+  });
+
   it('renders Hyperlink', async () => {
-    const { getByRole } = render(<Hyperlink {...props} />);
+    const { getByRole } = render(<Hyperlink {...props}>{content}</Hyperlink>);
     const wrapper = getByRole('link');
     expect(wrapper).toBeInTheDocument();
 
     expect(wrapper).toHaveClass('pgn__hyperlink');
+    expect(wrapper).toHaveClass('standalone-link');
     expect(wrapper).toHaveTextContent(content);
     expect(wrapper).toHaveAttribute('href', destination);
     expect(wrapper).toHaveAttribute('target', '_self');
@@ -36,8 +40,17 @@ describe('correct rendering', () => {
     expect(onClick).toHaveBeenCalledTimes(1);
   });
 
+  it('renders an underlined Hyperlink', async () => {
+    const { getByRole } = render(<Hyperlink isInline {...props}>{content}</Hyperlink>);
+    const wrapper = getByRole('link');
+    expect(wrapper).toBeInTheDocument();
+    expect(wrapper).toHaveClass('pgn__hyperlink');
+    expect(wrapper).not.toHaveClass('standalone-link');
+    expect(wrapper).toHaveClass('inline-link');
+  });
+
   it('renders external Hyperlink', () => {
-    const { getByRole, getByTestId } = render(<Hyperlink {...externalLinkProps} />);
+    const { getByRole, getByTestId } = render(<Hyperlink {...externalLinkProps}>{content}</Hyperlink>);
     const wrapper = getByRole('link');
     const icon = getByTestId('hyperlink-icon');
     const iconSvg = icon.querySelector('svg');
@@ -53,18 +66,16 @@ describe('correct rendering', () => {
 
 describe('security', () => {
   it('prevents reverse tabnabbing for links with target="_blank"', () => {
-    const { getByRole } = render(<Hyperlink {...externalLinkProps} />);
+    const { getByRole } = render(<Hyperlink {...externalLinkProps}>{content}</Hyperlink>);
     const wrapper = getByRole('link');
     expect(wrapper).toHaveAttribute('rel', 'noopener noreferrer');
   });
 });
 
 describe('event handlers are triggered correctly', () => {
-  let spy;
-  beforeEach(() => { spy = jest.fn(); });
-
   it('should fire onClick', async () => {
-    const { getByRole } = render(<Hyperlink {...props} onClick={spy} />);
+    const spy = jest.fn();
+    const { getByRole } = render(<Hyperlink {...props} onClick={spy}>{content}</Hyperlink>);
     const wrapper = getByRole('link');
     expect(spy).toHaveBeenCalledTimes(0);
     await userEvent.click(wrapper);
--- a/src/Hyperlink/index.tsx
+++ b/src/Hyperlink/index.tsx
@@ -7,7 +7,7 @@ import Icon from '../Icon';
 export const HYPER_LINK_EXTERNAL_LINK_ALT_TEXT = 'in a new tab';
 export const HYPER_LINK_EXTERNAL_LINK_TITLE = 'Opens in a new tab';
 
-interface Props extends Omit<React.ComponentPropsWithoutRef<'a'>, 'href' | 'target'> {
+interface Props extends Omit<React.ComponentPropsWithRef<'a'>, 'href' | 'target'> {
   /** specifies the URL */
   destination: string;
   /** Content of the hyperlink */
--- a/src/Hyperlink/index.tsx
+++ b/src/Hyperlink/index.tsx
@@ -1,29 +1,45 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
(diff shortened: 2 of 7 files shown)


openedx-webhooks:
Thanks for the pull request, @bradenmacdonald! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

src/Hyperlink/index.tsx
```diff
@@ -1,29 +1,45 @@
...
-  } = props;
+interface Props extends Omit<React.ComponentPropsWithoutRef<'a'>, 'href' | 'target'> {
+  /** specifies the URL */
+  destination: string;
+  /** Content of the hyperlink */
+  children: React.ReactNode;
+  /** Custom class names for the hyperlink */
+  className?: string;
+  /** Alt text for the icon indicating that this link opens in a new tab, if target="_blank". e.g. _("in a new tab") */
+  externalLinkAlternativeText?: string;
```
bradenmacdonald, line 18: I left this as an optional prop. There was propTypes code below to make `externalLinkAlternativeText` required when `target="_blank"` but this wasn't actually working as far as I can tell, especially since a default value is given.

If we want to make those props conditionally required (when target=`_blank`), I _can_ express that in the TypeScript types, but there are several instances in this repo alone that do not do that ([example](https://github.com/openedx/paragon/blob/9993fb3b938e50012b5aa220c1e75f4b8ff684c6/www/src/components/Menu.tsx#L218-L221)) and I suspect it would then show type errors in many other MFEs as well (if MFEs even use this component? Maybe they use react-router's Link instead).

We probably _should_ make that required at some point, because the default value is not internationalized. But the only good way to do that that I could think of is to require each MFE to subclass `Hyperlink` with a component that provides internationalized defaults for these fields, and that seems like too much work. Is there not some way to provide internationalized defaults in Paragon itself?

brian-smith-tcril (approved):

brian-smith-tcril:
@adamstankiewicz I'd like to merge this but I don't want to get in the way of https://github.com/openedx/paragon/issues/3050

Would this cause any problems with the work being done towards that?

src/Hyperlink/index.tsx
adamstankiewicz: > but this wasn't actually working as far as I can tell, especially since a default value is given.

Yeah, you're likely correct the `ifRequiredWhen` isn't really doing much due to the default value.

> I suspect it would then show type errors in many other MFEs as well (if MFEs even use this component? Maybe they use react-router's Link instead).
Yes, this is usually the case. See the recently filed issue https://github.com/openedx/paragon/issues/3050 regarding adapting `Hyperlink` to be compatible with `Link`, such that `Link` usage can have more consistent styling.

> We probably should make that required at some point, because the default value (HYPER_LINK_EXTERNAL_LINK_ALT_TEXT, above) is not internationalized. But the only good way to do that that I could think of is to require each MFE to subclass Hyperlink with a component that provides internationalized defaults for these fields, and that seems like too much work. Is there not some way to provide internationalized defaults in Paragon itself?

Paragon has had some i18n support in its components for a couple years now (see [ADR](https://github.com/openedx/paragon/blob/master/docs/decisions/0015-i18n-support.rst) and the [README](https://github.com/openedx/paragon/tree/9993fb3b938e50012b5aa220c1e75f4b8ff684c6?tab=readme-ov-file#internationalization)), except IIRC the `pull_translations` job routinely/currently fails due to an translation format error of some kind that's prevented the translations from updating in awhile. Related, I'm also not sure if Paragon has been migrated to using `openedx-atlas` / `openedx-translations` yet (unlikely), or if it intends to; @brian-smith-tcril might be able to share some insight about any plans to migrate.

It does seem by moving translations to `openedx-translations`, most MFEs are no longer importing/supplying `paragonMessages` anymore (e.g., removed from frontend-app-learning in [this commit](https://github.com/openedx/frontend-app-learning/commit/94d20833aaa635e17f3ff953a268553dab3638ef); I'm personally not sure if the Paragon i18n messages (and the shared header/footer component i18n messages) are somehow still passed to the MFEs, or if that support was (intentionally?) removed as part of the migration to `openedx-translations`?

If we want it to have a translated default prop (I agree, we should; looks like `Hyperlink` was missed in the original introduction of i18n for hardcoded English strings), then ideally we mark the strings for i18n via react-intl's `defineMessage` / `intl.formatMessage` or `FormattedMessage`, similar to a handful of other components in Paragon today; if the prop is provided, it gets used (where consumer might pass its own i18n string), otherwise falling back to the default i18n Paragon message.

src/Hyperlink/index.tsx
```diff
@@ -105,32 +121,20 @@ Hyperlink.propTypes = {
...
   /** specify if we need to show launch Icon. By default, it will be visible. */
   showLaunchIcon: PropTypes.bool,
 };
 
-export default withDeprecatedProps(Hyperlink, 'Hyperlink', {
-  /** specifies the text or element that a URL should be associated with */
-  content: {
-    deprType: DeprTypes.MOVED,
-    newName: 'children',
-  },
```
adamstankiewicz, line 135: 🎉 great to clean this up :) related to removing deprecated stuff, I'm looking forward to having all the deprecated UI components removed as part of the `alpha` design tokens release, too.

src/Hyperlink/index.tsx
```diff
@@ -1,29 +1,45 @@
...
-    target,
-    onClick,
-    externalLinkAlternativeText,
-    externalLinkTitle,
-    variant,
-    isInline,
-    showLaunchIcon,
-    ...attrs
-  } = props;
+interface Props extends Omit<React.ComponentPropsWithoutRef<'a'>, 'href' | 'target'> {
```
adamstankiewicz, line 10: [clarification] `Hyperlink` uses `forwardRef` below, yet the `Props` interface is defining it as `React.ComponentPropsWithoutRef`; should it be using `ComponentPropsWithRef`?

adamstankiewicz:
> @adamstankiewicz I'd like to merge this but I don't want to get in the way of #3050
> 
> Would this cause any problems with the work being done towards that?

I hadn't quite gotten to creating that PR last week; still hope to this week. This work shouldn't cause any problems. The fix for the #3050 is additive (i.e., would introduce a new `as` prop), where if this PR merges first, the PR for #3050 would ensure the `as` prop is appropriately typed.

Added a couple other review comments onto the PR.

src/Hyperlink/index.tsx
bradenmacdonald: Good question! In this case, the `forwardRef` wrapper is overriding the type and injecting the `ref` prop anyways, so it doesn't really matter whether we use `WithoutRef` or `WithRef` here - the type that will be seen elsewhere in the code is the same. But I think it makes sense to use `withRef` since we are dealing with refs, and it's perhaps less confusing that way. So I just pushed a small commit to change it.

src/Hyperlink/index.tsx
bradenmacdonald: Thanks for the context on i18n! I'm glad to see there is something in place, at least partially. I can look at fixing that in a follow-up PR.

Also great to see there's a plan to support `Link` :)

src/setupTest.ts
```diff
@@ -20,6 +21,6 @@ class ResizeObserver {
 
 window.ResizeObserver = ResizeObserver;
 
-window.crypto = {
-  getRandomValues: arr => crypto.randomBytes(arr.length),
+(window as any).crypto = {
+  getRandomValues: (arr: any) => crypto.randomBytes(arr.length),
```
bradenmacdonald, line 25: @adamstankiewicz I see you added this `getRandomValues` stub in https://github.com/openedx/paragon/commit/eac3925bcd7710e711f3988a504834f28a93060e , but I noticed just now that the test suite passes just fine if I completely remove this line. Is it still necessary/useful?

bradenmacdonald:
Thanks for the quick reviews @brian-smith-tcril  and @adamstankiewicz ! Could one of you please merge this at your convenience? :)

adamstankiewicz (approved):

src/Hyperlink/Hyperlink.test.tsx
```diff
@@ -4,30 +4,34 @@ import userEvent from '@testing-library/user-event';
...
-  target: '_blank',
+  target: '_blank' as const,
   externalLinkAlternativeText,
   externalLinkTitle,
   ...props,
 };
 
 describe('correct rendering', () => {
+  beforeEach(() => {
+    onClick.mockClear();
```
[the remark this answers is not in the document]
adamstankiewicz: [inform] When I open my PR to add support for `Link` usage with `Hyperlink`, I will plan to make this change (as well as remove the `window.crypto` stuff in `setupTest.js` per the other comment).

openedx-webhooks:
@bradenmacdonald 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/paragon&entry.867055334=https://github.com/openedx/paragon/pull/3077&entry.1484655318=https://github.com/bradenmacdonald&entry.752974735=2024-05-17+21:50&entry.1917517419=2024-05-22+13:00&entry.2133058324=Yes) so we can improve your experience in the future.

PR #3082, opened 2024-05-29
feat: introduce `as` prop for Hyperlink

### LINKED ISSUE #3050: [paragon-openedx.netlify.app] Feedback (on /components/hyperlink/)

adamstankiewicz:
## Description

Introduces `as` prop and types for `Hyperlink` to make it compatible with React Router's `Link`. The previously required `destination` prop is now only conditionally required when the `as` prop is a standard anchor tag (i.e., `a`).

If `Hyperlink` is used with `Link`, consumers would be passing the `to` prop, but Paragon shouldn't hardcode any router-specific props in its codebase.

### Deploy Preview

https://deploy-preview-3082--paragon-openedx.netlify.app/components/hyperlink

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/Hyperlink/Hyperlink.test.tsx
+++ b/src/Hyperlink/Hyperlink.test.tsx
@@ -1,12 +1,12 @@
-import React from 'react';
+import { IntlProvider } from 'react-intl';
 import { render } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
-import Hyperlink from '.';
+import Hyperlink, { HyperlinkProps } from '.';
 
-const destination = 'destination';
+const destination = 'http://destination.example';
 const content = 'content';
-const onClick = jest.fn();
+const onClick = jest.fn().mockImplementation((e) => e.preventDefault());
 const props = {
   destination,
   onClick,
@@ -20,13 +20,37 @@ const externalLinkProps = {
   ...props,
 };
 
+interface LinkProps extends HyperlinkProps {
+  to: string;
+}
+
+function Link({ to, children, ...rest }: LinkProps) {
+  return (
+    <a
+      data-testid="custom-hyperlink-element"
+      href={to}
+      {...rest}
+    >
+      {children}
+    </a>
+  );
+}
+
+function HyperlinkWrapper({ children, ...rest }: HyperlinkProps) {
+  return (
+    <IntlProvider locale="en">
+      <Hyperlink {...rest}>{children}</Hyperlink>
+    </IntlProvider>
+  );
+}
+
 describe('correct rendering', () => {
   beforeEach(() => {
-    onClick.mockClear();
+    jest.clearAllMocks();
   });
 
   it('renders Hyperlink', async () => {
-    const { getByRole } = render(<Hyperlink {...props}>{content}</Hyperlink>);
+    const { getByRole } = render(<HyperlinkWrapper {...props}>{content}</HyperlinkWrapper>);
     const wrapper = getByRole('link');
     expect(wrapper).toBeInTheDocument();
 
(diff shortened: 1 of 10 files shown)


src/Hyperlink/index.tsx
```diff
@@ -1,13 +1,17 @@
...
 import { Launch } from '../../icons';
 import Icon from '../Icon';
 
 export const HYPER_LINK_EXTERNAL_LINK_ALT_TEXT = 'in a new tab';
 export const HYPER_LINK_EXTERNAL_LINK_TITLE = 'Opens in a new tab';
 
-interface Props extends Omit<React.ComponentPropsWithRef<'a'>, 'href' | 'target'> {
+interface HyperlinkProps extends BsPrefixProps, Omit<React.ComponentPropsWithRef<'a'>, 'href' | 'target'> {
   /** specifies the URL */
   destination: string;
```
adamstankiewicz, line 17: When using `Hyperlink` with `Link` from React Router (or `GatsbyLink` on the Paragon docs site example), the `destination` prop will not suffice as it gets passed to an `href` prop.

Both `Link` and `GatsbyLink` utilize the standard hyperlink (i.e., `a`) when `href` prop is provided; to get it to work with internal routing, the `to` prop must be passed.

Because we use prop spreading of all attributes passed to the `Hyperlink` component, consumers _can_ pass `to` prop but would continue to need to pass `destination` as it's a required prop.

We will need to determine how consumers can pass `to` without needing to also pass an identical value to `destination` (i.e., `href`), e.g.:

```jsx
<Hyperlink
  as={GatsbyLink}
  // `destination` is still a required prop even though the `to` takes precedence.
  destination="/components/button"
  to="/components/button"
>
  Button
</Hyperlink>
```

adamstankiewicz: @bradenmacdonald Curious if you might have any thoughts on how to mitigate this concern around `to` vs. `href` vs. `destination`?

bradenmacdonald: Yeah, I was wondering about this.

I think one nice option is to change the Paragon `hyperlink` to accept `to` instead of `destination` and just treat `destination` as a deprecated alias for `to` ?

Otherwise to avoid the need to introspect `as` and determine what prop to pass based on some conditions, you could change the logic that currently converts `destination` to `href` and instead convert `destination` to _both_ `href` and `to`, and unconditionally pass both props to the inner `as` component. Though I think this may display a DOM warning in the console that `a` doesn't accept a `to` attribute. So maybe some conditional logic is actually needed.

src/Hyperlink/index.tsx
```diff
@@ -88,31 +117,27 @@ const Hyperlink = React.forwardRef<HTMLAnchorElement, Props>(({
...
-  variant: 'default',
-  isInline: false,
-  showLaunchIcon: true,
-};
-
 Hyperlink.propTypes = {
-  /** specifies the URL */
-  destination: PropTypes.string.isRequired,
+  /** specifies the component element type to render for the hyperlink */
+  // @ts-ignore
```
adamstankiewicz, line 132: [inform] It's unclear why TS throws an error on the `as` prop type; as its functional, opting to `@ts-ignore` for now.

src/Hyperlink/Hyperlink.test.tsx
```diff
@@ -66,19 +107,8 @@ describe('correct rendering', () => {
...
   it('prevents reverse tabnabbing for links with target="_blank"', () => {
-    const { getByRole } = render(<Hyperlink {...externalLinkProps}>{content}</Hyperlink>);
+    const { getByRole } = render(<HyperlinkWrapper {...externalLinkProps}>{content}</HyperlinkWrapper>);
     const wrapper = getByRole('link');
     expect(wrapper).toHaveAttribute('rel', 'noopener noreferrer');
   });
 });
-
-describe('event handlers are triggered correctly', () => {
-  it('should fire onClick', async () => {
```
brian-smith-tcril, line 76: It looks like `onClick` is still something we have an example for on the docs site https://deploy-preview-3082--paragon-openedx.netlify.app/components/hyperlink/#with-onclick

I think it might be worth keeping this test to cover the off chance someone accidentally removes `onClick={onClick}` from the `Hyperlink` component.

src/Menu/__snapshots__/Menu.test.jsx.snap
```diff
@@ -95,7 +95,6 @@ exports[`Menu Item renders correctly renders as expected with menu items 1`] = `
   <a
     className="pgn__hyperlink default-link standalone-link pgn__menu-item"
     href="https://en.wikipedia.org/wiki/Hyperlink"
-    onClick={[Function]}
```
brian-smith-tcril, line 98: While I think it's absolutely great that we don't have "bonus" `onClick`s anymore, it's not clear to me what caused this change. Any insight into this would be awesome!

brian-smith-tcril (commented):
Overall this is looking great! Super happy to see this so close to landing.

I left a couple little comments with questions about `onClick`, any thoughts on those would be great!

src/Hyperlink/Hyperlink.test.tsx
adamstankiewicz: I removed this test case, as its behavior is technically already covered within the "renders Hyperlink" test case, i.e.:

```jsx
// Clicking on the link should call the onClick handler
await userEvent.click(wrapper);
expect(onClick).toHaveBeenCalledTimes(1);
```

src/Menu/__snapshots__/Menu.test.jsx.snap
adamstankiewicz: The removal of the "bonus" `onClick` was caused by changing the `defaultProps.onClick` to `undefined` instead of the no-op function `() => {}`. I'm not sure having the no-op function is providing much value, so opted to remove it since its only a pass-thru prop, and isn't a breaking change.

For example, we don't do any conditional checks before calling it within `Hyperlink` (e.g., nothing like, `onClick.?(e)` implemented).

brian-smith-tcril (approved):
:shipit:

PR #3158, opened 2024-07-31
chore: merge alpha into master

PKulkoRaccoonGang:
## Description

- merged alpha into master
- fixed `--pgn-typography-font-weight-table-th` variable

## Merge Checklist

* [ ] If your update includes visual changes, have they been reviewed by a designer? Send them a link to the Netlify deploy preview, if applicable.
* [ ] Does your change adhere to the documented [style conventions](https://github.com/openedx/paragon/blob/master/docs/decisions/0012-css-styling-conventions)?
* [ ] Do any prop types have missing descriptions in the Props API tables in the documentation site (check deploy preview)?
* [ ] Were your changes tested using all available themes (see theme switcher in the header of the deploy preview, under the "Settings" icon)?
* [ ] Were your changes tested in the `example` app?
* [ ] Is there adequate test coverage for your changes?
* [ ] Consider whether this change needs to reviewed/QA'ed for accessibility (a11y). If so, please request an a11y review for the PR in the [#wg-paragon](https://openedx.slack.com/archives/C02NR285KV4) Open edX Slack channel.

## Post-merge Checklist

* [ ] Verify your changes were released to [NPM](https://www.npmjs.com/package/@openedx/paragon) at the expected version.
* [ ] If you'd like, [share](https://github.com/openedx/paragon/discussions/new?category=show-and-tell) your contribution in [#show-and-tell](https://github.com/openedx/paragon/discussions/categories/show-and-tell).
* [ ] 🎉 🙌 Celebrate! Thanks for your contribution.

PR #3186, opened 2024-08-20
feat: updating to Style Dictionary v4

PKulkoRaccoonGang:
## Description

This PR is an attempt to update the Style Dictionary v3 library to Style Dictionary v4 (including refactoring the design tokens to the DTCG format) + fixed some problems with typography.

## Merge Checklist

## Post-merge Checklist


openedx-webhooks:

### What's next?

*Please work through the following steps to get your changes ready for engineering review:*

#### :radio_button: Get product approval

#### :radio_button: Provide context

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation





#### :radio_button: Get a green build

#### :radio_button: Let us know that your PR is ready for review:

### Who will review my changes?


This repository is currently maintained by `@openedx/paragon-working-group`.  Tag them in a comment and let them know that your changes are ready for review.


### Where can I find more information?

### When can I expect my changes to be merged?

- The need for product review

PKulkoRaccoonGang:
**TODO**
- correct display of the header of generated files
- checking the correctness of variable generation (for some variables you can notice an extra indication of the unit of measurement, for example 0rem)
 ...

**ES6 modules**
Since the updated Style Dictionary v4 API is based on [ES6 modules](https://v4.styledictionary.com/version-4/migration/#file-headers), an open question is finding a solution for the correct operation of Style Dictionary v4 together with the existing CommonJS functionality of Gatsby. According to [the Gatsby documentation](https://www.gatsbyjs.com/docs/how-to/custom-configuration/es-modules/), support for ES6 modules is implemented in more recent releases of the framework. Most likely, for the current version of Gatsby (`4.23.1`), a potential solution could be to use asynchronous imports or update the Gatsby version.

brian-smith-tcril:
Thank you so much for putting this together! Considering the task of "do some discovery around v4" this is truly above and beyond!

> checking the correctness of variable generation (for some variables you can notice an extra indication of the unit of measurement, for example 0rem)

I think it's reasonable to update the `.json` files to include units where we don't yet have them as a solution here.

> Since the updated Style Dictionary v4 API is based on [ES6 modules](https://v4.styledictionary.com/version-4/migration/#file-headers), an open question is finding a solution for the correct operation of Style Dictionary v4 together with the existing CommonJS functionality of Gatsby. According to [the Gatsby documentation](https://www.gatsbyjs.com/docs/how-to/custom-configuration/es-modules/), support for ES6 modules is implemented in more recent releases of the framework. Most likely, for the current version of Gatsby (4.23.1), a potential solution could be to use asynchronous imports or update the Gatsby version.

Could you elaborate on this a bit? I worry that moving from Gatsby v4 to v5 would be quite a heavy lift, and I wonder how that compares to some workarounds people have tried https://github.com/gatsbyjs/gatsby/discussions/31599#discussioncomment-1239988

package.json
```diff
@@ -198,10 +198,10 @@
     ]
   },
   "lint-staged": {
-    "*.{js,jsx,ts,tsx}": [
+    "*.{mjs}": [
       "npx eslint"
     ],
-    "*.scss": [
+    "*.sass": [
```
adamstankiewicz, line 204: [curious] Can you clarify these changes a bit? Why are they needed?

styles/css/core/variables.css
```diff
@@ -70,26 +78,25 @@
   --pgn-typography-font-size-sm: .875rem;
   --pgn-typography-font-size-lg: 1.4063rem;
   --pgn-typography-font-size-base: 1.125rem;
-  --pgn-typography-font-family-monospace: SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono",  "Courier New", monospace;
+  --pgn-typography-font-family-monospace: SFMono-Regular, Menlo, Monaco, Consolas, '"Liberation Mono"', '"Courier New"', monospace;
```
adamstankiewicz, line 81: [clarification] Is having the already quoted font names now wrapped with single quotes, too intended/needed/work as expected?

tokens/src/core/components/ActionRow.json
```diff
@@ -1,9 +1,10 @@
 {
+  "$type": "dimension",
```
adamstankiewicz, line 2: [clarification] Will _all_ tokens defined for `ActionRow` be of `"$type": "dimension"`? Should this be associated more directly to the dimension-specific tokens, even though these are the only tokens in this file right now?

For example, if/when a non-dimension token is added to this file, is there a risk this `$type` won't be semantically correct/accurate anymore?

Similar question like applies to other components' tokens, too.

tokens/src/core/components/Card.json
```diff
@@ -1,95 +1,82 @@
...
-          "base": { "value": "{size.border.radius.base}", "type": "dimension", "source": "$card-border-radius" },
-          "logo": { "value": ".25rem", "type": "dimension", "source": "$card-logo-border-radius" },
-          "inner": {
-            "value": "calc({size.card.border.radius.base} - {size.card.border.width})",
-            "type": "dimension",
-            "source": "$card-inner-border-radius"
-          }
+          "base": { "source": "$card-border-radius", "$value": "{size.border.radius.base}" },
+          "logo": { "source": "$card-logo-border-radius", "$value": ".25rem" },
+          "inner": { "source": "$card-inner-border-radius", "$value": "calc({size.card.border.radius.base} - {size.card.border.width})" }
```
adamstankiewicz, line 49: [curious] Have we looked into whether we can avoid including CSS-specific syntax `calc` in the token itself, or how such tokens would be transformed outside of CSS variables?

For example, if there is arithmetic like subtraction as in this line, could the token `$value` be `{size.card.border.radius.base} - {size.card.border.width}` and have some sort of transform for the CSS variables output to wrap it with `calc(...)`?

That way, tokens like these would be more portable to other non-CSS based platforms like mobile.

tokens/src/core/components/general/link.json
```diff
@@ -1,27 +1,28 @@
 {
+  "$type": "textDecoration",
```
adamstankiewicz, line 2: [clarification] Similar to other comment about the `dimension` type defined more globally in a file, will _all_ tokens in this file be related to text-decoration? It would seem perhaps the `$type` should be defined within the `decoration` sub-object?

tokens/src/core/global/other.json
```diff
@@ -1,4 +1,8 @@
 {
+  "$type": "percentage",
```
adamstankiewicz, line 2: [clarification] only the `theme-interval` token specifically in this file is a percentage, right?

tokens/src/core/global/typography.json
```diff
@@ -2,89 +2,222 @@
...
       },
       "size": {
-        "base": { "value": "1.125rem", "type": "dimension", "source": "$font-size-base", "description": "Base font size." },
-        "lg": { "value": "1.4063rem", "type": "dimension", "source": "$font-size-lg", "description": "Large font size." },
-        "sm": { "value": ".875rem", "type": "dimension", "source": "$font-size-sm", "description": "Small font size." },
-        "xs": { "value": ".75rem", "type": "dimension", "source": "$font-size-xs", "description": "Extra-small font size." },
+        "base": {
+          "source": "$font-size-base",
+          "$value": "1.125rem",
+          "$type": "dimension",
```
adamstankiewicz, line 33: [suggestion] could the `"$type": "dimension"`, be moved up under `size` so it applies to all size permutations (e.g., base vs. lg vs. sm)?

tokens/src/themes/light/components/Avatar.json
```diff
@@ -1,7 +1,8 @@
 {
+  "$type": "color",
   "color": {
```
adamstankiewicz, line 3: [curious] Based on the [migration guidelines](https://styledictionary.com/version-4/migration/#cti-reliance):

image

It _seems_ the top-level `"color"` key here isn't actually needed anymore for `style-dictionary` to determine the CTI (Category-Type-Item), given the `$type` annotation. Is this something that was explored? It might help simplify JSON files slightly by removing possible layer(s) of nesting. If it is possible to nix the extra nesting layer, curious to hear your thoughts around scope in tackling that broadly.

tokens/style-dictionary.js
```diff
@@ -65,165 +80,188 @@ const colorTransform = (token, theme) => {
...
  * core tokens are built.
  */
-const createCustomCSSVariables = ({
+const createCustomCSSVariables = async ({
   formatterArgs,
   themeVariant,
-}) => {
+}, StyleDictionary) => {
+  // eslint-disable-next-line import/no-unresolved
+  const { sortByReference, usesReferences, getReferences } = (await import('style-dictionary/utils'));
```
adamstankiewicz, line 88: I wonder if it'd make sense to abstract a helper util function to make these utils accessible, e.g. 

```jsx
/* eslint-disable import/no-unresolved */
const getStyleDictionaryUtils = async () => import('style-dictionary/utils');
/* eslint-enabled import/no-unresolved */

// createCustomCSSVariables
const { sortByReference, usesReferences, getReferences } = await getStyleDictionaryUtils();
```

`getStyleDictionaryUtils` could also be used within `initializeStyleDictionary`, e.g.:

```jsx
const initializeStyleDictionary = async () => {
  // eslint-disable-next-line import/no-unresolved
  const StyleDictionary = (await import('style-dictionary')).default;
  const { getReferences } = await getStyleDictionaryUtils();
 
  // ...
}
```

tokens/src/core/global/spacing.json
```diff
@@ -1,36 +1,89 @@
...
+      "6": {
+        "$value": "calc({spacing.spacer.base} * 5)",
+        "$description": "Space value of level 6"
+      },
+      "base": {
+        "source": "$spacer",
+        "$value": "1rem",
+        "$description": "Basic space value"
+      },
+      "1,5": {
```
adamstankiewicz, line 38: [clarification] Is the comma in `"1,5"` and other spacing tokens intended, or should it be a period (i.e., `"1.5"`)?

tokens/style-dictionary.js
```diff
@@ -65,165 +80,188 @@ const colorTransform = (token, theme) => {
...
   const { dictionary, options, file } = formatterArgs;
-
   const outputTokens = themeVariant
     ? dictionary.allTokens.filter(token => token.filePath.includes(themeVariant))
     : dictionary.allTokens;
 
   const variables = outputTokens.sort(sortByReference(dictionary)).map(token => {
```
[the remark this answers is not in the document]
adamstankiewicz: Related, upon further investigation, it looks we _do_ rely on token-specific handling of `outputReferences`. For example, we opt-out of including references even when the formatter has them opted-in (e.g., for the case of always including HEX values in SVG `url` values).

That said, token-specific `outputReferences` is still supported with the out-of-the-box `formattedVariables`, by passing `outputReferences` a function instead of a boolean, e.g.:

```jsx
const variables = formattedVariables({
  format: 'css',
  dictionary,
  outputReferences: (token) => {
    // Formatter options suggest to never output references
    if (!outputReferences) {
      return false;
    }
    // Formatter options suggest to output references, but individual token might opt-out
    if (!token.outputReferences) {
      return false;
    }
    // Otherwise, output references.
    return true;
  },
  usesDtcg: true,
});
```

tokens/src/themes/light/components/Avatar.json
adamstankiewicz: Note: I might be misinterpreting this, where the top-level `color` is still needed from a grouping perspective; just not from a type pov.

package.json
PKulkoRaccoonGang: I needed to ignore the pre-commit hook)
This is a temporary change that I will remove 💯

tokens/style-dictionary.js
adamstankiewicz: [inform] Implemented in follow-up PR: https://github.com/openedx/paragon/pull/3203

tokens/style-dictionary.js
```diff
@@ -65,165 +80,188 @@ const colorTransform = (token, theme) => {
...
+   */
+  StyleDictionary.registerFileHeader({
+    name: 'customFileHeader',
+    fileHeader: () => {
+      const currentDate = new Date().toUTCString();
+      return [
+        '/*',
+        ' * IMPORTANT: This file is the result of assembling design tokens.',
+        ' * Do not edit directly.',
+        ` * Generated on ${currentDate}`,
```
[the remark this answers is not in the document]
adamstankiewicz: [inform] Implemented in follow-up PR: https://github.com/openedx/paragon/pull/3203

tokens/style-dictionary.js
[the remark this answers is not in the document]
adamstankiewicz: [inform] Implemented in follow-up PR: https://github.com/openedx/paragon/pull/3203

tokens/src/core/components/Card.json
adamstankiewicz: Dug into this a bit. It appears this still remains a complex problem within the `style-dictionary` community. There is support for a `resolveMath` transform, from `tokens-studio/sd-transforms`; however, when the token outputs references via `outputReferences: true | fn`, the transformed/resolved math is replaced with the original reference (i.e., a CSS variable with `var(--foo: 5px)` in this case).

I attempted to craft a solution for https://github.com/tokens-studio/sd-transforms/issues/203, but came to the same realization recently described [here](https://github.com/amzn/style-dictionary/issues/1055#issuecomment-2312072149):

> The big issue is that outputting refs happens on the format lifecycle and wrapping values in calc() statements usually happens in the transform lifecycle that happens before. So even if you wrap the values in calc() in transform, the format part will just undo that work by outputting refs by using the original value.
>
> Either the calc wrapping needs to happen on the format level or the outputting refs util needs to act on the transformed value somehow (keeping in mind the bugs that this used to cause in v3 and not regressing on this again)
>
> There are a couple of open issues to rethink how references are resolved and how values with references in them are transformed for a future v5 version of Style Dictionary, but this topic is rather complex and this issue won't be fixed until we have solutions to that broader topic.

tl;dr; Handling the wrapping of math expressions with the CSS `calc` syntax remains an open question, and likely will not land in `v4` of Style Dictionary per the above. Given this, it probably makes sense to keep the `calc` syntax for now, but when we begin transforming to non-CSS platforms (e.g., JavaScript, iOS, Android), having `calc` in the token value itself will present an issue. A possible workaround for if/when this becomes an issue might be to introduce custom transforms for the (future) non-CSS platforms that strips the wrapping `calc(...)` from the underlying math operations.

PKulkoRaccoonGang: This is definitely a very interesting question that we haven't thought about yet. Thanks for your research, now we know more!

styles/css/core/variables.css
PKulkoRaccoonGang: Resolved in https://github.com/openedx/paragon/pull/3207

tokens/src/core/components/ActionRow.json
PKulkoRaccoonGang: Resolved in https://github.com/openedx/paragon/pull/3207

tokens/src/core/components/general/link.json
PKulkoRaccoonGang: Resolved in https://github.com/openedx/paragon/pull/3207

tokens/src/core/global/other.json
PKulkoRaccoonGang: Resolved in https://github.com/openedx/paragon/pull/3207

tokens/src/core/global/typography.json
PKulkoRaccoonGang: [inform] Implemented in follow-up PR: https://github.com/openedx/paragon/pull/3203

tokens/src/core/global/spacing.json
PKulkoRaccoonGang: It doesn't matter for the final result. But I agree with you, I replaced the comma with a period (by analogy to the value in the description)

[inform] Implemented in follow-up PR: https://github.com/openedx/paragon/pull/3207

tokens/src/core/global/typography.json
```diff
@@ -2,90 +2,284 @@
...
+          "$value": "1.125rem",
+          "$description": "Base font size."
         },
-        "lead": {
-          "value": "calc({typography.font.size.base} * 1.25)",
-          "type": "dimension",
+        "lg": {
           "source": "$lead-font-size",
-          "description": "Lead text font size."
+          "$value": "{typography.font.size.base} * 1.25",
```
adamstankiewicz, line 38: D'oh, I think in my commit I lost the `calc(...)` that was previously wrapping this token's `$value` (this was the only token I experimented with). When previewing the docs site locally, the `.lead` utility class isn't working properly anymore, noticed first on the home page. Adding `calc(...)` back in will resolve.

[context] This was removed while exploring the `sd-transforms` transform for `ts/resolveMath` to see if we could abstract away `calc`, but as discussed in another comment, there is not currently a great solution for this, so we need to keep `calc` in here, at least for now :)

tokens/src/core/components/Code.json
```diff
@@ -1,32 +1,33 @@
 {
   "typography": {
     "code": {
-      "font-size": { "value": "87.5%", "type": "percentage", "source": "$code-font-size" },
+      "font-size": { "source": "$code-font-size", "$value": "87.5%", "$type": "percentage" },
```
adamstankiewicz, line 4: [suggestion] I believe this `87.5%` is equivalent to the `{typography.font.size.sm}` if we wanted to treat this token an alias to a semantic token (i.e., if/when `{typography.font.size.sm}` changes, this token will also change).

tokens/src/core/global/typography.json
```diff
@@ -2,90 +2,284 @@
...
+        },
+        "sm": {
+          "source": "$small-font-size",
+          "$value": ".875rem",
+          "$description": "Small font size."
+        },
+        "xs": {
+          "source": "$x-small-font-size",
+          "$value": ".75rem",
+          "$description": "X-small font size."
```
adamstankiewicz, line 49: Comparing the computed typography values in the docs for this PR's local checkout vs. published docs site, it looks like we might actually want/need to keep the `%` scale these tokens originally had.

For example, `x-small` used to be 75% but we migrated it to `.75rem`, which seems to result in slightly different font sizes. Note the `.small` and `.x-small` examples specifically (left: local; right: prod):

image

[aside] Related, you might notice the line heights are also slightly off for some of them as well. This seems to be due to the switch to `DataTable`, where some of the cell content is inheriting `DataTable`'s line-height styles, i.e. `.pgn__data-table td` (left: local; right: prod):

image

I suppose we could consider overriding `DataTable`'s default CSS styles for typography example tables by `unset`'ing the line-height related definition(s), or maybe cross-check whether the CSS utility classes are defining explicit `line-height` style properties vs. inheriting from its parent.

tokens/src/core/global/typography.json
```diff
@@ -2,90 +2,284 @@
...
+            "source": "$display2-size",
+            "$value": "4.875rem",
+            "$description": "Font size for heading of level 2."
+          },
+          "3": {
+            "source": "$display3-size",
+            "$value": "5.625rem",
+            "$description": "Font size for heading of level 3."
+          },
+          "4": {
```
adamstankiewicz, line 145: On a local checkout of the docs site from this branch, it looks like the migrate of typography example tables to use `DataTable` is inheriting its `overflow: hidden` on table cells, cutting off the text of "Display 4" on smaller screen sizes. Probably not a big deal, but figured I'd flag it as it could be perceived as a style bug potentially.

image

image

tokens/src/core/components/general/link.json
```diff
@@ -2,26 +2,27 @@
   "typography": {
     "link": {
       "decoration": {
-        "base": { "value": "none", "type": "textDecoration", "source": "$link-decoration" },
-        "hover": { "value": "underline", "type": "textDecoration", "source": "$link-hover-decoration" },
+        "$type": "textDecoration",
+        "base": { "source": "$link-decoration", "$value": "none" },
```
adamstankiewicz, line 6: [inform, docs site bug] The "Links" section on the typography foundations page is incorrectly showing `text-decoration: underline;` on "standalone links" (left: local; right: prod):

image

It appears this is due to the migration of the example tables to `DataTable`, where the `TextCell` is wrapping the cell content with a `p`, which according to the table description, "For links inside a `p`...", the standalone links are actually getting treated as inline links even without the explicit `.inline-link` class.

tokens/src/core/global/typography.json
PKulkoRaccoonGang: [inform] Implemented in follow-up PR: https://github.com/openedx/paragon/pull/3222

tokens/src/core/components/Code.json
PKulkoRaccoonGang: [inform] Implemented in follow-up PR: https://github.com/openedx/paragon/pull/3222

tokens/src/core/global/typography.json
PKulkoRaccoonGang: Yes, we will definitely need to do a full check of all tokens after the latest changes, as we can easily miss some small detail that can have a significant impact on both the components and the documentation site. Thanks! 👍

tokens/src/core/components/Code.json
PKulkoRaccoonGang: A small difference exists between `87.5%` and `0.875rem`. The `87.5%` value is calculated relative to the font size of the parent element. For example, if the parent has a font size of `16px`, then `87.5%` will be equivalent to `14px`. However, if the parent font size is different, the resulting pixel value will change accordingly.

In contrast, `0.875rem` is always calculated relative to the root element’s (`<html>`) base font size, which is typically `16px` by default in browsers. Thus, `0.875rem` will consistently be equivalent to `14px` as long as the html font size is `16px`, regardless of the parent element’s font size.

Currently, when using the `0.875rem` value from the token `{typography.font.size.sm}`, the value is indeed calculated relative to the root font size of the html tag (`16px` = `1rem`). The previous percentage value of `87.5%` was calculated based on the parent font size of `1.125rem` (the font size of the body).

I suggest using `87.5%`, what do you think?

tokens/src/core/components/general/link.json
PKulkoRaccoonGang: [inform] Implemented in follow-up commit: [`6676868` (#3186)](https://github.com/openedx/paragon/pull/3186/commits/6676868e6c46d70d91f0bdc42b7c2848e3b0aa7c)

tokens/src/core/global/typography.json
PKulkoRaccoonGang: [inform] Implemented in follow-up commit: [`6676868` (#3186)](https://github.com/openedx/paragon/pull/3186/commits/6676868e6c46d70d91f0bdc42b7c2848e3b0aa7c)

www/src/components/typography-page/Links.tsx
```diff
@@ -34,6 +31,9 @@ export default function Links() {
   return (
     <>
       <h2 className="mb-2">Links</h2>
+      <Alert variant="info" icon={Info}>
+        <a href="/#">Standalone Link</a> - the default style for <code>a</code> tag that appear outside of <code>p</code> tag.
+      </Alert>
```
PKulkoRaccoonGang, line 36: image

tokens/src/core/global/typography.json
PKulkoRaccoonGang: [inform] Implemented in follow-up commit: [6676868 (#3186)](https://github.com/openedx/paragon/pull/3186/commits/6676868e6c46d70d91f0bdc42b7c2848e3b0aa7c)

tokens/src/core/components/Code.json
PKulkoRaccoonGang: [inform] Implemented in follow-up commit: [6676868 (#3186)](https://github.com/openedx/paragon/pull/3186/commits/6676868e6c46d70d91f0bdc42b7c2848e3b0aa7c)

PKulkoRaccoonGang:
@brian-smith-tcril @adamstankiewicz In [this commit](https://github.com/openedx/paragon/pull/3186/commits/da00992533e1ac51b1e9bd53b43e8c50239e8a35#diff-1b8d91c59681705b700c37ce0d260f55ab525bf6174bb2686ecc37873e1897e0L1) we removed the brand-openedx theme. Do we want to remove the sidebar toggle button and any mention of the brand in Paragon?

image

tokens/src/themes/light/components/Carousel.json
```diff
@@ -1,43 +1,44 @@
...
       "control": {
         "bg": {
           "prev-icon": {
-            "value": "url(\"data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' fill='{color.carousel.control.base}' viewBox='0 0 8 8'%3e%3cpath d='M5.25 0l-4 4 4 4 1.5-1.5-2.5-2.5 2.5-2.5-1.5-1.5z'/%3e%3c/svg%3e\")",
-            "type": "file",
             "source": "$carousel-control-prev-icon-bg",
             "outputReferences": false,
-            "modify": [{ "type": "str-replace", "toReplace": "#", "replaceWith": "%23" }]
+            "modify": [{ "type": "str-replace", "toReplace": "#", "replaceWith": "%23" }],
+            "$value": "url(\"data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' fill='{color.carousel.control.base}' viewBox='0 0 8 8'%3e%3cpath d='M5.25 0l-4 4 4 4 1.5-1.5-2.5-2.5 2.5-2.5-1.5-1.5z'/%3e%3c/svg%3e\")"
```
brian-smith-tcril, line 23: I'm curious about the hardcoded `%3c` for `<` and `%3e` for `>` here. I'm not 100% sure how `str-replace` in `modify` works, could that be used for those similarly to how it's being used for `#`?

brian-smith-tcril (commented):
Overall this is looking really great! I left a few comments with questions, but it's awesome to see v4 so close to ready!

bin/paragon-scripts.js
```diff
@@ -85,11 +85,21 @@ const COMMANDS = {
         description: 'Include only source design tokens in the build.',
         defaultValue: false,
       },
+      {
+        name: '--output-references',
+        description: 'Include references in the build output.',
```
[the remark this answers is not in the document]
adamstankiewicz: @brian-smith-tcril Would it be helpful to include "Include **token** references in the build output"? This is getting at whether a token who's `$value` references another token should result in the CSS file output as a CSS variable or a hardcoded value. 

`--output-references` == include CSS variables for tokens with aliases to other tokens.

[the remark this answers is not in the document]
adamstankiewicz: Related, if that distinction of **token reference** is helpful, could rename the argument flag altogether to `--include-token-references`.

lib/build-tokens.js
```diff
@@ -41,33 +59,35 @@ async function buildTokensCommand(commandArgs) {
     platforms: {
       css: {
         prefix: 'pgn',
-        transformGroup: 'css',
```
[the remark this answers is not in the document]
adamstankiewicz: Confirmed, this conversation can be resolved 👍

lib/build-tokens.js
```diff
@@ -38,36 +57,42 @@ async function buildTokensCommand(commandArgs) {
...
+    },
     platforms: {
       css: {
         prefix: 'pgn',
-        transformGroup: 'css',
+        transformGroup: 'paragon-css',
         // NOTE: buildPath must end with a slash
         buildPath: buildDir.slice(-1) === '/' ? buildDir : `${buildDir}/`,
+        options: {
+          fileHeader: 'customFileHeader',
```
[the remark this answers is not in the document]
adamstankiewicz: The following is the `customFileHeader` config:

```
/**
 * Registers a custom fileHeader.
 */
StyleDictionary.registerFileHeader({
  name: 'customFileHeader',
  fileHeader: (defaultMessage) => [
    `${defaultMessage} while transforming design tokens.`,
    'See <root>/tokens/README.md for more details.',
  ],
});
```

IIRC, the current file header is where we want to be for now (i.e., not including timestamps, including a docs reference to learn more).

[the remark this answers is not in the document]
brian-smith-tcril: Awesome! Thanks for remembering/confirming!

styles/css/core/custom-media-breakpoints.css
```diff
@@ -1,10 +1,9 @@
 /**
- * IMPORTANT: This file is the result of assembling design tokens
- * Do not edit directly
- * Generated on Fri, 22 Sep 2023 08:20:17 GMT
+ * Do not edit directly, this file was auto-generated. while transforming design tokens.
+ * See <root>/tokens/README.md for more details.
```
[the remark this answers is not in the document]
brian-smith-tcril: Resolved as per https://github.com/openedx/paragon/pull/3186#discussion_r1823114970

bin/paragon-scripts.js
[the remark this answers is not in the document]
brian-smith-tcril: I think I like "Include CSS variables for tokens with aliases to other tokens." as the description. I'm fine with keeping the argument as `--output-references`, but `--include-token-references` or `--output-token-references` work too.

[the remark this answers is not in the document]
adamstankiewicz: My concern about mentioning "CSS variables" specifically is that the long term version is that we're not just outputting CSS variables, but in the future also formatted tokens for iOS/Android, etc. where "CSS variables" doesn't make much sense.

Similarly, if a consumer were to extend the default `style-dictionary` config to export SCSS variables from the tokens (beyond the default CSS variables), the `--output-tokens-references` is referencing SCSS variables, not CSS variables.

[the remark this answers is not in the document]
brian-smith-tcril: Good point, would just, "Include variables for tokens with aliases to other tokens." work?

styles/scss/core/_utilities.scss
```diff
@@ -21,13 +21,17 @@
 $color-levels: 100, 200, 300, 400, 500, 600, 700, 800, 900;
 
 .x-small {
-  font-size: $x-small-font-size;
+  font-size: $x-small-font-size !important;
```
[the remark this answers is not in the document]
adamstankiewicz: CSS utility classes in the design system _should_ have `!important`; check other existing CSS utility classes for reference, e.g.:

image

lib/build-tokens.js
```diff
@@ -85,49 +110,60 @@ async function buildTokensCommand(commandArgs) {
...
       css: {
         ...coreConfig.platforms.css,
         files: [
           {
             format: 'css/custom-variables',
             destination: `themes/${themeVariant}/variables.css`,
-            filter: hasSourceTokensOnly ? 'isSource' : undefined,
+            filter: hasSourceTokensOnly
+              ? `isSource.${themeVariant}`
+              : `isThemeVariant.${themeVariant}`,
```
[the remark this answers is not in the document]
adamstankiewicz: This specific `filter` pertains to theme variants, regardless of `hasSourceTokensOnly`.

For example, `isSource.${themeVariant}` is filtering for source tokens of a specific theme variant; similarly, `isThemeVariant.${themeVariant}` is filtering for all tokens (source or otherwise) for a specific theme variant.

`source` tokens are the ones from custom brand packages, i.e. the tokens to extend/modify the default tokens defined within the Paragon repo (passed via style-dictionary's `include`).

When `hasSourceTokensOnly`, it's telling style-dictionary to only transform/output tokens from the `source` tokens (aka only the custom brand package tokens, not the full set of tokens in upstream Paragon).

styles/css/core/custom-media-breakpoints.css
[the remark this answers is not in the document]
adamstankiewicz: Yep, this is the intended custom file header we landed on after a few iterations (i.e., no timestamp, include reference to README docs for more details).

tokens/src/core/components/Avatar.json
```diff
@@ -1,16 +1,16 @@
...
-      "xl": { "value": "6rem", "type": "dimension", "source": "$avatar-size-xl" },
-      "xxl": { "value": "11.5rem", "type": "dimension", "source": "$avatar-size-xxl" },
-      "huge": { "value": "18.75rem", "type": "dimension", "source": "$avatar-size-huge" },
+      "base": { "source": "$avatar-size", "$value": "3rem", "$type": "dimension" },
+      "xs": { "source": "$avatar-size-xs", "$value": "1.5rem", "$type": "dimension" },
+      "sm": { "source": "$avatar-size-sm", "$value": "2.25rem", "$type": "dimension" },
+      "lg": { "source": "$avatar-size-lg", "$value": "4rem", "$type": "dimension" },
+      "xl": { "source": "$avatar-size-xl", "$value": "6rem", "$type": "dimension" },
+      "xxl": { "source": "$avatar-size-xxl", "$value": "11.5rem", "$type": "dimension" },
+      "huge": { "source": "$avatar-size-huge", "$value": "18.75rem", "$type": "dimension" },
```
adamstankiewicz, line 10: super nit: might recommend keeping `source` the last property defined, as the other properties like `$value`, and `$type` are more critical.

bin/paragon-scripts.js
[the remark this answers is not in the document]
adamstankiewicz: I'd be OK with that, though the slight tradeoff is that we'd be slightly moving away from style-dictionary's official language around "references".

> Include variables for tokens with references to other tokens."

🤔

tokens/src/core/components/CloseButton.json
```diff
@@ -2,12 +2,8 @@
   "typography": {
     "close-button": {
       "font": {
-        "size": {
-          "value": "calc({typography.font.size.base} * 1.5)", "type": "dimension", "source": "$close-font-size"
-        },
-        "weight": {
-          "value": "{typography.font.weight.bold}", "type": "fontWeight", "source": "$close-font-weight"
-        }
+        "size": { "source": "$close-font-size", "$value": "calc({typography.font.size.base} * 1.5)", "$type": "dimension" },
```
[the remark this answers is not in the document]
adamstankiewicz: See https://github.com/openedx/paragon/pull/3186/files#r1739426413 for details on where we landed regarding `calc.`

tl;dr; there's no great alternative right now to avoid `calc` in style-dictionary tokens.

bin/paragon-scripts.js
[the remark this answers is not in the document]
brian-smith-tcril: I was actually thinking that too. My main problem with "Include references in the build output." is that it doesn't explain what references mean or what said references may be referring to. I'd be happy with "Include references for tokens with aliases to other tokens in the build output."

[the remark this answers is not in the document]
adamstankiewicz: Sold, let's roll with that :) 

> Include references for tokens with aliases to other tokens in the build output.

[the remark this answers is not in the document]
brian-smith-tcril: So now the only question is the argument name itself, any strong feelings between
* `--output-references`
* `--include-token-references`
* `--output-token-references`
or maybe even
* `--include-references`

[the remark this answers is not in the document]
adamstankiewicz: I think `--output` makes sense since it's controlling the output build files, whereas `--include` _could_ suggest including references pre-output.

I tend to prefer `--output-token-references`  the most, but don't feel too strongly.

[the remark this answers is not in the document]
brian-smith-tcril: Sounds good to me, let's go with `--output-token-references`!

tokens/src/core/components/Avatar.json
PKulkoRaccoonGang: Agreed, that's a good idea 👍 
I would also leave some documentation somewhere about what rules to follow when creating/editing design tokens. Since it's not entirely obvious what sequence should be present in design tokens. I think we should stop at something like this:
1. `$value`
2. `description`
3. `modify`
4. `source`

tokens/src/themes/light/components/Carousel.json
PKulkoRaccoonGang: This is a [modifier](https://github.com/openedx/paragon/blob/alpha/tokens/style-dictionary.js#L112) for replacing one value in a string with another. After playing with this modifier a bit, I realized that it doesn't support multiple replacement parameters, for example, you can't use something like:

```
            "modify": [
              { "type": "str-replace", "toReplace": "#", "replaceWith": "%23" },
              { "type": "str-replace", "toReplace": ">", "replaceWith": "%3E" },
              { "type": "str-replace", "toReplace": "<", "replaceWith": "%3C" }
            ],
            "$value": "url(\"data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' fill='{color.carousel.control.base}' viewBox='0 0 8 8'><path d='M2.75 0l-1.5 1.5 2.5 2.5-2.5 2.5 1.5 1.5 4-4-4-4z'/></svg>\")"
```

I'm not sure if we need to handle any cases right now other than the `#` symbol (we're currently replacing only that symbol), but I think the option of multiple parameters for replacement would be useful in the future 💯

tokens/src/core/components/CloseButton.json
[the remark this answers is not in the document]
PKulkoRaccoonGang: How about creating an issue to track a specific CSS `calc()`?
This way, we won’t lose track of it in the future.

bin/paragon-scripts.js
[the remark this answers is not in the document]
PKulkoRaccoonGang: Nice! This will definitely make more sense now 💯

lib/build-tokens.js
```diff
@@ -38,36 +57,42 @@ async function buildTokensCommand(commandArgs) {
...
           },
         ],
-        transforms: StyleDictionary.transformGroup.css.filter(item => item !== 'size/rem').concat('color/sass-color-functions', 'str-replace'),
-        options: {
-          fileHeader: 'customFileHeader',
-        },
       },
     },
+    log: {
+      verbosity: verbose ? 'verbose' : 'default',
```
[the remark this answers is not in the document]
PKulkoRaccoonGang: There is no significant difference in the case of displaying errors. I would leave default as the default value (with the option to show more verbose logs)

Default example
Screenshot 2024-11-01 at 11 20 43

Verbose example
image

bin/paragon-scripts.js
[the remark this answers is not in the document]
PKulkoRaccoonGang: [inform] Implemented in follow-up commit: [`036f28c` (#3186)](https://github.com/openedx/paragon/pull/3186/commits/036f28cf6023eb750ee198dc010186f18101b533)

src/Card/CardDeck.jsx
```diff
@@ -1,7 +1,6 @@
 import React, { Children, useMemo } from 'react';
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
-import BaseCardDeck from 'react-bootstrap/CardDeck';
```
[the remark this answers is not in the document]
PKulkoRaccoonGang: Do you mean adding the removal of the missed deprecated Card components to an [existing commit](https://github.com/openedx/paragon/pull/3158/commits/5a23e5740dd35fa15b6166f34079620e231e9fee) that was made earlier or moving this change to a separate PR/commit?

lib/build-tokens.js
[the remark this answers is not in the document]
brian-smith-tcril: Sounds good to me. If someone comes up with a use case for needing to silence logs (none come to mind), or `style-dictionary` is updated to have more verbosity options, we can look into changing this in the future.

lib/build-tokens.js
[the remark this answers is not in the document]
brian-smith-tcril: > When `hasSourceTokensOnly`, it's telling style-dictionary to only transform/output tokens from the source tokens (aka only the custom brand package tokens, not the full set of tokens in upstream Paragon).

I guess the word that feels a little odd to me is "has" - maybe `transformSourceTokensOnly` or `outputSourceTokensOnly`?

src/Card/CardDeck.jsx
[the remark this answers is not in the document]
brian-smith-tcril: At first I thought there were more of these in this PR. Since there aren't a ton I'm fine with keeping this in this PR.

lib/build-tokens.js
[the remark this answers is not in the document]
adamstankiewicz: Fair enough. I like the transform prefix suggestion instead 🙂

styles/scss/core/_utilities.scss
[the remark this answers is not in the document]
brian-smith-tcril: > CSS utility classes in the design system should have `!important`

That's what I thought - I guess I'm just not sure why that wasn't the case *before*, and if there might be something that currently relies on them unconventionally *not* having `!important`. I'm just worried about a possible "if it's a bug people rely on it's a feature" scenario.

tokens/src/themes/light/components/Carousel.json
brian-smith-tcril: That's unfortunate. It would be really nice to support multiple `str-replace`s. Not a dealbreaker though.

lib/build-tokens.js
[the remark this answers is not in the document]
brian-smith-tcril: I like `transformSourceTokensOnly` the best so far too! Let's roll with that!

styles/scss/core/_utilities.scss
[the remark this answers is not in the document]
PKulkoRaccoonGang: I searched on GitHub for the use of certain utilities in the OpenEdX repositories and did not find any user overrides that could be impacted by the use of `!important`. My vote is to keep `!important`.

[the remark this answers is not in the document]
brian-smith-tcril: Thanks for going above and beyond and looking into that @PKulkoRaccoonGang!

lib/build-tokens.js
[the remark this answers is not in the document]
brian-smith-tcril: https://github.com/openedx/paragon/pull/3186/commits/b4d560bee10f28a9ea2c9d26275b647334a00111

styles/scss/core/_utilities.scss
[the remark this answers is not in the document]
adamstankiewicz: Appreciate the sanity check, @PKulkoRaccoonGang!

tokens/src/core/components/CloseButton.json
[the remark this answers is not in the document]
adamstankiewicz: Thanks @brian-smith-tcril!

tokens/src/core/global/elevation.json
```diff
@@ -1,19 +1,61 @@
 {
   "elevation": {
+    "$type": "ratio",
```
adamstankiewicz, line 3: [sanity check] Does `ratio` make sense as the type (see [DTCG format spec](https://tr.designtokens.org/format/#additional-types)'s description of the `ratio` type)? Perhaps we want this to be `"$type": "number"` since it's really just a number ([docs](https://tr.designtokens.org/format/#number))?

tokens/src/core/alias/size.json
```diff
@@ -1,15 +1,16 @@
 {
   "size": {
+    "$type": "dimension",
     "border": {
-      "width": { "value": "1px", "type": "dimension", "source": "$border-width", "description": "Default border width." },
+      "width": { "source": "$border-width", "$value": "1px", "$description": "Default border width." },
```
adamstankiewicz, line 5: [curious] Cross-checking the `dimension` type with the DTCG spec again, I noticed that it seems you can also define the dimension value and its unit separately, e.g.:

```json
{
  "size": {
    "$type": "dimension",
    "border": {
      "width": {
        "$value": {
          "value": 1,
          "unit": "px"
        }
      }
     }
  }
}
```

Not sure if this is something `style-dictionary` v4 supports out-of-the-box. Do we think it's worth exploring defining dimension token values and units separately vs. combined as we have it today (i.e., `"$value": "1px"`).

brian-smith-tcril: Since this is what is documented in the DTCG spec I'd think we'd want to move forward with this change. Let's verify this works with style-dictionary v4 and figure out next steps from there.

tokens/src/core/components/Avatar.json
PKulkoRaccoonGang: [inform]: We have a draft PR providing additions of a linter for JSON files and a sorter script
https://github.com/openedx/paragon/pull/3285

tokens/src/core/global/elevation.json
PKulkoRaccoonGang: [inform] Implemented in follow-up PR: https://github.com/openedx/paragon/pull/3286

brian-smith-tcril (approved):
I believe all my previous comments have been addressed. This looks great!

Super excited to see all of this come together. Amazing work @PKulkoRaccoonGang!

tokens/src/core/alias/size.json
adamstankiewicz: For posterity, we tried to update the token definition to adopt the DTCG format spec for `$dimension` tokens, but it seems to result in unexpected output, e.g.:

```css
--pgn-size-border-width-value: 1; /* Default border width. */
--pgn-size-border-width-unit: 0px; /* Default border width. */
```

When we'd really want something like:

```css
--pgn-size-border-width-value: 1px; /* Default border width. */
```

I've opened an issue upstream in the `style-dictionary` repository to try to drive clarity in the expected output for `$dimension` tokens using the DTCG spec / report a bug, if applicable: https://github.com/amzn/style-dictionary/issues/1398

Given this, for now, we will keep the Paragon `$dimension` tokens using the combined `$value` vs. following the DTCG spec.

adamstankiewicz (approved):
Looks great! Appreciate all your iterations on this PR, @PKulkoRaccoonGang! Will be great to see this land.

Issue #3057, opened 2024-05-15
Design tokens improvement


PKulkoRaccoonGang:
To make working with design tokens easier in the future, it would be great to add a few helpers:

1. Set up a linter to validate the formatting of [files containing design tokens](https://github.com/openedx/paragon/tree/alpha/tokens/src) (`tokens` folder).

2. Script for checking CSS variables declaration.
Previously, when using SCSS variables, if the SCSS variable was not declared but was used, we could see an error during assembly and fix this problem. When using CSS variables, if the variable is used in Paragon components and has been deleted, we will not know about it. Thus, the browser is the only tool for validating variable declarations, which is not very convenient.

<img width="1486" alt="image" src="https://github.com/openedx/paragon/assets/93188219/33044ca7-875f-4773-ab59-fb2a436d2a08">

PR #3203, opened 2024-08-31
feat: expose --output-references CLI arg for building tokens, registers filters, and updates CSS vars format

adamstankiewicz:
_Note: Amongst other changes, addresses some of my feedback on the related, base PR: https://github.com/openedx/paragon/pull/3186_

#### CLI Enhancements

* Exposes `--output-references` CLI argument for `build-tokens` command. Defaults to `true`. Ensures brand package output with the CLI includes references in build output out-of-the-box.

#### Token Management Improvements

* Registers filter(s) `isThemeVariant.{'light'}`, handling future theme variants when implemented (e.g., `isThemeVariant.dark`).
* Migrates `createCustomCSSVariables` to use `formattedVariables` to rely on out-of-the-box CSS variable formatting. The formatter still supports token-specific overrides of `outputReferences`. If a token has modifications via `modify`, the modified base reference is not included in the output.
* Updates custom fileHeader implementation, including a relative path to design tokens documentation.
* Fixes bug with line-height tokens, switching their `$type` from `dimension` to `number` to resolve typography style regressions.
* Updates typography tokens related to font size, font weight, and line-height for more consistent naming structure when taking into account mobile.
* Updates `@mobile-type` SCSS mixin to support level-specific customization of mobile typography styles for display 1-4.
* Renames `"description"` field in tokens to `"$description""` per the DTCG format.

#### Documentation Site Enhancements

* Ensures the "Typography" foundations page properly previews the correct font size for regular "Body" text and includes the missing "HEADING LABEL" example.
* Updates to "Colors" page in docs site:
  * Displays token name instead of CSS variable in the color swatch previews (see screenshot below).
  * Include `accent-a` and `accent-b` alongside other color names, rather than manually rendering `Swatch` for the accents.
  * Modifies the grid styles for color swatch preview to be more responsive.
* Resolves `NaNpx` bug in `MeasuredItem` component on docs site, while computing the measurements to display for an element (e.g., font size). Instead, it renders an empty block while measurements are resolved.
* Updates `CodeBlock` styles on docs site to add its border and background color only to the `LivePreview`, not the entire `CodeBlock` example.
* Reduces whitespace on docs site homepage.
* Simplifies columns on docs site header, ensuring `SiteTitle` is horizontally aligned in the center.

#### Informational Notes

* [inform] Related to CSS `calc` defined within tokens, while we could use `sd-transforms` to register a transform `ts/resolveMath` ([source](https://github.com/tokens-studio/sd-transforms/blob/1ad17e20dc0666756a4cc1be643837ca82b3a725/src/checkAndEvaluateMath.ts#L154)), it does not wrap math calculations with `calc(...)`, effectively not evaluating the operation in CSS. There's several related issues ([example](https://github.com/tokens-studio/sd-transforms/issues/13), [example](https://github.com/tokens-studio/sd-transforms/issues/121)) in `style-dictionary` and `sd-transforms` about this issue, with no current resolution; only hacky workarounds.

### Deploy Preview

N/A

**Updates "Colors" page color swatch previews:**

Note: screenshot depicts a local checkout`@edx/elm-theme`

image

## Merge Checklist

## Post-merge Checklist


lib/build-tokens.js
```diff
@@ -104,17 +123,25 @@ async function buildTokensCommand(commandArgs) {
           {
             format: 'css/custom-variables',
             destination: `themes/${themeVariant}/variables.css`,
-            filter: hasSourceTokensOnly ? 'isSource' : undefined,
+            filter: hasSourceTokensOnly
+              ? `isThemeVariant${themeVariant.charAt(0).toUpperCase()}${themeVariant.slice(1)}SourceOnly`
+              : `isThemeVariant${themeVariant.charAt(0).toUpperCase()}${themeVariant.slice(1)}`,
```
adamstankiewicz, line 128: [inform] Currently, the registered format `css/custom-variables` performs filtering within its own logic (i.e., filtering on `dictionary.allTokens`), rather than treating the filter as a legit style-dictionary filter here. New filters are now registered to filter tokens by the appropriate `themeVariant` file path within the `filter` property here versus within the definition of  `css/custom-variables`.

package.json
```diff
@@ -52,7 +52,8 @@
     "prepare": "husky || true",
     "build-tokens": "./bin/paragon-scripts.js build-tokens --build-dir ./styles/css",
     "replace-variables-usage-with-css": "./bin/paragon-scripts.js replace-variables -p src -t usage",
-    "replace-variables-definition-with-css": "./bin/paragon-scripts.js replace-variables -p src -t definition"
+    "replace-variables-definition-with-css": "./bin/paragon-scripts.js replace-variables -p src -t definition",
+    "cli:help": "./bin/paragon-scripts.js help"
```
adamstankiewicz, line 56: [inform] Adds a way to run the CLI `help` command from Paragon root.

styles/css/core/variables.css
```diff
@@ -1,28 +1,28 @@
...
-  --pgn-elevation-zindex-1600: 1600;
-  --pgn-elevation-zindex-1400: 1400;
-  --pgn-elevation-zindex-1200: 1200;
-  --pgn-elevation-zindex-1000: 1000;
-  --pgn-elevation-zindex-800: 800;
-  --pgn-elevation-zindex-600: 600;
-  --pgn-elevation-zindex-400: 400;
-  --pgn-elevation-zindex-200: 200;
-  --pgn-elevation-zindex-0: 0;
+  --pgn-elevation-zindex-fixed: 1030; /* z-index of for fixed element. */
```
adamstankiewicz, line 12: By relying on the `formattedVariables` helper function from `style-dictionary/utils`, we get inline comments from the token's `description` field, if any. Much of the other logic we had implemented is provided out-of-the-box (e.g., ordering of CSS variables).

styles/css/core/custom-media-breakpoints.css
```diff
@@ -1,7 +1,6 @@
-/*
- * IMPORTANT: This file is the result of assembling design tokens.
- * Do not edit directly.
- * Generated on Tue, 27 Aug 2024 17:14:44 GMT
+/**
+ * Do not edit directly, this file was auto-generated.
+ * Generated on Sat, 31 Aug 2024 19:26:11 GMT
  */
```
adamstankiewicz, line 4: [inform] This is the newly introduced default `style-dictionary` file header, with its optional auto-generated timestamp.

tokens/style-dictionary.js
```diff
@@ -80,45 +84,41 @@ const colorTransform = (token, theme) => {
...
-        $value = $value.replace(ref.$value, `var(--${ref.name})`);
-      });
-    }
-
-    return `  --${token.name}: ${$value};`;
-  }).join('\n');
-
-  return `${createCustomHeader(StyleDictionary, file).join('\n')}\n:root {\n${variables}\n}\n`;
+  const { outputReferences, formatting } = options;
+  const variables = formattedVariables({
```
adamstankiewicz, line 91: [inform] Relies on helper function `formattedVariables` provided by `style-dictionary/utils` to format tokens as CSS variables instead of re-implementing much of its same logic here.

www/src/scss/base.scss
```diff
@@ -27,7 +27,3 @@ body {
 .container-fluid {
   max-width: none;
 }
-
-.fs-16 {
```
adamstankiewicz, line 31: [inform] removed in favor of new `.font-size-normal` CSS utility class.

www/src/scss/edxorg-theme.scss
```diff
@@ -1,3 +1,3 @@
-@import url("https://cdn.jsdelivr.net/npm/@edx/brand-edx.org@alpha/dist/core.min.css");
-@import url("https://cdn.jsdelivr.net/npm/@edx/brand-edx.org@alpha/dist/light.min.css");
+@import url("https://cdn.jsdelivr.net/npm/@edx/elm-theme@1.8.1/dist/core.min.css");
+@import url("https://cdn.jsdelivr.net/npm/@edx/elm-theme@1.8.1/dist/light.min.css");
```
adamstankiewicz, line 2: As of this comment, 2U no longer has plans to migrate the existing `@edx/brand-edx.org` in favor of cutting MFEs directly to `@edx/elm-theme` once MFEs support tokens-compatible Paragon, so could opt to use `@edx/elm-theme` for the docs site preview now.

www/src/scss/edxorg-theme.scss
```diff
@@ -1,3 +1,5 @@
-@import url("https://cdn.jsdelivr.net/npm/@edx/brand-edx.org@alpha/dist/core.min.css");
-@import url("https://cdn.jsdelivr.net/npm/@edx/brand-edx.org@alpha/dist/light.min.css");
+@import "~paragon-style/scss/core/core";
 @import "base";
+
+@import url("https://cdn.jsdelivr.net/npm/@edx/elm-theme@1.8.1/dist/core.min.css");
+@import url("https://cdn.jsdelivr.net/npm/@edx/elm-theme@1.8.1/dist/light.min.css");
```
adamstankiewicz, line 5: [inform] At the moment, 2U has no intentions to migrate the existing `@edx/brand-edx.org`. Instead, MFEs will cutover directly to our new tokens-compatible theme `@edx/elm-theme` upon upgrade to a Paragon version that relies on tokens.

lib/build-tokens.js
```diff
@@ -40,6 +58,10 @@ async function buildTokensCommand(commandArgs) {
     source: tokensSource
       ? [`${tokensSource}/core/**/*.json`, `${tokensSource}/core/**/*.toml`]
       : [],
+    preprocessors: ['tokens-studio'],
+    expand: {
+      typesMap: (await getTokensStudioStyleDictionaryTransforms()).expandTypesMap,
+    },
```
adamstankiewicz, line 64: [inform] Largely a no-op change right now (i.e., except for some minor formatting in the output, e.g. including leading zeros in `0.625rem`), but these changes should make it possible to support/define tokens exported from the Figma plugin Token Studio.

bin/paragon-scripts.js
```diff
@@ -85,11 +85,21 @@ const COMMANDS = {
         description: 'Include only source design tokens in the build.',
         defaultValue: false,
       },
+      {
+        name: '--output-references',
+        description: 'Include references in the build output.',
+        defaultValue: true,
+      },
```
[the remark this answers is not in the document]
brian-smith-tcril: 100% agree. Good call/addition!

lib/build-tokens.js
```diff
@@ -52,24 +65,30 @@ async function buildTokensCommand(commandArgs) {
             destination: 'core/variables.css',
             filter: hasSourceTokensOnly ? 'isSource' : undefined,
             options: {
-              outputReferences: !hasSourceTokensOnly,
+              outputReferences,
+              formatting: {
+                fileHeaderTimestamp: true,
+              },
```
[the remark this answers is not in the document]
brian-smith-tcril: Good call! The default header looks like it covers everything we need, and I'm all for reducing complexity (even in small ways like this!)

```css
/**
 * Do not edit directly, this file was auto-generated.
 * Generated on Tue, 03 Sep 2024 13:46:26 GMT
 */
```

lib/build-tokens.js
```diff
@@ -91,45 +112,62 @@ async function buildTokensCommand(commandArgs) {
...
       css: {
         ...coreConfig.platforms.css,
         files: [
           {
             format: 'css/custom-variables',
             destination: `themes/${themeVariant}/variables.css`,
-            filter: hasSourceTokensOnly ? 'isSource' : undefined,
+            filter: hasSourceTokensOnly
+              ? `isSource.${themeVariant}`
+              : `isThemeVariant.${themeVariant}`,
```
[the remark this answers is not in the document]
brian-smith-tcril: Just trying to wrap my head around this change a little bit. Before this change if `hasSourceTokensOnly` was `false` we'd have filter as `undefined` - does that mean it used to apply to all variants?

lib/build-tokens.js
```diff
@@ -91,45 +112,62 @@ async function buildTokensCommand(commandArgs) {
...
             options: {
-              outputReferences: !hasSourceTokensOnly,
+              outputReferences,
+              formatting: {
+                fileHeaderTimestamp: true,
+              },
```
brian-smith-tcril, line 129: I see this block now exists in a couple spots in this file. It's only 3 lines so I'm not *worried* about the copypasta, but I am curious as to if there's a "built-in" way to have a "global" formatting setting. If there isn't a "built-in" way I'm thinking it might be cool to move this out to a const like we have with `defaultParams` so if anyone wants to change the formatting they can do it in one place.

package.json
```diff
@@ -114,7 +115,7 @@
     "@babel/preset-typescript": "^7.16.7",
     "@edx/eslint-config": "^3.2.0",
     "@edx/stylelint-config-edx": "^2.3.0",
-    "@edx/typescript-config": "^1.0.1",
+    "@edx/typescript-config": "^1.1.0",
```
brian-smith-tcril, line 118: Just curious: is this change necessary for this PR? No doubt it's a *good* change, just wondering if was a nice-to-have or a need-to-have.

styles/scss/core/_typography.scss
```diff
@@ -1,11 +1,23 @@
...
+  .display-3 {
+    font-size: $display3-mobile-size;
+    line-height: $display-mobile-line-height;
+  }
+
   .display-4 {
-    font-size: $display-mobile-size;
+    font-size: $display4-mobile-size;
     line-height: $display-mobile-line-height;
   }
```
brian-smith-tcril, line 22: Looking at this it seems we used to use the same size and line height for `.display-1` through `.display-4`, and this updates it so we have different sizes but not different line heights? Just curious as to the reasoning behind that.

styles/scss/core/_utilities.scss
```diff
@@ -21,13 +21,17 @@
 $color-levels: 100, 200, 300, 400, 500, 600, 700, 800, 900;
 
 .x-small {
-  font-size: $x-small-font-size;
+  font-size: $x-small-font-size !important;
```
brian-smith-tcril, line 24: I'm always slightly wary of `!important`, could you explain why it's needed here? (Depending on the reasoning it might be nice to add that as a comment in the file itself)

brian-smith-tcril (commented):
This looks great! I left quite a few comments but they're mostly just questions. Thank you so much for all of this!

lib/build-tokens.js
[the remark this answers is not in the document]
adamstankiewicz: Previously, `hasSourceTokensOnly: false` would rely on filtering implemented _within the format function_, instead of as a true `style-dictionary` filter implemented with `.registerFilter`. That is, formatters should assume any token filtering happens _before_ the format is applied, which already limits the contents of `dictionary.allTokens` we previously filtered on explicitly.

There's some more detailed rationale about this change provided on my review of Peter's PR [here](https://github.com/openedx/paragon/pull/3186#discussion_r1739634704), notably:

> IMHO, having `unfilteredTokens` in the `dictionary` here suggests a possible code smell that we might want to be doing such filtering as a legit custom filter registered with `style-dictionary` (similar to how we seem to define an `isSource` filter today via `registerFilter`).

See the current implementation of `createCustomCSSVariables` ([source](https://github.com/openedx/paragon/blob/0e16dbbff83b7e1ee72b93205b193feb3a99603a/tokens/style-dictionary.js#L74-L76)).

styles/scss/core/_typography.scss
adamstankiewicz: Great question! So, the new 2U Paragon Elm Theme has distinct sizes for mobile `.display-1` through `.display-4`, but Paragon `alpha` doesn't currently actually have the tokens/CSS support for level-specific mobile display font sizes. As such, this PR introduces tokens for distinct levels of display size for mobile. 

Regarding why there's not currently level-specific mobile line heights is that the Elm Theme continues to rely on a single line height across all mobile display levels (i.e., `line-height: 1`), which also happens to be the non-mobile default line height as well.

Paragon base theme defines `display-mobile-line-height: 3.5rem`, and Elm Theme overrides this to be `display-mobile-line-height: 1`.

There is also only a single line height value configured for all non-mobile display levels, too.

Note that using `line-height: 1` (i.e., without a unit like `rem`) defines the line height to be equal to the font-size which is distinct per level. Given this, `line-height: 1` is technically still results in distinct _computed_ line heights per mobile display level.

styles/scss/core/_utilities.scss
adamstankiewicz: CSS utility class definitions generally _should_ always have `!important`. See list of other [existing utility classes](https://paragon-openedx.netlify.app/foundations/css-utilities/) on the Paragon docs site to observe their inclusion of `!important`.

See example(s) of CSS utility class definitions from Bootstrap v4 ([source](https://github.com/twbs/bootstrap/blob/e5643aaa89eb67327a5b4abe7db976f0ea276b70/scss/utilities/_sizing.scss#L11-L20)) as a reference, too.

styles/scss/core/_typography.scss
brian-smith-tcril: That sounds very reasonable! I do wonder if _supporting_ different custom line heights for each is something we want to do. I'd lean towards thinking we _should_, but I could definitely be convinced otherwise.

adamstankiewicz: Yeah, supporting that as a use case is how I'm leaning, FWIW. Though, maybe we roll with it as is for this PR and @PKulkoRaccoonGang's related PR(s) to do the style-dictionary upgrade to v4, and treat level-specific display line heights as a standalone fast follow to these PRs before `master` -> `alpha`, given it's not necessarily related to the v4 upgrade itself per se and would probably be a reasonably scoped individual contribution.

styles/scss/core/_utilities.scss
brian-smith-tcril: Ah yes, I was getting towards the end of the diff and missed the filename.

It looks like *almost* everything else in the file has `!important` (only `.icon-spin` doesn't) - is the lack of `!important` in `.icon-spin` intentional?

adamstankiewicz: Hmm, in additional `icon-spin`, the `font-weight: normal` property on `.micro` doesn't include `!important` either, though I don't it should for `font-weight` (you could want micro-sized copy with `font-weight-bold`, etc.). IMHO, `.micro` shouldn't be defining `font-weight` at all.

Regarding `.icon-spin` specifically, it appears to only be used with the `StatefulButton` component across all of the `openedx` GitHub organization. I actually wonder whether we truly want to consider this a global utility class or not, given its specific to icons.

For example, alternatively, should the `Icon` component expose a `spin` prop that extends the rendered icon to add the `icon-spin` class conditionally?

E.g., proposed usage from `StatefulButton`:

```jsx
icons: {
  pending: <Icon src={SpinnerSimple} spin />,
},
```

I'm struggling to identify use cases for `.icon-spin` outside of the `Icon` component. With `icon-spin` as a global CS utility class, I could also see it getting abused to make things spin/animate on the page that shouldn't (e.g., non-icons).

My 2c recommendation: 
* Remove `.icon-spin` as a CSS utility class, implementing it instead of a class name specific to `Icon` component.
* Expose optional `spin` prop on `Icon` component.
* Update `StatefulButton` to pass `spin` prop on its `pending` icon.

brian-smith-tcril: > My 2c recommendation:
>
> * Remove .icon-spin as a CSS utility class, implementing it instead of a class name specific to Icon component.
> * Expose optional spin prop on Icon component.
> * Update StatefulButton to pass spin prop on its pending icon.

This absolutely sounds like the right move to me!

> IMHO, `.micro` shouldn't be defining `font-weight` at all.

Do we have a sense of what might break if it *didn't*?

brian-smith-tcril: That being said, I don't think either of those should be part of this PR. Just making a couple follow-up issues to address them seems like the right move to me.

adamstankiewicz: > > IMHO, .micro shouldn't be defining font-weight at all.
>
> Do we have a sense of what might break if it didn't?

It shouldn't break anything. The font-weight in `.micro` is font-weight: normal`, which corresponds to font-weight 400, the default `body` font weight used elsewhere.

> That being said, I don't think either of those should be part of this PR. Just making a couple follow-up issues to address them seems like the right move to me.

100% 😄

lib/build-tokens.js
brian-smith-tcril: @adamstankiewicz I think this is the last lingering question I have before giving this a :heavy_check_mark:!

adamstankiewicz: Plan to investigate later this afternoon. I do recall seeing somewhere in the style-dictionary docs about global settings, so it might be possible. If the global approach isn't possible, I agree abstracting out a `defaultParams` or similar would be worthwhile to be more DRY. I'll share an update later :)

src/Form/_FormText.scss
```diff
@@ -1,5 +1,5 @@
 .pgn__form-text {
-  font-size: var(--pgn-typography-font-size-small-base);
+  font-size: var(--pgn-typography-font-size-sm);
```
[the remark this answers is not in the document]
adamstankiewicz: @PKulkoRaccoonGang I have a [draft PR](https://github.com/edx/elm-theme/pull/24) brewing to account for `edx/elm-theme`, based on the changes introduced in this PR and the broader v3 -> v4 upgrade (e.g., adopting the DTCG format).

Note: this linked PR also accounts for a few style bug fixes as well that I noticed while QA'ing `elm-theme` against the Paragon docs site locally.

--- 

> I have described https://github.com/openedx/paragon/issues/3057 before, maybe we should come back to it in the future. Please let me know your thoughts.

100% agreed we should explore opportunities for linting and/or other validation of tokens, but as a future initiative.

lib/build-tokens.js
adamstankiewicz: Just pushed up a commit to re-introduce a custom file header, and apply it globally across all files in the `css` platform.

```css
/**
 * Do not edit directly, this file was auto-generated. while transforming design tokens.
 * See <root>/tokens/README.md for more details.
 */
```

brian-smith-tcril (approved):
:shipit:

PKulkoRaccoonGang (approved):
This is pretty cool! LGTM!

PR #3219, opened 2024-09-07
feat: registers sd-transforms; includes refs in source-tokens-only output; resolves refs warnings

adamstankiewicz:
# Description

1. Integrates `style-dictionary` with `sd-transforms` ([GitHub](https://github.com/tokens-studio/sd-transforms)) from [Tokens Studio](https://tokens.studio/). Enables support for copying exported tokens from the Tokens Studio plugin in Figma and adding them to the Paragon tokens schema.
1. Registers a custom transform group `paragon-css` to abstract away our custom transforms, `sd-transforms`, and the default CSS transforms. The previously excluded `size/rem` filter for the default CSS transforms is added back in.
1. Resolves noisy warnings regarding missing output references in resulting token transforms, by now including the reference token(s) in the output. See screenshots below.
    * In Paragon, this use case applies when transforming `light` theme variant tokens that reference `core` tokens. Now, these handful referenced `core` tokens are included in the `light` theme CSS variables, although duplicative of their definitions in `core.css`; this is an acceptable tradeoff to remove noisy, disruptive warnings as it only expands the contents of output files by the number of _referenced_ tokens.
    * In brand packages (e.g., `@edx/elm-theme`), this use case applies when either `core` and/or `light` tokens reference base, default Paragon tokens without redefining the base token within the brand package. Now, the referenced default Paragon token(s) are included in the brand package output.
1. Adds `build-tokens:watch` NPM script to automatically execute the `build-tokens` CLI on file changes, enabling more rapid development when changing tokens and verifying output without having to manually re-run `build-tokens` yourself.

### Resolves output references warning in build output

Related to (3) above. Paragon tokens output on left; brand package tokens output on right (see related `@edx/elm-theme` [PR](https://github.com/edx/elm-theme/pull/24)).

#### Before

image

#### After

image

## Merge Checklist

* [ ] If your update includes visual changes, have they been reviewed by a designer? Send them a link to the Netlify deploy preview, if applicable.
* [ ] Does your change adhere to the documented [style conventions](https://github.com/openedx/paragon/blob/master/docs/decisions/0012-css-styling-conventions)?
* [ ] Do any prop types have missing descriptions in the Props API tables in the documentation site (check deploy preview)?
* [ ] Were your changes tested using all available themes (see theme switcher in the header of the deploy preview, under the "Settings" icon)?
* [ ] Were your changes tested in the `example` app?
* [ ] Is there adequate test coverage for your changes?
[...]

PR #3222, opened 2024-09-08
feat: added expanded tokens

PKulkoRaccoonGang:
## Description

- added expanded tokens
- fixed some bugs
- created an abstracted CSS variables

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -34,6 +34,7 @@
     "color-function-notation": "legacy",
     "value-keyword-case": ["lower", {
       "ignoreProperties": ["/font-family/"]
-    }]
+    }],
+    "custom-property-empty-line-before": null
   }
 }
--- a/lib/build-tokens.js
+++ b/lib/build-tokens.js
@@ -154,13 +154,13 @@ async function buildTokensCommand(commandArgs) {
   ];
 
   // Build tokens for each configuration
-  await Promise.all(configs.map(async ({ config, isThemeVariant, themeVariant }) => {
+  await Promise.all(configs.map(async ({ config, themeVariant }) => {
     const sd = new StyleDictionary(config);
     await sd.cleanAllPlatforms();
     await sd.buildAllPlatforms();
     createIndexCssFile({
       buildDir,
-      isThemeVariant: !!isThemeVariant,
+      isThemeVariant: !!themeVariant,
       themeVariant,
     });
   }));
--- a/src/Annotation/index.scss
+++ b/src/Annotation/index.scss
@@ -1,5 +1,21 @@
 @import "mixins";
 
+:root {
+  --pgn-elevation-annotation-box-shadow:
+    drop-shadow(
+      var(--pgn-elevation-annotation-box-shadow-1-offset-x)
+      var(--pgn-elevation-annotation-box-shadow-1-offset-y)
+      var(--pgn-elevation-annotation-box-shadow-1-blur)
+      var(--pgn-elevation-annotation-box-shadow-1-color)
+    )
+    drop-shadow(
+      var(--pgn-elevation-annotation-box-shadow-2-offset-x)
+      var(--pgn-elevation-annotation-box-shadow-2-offset-y)
+      var(--pgn-elevation-annotation-box-shadow-2-blur)
+      var(--pgn-elevation-annotation-box-shadow-2-color)
+    );
+}
+
 .pgn__annotation {
   padding: var(--pgn-spacing-annotation-padding);
   border-radius: var(--pgn-size-annotation-border-radius);
--- a/src/Carousel/index.scss
+++ b/src/Carousel/index.scss
@@ -1,3 +1,26 @@
+:root {
+  --pgn-transition-carousel-base:
+    var(--pgn-transition-carousel-base-property)
+    var(--pgn-transition-carousel-base-duration)
+    var(--pgn-transition-carousel-base-timing-function)
+    var(--pgn-transition-carousel-base-delay)
+    var(--pgn-transition-carousel-base-behavior);
+
+  --pgn-transition-carousel-indicator:
+    var(--pgn-transition-carousel-indicator-property)
+    var(--pgn-transition-carousel-indicator-duration)
+    var(--pgn-transition-carousel-indicator-timing-function)
+    var(--pgn-transition-carousel-indicator-delay)
+    var(--pgn-transition-carousel-indicator-behavior);
+
+  --pgn-transition-carousel-control:
+    var(--pgn-transition-carousel-control-property)
+    var(--pgn-transition-carousel-control-duration)
+    var(--pgn-transition-carousel-control-timing-function)
+    var(--pgn-transition-carousel-control-delay)
+    var(--pgn-transition-carousel-control-behavior);
+}
+
 .carousel {
   position: relative;
 }
@@ -58,7 +81,7 @@
   .active.carousel-item-right {
     z-index: 0;
     opacity: 0;
-    transition: opacity 0s var(--pgn-transition-carousel-base);
+    transition: var(--pgn-transition-carousel-base);
   }
 }
 
--- a/src/DataTable/index.scss
+++ b/src/DataTable/index.scss
@@ -1,3 +1,15 @@
+:root {
+  --pgn-elevation-data-table-box-shadow:
+    var(--pgn-elevation-data-table-box-shadow-offset-x)
+    var(--pgn-elevation-data-table-box-shadow-offset-y)
+    var(--pgn-elevation-data-table-box-shadow-blur)
+    var(--pgn-elevation-data-table-box-shadow-color);
+
+  --pgn-spacing-data-table-padding-cell:
+    var(--pgn-spacing-data-table-padding-cell-x)
+    var(--pgn-spacing-data-table-padding-cell-y);
+}
+
 .pgn__data-table-wrapper {
   font-size: var(--pgn-typography-font-size-sm);
   border-radius: var(--pgn-size-border-radius-base);
--- a/styles/css/core/abstraction-variables.css
+++ b/styles/css/core/abstraction-variables.css
@@ -1,0 +1,44 @@
+:root {
+  /**
+   * TRANSITION
+   */
+  --pgn-transition-base:
+    var(--pgn-transition-base-property)
+    var(--pgn-transition-base-duration)
+    var(--pgn-transition-base-timing-function)
+    var(--pgn-transition-base-delay)
+    var(--pgn-transition-base-behavior);
+
+  --pgn-transition-fade:
+    var(--pgn-transition-fade-property)
+    var(--pgn-transition-fade-duration)
+    var(--pgn-transition-fade-timing-function)
+    var(--pgn-transition-fade-delay)
+    var(--pgn-transition-fade-behavior);
+
+  --pgn-transition-collapse-height:
+    var(--pgn-transition-collapse-height-property)
+    var(--pgn-transition-collapse-height-duration)
+    var(--pgn-transition-collapse-height-timing-function)
+    var(--pgn-transition-collapse-height-delay)
+    var(--pgn-transition-collapse-height-behavior);
+
+  --pgn-transition-collapse-width:
+    var(--pgn-transition-collapse-width-property)
+    var(--pgn-transition-collapse-width-duration)
+    var(--pgn-transition-collapse-width-timing-function)
+    var(--pgn-transition-collapse-width-delay)
+    var(--pgn-transition-collapse-width-behavior);
+
+  --pgn-transition-form-input:
+    var(--pgn-transition-form-input-1-property)
+    var(--pgn-transition-form-input-1-duration)
+    var(--pgn-transition-form-input-1-timing-function)
+    var(--pgn-transition-form-input-1-delay)
+    var(--pgn-transition-form-input-1-behavior),
+    var(--pgn-transition-form-input-2-property)
+    var(--pgn-transition-form-input-2-duration)
+    var(--pgn-transition-form-input-2-timing-function)
+    var(--pgn-transition-form-input-2-delay)
+    var(--pgn-transition-form-input-2-behavior);
+}
--- a/styles/css/core/variables.css
+++ b/styles/css/core/variables.css
@@ -30,18 +30,76 @@
   --pgn-elevation-modal-zindex: 1050;
   --pgn-elevation-modal-backdrop-zindex: 1040;
   --pgn-elevation-dropdown-zindex: 1000;
-  --pgn-transition-collapse-width: width .35s ease; /* Collapse transition for width that takes 350ms */
-  --pgn-transition-collapse-height: height .35s ease; /* Collapse transition for height that takes 350ms */
-  --pgn-transition-fade: opacity .15s linear; /* Opacity transition that takes 150ms */
-  --pgn-transition-base: all .2s ease-in-out; /* Generic transition for any property change */
-  --pgn-transition-progress-bar-bar-transition: width .6s ease;
-  --pgn-transition-progress-bar-bar-animation-timing: 1s linear infinite;
-  --pgn-transition-form-control: background-color 0.15s ease-in-out, border-color 0.15s ease-in-out, box-shadow 0.15s ease-in-out;
-  --pgn-transition-form-input: border-color 0.15s ease-in-out, box-shadow 0.15s ease-in-out;
-  --pgn-transition-carousel-control: opacity .15s ease;
-  --pgn-transition-carousel-indicator: opacity .6s ease;
+  --pgn-transition-collapse-width-behavior: normal; /* Collapse transition for width that takes 350ms */
+  --pgn-transition-collapse-width-delay: 0s; /* Collapse transition for width that takes 350ms */
+  --pgn-transition-collapse-width-timing-function: ease; /* Collapse transition for width that takes 350ms */
+  --pgn-transition-collapse-width-duration: 0.35s; /* Collapse transition for width that takes 350ms */
+  --pgn-transition-collapse-width-property: width; /* Collapse transition for width that takes 350ms */
+  --pgn-transition-collapse-height-behavior: normal; /* Collapse transition for height that takes 350ms */
+  --pgn-transition-collapse-height-delay: 0s; /* Collapse transition for height that takes 350ms */
+  --pgn-transition-collapse-height-timing-function: ease; /* Collapse transition for height that takes 350ms */
+  --pgn-transition-collapse-height-duration: 0.35s; /* Collapse transition for height that takes 350ms */
+  --pgn-transition-collapse-height-property: height; /* Collapse transition for height that takes 350ms */
+  --pgn-transition-fade-behavior: normal; /* Opacity transition that takes 150ms */
+  --pgn-transition-fade-delay: 0s; /* Opacity transition that takes 150ms */
+  --pgn-transition-fade-timing-function: linear; /* Opacity transition that takes 150ms */
+  --pgn-transition-fade-duration: 0.15s; /* Opacity transition that takes 150ms */
+  --pgn-transition-fade-property: opacity; /* Opacity transition that takes 150ms */
+  --pgn-transition-base-behavior: normal; /* Generic transition for any property change */
+  --pgn-transition-base-delay: 0s; /* Generic transition for any property change */
+  --pgn-transition-base-timing-function: ease-in-out; /* Generic transition for any property change */
+  --pgn-transition-base-duration: 0.2s; /* Generic transition for any property change */
+  --pgn-transition-base-property: all; /* Generic transition for any property change */
+  --pgn-transition-progress-bar-transition-behavior: normal;
+  --pgn-transition-progress-bar-transition-delay: 0s;
+  --pgn-transition-progress-bar-transition-timing-function: ease;
+  --pgn-transition-progress-bar-transition-duration: 0.6s;
+  --pgn-transition-progress-bar-transition-property: width;
+  --pgn-transition-progress-bar-animation-timing-iteration-count: infinite;
+  --pgn-transition-progress-bar-animation-timing-delay: 0s;
+  --pgn-transition-progress-bar-animation-timing-timing-function: linear;
+  --pgn-transition-progress-bar-animation-timing-duration: 1s;
+  --pgn-transition-form-control-3-behavior: normal;
+  --pgn-transition-form-control-3-delay: 0s;
+  --pgn-transition-form-control-3-timing-function: ease-in-out;
+  --pgn-transition-form-control-3-duration: 0.15s;
+  --pgn-transition-form-control-3-property: box-shadow;
+  --pgn-transition-form-control-2-behavior: normal;
+  --pgn-transition-form-control-2-delay: 0s;
+  --pgn-transition-form-control-2-timing-function: ease-in-out;
+  --pgn-transition-form-control-2-duration: 0.15s;
+  --pgn-transition-form-control-2-property: border-color;
+  --pgn-transition-form-control-1-behavior: normal;
+  --pgn-transition-form-control-1-delay: 0s;
+  --pgn-transition-form-control-1-timing-function: ease-in-out;
+  --pgn-transition-form-control-1-duration: 0.15s;
+  --pgn-transition-form-control-1-property: background-color;
+  --pgn-transition-form-input-2-behavior: normal;
+  --pgn-transition-form-input-2-delay: 0s;
+  --pgn-transition-form-input-2-timing-function: ease-in-out;
+  --pgn-transition-form-input-2-duration: 0.15s;
+  --pgn-transition-form-input-2-property: box-shadow;
+  --pgn-transition-form-input-1-behavior: normal;
+  --pgn-transition-form-input-1-delay: 0s;
+  --pgn-transition-form-input-1-timing-function: ease-in-out;
+  --pgn-transition-form-input-1-duration: 0.15s;
+  --pgn-transition-form-input-1-property: border-color;
+  --pgn-transition-carousel-control-behavior: normal;
+  --pgn-transition-carousel-control-delay: 0ms;
+  --pgn-transition-carousel-control-timing-function: ease;
+  --pgn-transition-carousel-control-duration: 0.15s;
+  --pgn-transition-carousel-control-property: opacity;
+  --pgn-transition-carousel-indicator-behavior: normal;
+  --pgn-transition-carousel-indicator-delay: 0ms;
+  --pgn-transition-carousel-indicator-timing-function: ease;
+  --pgn-transition-carousel-indicator-duration: var(--pgn-transition-carousel-duration);
+  --pgn-transition-carousel-indicator-property: opacity;
   --pgn-transition-carousel-duration: 0.6s;
-  --pgn-transition-carousel-base: transform var(--pgn-transition-carousel-duration) ease-in-out;
+  --pgn-transition-carousel-base-behavior: normal;
+  --pgn-transition-carousel-base-delay: 0ms;
+  --pgn-transition-carousel-base-timing-function: ease-in-out;
+  --pgn-transition-carousel-base-duration: var(--pgn-transition-carousel-duration);
+  --pgn-transition-carousel-base-property: transform;
   --pgn-transition-btn: none;
   --pgn-transition-badge: none;
   --pgn-typography-print-page-size: a3;
@@ -87,7 +145,7 @@
   --pgn-typography-font-size-micro: 0.688rem; /* Micro utils text font size. */
   --pgn-typography-font-size-xs: 0.75rem; /* X-small font size. */
   --pgn-typography-font-size-sm: 0.875rem; /* Small font size. */
-  --pgn-typography-font-size-lg: var(--pgn-typography-font-size-base) * 1.25; /* Lead text font size. */
+  --pgn-typography-font-size-lg: calc(var(--pgn-typography-font-size-base) * 1.25); /* Lead text font size. */
   --pgn-typography-font-size-base: 1.125rem; /* Base font size. */
   --pgn-typography-font-family-monospace: SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier New', monospace; /* Monospace font family. */
   --pgn-typography-font-family-serif: serif; /* Serif font family. */
@@ -159,7 +217,7 @@
   --pgn-typography-dropdown-font-size: var(--pgn-typography-font-size-base);
   --pgn-typography-code-kbd-nested-font-weight: var(--pgn-typography-font-weight-bold);
   --pgn-typography-code-kbd-font-size: var(--pgn-typography-code-font-size);
-  --pgn-typography-code-font-size: 87.5%;
+  --pgn-typography-code-font-size: var(--pgn-typography-font-size-sm);
   --pgn-typography-close-button-font-weight: var(--pgn-typography-font-weight-bold);
   --pgn-typography-close-button-font-size: calc(var(--pgn-typography-font-size-base) * 1.5);
   --pgn-typography-footer-text-font-size: var(--pgn-typography-font-size-xs);
@@ -258,9 +316,11 @@
   --pgn-spacing-nav-link-padding-y: 0.5rem;
   --pgn-spacing-modal-dialog-margin: 1.5rem;
   --pgn-spacing-modal-header-padding-y: 1rem;
-  --pgn-spacing-modal-header-padding-base: var(--pgn-spacing-modal-header-padding-y) 1.5rem;
+  --pgn-spacing-modal-header-padding-base-y: var(--pgn-spacing-modal-header-padding-y);
+  --pgn-spacing-modal-header-padding-base-x: 1.5rem;
   --pgn-spacing-modal-footer-padding-y: 1rem;
-  --pgn-spacing-modal-footer-padding-base: var(--pgn-spacing-modal-footer-padding-y) 1.5rem;
+  --pgn-spacing-modal-footer-padding-base-y: var(--pgn-spacing-modal-footer-padding-y);
+  --pgn-spacing-modal-footer-padding-base-x: 1.5rem;
   --pgn-spacing-modal-inner-padding-bottom: 0.7rem;
   --pgn-spacing-modal-inner-padding-base: 1.5rem;
   --pgn-spacing-menu-item-icon-margin-right: var(--pgn-spacing-menu-item-icon-margin-left);
@@ -274,7 +334,10 @@
   --pgn-spacing-form-control-select-feedback-tooltip-padding-x: 0.5rem;
   --pgn-spacing-form-control-select-feedback-tooltip-padding-y: 0.25rem;
   --pgn-spacing-form-control-select-feedback-margin-top: var(--pgn-spacing-form-text-margin-top);
-  --pgn-spacing-form-control-select-feedback-icon-position: center right calc(var(--pgn-spacing-form-control-select-padding-x-base) + var(--pgn-spacing-form-control-select-indicator-padding));
+  --pgn-spacing-form-control-select-feedback-icon-position-offset-y: 0;
+  --pgn-spacing-form-control-select-feedback-icon-position-offset-x: calc(var(--pgn-spacing-form-control-select-padding-x-base) + var(--pgn-spacing-form-control-select-indicator-padding));
+  --pgn-spacing-form-control-select-feedback-icon-position-position-x: right;
+  --pgn-spacing-form-control-select-feedback-icon-position-position-y: center;
   --pgn-spacing-form-control-select-feedback-icon-padding-right: calc((1em + 2 * var(--pgn-spacing-form-control-select-padding-y-base)) * 3 / 4 + var(--pgn-spacing-form-control-select-padding-x-base) + var(--pgn-spacing-form-control-select-indicator-padding));
   --pgn-spacing-form-control-select-indicator-padding: 1rem;
   --pgn-spacing-form-control-select-padding-x-lg: var(--pgn-spacing-form-input-padding-x-lg);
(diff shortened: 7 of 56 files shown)


openedx-webhooks:

### What's next?

*Please work through the following steps to get your changes ready for engineering review:*

#### :radio_button: Get product approval

#### :radio_button: Provide context

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation





#### :radio_button: Get a green build

#### :radio_button: Let us know that your PR is ready for review:

### Who will review my changes?


This repository is currently maintained by `@openedx/paragon-working-group`.  Tag them in a comment and let them know that your changes are ready for review.


### Where can I find more information?

### When can I expect my changes to be merged?

- The need for product review

tokens/src/themes/light/components/Annotation.json
```diff
@@ -4,7 +4,22 @@
...
       "box-shadow": {
         "source": "$annotation-box-shadow",
-        "$value": "drop-shadow(0 2px 4px rgba(0, 0, 0, .15)) drop-shadow(0 2px 8px rgba(0, 0, 0, .15))"
+        "$value": [
+          {
+            "color": "rgba(0, 0, 0, .15)",
+            "offsetX": "0",
+            "offsetY": "2px",
+            "blur": "4px",
+            "dropShadow": true
```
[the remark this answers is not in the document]
adamstankiewicz: Also, an inform on a related bit to `expand`. See https://github.com/openedx/paragon/pull/3219 (specifically [this comment](https://github.com/openedx/paragon/pull/3219#discussion_r1748101859)), which extends the `expand.typesMap` to include the types map from `sd-transforms` ([GitHub](https://github.com/tokens-studio/sd-transforms)).

Ideally, the solution will be compatible with `sd-transforms` as well 🤞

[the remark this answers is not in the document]
PKulkoRaccoonGang: Thank you! Added expanded tokens, please take a look

src/Card/index.scss
```diff
@@ -353,7 +353,15 @@ a.pgn__card {
...
-      box-shadow: var(--pgn-elevation-box-shadow-level-1);
+      box-shadow:
+        var(--pgn-elevation-box-shadow-level-1-1-offset-y)
+        var(--pgn-elevation-box-shadow-level-1-1-offset-x)
+        var(--pgn-elevation-box-shadow-level-1-1-blur)
+        var(--pgn-elevation-box-shadow-level-1-1-color),
+        var(--pgn-elevation-box-shadow-level-1-2-offset-y)
+        var(--pgn-elevation-box-shadow-level-1-2-offset-x)
+        var(--pgn-elevation-box-shadow-level-1-2-blur)
+        var(--pgn-elevation-box-shadow-level-1-2-color);
```
adamstankiewicz, line 364: In this case, we could something similar as above comment (i.e., create an abstracted CSS variable that consolidates each of these independent CSS variables into one).

However, something I'm still contemplating (curious for your input!) is whether such a abstracted CSS variable should be manually defined here more like an CSS variable, e.g. `--pgn-elevation-card-box-shadow`, which would contain the specific parts/layered box-shadow as exists currently.

While this would give theme authors the ability to override box-shadow based on `--pgn-elevation-card-box-shadow` OR each individual CSS variable part, I might wonder if there should be the abstracted group of `--pgn-elevation-box-shadow-level-1` created for each elevation level out-of-the-box, whereby this `box-shadow` could continue to use `--pgn-elevation-box-shadow-level-1`.

The challenge here is how/where to generate the abstracted CSS variable that acts as the grouping for the independent CSS variable parts. 

For example, is there a way to get `--pgn-elevation-box-shadow-level-1` programmatically generated during `build-tokens`, or will we need to manually create these abstraction variables ourselves? 🤔

brian-smith-tcril: My mind goes to either:
* Manually creating the abstraction instead of trying to create the abstraction based on the expanded set

or

* Making the expanded set based on the abstraction (adding `--pgn-elevation-box-shadow-level-1` automatically adds all the expanded ones).

That being said, I think both trying to create the abstraction from the expanded or create the expanded from the abstraction will add complexity.

I believe handling it manually is likely the solution that will lead to the fewest snags.

adamstankiewicz: This is also where I'm leaning (i.e., manually created abstractions), as I'm not sure whether `style-dictionary` actually could support doing something like this programmatically.

Regarding _where_ we manually define these CSS variable abstractions (e.g., `--pgn-elevation-box-shadow-level-1`), it probably shouldn't go in the generated output from `style-dictionary` (i.e., the `variables.css` files) as those should remain only for automated output to prevent overriding of custom changes when running `build-tokens`.

Maybe we define a new file under `core` and `themes/light` to define these abstractions for more "global" variables (e.g., something like `abstraction-variables.css`), which gets included in the `index.css` where we currently import `variables.css`).

If the abstraction is specific to an individual component (e.g., `Annotation`'s `filter` style property), the abstraction could be defined closer to where it's used as described in an [above comment](https://github.com/openedx/paragon/pull/3222#discussion_r1759732065).

brian-smith-tcril: > something like `abstraction-variables.css`

I like this a lot!

> the abstraction could be defined closer to where it's used as described

I'm torn on this idea. On one side it feels like a cleaner way to structure things, but on the other side it feels like it'd be harder for someone trying to make overrides to know all the abstractions available to them. Maybe there's a way we can have our cake and eat it too here I'm missing :thinking:

src/Carousel/index.scss
```diff
@@ -58,7 +62,11 @@
   .active.carousel-item-right {
     z-index: 0;
     opacity: 0;
-    transition: opacity 0s var(--pgn-transition-carousel-base);
+    transition:
+      opacity 0s var(--pgn-transition-carousel-base-property)
+      var(--pgn-transition-carousel-base-duration)
```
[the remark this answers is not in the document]
PKulkoRaccoonGang: 0s in this context is the delay time before the transition starts for the opacity property. I think we shouldn't use the `var(--pgn-transition-carousel-base-delay)` variable here, nor should we use `var(--pgn-transition-carousel-base-duration)`, since this variable refers to the delay time for the `transition` property. I think if we want to get rid of the hardcoded `0s` value, we'll need to create a new token for that. Or we can leave the value as it is now, since there was no variable for it before.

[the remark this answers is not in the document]
brian-smith-tcril: I'm not sure I'm fully following how this fits into the css `transition` syntax. On ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/transition)) the closest I can find is

```
/* property name | duration | easing function | delay */
transition: margin-right 4s ease-in-out 1s;
```

which before this change would be

|  |  |
| - | - |
| property name | `opacity` |
| duration | `0s` |
| easing function | part of `var(--pgn-transition-carousel-base)` |
| delay | part of `var(--pgn-transition-carousel-base)` |

but after this change we have

|  |  |
| - | - |
| property name | `opacity` |
| duration | `0s` |
| easing function | `var(--pgn-transition-carousel-base-property)` |
| delay | `var(--pgn-transition-carousel-base-duration)` |
| ? | `var(--pgn-transition-carousel-base-timing-function)` |
| ? | `var(--pgn-transition-carousel-base-delay)` |

is there something I'm missing here?

.stylelintrc.json
```diff
@@ -34,6 +34,7 @@
     "color-function-notation": "legacy",
     "value-keyword-case": ["lower", {
       "ignoreProperties": ["/font-family/"]
-    }]
+    }],
+    "custom-property-empty-line-before": null
```
PKulkoRaccoonGang, line 38: [inform]: Disabled this rule because I think adding indents between abstract variables improves readability.

lib/build-tokens.js
```diff
@@ -154,13 +154,13 @@ async function buildTokensCommand(commandArgs) {
...
   // Build tokens for each configuration
-  await Promise.all(configs.map(async ({ config, isThemeVariant, themeVariant }) => {
+  await Promise.all(configs.map(async ({ config, themeVariant }) => {
     const sd = new StyleDictionary(config);
     await sd.cleanAllPlatforms();
     await sd.buildAllPlatforms();
     createIndexCssFile({
       buildDir,
-      isThemeVariant: !!isThemeVariant,
+      isThemeVariant: !!themeVariant,
```
PKulkoRaccoonGang, line 163: I noticed that `isThemeVariant` always has the value `undefined` and in `styles/css/themes/light` the `index.css` file is not generated

src/Annotation/index.scss
```diff
@@ -4,7 +4,19 @@
...
+      var(--pgn-elevation-annotation-box-shadow-1-offset-y)
+      var(--pgn-elevation-annotation-box-shadow-1-blur)
+      var(--pgn-elevation-annotation-box-shadow-1-color)
+    )
+    drop-shadow(
+      var(--pgn-elevation-annotation-box-shadow-2-offset-x)
+      var(--pgn-elevation-annotation-box-shadow-2-offset-y)
+      var(--pgn-elevation-annotation-box-shadow-2-blur)
+      var(--pgn-elevation-annotation-box-shadow-2-color)
+    );
```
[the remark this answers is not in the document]
PKulkoRaccoonGang: Thanks for the clarification 👍

src/Card/index.scss
PKulkoRaccoonGang: Thanks, added changes in the last commit, please take a look

tokens/src/core/components/Carousel.json
```diff
@@ -25,11 +25,20 @@
...
+          "duration": "{transition.carousel.duration}",
+          "delay": "0ms",
+          "timingFunction": "ease-in-out"
+        }
+      },
       "duration": { "source": "$carousel-transition-duration", "$value": ".6s", "$type": "duration" },
-      "indicator": { "source": "$carousel-indicator-transition", "$value": "opacity .6s ease", "$type": "transition" },
-      "control": { "source": "$carousel-control-transition", "$value": "opacity .15s ease", "$type": "transition" }
+      "indicator": { "source": "$carousel-indicator-transition", "$value": "opacity .6s ease" },
+      "control": { "source": "$carousel-control-transition", "$value": "opacity .15s ease" }
```
adamstankiewicz, line 41: [curious/clarification] Are these tokens related to `transition` also composite tokens that should define the transition in their individual parts?

Something like:

```json
{
  "transition": {
    "carousel": {
      "indicator": { 
        "source": "$carousel-indicator-transition",
        "$value": {
          "property": "opacity",
          "duration": "{transition.carousel.duration}",
          "delay": 0,
          "timingFunction": "ease",
        }
      }
    }
  }
}
```

src/Card/index.scss
brian-smith-tcril: That looks great! I'm still not 100% sure what the long term plan for where abstractions live should be, but for now with `abstraction-variables.css` being ~300 lines having everything directly in there makes sense. Thanks for putting it together!

tokens/src/core/components/Carousel.json
brian-smith-tcril: This got me curious about other possible composite values so I did a regex search on this branch for `"\$value":\s*"(?![^"]*\s[^"]*$)([^"]*\s[^"]*)"` - it caught the `calc` ones but I also found

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/components/DataTable.json#L24

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/components/Dropdown.json#L40

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/components/Modal.json#L30

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/components/Modal.json#L34

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/components/ProgressBar.json#L34

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/components/ProgressBar.json#L35

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/components/Form/spacing.json#L80

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/components/Form/transition.json#L7

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/components/Form/transition.json#L11

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/global/transition.json#L6

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/global/transition.json#L11

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/global/transition.json#L17

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/global/transition.json#L22

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/global/typography.json#L14

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/global/typography.json#L25

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/themes/light/components/Nav.json#L10

https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/themes/light/components/Form/other.json#L102

I don't think all of these are problematic but it'd be good to think through which ones are OK and which ones need to be updated.

src/DataTable/index.scss
```diff
@@ -94,7 +102,6 @@
     background-color: var(--pgn-color-data-table-bg-base);
     border-radius: var(--pgn-size-border-radius-base);
     box-shadow: var(--pgn-elevation-data-table-box-shadow);
-    padding: var(--pgn-spacing-data-table-padding-small);
```
brian-smith-tcril, line 97: Is this removal intentional?

styles/css/themes/light/abstraction-variables.css
```diff
@@ -0,0 +1,301 @@
+:root {
+  --pgn-elevation-box-shadow-level-1:
```
adamstankiewicz, line 2: [nit/suggestion] I might structure this file with comments to split out the different "groups" of variables to make the file more parseable/readable.

For example, all the `elevation` related tokens could be grouped under a section heading like:

```css
/**
 * ELEVATION
 */
```

or something similar :)

PKulkoRaccoonGang: Added, good idea 💯

styles/css/themes/light/index.css
```diff
@@ -1,2 +1,3 @@
 @import "variables.css";
 @import "utility-classes.css";
+@import "abstraction-variables.css";
```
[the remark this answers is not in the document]
PKulkoRaccoonGang: Yeah, it's make sense, added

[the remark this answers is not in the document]
PKulkoRaccoonGang: Based on Brian's suggestions, I added an updated import order. Thank you for noticing

light/index.css
```
@import "variables.css";
@import "abstraction-variables.css";
@import "utility-classes.css";
```

core/index.css
```
@import "variables.css";
@import "abstraction-variables.css";
@import "custom-media-breakpoints.css";
```

src/DataTable/index.scss
PKulkoRaccoonGang: No, I returned the property back

tokens/src/core/components/Carousel.json
PKulkoRaccoonGang: > [curious/clarification] Are these tokens related to transition also composite tokens that should define the transition in their individual parts?

Yes, thanks
Added additional composite values

> This got me curious about other possible composite values so I did a regex search on this branch for "\$value":\s*"(?![^"]*\s[^"]*$)([^"]*\s[^"]*)" - it caught the calc ones but I also found

Thank you for providing this list, now we will definitely not miss anything! 💯 

> Find where we're using the consolidated token
`"header": { "source": "$dropdown-header-padding", "$value": "{spacing.dropdown.padding.y.base} {spacing.dropdown.padding.x.item}" } `

This variable came to us from bootstrap 4, during the creation of tokens we transferred [the file](https://github.com/openedx/paragon/blob/Peter_Kulko/expanded-tokens-value/src/Dropdown/dropdown-bootstrap.scss#L171) from bootstrap to determine its value from Paragon.

> Not sure... sidenote: does this need a calc?
"$value": "{other.content.form.control.select.indicator.icon} no-repeat right {spacing.form.input.padding.y.base} center / {color.form.control.select.bg.base}", 

`/` is a part of the syntax that allows you to specify both the position of the image and its size at the same time (it is not a mathematical operator).

src/Carousel/index.scss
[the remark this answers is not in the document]
PKulkoRaccoonGang: Good catch! There is a missing comma here because we have two different transitions: a transition for the `opacity` property and a transition for `transform`.

Added `transition: opacity 0s, var(--pgn-transition-carousel-base);`

styles/css/core/abstract-variables.css
brian-smith-tcril, line 1: I'm fine either way, but in previous comments we used the name `abstraction-variables` instead of `abstract-variables`. @adamstankiewicz do you have strong feelings about the filename?

styles/css/themes/light/abstraction-variables.css
brian-smith-tcril, line 1: To go along with my previous comment, if nothing else I think we should use the *same* filename in `core` and here. So either rename this file to `abstract-variables` or rename the core file to `abstraction-variables`.

styles/css/themes/light/index.css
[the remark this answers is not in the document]
brian-smith-tcril: The diff is still showing this as coming after `utility-classes`.

styles/scss/core/core.scss
```diff
@@ -16,6 +16,7 @@
 // Paragon core CSS
 @import "../../css/core/variables";
 @import "../../css/core/custom-media-breakpoints";
+@import "../../css/core/abstract-variables";
```
brian-smith-tcril, line 19: Looks like the change mentioned [here](https://github.com/openedx/paragon/pull/3222#discussion_r1770577421) didn't make it in?

www/src/scss/openedx-theme.scss
```diff
@@ -1,4 +1,5 @@
 @import "~paragon-style/scss/core/core";
 @import "~paragon-style/css/themes/light/variables";
 @import "~paragon-style/css/themes/light/utility-classes";
+@import "~paragon-style/css/themes/light/abstraction-variables";
```
brian-smith-tcril, line 4: Should this be rearranged as per https://github.com/openedx/paragon/pull/3222#discussion_r1770577421 ?

brian-smith-tcril (commented):
This is coming together beautifully! Left a couple comments where the diff doesn't match up with what was being discussed in prior comment threads, but overall it's looking great!

styles/css/core/abstract-variables.css
PKulkoRaccoonGang: This is a typo. We already have an `abstraction-variables.css` file for theme variables, in core variables the file should be named the same. Fixed

styles/css/themes/light/abstraction-variables.css
PKulkoRaccoonGang: I suggest that both files be called `abstraction-variables.css`, if you don't mind :)

styles/scss/core/core.scss
PKulkoRaccoonGang: Corrected to `abstraction-variables`

www/src/scss/openedx-theme.scss
PKulkoRaccoonGang: We now use the name `abstraction-variables` for core variables and theme-related variables.

styles/css/themes/light/abstraction-variables.css
brian-smith-tcril: Sounds great to me!

styles/scss/core/core.scss
brian-smith-tcril: This comment was about the import order, not the naming

www/src/scss/openedx-theme.scss
brian-smith-tcril: This comment was about the import order, not the naming

brian-smith-tcril: GitHub didn't note this as outdated, looks like it's in the proper order now! Thanks!

src/Carousel/index.scss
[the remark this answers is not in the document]
adamstankiewicz: With the token $value as an array, it'd end up looking something like this:

```scss
  :root {
    --pgn-transition-carousel-base):
      var(--pgn-transition-carousel-base-1-property) --pgn-transition-carousel-base-1-duration,
      var(--pgn-transition-carousel-base-2-property) --pgn-transition-carousel-base-2-duration --pgn-transition-carousel-base-2-timing-function --pgn-transition-carousel-base-2-delay;
  }
    
  transition: var(--pgn-transition-carousel-base);
```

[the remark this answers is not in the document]
PKulkoRaccoonGang: Added in [`24fd229` (#3222)](https://github.com/openedx/paragon/pull/3222/commits/24fd2297063f38641f376bbeadf4e8b98d16bfd8)

Also added automatic generation of the `abstraction-variables.css` file and added the correct order of imports in the `index.css` files for core and themes tokens.

tokens/src/core/components/Dropdown.json
```diff
@@ -37,7 +37,13 @@
           "base": { "source": "$dropdown-padding-y", "$value": ".5rem" },
           "item": { "source": "$dropdown-item-padding-y", "$value": ".25rem" }
         },
-        "header": { "source": "$dropdown-header-padding", "$value": "{spacing.dropdown.padding.y.base} {spacing.dropdown.padding.x.item}" }
+        "header": {
+          "source": "$dropdown-header-padding",
+          "$value": {
+            "x": "{spacing.dropdown.padding.x.item}",
+            "y": "{spacing.dropdown.padding.y.base}"
+          }
```
brian-smith-tcril, line 45: Having `x.item` as `x` and `y.base` as `y` seems a little odd to me. We have both `base` and `item` in `x` and `y` for the general `padding` token right above this, should we do that here too?

tokens/src/core/global/transition.json
```diff
@@ -3,23 +3,47 @@
     "$type": "transition",
     "base": {
       "source": "$transition-base",
-      "$value": "all .2s ease-in-out",
+      "$value": {
+        "property": "all",
+        "duration": ".2s",
+        "timing-function": "ease-in",
```
brian-smith-tcril, line 9: It looks like this has changed from what was here before

```suggestion
        "timing-function": "ease-in-out",
```

tokens/src/core/global/typography.json
```diff
@@ -11,7 +11,16 @@
...
+            "$value": {
+              "primary": "-apple-system",
+              "secondary": "BlinkMacSystemFont",
+              "tertiary": "'Segoe UI'",
+              "alternate-1": "Roboto",
+              "alternate-2": "'Helvetica Neue'",
+              "alternate-3": "Arial",
+              "alternate-4": "'Noto Sans'",
+              "fallback": "sans-serif"
+            },
```
brian-smith-tcril, line 23: It seems this no longer has the following
```
'Apple Color Emoji',  'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji'
```

Should we add all of those as additional alternates? Is there a better way to handle this?

brian-smith-tcril (commented):
I went through all of the places mentioned in https://github.com/openedx/paragon/pull/3222#discussion_r1763520372 and added checkmarks. Most of them look perfect! I left comments on the ones that I noticed issues with:

#### Find where we're using the consolidated token
- [x] Done?
https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/components/Dropdown.json#L40

#### Make composite token
- [x] Done?
https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/global/transition.json#L6

#### Make composite token
- [ ] Done?
https://github.com/openedx/paragon/blob/372df044a9688670784a82d68a32441ecafc7b5a/tokens/src/core/global/typography.json#L14

tokens/src/core/components/Dropdown.json
PKulkoRaccoonGang: Probably yes, since we had the SCSS variable [$dropdown-header-padding](https://github.com/openedx/paragon/blob/1a6b9b33cf09423a9fe760c3f2c0615acc8f31ef/src/Dropdown/_variables.scss#L33) before

tokens/src/core/global/transition.json
PKulkoRaccoonGang: Corrected in [(#3222)](https://github.com/openedx/paragon/pull/3222/commits/cf989344f882f25d15ee2c56b0c03cfb1df2a1d9)

tokens/src/core/global/typography.json
PKulkoRaccoonGang: After reviewing the Token Studio library documentation, I noticed that for properties like box shadow, there is an interface that allows users to adjust each value of the composite shadow token (e.g., x-axis, y-axis, etc.). In the context of the shadows provided by Paragon, this makes perfect sense, and consumers will be able to fully utilize this functionality.

tokens-boxShadow-form-empty-2-01

Similarly, on [the Typography compositional tokens](https://docs.tokens.studio/token-types/token-type-typography-composite) description page, we can see that fonts are also composed of several elements, and an interface is available for editing them.

tokens-typography-form-annotated-2-01

However, I observed that for the fontFamily property, there is only a single selection field without a clear distinction between alternative fonts and fallbacks. Should we consider splitting the fontFamily property into separate tokens for each font (e.g., primary, alternative, fallback), or does that approach not add value? I think we should keep the fontFamily values ​​as a single (non-divisible) value. What are your thoughts?

brian-smith-tcril: I'd be fine with keeping it as one long string with not splitting it out and keeping what we had before, so just setting the `fontFamily` value to
```
"-apple-system, BlinkMacSystemFont, 'Segoe UI',  Roboto, 'Helvetica Neue', Arial, 'Noto Sans', sans-serif, 'Apple Color Emoji',  'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji'"
```

I'd like to hear @adamstankiewicz's thoughts on this though

tokens/src/core/components/Dropdown.json
```diff
@@ -37,7 +37,13 @@
           "base": { "source": "$dropdown-padding-y", "$value": ".5rem" },
           "item": { "source": "$dropdown-item-padding-y", "$value": ".25rem" }
         },
-        "header": { "source": "$dropdown-header-padding", "$value": "{spacing.dropdown.padding.y.base} {spacing.dropdown.padding.x.item}" }
+        "header": {
+          "source": "$dropdown-header-padding",
+          "$value": {
+            "x": "{spacing.dropdown.padding.x.item}",
+            "y": "{spacing.dropdown.padding.y.base}"
```
brian-smith-tcril, line 44: After playing around with this on the docs site a bit it seems like having the y padding tied to the menu base y padding is a bit confusing. I wouldn't expect the padding above the header to jump from `1rem` when `dropdown-padding-y` is set to `0.5rem` to `2rem` when setting `dropdown-padding-y` to `1rem`.

To avoid confusion I think the best path forward here is to just have the `y` value for header padding defined as an independent value.

```suggestion
            "y": ".5rem"
```

tokens/src/core/global/typography.json
brian-smith-tcril, line 23: Based on the example [here](https://tr.designtokens.org/format/#example-19)

```suggestion
            "$value": ["-apple-system", "BlinkMacSystemFont", "Segoe UI", "Roboto", "Helvetica Neue", "Arial", "Noto Sans", "sans-serif", "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji"],
```

styles/css/core/abstraction-variables.css
```diff
@@ -0,0 +1,76 @@
...
+    var(--pgn-transition-form-input-2-property)
+    var(--pgn-transition-form-input-2-duration)
+    var(--pgn-transition-form-input-2-timing-function)
+    var(--pgn-transition-form-input-2-delay)
+    var(--pgn-transition-form-input-2-behavior);
+
+  /**
+   * TYPOGRAPHY
+   */
+  --pgn-typography-font-family-sans-serif:
```
adamstankiewicz, line 48: [curious] Related to @brian-smith-tcril's previous comment about adopting the `$value` for typography font-family as an array, I wonder if the need to explicitly create these abstraction variable for typography is needed vs. relying on the standard output from style-dictionary.

styles/css/core/variables.css
```diff
@@ -30,27 +30,27 @@
   --pgn-elevation-modal-zindex: 1050;
   --pgn-elevation-modal-backdrop-zindex: 1040;
   --pgn-elevation-dropdown-zindex: 1000;
-  --pgn-transition-collapse-width-behavior: unset; /* Collapse transition for width that takes 350ms */
+  --pgn-transition-collapse-width-behavior: normal; /* Collapse transition for width that takes 350ms */
```
PKulkoRaccoonGang, line 33: [inform]: When rechecking the documentation site I noticed that some transitions are not working. CSS transition is incorrect due to the use of `unset` value, replaced `unset` with `normal `

styles/scss/core/_variables.scss
```diff
@@ -468,6 +468,7 @@ $border-radius-sm: var(--pgn-size-border-radius-sm) !default;
 $rounded-pill: var(--pgn-size-rounded-pill) !default;
 
 $level-1-box-shadow: var(--pgn-elevation-box-shadow-level-1) !default;
+
```
adamstankiewicz, line 471: nit: unnecessary new line (might make sense to keep `$level-1-box-shadow` co-located with the other `$level-*` SCSS variables below.

tokens/src/core/components/Carousel.json
```diff
@@ -29,38 +29,33 @@
...
-            "timingFunction": "ease-in-out"
-          }
-        ]
+        "$value": {
+          "property": "transform",
+          "duration": "{transition.carousel.duration}",
+          "timingFunction": "ease-in-out",
+          "delay": "0ms",
+          "behavior": "normal"
+        }
```
PKulkoRaccoonGang, line 38: [inform]: According to the original token, the `$carousel-transition` has one value:
```
"base": { "source": "$carousel-transition", "$value": "transform {transition.carousel.duration} ease-in-out", "$type": "transition" },
```

tokens/src/core/global/typography.json
```diff
@@ -22,7 +23,7 @@
         },
         "monospace": {
           "source": "$font-family-monospace",
-          "$value": "SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono',  'Courier New', monospace",
+          "$value": ["SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono',  'Courier New', monospace"],
```
adamstankiewicz, line 26: It looks like this `$value` is an array with a single item vs. an array containing items for each font-family. Should it be something like the following?

```suggestion
          "$value": ["SFMono-Regular", "Menlo", "Monaco", "Consolas", "Liberation Mono",  "Courier New", "monospace"],
```

tokens/src/core/components/Carousel.json
```diff
@@ -29,38 +29,33 @@
...
       "duration": { "source": "$carousel-transition-duration", "$value": ".6s", "$type": "duration" },
       "indicator": {
         "source": "$carousel-indicator-transition",
         "$value": {
           "property": "opacity",
           "duration": "{transition.carousel.duration}",
+          "timingFunction": "ease",
           "delay": "0ms",
-          "timingFunction": "ease"
+          "behavior": "normal"
```
PKulkoRaccoonGang, line 48: [inform]: Some transitions tokens were missing the `behavior` parameter, now all tokens that are used to add a transitive have the same parameters:
- property
- duration
- timingFunction
- delay
- behavior

tokens/src/core/global/typography.json
PKulkoRaccoonGang: Added in new commit [`2786dd4` (#3222)](https://github.com/openedx/paragon/pull/3222/commits/2786dd4056cd5e62ee5aae402788a7008d0ae8b4#diff-c8641349d8aa8a3b4b6654afc986d0d0d1b50d60899deb03e3b91c688e68fcd1L23)

styles/scss/core/_variables.scss
PKulkoRaccoonGang: Agree, removed in [`0905115` (#3222)](https://github.com/openedx/paragon/pull/3222/commits/09051155b1aa93b654844015ef9289b0387736d2#diff-047cf3e37a3d52f4ec293da65abc4a20d5364ed9effd900f64b33c7981110976L471)

tokens/src/core/components/Dropdown.json
PKulkoRaccoonGang: Thanks, removed in [`0905115` (#3222)](https://github.com/openedx/paragon/pull/3222/commits/09051155b1aa93b654844015ef9289b0387736d2#diff-ad58aa35fe3f66fc8148f05a3e6dfd4bf47b9a943f7bb407521daedd23d6c573L44)

PKulkoRaccoonGang:
@brian-smith-tcril @adamstankiewicz there seems to be a small issue with the fonts after the latest changes to the documentation site. It may have happened after this [`da00992` (#3186)](https://github.com/openedx/paragon/pull/3186/commits/da00992533e1ac51b1e9bd53b43e8c50239e8a35) commit.
Is this the expected font size?

**Master**
image

**Current**
image

image

tokens/src/themes/light/components/ProgressBar.json
```diff
@@ -3,7 +3,9 @@
     "$type": "color",
     "progress-bar": {
       "bg": { "source": "$progress-bg", "$value": "transparent" },
-      "border": { "source": "$progress-bar-border-color", "$value": "{color.gray.500}" },
+      "border": { "source": "$progress-bar-border-color", "$value": "{color.gray.500}" }
+    },
+    "progress": {
       "bar": {
```
adamstankiewicz, line 9: [curious/clarification] Can you help me understand why there's a token key of `progress-bar` on line 4, and another `progress.bar` defined on lines 8-9? I believe both of these result in `color.progress-bar.*` in the resulting output CSS variables.

Would it potentially make sense to co-locate tokens under `progress-bar` or `progress.bar` to avoid having 2 different keys result in the same structure within the output CSS variables? Or is there a reason why these are split out separately?

adamstankiewicz:
> @brian-smith-tcril @adamstankiewicz there seems to be a small issue with the fonts after the latest changes to the documentation site. It may have happened after this [`da00992` (#3186)](https://github.com/openedx/paragon/pull/3186/commits/da00992533e1ac51b1e9bd53b43e8c50239e8a35) commit. Is this the expected font size?

@PKulkoRaccoonGang I believe this style discrepancy might be resolved by addressing [this comment](https://github.com/openedx/paragon/pull/3222#discussion_r1805061707). By ensuring the token value is an array of multiple elements (one per font) vs. a single array element containing all fonts, it seems like the font size and typography style will be fixed, as seen below:

image

tokens/src/core/global/typography.json
brian-smith-tcril: That matches up with the syntax used in https://github.com/openedx/paragon/pull/3222#discussion_r1803396628

styles/css/core/variables.css
brian-smith-tcril: Looks good. I am curious about which transitions you found that weren't working before the change. I'm wondering if the shorthand we were using before led to some having `transition-behavior` set to `allow-discrete` instead of `normal` ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/transition-behavior))

tokens/src/themes/light/components/ProgressBar.json
PKulkoRaccoonGang: Most likely, the reason for the confusion is separate `bg` tokens for `progress-bar` and `progress`. Moved the `bg` color separately for `progress`. Nice catch!

```
.progress {
  display: flex;
  height: var(--pgn-size-progress-bar-height-base);
  overflow: hidden;
  line-height: 0;
  background-color: var(--pgn-color-progress-bg);
  box-shadow: var(--pgn-elevation-progress-bar-box-shadow);

  @include font-size(var(--pgn-typography-progress-bar-font-size));
  @include border-radius(var(--pgn-size-progress-bar-border-radius));
}

.progress-bar {
  display: flex;
  flex-direction: column;
  justify-content: center;
  overflow: hidden;
  color: var(--pgn-color-progress-bar-base);
  text-align: center;
  white-space: nowrap;
  background-color: var(--pgn-color-progress-bar-bg-base);

  @include transition(var(--pgn-transition-progress-bar-transition));
}
```

styles/css/core/variables.css
PKulkoRaccoonGang: While checking the Form.Control component, I noticed issues with transitions. It seemed like the transition was missing, so I decided to copy the value of the CSS variable var(--pgn-transition-form-input) and manually apply it to the transition property. However, this resulted in an incorrect value, indicating that the var(--pgn-transition-form-input) variable holds the wrong value.

image

After this experiment, I realized that the value `unset` is not valid for transitions. I have replaced it in all tokens related to transitions.

tokens/src/core/global/typography.json
PKulkoRaccoonGang: Corrected, thanks!

styles/css/core/variables.css
brian-smith-tcril: Thanks for the added context! I checked on the docs site to see the pre-design-tokens transition value for `Form.Control` and it does appear to be using `transition-behavior: normal;`.

image

 I don't know of any places we use `transition-behavior: allow-discrete` so I think explicitly setting it to `normal` here should be fine.

brian-smith-tcril (approved):
Thanks for sticking through the wild ride of a PR review this was! Let's land this!

adamstankiewicz (approved):

PR #3320, opened 2024-12-05
chore: add typescript types for `BaseCard`

brian-smith-tcril:
Hopefully this does the trick for getting https://github.com/openedx/paragon/pull/3318 into `.tsx` land

### DIFF
--- a/src/Card/BaseCard.tsx
+++ b/src/Card/BaseCard.tsx
@@ -2,11 +2,43 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
+import type { ComponentWithAsProp, BsPropsWithAs } from '../utils/types/bootstrap';
+
+// @ts-ignore
 import CardBody from './CardBody';
 
 const BASE_CARD_CLASSNAME = 'card';
 
-const BaseCard = React.forwardRef(
+const colorVariants = [
+  'primary',
+  'secondary',
+  'success',
+  'danger',
+  'warning',
+  'info',
+  'dark',
+  'light',
+] as const;
+
+const textVariants = [
+  'white',
+  'muted',
+] as const;
+
+type ColorVariant = typeof colorVariants[number];
+type TextVariant = typeof textVariants[number];
+interface Props extends BsPropsWithAs {
+  prefix?: string;
+  bgColor?: ColorVariant;
+  textColor?: ColorVariant | TextVariant;
+  borderColor?: ColorVariant;
+  hasBody?: boolean;
+  className?: string;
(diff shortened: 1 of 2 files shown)


src/Card/BaseCard.tsx
```diff
@@ -2,21 +2,53 @@ import React from 'react';
...
+const BaseCard : BaseCardType = React.forwardRef<HTMLDivElement, Props>(
   (
     {
       prefix,
       className,
       bgColor,
       textColor,
       borderColor,
-      hasBody,
+      hasBody = false,
```
brian-smith-tcril, line 49: I'm not sure what codecov wants from me here

https://app.codecov.io/gh/openedx/paragon/pull/3320?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=openedx

I'd think
https://github.com/openedx/paragon/blob/f3beb20b2cde1750de1c15b15160f80b73df9013/src/Card/tests/BaseCard.test.jsx#L32-L52
would cover this

PKulkoRaccoonGang: I also encountered this problem when writing tests for this component. I managed to solve it by removing the default value for hasBody :)

PKulkoRaccoonGang (approved):
Thanks for this!

src/Card/BaseCard.tsx
brian-smith-tcril: > I managed to solve it by removing the default value for `hasBody`

In the component or in the tests? Removing default values just to make codecov happy doesn't feel like a good call to me. I'd love to figure out what codecov is looking for here, but I'm also open to letting it slide as-is considering the overall (non-diff) coverage is still fine.

bradenmacdonald: I think: Codecov is technically correct. The problem is that the default value is never used, because the legacy `BaseCard.defaultProps = { ..., hasBody: false, ... }` takes effect, and the parameter default never has a chance to get used. I recommend getting rid of the `BaseCard.defaultProps` entirely, and moving all the defaults to be ES6 parameter defaults, and it should solve this problem. (`defaultProps` is deprecated for function components anyways)

brian-smith-tcril: Great call! Updated in https://github.com/openedx/paragon/compare/d68b2ff3e4ebfdc8bacc039b6b515c3150d3a8b9..f95bac1ce85e921a7aa79f6200074b506531712d and codecov is happy now!

jciasenza:
Great Job !!!

PR #3323, opened 2024-12-06
feat: use data uri for card fallback image

brian-smith-tcril:
### Deploy Preview

## Merge Checklist

## Post-merge Checklist


bradenmacdonald (approved):
Wow, that was fast! Thanks :)
I agree "no image at all" would be better, but this is a nice improvement for now.

brian-smith-tcril:
So this is failing on
https://github.com/openedx/paragon/blob/1b90fdacf3dd6738c61a7dfdfe3de5683ecf8faf/src/Card/tests/CardImageCap.test.jsx#L122-L130
now, but I'm not sure what "hiding component if it isn`t fallbackSrc and src don`t work" is supposed to be testing, or if it's accomplishing it's intended purpose.

It was passing before because the fallback png would lead to
https://github.com/openedx/paragon/blob/1b90fdacf3dd6738c61a7dfdfe3de5683ecf8faf/__mocks__/fileMock.js#L1

I tried throwing
```jsx
jest.mock('../CardFallbackDefaultImage', () => ({ cardSrcFallbackImg: 'test-file-stub' }));
```
into the test file to see if I could fix the error really quickly, but it still failed.

Before spending any real amount of time fighting this test I'd like to understand what it is actually trying to verify.

brian-smith-tcril:
@PKulkoRaccoonGang I checked the git blame on the test and it looks like you added it last year as part of https://github.com/openedx/paragon/pull/1994.

Any chance you remember what
```jsx
expect(srcImg.src.endsWith('test-file-stub')).toEqual(true);
```
is verifying?

PKulkoRaccoonGang:
@brian-smith-tcril, thank you for these changes!

In the [package.json](https://github.com/openedx/paragon/blob/master/package.json#L157), we have the following configuration:

```json
"moduleNameMapper": {
  "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/__mocks__/fileMock.js",
  "\\.(css|scss)$": "identity-obj-proxy"
}
```

This setup is necessary because Jest does not natively handle media files. Instead, a stub is provided to prevent errors during test execution. We use this approach in the [Avatar component](https://github.com/openedx/paragon/blob/master/src/Avatar/Avatar.test.jsx#L25), as it requires mocking a .svg image.

```javascript
expect(srcImg.src.endsWith('test-file-stub')).toEqual(true);
```

This check verifies that the mocked file (test-file-stub) is used as the source for the component’s src attribute when the actual image loading fails.

Since we now want to use an image in base64 format, there is no longer a need to mock the image. I propose updating the test as follows:

```javascript
it('renders the default image if both src and fallbackSrc fail to load', () => {
  render(<CardImageCapWrapper src="fakeURL" fallbackSrc="fakeURL" srcAlt="Src alt text" />);

  const srcImg = screen.getByAltText('Src alt text');
  fireEvent.load(srcImg);
  fireEvent.error(srcImg);
  expect(srcImg.src).toEqual(cardSrcFallbackImg);
});
```

I tested this change locally, and it works as expected.

brian-smith-tcril:
Thanks for the advice @PKulkoRaccoonGang! I've updated the test and everything looks good now!

PKulkoRaccoonGang (approved):

adamstankiewicz (approved):

brian-smith-tcril:
trying close/reopen to trigger the required deploy preview (switched base from `master` to `release-22.x`)

PR #3284, opened 2024-11-13
Reduce dist size by 500 KB - reinstate browserslist, removing transforms for ancient browsers

bradenmacdonald:
## Description

Currently, Paragon is built using Babel's [`preset-env`](https://babeljs.io/docs/babel-preset-env) but _without_ specifying any `targets` nor `.browserslistrc` config file. This [has the unfortunate effect](https://babeljs.io/docs/options#no-targets) that:

> When no targets are specified: Babel will assume you are targeting the oldest browsers possible. For example, @babel/preset-env will transform all ES2015-ES2020 code to be ES5 compatible.
>
> We recognize this isn’t ideal and will be revisiting this in Babel v8.
>
> We recommend setting `targets` to reduce the output code size.

So if you look at any of the files in `dist/`, you'll see that they are being transformed in an extremely inefficient way in order to support pre-ES5 browsers (that are more than a decade old!).

In this PR, I am proposing we change the build to use the recommended `targets: defaults` option. This small change to the config results in a _big_ 500KB savings to the transformed files in `dist` when `npm run build` is run:

Before: 2,919,906 bytes (4.5 MB on disk) for 740 items
After: 2,401,170 bytes (4.1 MB on disk) for 740 items

I have also removed [`@babel/plugin-proposal-class-properties`](https://www.npmjs.com/package/@babel/plugin-proposal-class-properties) and [`@babel/plugin-proposal-object-rest-spread`](https://www.npmjs.com/package/@babel/plugin-proposal-object-rest-spread) because they're both deprecated (click their links for details), and the replacements are now included in `preset-env` so don't need to be separately installed.

See also https://github.com/openedx/paragon/pull/3283 which makes a similar change for the icons.

## What about compatibility?

[...]

PR #2774, opened 2023-11-06
[BD-46] feat: updates to React 18

PKulkoRaccoonGang:
## Description

Include a description of your changes here, along with a link to any relevant Jira tickets and/or Github issues.

### Deploy Preview

Include a direct link to your changes in this PR's deploy preview here (e.g., a specific component page).

## Merge Checklist

* [ ] If your update includes visual changes, have they been reviewed by a designer? Send them a link to the Netlify deploy preview, if applicable.
* [ ] Does your change adhere to the documented [style conventions](https://github.com/openedx/paragon/blob/master/docs/decisions/0012-css-styling-conventions)?
* [ ] Do any prop types have missing descriptions in the Props API tables in the documentation site (check deploy preview)?
* [ ] Were your changes tested using all available themes (see theme switcher in the header of the deploy preview, under the "Settings" icon)?
* [ ] Were your changes tested in the `example` app?
* [x] Is there adequate test coverage for your changes?
* [ ] Consider whether this change needs to reviewed/QA'ed for accessibility (a11y). If so, please add `wittjeff` and `adamstankiewicz` as reviewers on this PR.

## Post-merge Checklist

* [ ] Verify your changes were released to [NPM](https://www.npmjs.com/package/@edx/paragon) at the expected version.
* [ ] If you'd like, [share](https://github.com/openedx/paragon/discussions/new?category=show-and-tell) your contribution in [#show-and-tell](https://github.com/openedx/paragon/discussions/categories/show-and-tell).
* [ ] 🎉 🙌 Celebrate! Thanks for your contribution.

PR #2767, opened 2023-11-01
[BD-46] [WIP] docs: upgrade gatsby to v5

viktorrusakov:
## Description

An attempt to upgrade gatsby to v5. Spoiler alert: everything breaks.

### Deploy Preview

Include a direct link to your changes in this PR's deploy preview here (e.g., a specific component page).

## Merge Checklist

* [ ] If your update includes visual changes, have they been reviewed by a designer? Send them a link to the Netlify deploy preview, if applicable.
* [ ] Does your change adhere to the documented [style conventions](https://github.com/openedx/paragon/blob/master/docs/decisions/0012-css-styling-conventions)?
* [ ] Do any prop types have missing descriptions in the Props API tables in the documentation site (check deploy preview)?
* [ ] Were your changes tested using all available themes (see theme switcher in the header of the deploy preview, under the "Settings" icon)?
* [ ] Were your changes tested in the `example` app?
* [ ] Is there adequate test coverage for your changes?
* [ ] Consider whether this change needs to reviewed/QA'ed for accessibility (a11y). If so, please add `wittjeff` and `adamstankiewicz` as reviewers on this PR.

## Post-merge Checklist

* [ ] Verify your changes were released to [NPM](https://www.npmjs.com/package/@edx/paragon) at the expected version.
* [ ] If you'd like, [share](https://github.com/openedx/paragon/discussions/new?category=show-and-tell) your contribution in [#show-and-tell](https://github.com/openedx/paragon/discussions/categories/show-and-tell).
* [ ] 🎉 🙌 Celebrate! Thanks for your contribution.

Issue #3329, opened 2024-12-09
"Build Docs" failing silently


brian-smith-tcril:
https://github.com/openedx/paragon/actions/runs/12241372269/job/34146296486 led to the docs site going down, but the job still passed. This shouldn't happen.

bradenmacdonald:
https://github.com/openedx/paragon/actions/runs/12816523865/job/35737656634 shows the "tests > Lint" check as passing when it clearly failed.

bradenmacdonald:
Same again on https://github.com/openedx/paragon/actions/runs/15402502375/job/43338207703 - the "tests > Lint" is failed, but the CI shows the overall "tests" build is passing.

brian-smith-tcril:
Digging into this a bit and it seems related to a few node/npm issues I found
* https://github.com/npm/cli/issues/6399
* https://github.com/npm/cli/issues/6506

I'm continuing to investigate.

brian-smith-tcril:
Trying to isolate the issue I found:

```diff
diff --git a/package.json b/package.json
index b95b7a3bd8..59a171d141 100644
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
   ],
   "sideEffects": false,
   "scripts": {
+    "fail": "npm run fail --workspace=www",
     "build": "make build",
     "build-docs": "npm run build --workspace=www",
     "commit": "commit",
diff --git a/www/package.json b/www/package.json
index 8c65d928c4..f85f81f7b7 100644
--- a/www/package.json
+++ b/www/package.json
@@ -55,6 +55,7 @@
   ],
   "license": "MIT",
   "scripts": {
+    "fail": "exit 1",
     "build": "gatsby build --prefix-paths && npm run build-themes && playroom build",
     "clean": "gatsby clean",
     "develop": "npm run build-themes && gatsby develop",
```

```sh
bsmith@aximdev:~/code/paragon$ npm run fail; echo $?

> @openedx/paragon@20.18.1 fail
> npm run fail --workspace=www


> paragon-pattern-library-documentation@1.0.0 fail
> exit 1

npm ERR! Lifecycle script `fail` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: paragon-pattern-library-documentation@1.0.0 
npm ERR!   at location: /home/bsmith/code/paragon/www 
0
```

```sh
bsmith@aximdev:~/code/paragon$ npm run fail --workspace=www; echo $?

[...]

PR #3367, opened 2024-12-16
Enable support for React 18, bump Gatsby to v5

bradenmacdonald:
## Description

We are long overdue for supporting React 18. The challenge is that bumping to React 18 and testing with it on our docs site also requires bumping Gatsby from v4 to v5 and `gatsby-plugin-mdx` from v3 to v5, which introduces [a ton of breaking changes to how Gatsby and MDX work](https://www.gatsbyjs.com/plugins/gatsby-plugin-mdx/#migrating-from-v3-to-v4).

These upgrades have been attempted before in https://github.com/openedx/paragon/pull/2774 and https://github.com/openedx/paragon/pull/2767 but so far nobody has been able to finish the work up. I'm thinking we need to bite the bullet sooner or later and get it sort out.

### Deploy Preview

https://deploy-preview-3367--paragon-openedx-v22.netlify.app/

Note: Compare to https://paragon-openedx-v22.netlify.app , not to https://paragon-openedx.netlify.app/ !

## Status

Working:
* Paragon build (`npm run build`)
* Paragon in general (the components on the docs site work as normal - see preview above)
* The docs site (except for some minor issues; see below)
* Test suite
* Lint checks
* Fully backwards compatible (still supports React 16 & 17)

## Issues and solutions

* [x] **Need to test this with an existing MFE to confirm backwards compatibility**
* [x] Modals seem to be throwing a JS error on the docs site and the `useIsVisible` hook ([link](https://deploy-preview-3367--paragon-openedx-v22.netlify.app/components/hooks/useisvisible/#sample-usage)) doesn't seem to be working properly anymore.
   - Fixed with 89791077c172b9e3b4c9171e82ea4a1b9cdb846e
* [x] The `icons` are showing a lot of babel compatibility boilerplate again, i.e. the changes from #3284 seem to be undone after upgrading to React 18.
   - Fixed by pinning `@edx/browserslist-config` to v1.3.0 to avoid https://github.com/openedx/browserslist-config/pull/16 which seems to be the source of the problem
* [x] [The home page of the docs site](https://deploy-preview-3367--paragon-openedx-v22.netlify.app/) lists pages like "Icons" in the Components list ("Layout" appears twice - once as a component and once as a page).
    - Fixed with c747946185450596ea48847c68210bfdbd83e765
* [x] The ["Components" part of the "Foundations/Layout" page](https://deploy-preview-3367--paragon-openedx-v22.netlify.app/foundations/layout/#components) uses a `jsx live` area that is just appearing as plain code, not rendered components.
    - Fixed with b77c6a90651282cd6c813aaa1613d58098826dc4
* [x] Components like Button are missing "Theme Variables (SCSS)" - compare new [Button page](https://deploy-preview-3367--paragon-openedx-v22.netlify.app/components/button/) vs. [existing v22 Button page](https://paragon-openedx-v22.netlify.app/components/button/)
    - Fixed with 64ccf2be81651bcec4be04a86a26bb11bdd4ac54
* [x] [The "Playground" page](https://deploy-preview-3367--paragon-openedx-v22.netlify.app/playground/) is not working.
    - Fixed by @brian-smith-tcril via https://github.com/openedx/paragon/pull/3381
* [x] Horizontal padding, font sizes, and other minor changes are seen on pages like the Button component docs. 
   - Turns out this is from comparing v22 to v23. Lots of minor styling changes are visible on the v23 docs site.
* [x] Some of the "Foundations" pages (the ones that use MDX and `pageQuery` data - specifically "Layout" and "CSS Utilities" , have errors and aren't loading properly. Fixed with aed4ef4e03d1e5e275ef70531bf4b820fd17e92f and 8abf42c0cb73d7f4227cb1b156ac3b9a2f256a22
* [x] Some of the "Foundations" pages (the ones that use MDX, not TSX, like [Layout](https://deploy-preview-3367--paragon-openedx-v23.netlify.app/foundations/layout/) , are not rendering with the correct layout. Fixed by @brian-smith-tcril via e0a35392068a4361337ba99a182bc4b48f6b24bb
* [x] Currently, gatsby is [running out of memory](https://www.gatsbyjs.com/docs/how-to/performance/resolving-out-of-memory-issues/) when trying to run `npm run start`. I thought I had solved this with `GATSBY_CPU_COUNT=4` (f67d1f75262c2632288428ed132aeaf6c35a93e9) but it no longer seems to be helping (even if lowered down to 1). I think the memory usage depends on how much has been cached, so clearing the cache always requires a lot more memory. For now the only workaround is to run it with `NODE_OPTIONS=--max-old-space-size=16384 npm run start` but allocating 16GB each time is not feasible.
    - Note that `npm run build-docs` seems to be working fine, and the "deploy preview" build of this PR works fine too.
    - Solved! By @brian-smith-tcril who found that adding `DEV_SSR`/`FAST_DEV` will avoid the issue.
* [x] This PR necessarily bumps the version of `sass` used via transitive dependencies. Will this version bump cause issues for Paragon consumers (if we change the sass code to the new version)? e.g. see 958f61708796741bb87e37b9ba20610155556316
    - Seems like it should be fine - see [sass discussion comment below](https://github.com/openedx/paragon/pull/3367#issuecomment-2554826278)
* [x] Upgrading `sass` means instead of e.g. `map-get(...)` we are encouraged to use `@use "sass:map";` and `map.get(...)` - done via 958f61708796741bb87e37b9ba20610155556316 but this may need to be reverted, see previous point.
* [x] Stylelint is not passing
* [x] Test cases need to be updated - done with e.g. c5ca0a5f83116b18a37df5ac617d49c703c0463d , 1942f03acc34335b20b69ae3e5690f04e400be32 and be8e10a85eafd3515c73e0635034a8d3d70fa43c
* [x] Many other sass deprecation warnings related to `import` etc. have been suppressed for now: 6946d67e0b64bf21d07ec33e1c9c3baea1064607 **we probably can't address this properly without a major bootstrap upgrade**
* [x] GraphQL syntax has changed in some cases, resulting in a bunch of deprecation warnings - solved with 2dd6935c5d5236704bee9a0ddebdf5c6c00af97c
* [x] The `{` and `<` characters must now be escaped in markdown - solved with 0f6004fd8d124eb0ae361b68ff55150e8d5af8d0
* [x] The `childMdx` API is no longer available - `gatsby-plugin-mdx` [no longer allows rendering multiple MDX documents in one page](https://github.com/gatsbyjs/gatsby/discussions/37301). I worked around this by adding `react-markdown` and rendering the props descriptions as plain markdown (they don't use MDX anyways) - 48aa67020e9154ef99439b769b1e0fe126d4ce62
* [x] Some Gatsby/rehype plugins can only be imported as ESM - so I had to convert the Gatsby config to ESM 60e534c082a1254af99b68c9be59c7e586ab99fb
* [x] Code blocks in Markdown no longer pass in attributes like [`live`](https://github.com/openedx/paragon/blob/92c43c36c14b86b7fd8cb85f751ba2de9df2157a/src/Nav/README.md?plain=1#L29) automatically, so I had to add `rehype-mdx-code-props` in 11b760541652dcafefb91e99bf6cfa7c180b7e18
* [x] Supporting `<PropsTable>` as an MDX component was tricky because it requires some context from the parent JSX component, which was being inlined as a local variable resulting in lint warnings: "Do not define components during render". It turns out that the props tables in question, on [this Data Views page](https://paragon-openedx.netlify.app/components/datatable/dataviews/) are already present in the usual place at the end of the document anyways. So I think it's more consistent to just remove support for this element from that page (remove the duplicate props tables) and avoid having to re-engineer `PropsTable` to better support context for this one very questionable use case. Done with c92a1a2d20249e5c8c59dc3743994bf6882d080e
* [x] update imports for 'renderHook' (now part of @testing-library/react) - 53500ee19c3f6d3760a5bdf556ec11c6282a82aa


openedx-webhooks:


This repository is currently maintained by @openedx/paragon-working-group.


<details><summary><h4>:radio_button: Get product approval</h4></summary>

</details><details>
<summary><h4>:radio_button: Provide context</h4></summary>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details><details>




<summary><h4>:radio_button: Get a green build</h4></summary>

</details><details>


---

<summary><h3>Where can I find more information? </h3></summary>

</details><details>
<summary><h3>When can I expect my changes to be merged?</h3></summary>

- The need for product review

</details>

brian-smith-tcril:
After reading through a ton of https://github.com/gatsbyjs/gatsby/issues/36899 and trying a few things I found https://github.com/gatsbyjs/gatsby/issues/36899#issuecomment-1318680725 and managed to get the dev server running.

```diff
bsmith@aximdev:~/code/paragon$ git diff
diff --git a/www/gatsby-config.mjs b/www/gatsby-config.mjs
index d0ab6c658e..aaf5daab39 100644
--- a/www/gatsby-config.mjs
+++ b/www/gatsby-config.mjs
@@ -130,4 +130,7 @@ export default {
   // Match the location of the site on github pages if no path prefix is specified
   pathPrefix: process.env.PATH_PREFIX || '',
   plugins,
+  flags: {
+    DEV_SSR: true
+  },
 };

```

Edit: I looked into this a bit more and found https://www.gatsbyjs.com/docs/reference/release-notes/v2.28/#feature-flags-in-gatsby-configjs. Setting the `FAST_DEV` flag also works (and seems like a good thing to do!)

brian-smith-tcril:
Re:

> Some of the "Foundations" pages (the ones that use MDX, not TSX, like [Layout](https://deploy-preview-3367--paragon-openedx-v23.netlify.app/foundations/layout/) , are not rendering with the correct layout. Some fix is required (see related work in https://github.com/openedx/paragon/pull/2767 which has some code related to this)

With [this change](https://github.com/openedx/paragon/pull/3367/files#diff-28bc87878c5d37ffbc134d02c45154d4855ad72b90d42f73cf25174ed07539b8L73-L75)

```diff
-        default: require.resolve(
-          './src/templates/default-mdx-page-template.tsx',
-        ),
```

It seems the `default-mdx-page-template` layout is just not being used at all. I know `defaultLayouts` were [removed in v4](https://www.gatsbyjs.com/plugins/gatsby-plugin-mdx/#v3-to-v4-breaking-changes), and the migration of `component-page-template` seems to have gone smoothly, but I'm a little lost as to how to properly migrate the other layout.

bradenmacdonald:
> It seems the default-mdx-page-template layout is just not being used at all. I know defaultLayouts were [removed in v4](https://www.gatsbyjs.com/plugins/gatsby-plugin-mdx/#v3-to-v4-breaking-changes), and the migration of component-page-template seems to have gone smoothly, but I'm a little lost as to how to properly migrate the other layout.

Yeah, me too. I tried a couple things based on the docs but didn't have much luck yet. There is some perhaps complication from the fact that Gatsby includes everything in `src/pages/` by default, but I've _also_ added those pages explicitly using a `gatsby-source-filesystem` in `gatsby-config.mjs` because I was seeing some `Could not locate File node` errors if I didn't include that in the config (though it wasn't needed with the previous version). 

Clearly, there is somewhere that we need to specify that `default-mdx-page-template` should be used, but I haven't found where yet, nor figured out why putting it in `www/utils/createPages.js` similarly to how https://github.com/openedx/paragon/pull/2767 did it isn't working.

brian-smith-tcril:
So I managed to get the layout page to render, I had to comment some stuff in the `.mdx` out and there are still a few errors, but it's at least a jumping off point

https://github.com/openedx/paragon/compare/braden/react-gatsby-bump...5589454dbefa3390097e83c5202a73c0b01f82e6

bradenmacdonald:
@brian-smith-tcril Nice! I pulled in most of that change for now. It seems like pages like `icons.mdx` are now working perfectly, so the layout part is correct - just need to solve the random errors on the other pages.

bradenmacdonald:
For some reason, either when I changed the target branch to `release-22.x` or when I pushed that fix from @brian-smith-tcril, the "v22" deploy preview is now broken and can't load the theme CSS: https://deploy-preview-3367--paragon-openedx-v22.netlify.app/

brian-smith-tcril:
@bradenmacdonald the build step of the latest deploy of this branch on netlify failed. I pulled the latest version of this branch and tried running `npm run build-docs` and it failed in the same place locally for me.

I commented out a couple things in `layout.mdx` and `css-utilities.mdx` and got it to build https://github.com/brian-smith-tcril/paragon/commit/e42915ba42d3d2a9b3018b0f3416143dff7c39ff

bradenmacdonald:
Ah, thanks @brian-smith-tcril :) I'll hold off on committing that for now until I see if we can actually get that `< CSSUtilitiesTable>` component working.

brian-smith-tcril:
re:
> * This PR necessarily bumps the version of `sass` used via transitive dependencies. Will this version bump cause issues for Paragon consumers (if we change the sass code to the new version)? e.g. see 958f61708796741bb87e37b9ba20610155556316
>     - ‼️ **help answering this would be appreciated** ‼️

We discussed this in the WG meeting yesterday. It seems the version bump is a minor version bump as opposed to a major one:
* Since it's a minor bump, that means existing `scss` in MFEs should still work as expected
* It seems using new features that don't exist in earlier minor versions of `sass` *in Paragon* also shouldn't be an issue. MFEs that upgrade to the new version of Paragon that has the new version of `sass` should automatically get the new version of `sass` in their package locks.

If there's some nuance here I'm missing please let me know!

bradenmacdonald:
@brian-smith-tcril FYI, I got the `<CSSUtilitiesTable>` working now, so the build/preview is back:

https://deploy-preview-3367--paragon-openedx-v22.netlify.app/

bradenmacdonald:
@brian-smith-tcril When I pulled in the changes from https://github.com/openedx/paragon/pull/3125 into this branch, I encountered the same problem with webpack confusing two files like `useToggle.tsx` and `useToggle.mdx`. It seems that your workaround of explicitly adding `gatsby-plugin-typescript` doesn't work anymore in Gatsby 5 (well actually, it does work, but it breaks the MDX plugin).

So what I've done for now is renamed e.g. `useToggle.tsx` to `useToggleHook.tsx` (as opposed to your suggestion of renaming `useToggle.mdx` to `useToggleDoc.mdx`). This should maintain both API compatibility and keep the URLs unchanged, unless I'm missing something. Does that make sense?

brian-smith-tcril:
> So what I've done for now is renamed e.g. `useToggle.tsx` to `useToggleHook.tsx` (as opposed to your suggestion of renaming `useToggle.mdx` to `useToggleDoc.mdx`). This should maintain both API compatibility and keep the URLs unchanged, unless I'm missing something. Does that make sense?

That sounds like a very sensible solution to me!

brian-smith-tcril:
@bradenmacdonald I see there are only a couple things left unchecked:
> - [ ] **Need to test this with an existing MFE to confirm backwards compatibility**
> - [ ] [The home page of the docs site](https://deploy-preview-3367--paragon-openedx-v22.netlify.app/) lists pages like "Icons" in the Components list ("Layout" appears twice - once as a component and once as a page).
> - [ ] Components like Button are missing "Theme Variables (SCSS)" - compare new [Button page](https://deploy-preview-3367--paragon-openedx-v22.netlify.app/components/button/) vs. [existing v22 Button page](https://paragon-openedx-v22.netlify.app/components/button/)
> - [ ] [The "Playground" page](https://deploy-preview-3367--paragon-openedx-v22.netlify.app/playground/) is not working.

I'd be more than happy to dive in and try to figure some of those out - is there one in particular you'd like another set of eyes on?

bradenmacdonald:
@brian-smith-tcril I don't know much about the Playground page, so that would probably be the one I could most use help with - thanks!

example/package.json
```diff
@@ -17,14 +17,10 @@
   "license": "ISC",
   "dependencies": {
     "@edx/brand": "npm:@openedx/brand-openedx@^1.2.2",
-    "@edx/frontend-platform": "^4.2.0",
+    "@edx/frontend-build": "^13.0.14",
```
adamstankiewicz, line 20: [clarification] Why is `@edx/frontend-build` moved to a regular dependency vs. keeping it as a dev dependency?

www/src/components/CodeBlock.tsx
```diff
@@ -136,6 +135,11 @@ function CodeBlock({
     setShowToast(true);
   };
 
+  if (className === '' && typeof children === 'string' && !children.includes('\n')) {
+    // This is an inline code block. Don't use syntax highlighting.
+    return <code>{children}</code>;
+  }
```
adamstankiewicz: [curious] When/where does this conditional for an inline code block get used?

www/src/components/LinkedHeading.tsx
```diff
@@ -1,10 +1,9 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 
 export interface ILinkedHeading {
   h: string,
-  children: React.ReactNode,
-  id: string,
+  children?: React.ReactNode,
```
adamstankiewicz: [clarification] Why is `children`, and `id` considered optional in this interface, when the prop types previously denoted them as required? I'm not sure why we'd have a use case for rendering `LinkedHeading` _without_ `children` (a heading without content?). Similarly, the `id` is needed for heading to be skip-linkable; if `id` is not required, wouldn't that largely differ from the intent of this `LinkedHeading` component?

www/src/components/Menu.tsx
```diff
@@ -35,15 +35,15 @@ const menuQuery = graphql`
...
     ) {
-      categories: group(field: frontmatter___categories) {
+      categories: group(field: {frontmatter: {categories: SELECT}}) {
         nodes {
           ...ComponentPage
         }
         fieldValue
       }
-      types: group(field: frontmatter___type) {
+      types: group(field: {frontmatter: {type: SELECT}}) {
```
adamstankiewicz, line 46: I'm not sure if these are the specific changes causing the change, but I noticed on the PR deploy preview contains additional links in the footer menu compared to the v22 docs site.

The PR deploy preview seems to include static content pages instead of the intended exports (i.e., components, hooks, etc.). For example, the footer menu now contains things like "Brand-Icons", "Component generation", "Getting started", etc.).

Is this intentional?

Also worth noting, clicking at least the "CSS Utilities" in the footer menu leads to a 404 not found page ([link](https://deploy-preview-3367--paragon-openedx-v22.netlify.app/css-utilities-content/)).

www/src/components/Search/Search.scss
```diff
@@ -53,7 +55,7 @@
     margin: 6.875rem auto auto;
     cursor: auto;
 
-    @media (max-width: map-get($grid-breakpoints, "md")) {
+    @media (max-width: map.get($grid-breakpoints, "md")) {
```
adamstankiewicz, line 58: [curious] I noticed some `@media` queries were modified to replace things like `max-width` with `width >=` (e.g., `@media only screen and (width <= 768px) {` in `Toast`).

1. Should we be consistent about using `max-width` and `min-width` vs. just `width` in our media queries?
2. Should any hardcoded pixel values in media queries (like the `Toast` example above) be using the `$grid-breakpoints` via `map.get`?

www/utils/createPages.js
```diff
@@ -39,39 +44,57 @@ async function createPages(graphql, actions, reporter) {
...
   // you'll call `createPage` for each result
-  // eslint-disable-next-line no-restricted-syntax
-  for (const { node } of components) {
-    const componentDir = node.slug.split('/')[0];
-    const variablesPath = path.resolve(__dirname, `../../src/${componentDir}/_variables.scss`);
-    const githubEditPath = `https://github.com/openedx/paragon/edit/master/src${node.fileAbsolutePath.split('src')[1]}`;
-    let scssVariablesData = {};
+  for (const { node } of pages) {
+    const githubEditPath = `https://github.com/openedx/paragon/edit/master/src${node.internal.contentFilePath.split('src')[1]}`;
+
```
adamstankiewicz: nit: linting around "trailing spaces not allowed"

adamstankiewicz (commented):
Thanks for tackling these changes, @bradenmacdonald! Left a few comments throughout. Here's a couple other issues I noticed when comparing this PR's deploy preview vs. the v22 docs site:

1. Modals seem to be throwing a JS error on the docs site when trying to open them. For example, see `AlertModal` ([link](https://deploy-preview-3367--paragon-openedx-v22.netlify.app/components/modal/alert-modal/#basic-usage)) and click the "Open alert modal" button. Observe a JS error `TypeError: object is not iterable (cannot read property Symbol(Symbol.iterator))`. This seems to occur on each modal variant that extends `ModalDialog`.
2. The `useIsVisible` hook ([link](https://deploy-preview-3367--paragon-openedx-v22.netlify.app/components/hooks/useisvisible/#sample-usage)) doesn't seem to be working properly anymore. The "Is element visible? yes/no" in the live code example is not reflecting whether the element is actually visible or not (consistently says "no" even when the element is visible).

example/package.json
bradenmacdonald: This is for the example MFE (not Paragon itself), and [I'm following the best practice I established for MFEs](https://github.com/openedx/open-edx-proposals/pull/615) that "dependencies" are what you need to build it (before deploying), and "devDependencies" are for development-only things like tests and linting.

If we had a part of the MFE that runs on the server, it would be different, but there's no server-side code here.

www/src/components/CodeBlock.tsx
bradenmacdonald: Anywhere you have an \`inline code node\` like this:

Screenshot

www/src/components/Menu.tsx
bradenmacdonald: Yes, those additional links are a bug and I had already noted them in the list of issues above that I need to work through. I did not intend to make any changes to the links (or really any visual/content changes to the docs site at all - any that you see are a bug).

www/src/components/Search/Search.scss
bradenmacdonald: > I noticed some @media queries were modified to replace things like max-width with width >= (e.g., @media only screen and (width <= 768px) { in Toast).

Those changes were done automatically when I ran `sass-migrator` on the codebase (though I ended up committing only a subset of the various changes it made). So I _assumed_ that they're equivalent syntax, that this way is the new best practice, and that linters [will] warn about the previous way. But I didn't check those assumptions or look into it in much detail. The `width <= 768px` seems simple and reasonable to me. That change isn't necessary as far as I know though, so let me know if you see any issues.

> Should any hardcoded pixel values in media queries (like the Toast example above) be using the $grid-breakpoints via map.get?

Probably, but let's do that in a follow-up PR as this one is already huge, and I'm just trying to do the bare minimum necessary for compatibility with the new React/Sass/Gatsby versions.

www/src/components/CodeBlock.tsx
bradenmacdonald: (I shouldn't have used the word "block")

www/src/components/LinkedHeading.tsx
bradenmacdonald: Because here where that LinkedHeading is used, TypeScript didn't know that `id` and `children` are present and was showing an error:

Screenshot

But I found a better way to deal with this in 6be0094443138440e25b687655dbf663966b4655

bradenmacdonald:
> * Modals seem to be throwing a JS error on the docs site when trying to open them. For example, see `AlertModal` ([link](https://deploy-preview-3367--paragon-openedx-v22.netlify.app/components/modal/alert-modal/#basic-usage)) and click the "Open alert modal" button. Observe a JS error `TypeError: object is not iterable (cannot read property Symbol(Symbol.iterator))`. This seems to occur on each modal variant that extends `ModalDialog`.
> * The `useIsVisible` hook ([link](https://deploy-preview-3367--paragon-openedx-v22.netlify.app/components/hooks/useisvisible/#sample-usage)) doesn't seem to be working properly anymore. The "Is element visible? yes/no" in the live code example is not reflecting whether the element is actually visible or not (consistently says "no" even when the element is visible).

Ah, these were actually two more symptoms of the same general problem of webpack getting confused between `.mdx` and `.tsx` files with the same name. I should have been more thorough in my earlier fix. In any case, it's fixed now with 89791077c172b9e3b4c9171e82ea4a1b9cdb846e

www/src/components/CodeBlock.tsx
adamstankiewicz: Ah, I see. Makes sense, though somewhat odd that on `master`, the `CodeBlock` was not previously used for these inline `code` nodes, despite having `code: CodeBlock` configured within the previous implementation's `shortCodes` ([source](https://github.com/openedx/paragon/blob/81a0abfa44c6a802b650ff092180cb596f7ad4e1/www/src/templates/component-page-template.tsx#L86)). For example, adding a 

With `CodeBlock` actually being used now for these inline `code` nodes, makes sense we'd need to handle it explicitly. Wonder why it wasn't used before. Either way, thanks for clarifying!

www/src/components/Search/Search.scss
adamstankiewicz: > The width <= 768px seems simple and reasonable to me. That change isn't necessary as far as I know though, so let me know if you see any issues.

No issues immediately observed; I agree the change should be functionally equivalent. The newer syntax is easier definitely easier to grok.

> Probably, but let's do that in a follow-up PR as this one is already huge

Totally, just wanted to call it out for now :)

www/src/components/CodeBlock.tsx
bradenmacdonald: Yeah, I am also wondering why. Those inline nodes are still using `code` as their tag before this PR, so it's unclear to me why they weren't using `CodeBlock`. In any case, it was an easy workaround.

brian-smith-tcril:
> @brian-smith-tcril I don't know much about the Playground page, so that would probably be the one I could most use help with - thanks!

https://github.com/openedx/paragon/pull/3381

brian-smith-tcril:
Tested both `frontend-app-learner-dashboard` and `frontend-app-learning` locally and everything seems to be working as expected. I made draft PRs on each repo to spin up sandboxes for some extra verification - I'll update this comment when the sandboxes are live:
* https://github.com/openedx/frontend-app-learner-dashboard/pull/549
  * edit: looking good! https://app.pr-549-550b00.sandboxes.opencraft.hosting/learner-dashboard/ 
* https://github.com/openedx/frontend-app-learning/pull/1576
  * edit: looking good! https://app.pr-1576-fadd12.sandboxes.opencraft.hosting/learning/course/course-v1:demo+Course+1/home

bradenmacdonald:
I'm seeing the same in `authoring` - everything looks good. The main thing I wanted to check was that the SCSS builds and works the same, which seems to be the case. (There's not really any changes to the Paragon components JS code in this PR.)

bradenmacdonald:
@brian-smith-tcril @adamstankiewicz Alright, from my perspective this PR is ready to go! Could I please get a review/approval from you folks? And then (assuming no further issues arise), let me know how/when we should merge this. I'm thinking it will be good to preserve some of the commit history for future reference and to make it easier to deal with porting to the v23 branch, but I can definitely squash some things down.

www/src/components/Menu.tsx
bradenmacdonald: Fixed!

src/Bubble/index.tsx
```diff
@@ -37,7 +37,7 @@ const Bubble = React.forwardRef<HTMLDivElement, BubbleProps>(({
 ));
 
 Bubble.propTypes = {
-  /** Specifies contents of the component. */
+  /** Specifies contents of the component. */ // @ts-ignore
```
brian-smith-tcril: Having `@ts-ignore` on the same line as the descriptive comment feels a bit odd to me

src/Chip/index.tsx
```diff
@@ -93,7 +93,7 @@ const Chip = React.forwardRef(({
 });
 
 Chip.propTypes = {
-  /** Specifies the content of the `Chip`. */
+  /** Specifies the content of the `Chip`. */ // @ts-ignore
```
brian-smith-tcril: Same comment as the one I left on `Bubble/index.tsx`. Can we move these to not be on the same line as the description?

src/DataTable/dataviews.mdx
```diff
@@ -110,16 +110,11 @@ It displays the data provided by the `DataTableContext` as an html table.
...
-
 ## Table Subcomponents
 
 Subcomponents of `DataTable.Table` can be used independently of the main component. They are designed for use with a `react-table` instance.
 
-<PropsTable displayName="TableRow" />
-<PropsTable displayName="TableCell" />
-<PropsTable displayName="TableHeaderCell" />
-<PropsTable displayName="TableHeaderRow" />
+See the available props of `<TableRow>`, `<TableCell>`, `<TableHeaderCell>`, and `<TableHeaderRow>` below.
```
brian-smith-tcril, line 117: I looked at the deploy preview for this page and everything seems to make sense as it stands now, so I have no problem keeping this change, but it's not clear to me why it was made. Could you provide some context?

src/Hyperlink/index.tsx
```diff
@@ -113,7 +113,7 @@ Hyperlink.defaultProps = {
 Hyperlink.propTypes = {
   /** specifies the URL */
   destination: PropTypes.string.isRequired,
-  /** Content of the hyperlink */
+  /** Content of the hyperlink */ // @ts-ignore
```
brian-smith-tcril: Another one of these

src/Modal/index.jsx
```diff
@@ -50,6 +50,8 @@ class Modal extends React.Component {
 
   componentWillUnmount() {
     if (this.parentElement) {
+      // TODO: update this to use the new createRoot() compatible APIs.
```
brian-smith-tcril, line 53: If this `TODO` is one that's going to stick around after this PR lands it'd probably be good to make a GH issue for it

brian-smith-tcril (approved):
Left a few comments but nothing blocking!

Super exciting to see this all come together! This is huge (in a good way)!

brian-smith-tcril:
@bradenmacdonald re:

> I'm thinking it will be good to preserve some of the commit history for future reference and to make it easier to deal with porting to the v23 branch, but I can definitely squash some things down.

I agree keeping the history here could be nice, but I'm also not opposed to squashing the whole thing. There are definitely at least a couple commits that should be squashed if we keep the history (`chore: fixup previous commit (fix whitespace lint)` comes to mind as an example)

edit: looks like this would require some fancy rebasing to keep history
image

edit2: it looks like the "you can't rebase" message is because of https://github.com/openedx/paragon/pull/3367/commits/505c4ba55d99dee624abad0f945453569a413929 (merge commit). With that in mind I'm leaning towards just squashing this in and referencing the PR when porting to 23. I'm not opposed to the 23 history having:
* this PR as 1 commit
* revert of that 1 commit
* reimplementation of this PR

But I'm very open to other opinions on the matter.

src/DataTable/dataviews.mdx
bradenmacdonald: For sure. I did already try to explain that in the PR issue list above:

> Supporting `<PropsTable>` as an MDX component was tricky because it requires some context from the parent JSX component, which was being inlined as a local variable resulting in lint warnings: "Do not define components during render". It turns out that the props tables in question, on [this Data Views page](https://paragon-openedx.netlify.app/components/datatable/dataviews/) are already present in the usual place at the end of the document anyways. So I think it's more consistent to just remove support for this element from that page (remove the duplicate props tables) and avoid having to re-engineer PropsTable to better support context for this one very questionable use case.

It will probably make more sense if you look at the changes in commit c92a1a2d20249e5c8c59dc3743994bf6882d080e too.

I can't simply hoist that specific `PropsTable` variation outside of the function it was originally declared in, becuase then it doesn't have the `component` data it needs. And what's in the MDX file, e.g. `<PropsTable displayName="CardView" />` is obviously insufficient to render the table without that additional context.

src/Bubble/index.tsx
bradenmacdonald: Sure, 4b6082ea22e3d60229a7c307526b5a97c30b1a1c seems to work fine, and moves it to the following line.

src/Modal/index.jsx
bradenmacdonald: Created issue: https://github.com/openedx/paragon/issues/3382

bradenmacdonald:
> edit: looks like this would require some fancy rebasing to keep history

@brian-smith-tcril I've never been afraid of fancy rebasing :p I've now squashed this down to 6 commits and I think each of them makes sense, but I can squash them further if you want.

The diff is exactly identical except I made one tiny correction while I was at it:
Screenshot 2025-01-16 at 11 27 00 AM

bradenmacdonald:
I just noticed that some component pages like "Marketing Modal" were missing the SCSS variables that they're supposed to show, so I fixed it with b4a4bb3b9f799a4aacf79e1b836effe86495b768. I'm going to squash that commit into the other big commits though.

www/utils/createPages.js
```diff
@@ -39,39 +44,56 @@ async function createPages(graphql, actions, reporter) {
...
+    if (node.fields.source === 'components') {
+      // Check for a _variables.scss file for this component, e.g. src/Button/_variables.scss.
+      // If it exists, load the data:
+      let scssVariablesData = {};
+      const componentDir = path.dirname(node.internal.contentFilePath);
+      const variablesPath = componentDir + '/_variables.scss';
```
adamstankiewicz: nit: there is a GitHub annotation for this line around "Unexpected string concatenation". Should it be addressed (i.e., using a template literal instead)? Also curious why CI isn't failing if there's a check failure annotation?

adamstankiewicz (approved):

www/utils/createPages.js
bradenmacdonald: Weird. It clearly failed in the CI but the "Test > Lint" check overall is still marked as passing:

Screenshot

Could be related to https://github.com/openedx/paragon/issues/3329 ?

Anyhow, I addressed the lint issue and squashed it into the relevant existing commit.

PR #3390, opened 2025-01-23
sync `release-22.x` -> `release-23.x`

brian-smith-tcril:
## Description

This is just a new branch created off of b2da298b69f96cb6653ecf3971413f51eaa2b994

The idea here is to get `release-23.x` updated to include everything from `release-22.x` before https://github.com/openedx/paragon/pull/3367 landed, so the conflict resolution needed to land those changes in `release-23.x` can be isolated.


src/hooks/tests/useToggle.test.tsx
```diff
@@ -19,7 +20,7 @@ const resetHandlerMocks = () => {
 };
 
 // eslint-disable-next-line react/prop-types
```
adamstankiewicz, line 22: nit: is this eslint disable comment still relevant?

adamstankiewicz (approved):

src/hooks/tests/useToggle.test.tsx
brian-smith-tcril: Good question! It looks like it's still there on latest `release-22.x` as of writing https://github.com/openedx/paragon/blob/c5c48aa2aed9c4e583d83bbe58783a93db40977a/src/hooks/tests/useToggle.test.tsx#L22

I made an issue to go through and clean up any of these we don't need https://github.com/openedx/paragon/issues/3391

PR #3451, opened 2025-02-26
fix: fixed Logistration page display issue

sundasnoreen12:
### Description
[INF-1786](https://2u-internal.atlassian.net/browse/INF-1786)

The email and password input fields on the login form are overlapping.
Expected Behaviour: When a user enters text into these fields, the entered characters should ideally replace any placeholder text. 
Current Behaviour: Currently, the entered text overlaps  the original placeholder input.

image

**Solution**

Fixed Logistration page display issue due to auto fill.

#### Merge Checklist

* [ ] If your update includes visual changes, have they been reviewed by a designer? Send them a link to the Sandbox, if applicable.

#### Post-merge Checklist

* [ ] Deploy the changes to prod after verifying on stage or ask **@openedx/2u-infinity** to do it. 

### DIFF
--- a/src/Form/_index.scss
+++ b/src/Form/_index.scss
@@ -265,8 +265,8 @@ $select-icon-padding: .5625rem !default;
   }
 
   .form-control:focus ~ &,
-  // select.form-control ~ &,
-  .form-control.has-value ~ & {
+  .form-control.has-value ~ &,
+  .form-control:is(:-webkil-autofill, :autofill) ~ & {
     .pgn__form-control-floating-label-text {
       background-color: $input-bg;
     }
--- a/src/Form/_index.scss
+++ b/src/Form/_index.scss
@@ -266,7 +266,7 @@ $select-icon-padding: .5625rem !default;
 
   .form-control:focus ~ &,
   .form-control.has-value ~ &,
-  .form-control:is(:-webkil-autofill, :autofill) ~ & {
+  .form-control:is(:-webkit-autofill, :autofill) ~ & {
     .pgn__form-control-floating-label-text {
       background-color: $input-bg;
     }
--- a/src/Form/FormControl.jsx
+++ b/src/Form/FormControl.jsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useCallback, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import RBFormControl from 'react-bootstrap/FormControl';
@@ -30,28 +30,11 @@ const FormControl = React.forwardRef(({
   const inputRef = React.useRef();
   const resolvedRef = ref || inputRef;
   const size = props.size || formGroupContext.size;
-  const [isAutoFill, setIsAutoFill] = useState(false);
 
   const [hasValue, checkInputEventValue] = useHasValue({
     defaultValue: props.defaultValue,
     value: props.value,
   });
-
-  useEffect(() => {
-    if (!inputRef.current) { return; }
-
-    const checkAutoFill = () => {
-      setIsAutoFill(inputRef.current.matches(':autofill, :-webkit-autofill'));
-    };
-
-    checkAutoFill(); // Check immediately on mount
-
-    // Fallback check after a short delay
-    const timeoutId = setTimeout(checkAutoFill, 100);
-
-    // eslint-disable-next-line consistent-return
-    return () => clearTimeout(timeoutId);
-  }, []);
 
   const handleResize = useCallback(() => {
     if (as === 'textarea' && autoResize) {
@@ -96,7 +79,7 @@ const FormControl = React.forwardRef(({
         isInvalid={isInvalid}
         isValid={isValid}
         className={classNames(controlClassName, {
-          'has-value': hasValue ? true : isAutoFill,
+          'has-value': hasValue,
         })}
         onChange={handleOnChange}
         mask={inputMask}
--- a/src/Form/FormControl.jsx
+++ b/src/Form/FormControl.jsx
@@ -19,6 +19,7 @@ const FormControl = React.forwardRef(({
   autoResize,
   onChange,
   inputMask,
+  isAutoFill,
   ...props
 }, ref) => {
   const {
@@ -79,7 +80,7 @@ const FormControl = React.forwardRef(({
         isInvalid={isInvalid}
         isValid={isValid}
         className={classNames(controlClassName, {
-          'has-value': hasValue,
+          'has-value': hasValue || isAutoFill,
         })}
         onChange={handleOnChange}
         mask={inputMask}
@@ -127,6 +128,8 @@ FormControl.propTypes = {
   autoResize: PropTypes.bool,
   /** Specifies what format to use for the input mask. */
   inputMask: PropTypes.string,
+  /** Indicates whether the field contains an autofill value. */
+  isAutoFill: PropTypes.bool,
 };
 
 FormControl.defaultProps = {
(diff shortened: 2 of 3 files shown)


awais-ansari (approved):

ayesha-waris (approved):

adamstankiewicz:
I'm having trouble replicating this issue. Can you outline some testing instructions to help reviewers replicate the issue? Presumably this is related to browsers' native auto-fill support of form fields, given the addition of an `isAutoFill` prop?

It would also be helpful to understand how the proposed `isAutoFill` prop is intended to be used, in context. For instance, when would the MFE pass `isAutoFill` vs. not including it, if this is related to browsers' native auto-fill support? For example, how would the MFE know when browsers' auto-fill is being used or not?

Using Chrome's auto-fill (see below screen recording), for example, when I hover over the password I want to auto-fill, I see that the auto-filled password overlays the input's placeholder label. However, when I apply the auto-fill, the `Form.Control` properly responds to having a new value, and the placeholder label properly moves to the top of the input field.

More generally, I'm also curious if you've dug into why `hasValue` might not be evaluating as truthy on its own for this use case, without the addition of an `isAutoFill` prop.

https://github.com/user-attachments/assets/5174339b-adbd-4c16-a8bc-fc2eabd17a5e

sundasnoreen12:
> I'm having trouble replicating this issue. Can you outline some testing instructions to help reviewers replicate the issue? Presumably this is related to browsers' native auto-fill support of form fields, given the addition of an `isAutoFill` prop?
> 
> It would also be helpful to understand how the proposed `isAutoFill` prop is intended to be used, in context. For instance, when would the MFE pass `isAutoFill` vs. not including it, if this is related to browsers' native auto-fill support? For example, how would the MFE know when browsers' auto-fill is being used or not?
> 
> Using Chrome's auto-fill (see below screen recording), for example, when I hover over the password I want to auto-fill, I see that the auto-filled password overlays the input's placeholder label. However, when I apply the auto-fill, the `Form.Control` properly responds to having a new value, and the placeholder label properly moves to the top of the input field.
> 
> More generally, I'm also curious if you've dug into why `hasValue` might not be evaluating as truthy on its own for this use case, without the addition of an `isAutoFill` prop.
> 
>  paragon_form_control_autofill_bug_replication_attempt.mov


Hi @adamstankiewicz ,

The issue occurs when a user has previously saved their email and password for the login screen. Upon refreshing the login page, the browser automatically fills in the saved credentials. However, since the page is not focused on load, the text and placeholders overlap. The issue resolves once the user clicks anywhere on the screen.

On initial page load, the <input> tag does not have a value attribute until the user interacts with the page. The hasValue check does not evaluate as truthy in this scenario because its value depends on props.value. Since autofill populates the fields without triggering a focus event, props.value remains empty initially.

I am attaching the frontend PR (https://github.com/openedx/frontend-app-authn/pull/1431) to provide more context on how we are passing the isAutoFill boolean variable.

Let me know if you have any further questions.

adamstankiewicz:
Hi @sundasnoreen12. I brought this proposed approach to this week's Paragon Working Group meeting for additional feedback. See meeting notes here: https://openedx.atlassian.net/wiki/spaces/BPL/pages/4864081939/2025-03-12+Meeting+notes

Overall, we agreed this autofill style issue is something that should be addressed. However, given the problem is generally applicable across any `Form.Control`, not just the usages within the Logistration MFE, we believe the fix should live solely within Paragon itself. That is, consuming MFEs should not have to worry about the autofill use case; the `Form.Control` component from Paragon should "just work".

As is, this proposed approach requires the consuming MFE to implement its own autofill detection. Each MFE that wants to handle autofill would need to re-implement its own autofill detection logic, risking different implementations across the platform. By keeping the autofill detection logic solely within Paragon, consumers would not need to pass an `isAutoFill` prop.

We also discussed the _how_ of autofill detection. We're curious if you've explored other approaches beyond the `MutationObserver` proposed via https://github.com/openedx/frontend-app-authn/pull/1431. For instance, it seems a more commonly used/accepted practice is to rely on a void CSS transition and setting up a JavaScript event listener for `animationstart`, as described in [this StackOverflow post](https://stackoverflow.com/a/41530164/9578634) and seen in the `detect-autofill-react` NPM package's implementation ([source](https://github.com/sundarshahi/detect-autofill-react/tree/main/lib/hooks)).

Lunyachek:
Hi everyone,

I’d like to propose a simple CSS-only fix for the autofill styling issue in Form.Control. This approach ensures that floating labels behave correctly when autofill is applied, without requiring additional JavaScript logic.

```
input.form-control:-webkit-autofill ~ .pgn__form-control-floating-label {
    @include form-control-floating-label-float(
        $input-padding-x,
        $input-padding-y,
        $input-line-height,
        $input-font-size,
        $input-border-width
    );

    .pgn__form-control-floating-label-text {
        background-color: $input-bg;
        font-size: $input-font-size;
    }
}
```
By leveraging :-webkit-autofill, this solution allows Paragon to handle autofill styling natively within CSS, without requiring consuming MFEs to implement autofill detection themselves.

Let me know your thoughts!

adamstankiewicz:
@Lunyachek Thanks for the alternative proposal for a CSS-only solution! I think your proposal makes sense, and it does simplify the overall approach since it doesn't need to be coupled with a JavaScript implementation of `Form.Control` (or MFE-implemented autofill detection).

Based on your suggestion, looks like existing SCSS selectors using `@include form-control-floating-label-float` ([source](https://github.com/openedx/paragon/blob/f3ab69aac949822c5fe30ca399d9e8f1846a6576/src/Form/_index.scss#L297-L309)) could be extended to include the `:webkit-autofill` and `:autofill` pseudo-class, e.g.:

```scss
.form-control:focus ~ &,
.form-control.has-value ~ &,
.form-control:is(:-webkil-autofill, :autofill) ~ & {
  .pgn__form-control-floating-label-text {
      background-color: $input-bg;
    }

    @include form-control-floating-label-float(
      $input-padding-x,
      $input-padding-y,
      $input-line-height,
      $input-font-size,
      $input-border-width
    );
}
```

I'd be in favor of this CSS-only solution within Paragon itself! 👍 Curious to hear your thoughts @sundasnoreen12

src/Form/_index.scss
```diff
@@ -265,8 +265,8 @@ $select-icon-padding: .5625rem !default;
   }
 
   .form-control:focus ~ &,
-  // select.form-control ~ &,
-  .form-control.has-value ~ & {
+  .form-control.has-value ~ &,
+  .form-control:is(:-webkil-autofill, :autofill) ~ & {
```
adamstankiewicz: typo:

```suggestion
  .form-control:is(:-webkit-autofill, :autofill) ~ & {
```

adamstankiewicz (commented):
Thanks for iterating on the approach, @sundasnoreen12! I left one typo-related comment that should be addressed before approval.

peterkulko:
Hello everyone! Thanks for the fix. During testing, I noticed a minor issue with the font size of the autocomplete text. After [some research](https://stackoverflow.com/questions/57195620/css-autocomplete-font-size), I wasn’t able to find a stable or reliable solution for changing it. Given that, I suggest we ignore this issue for now. Still, if anyone has ideas or suggestions, I’d be glad to hear them.

https://github.com/user-attachments/assets/8445f2c3-adaa-4bf4-bffb-198568d63f02

adamstankiewicz:
@PKulkoRaccoonGang Thanks for the additional QA and the screen capture of the font size issue. I agree, we should defer on the font size discrepancy for now. Do you think the font size issue warrants backlogging a GitHub issue with your findings so far?

peterkulko:
@adamstankiewicz good idea! I created a separate issue for the font size https://github.com/openedx/paragon/issues/3499

adamstankiewicz (approved):

Issue #3391, opened 2025-01-23
clean up outdated `eslint-disable`


brian-smith-tcril:
We should do a pass to find any `eslint-disable` lines we don't need any more and remove them.

Inspired by https://github.com/openedx/paragon/pull/3390/files/b2da298b69f96cb6653ecf3971413f51eaa2b994#r1927114654

Example line in question
https://github.com/openedx/paragon/blob/c5c48aa2aed9c4e583d83bbe58783a93db40977a/src/hooks/tests/useToggle.test.tsx#L22

PR #3488, opened 2025-03-19
chore: clean up outdated eslint-disable

PKulkoRaccoonGang:
## Description

This PR focuses on code cleanup by removing or modifying ESLint disable comments across various files.

**Issue:** https://github.com/openedx/paragon/issues/3391

## Merge Checklist

## Post-merge Checklist


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

example/src/MyComponent.jsx
```diff
@@ -1,6 +1,8 @@
 import React, { useState } from 'react';
-import { Button, Form, Icon, Bubble, Stack, Container } from '@openedx/paragon'; // eslint-disable-line
-import { FavoriteBorder } from '@openedx/paragon/icons'; // eslint-disable-line
+// eslint-disable-next-line import/no-extraneous-dependencies, object-curly-newline
+import { Button, Form, Icon, Bubble, Stack, Container } from '@openedx/paragon';
```
adamstankiewicz: We can most likely remove the `object-curly-newline` in favor of the following:

```suggestion
import {
  Button, Form, Icon, Bubble, Stack, Container
} from '@openedx/paragon';
```

But it is good to be explicit about the `import/no-extraneous-dependencies` rule vs. the `eslint-disable-line`.

src/CheckBox/index.jsx
```diff
@@ -15,7 +15,6 @@ class Check extends React.Component {
     };
   }
 
-  /* eslint-disable react/no-did-update-set-state */
   componentDidUpdate(prevProps) {
```
adamstankiewicz, line 18: [curious] With this `componentDidUpdate` still using `setState` below, I'm wondering how we can remove this disabled rule without removing the `setState` usage?

src/ChipCarousel/ChipCarousel.test.jsx
```diff
@@ -1,16 +1,45 @@
...
-  <div onClick={jest.fn()} data-testid="chip" />,
+  <div
+    onClick={jest.fn()}
+    onKeyDown={(e) => e.key === 'Enter' && e.target.click()}
+    role="button"
+    tabIndex={0}
+    data-testid="chip"
+  >
+    Item 1
+  </div>,
```
adamstankiewicz: [curious] Should these be using the `Chip` component directly, vs. creating a mock `div` implementation of the `Chip`?

src/ChipCarousel/ChipCarousel.test.jsx
```diff
@@ -1,16 +1,45 @@
...
+  <div
+    onClick={jest.fn()}
+    onKeyDown={(e) => e.key === 'Enter' && e.target.click()}
+    role="button"
+    tabIndex={0}
+    data-testid="chip"
+  >
+    Item 3
+  </div>,
+  // eslint-disable-next-line jsx-a11y/control-has-associated-label
```
adamstankiewicz: [curious] This is a new ESLint rule, that wasn't previously disabled above. Does this suggest a possible a11y bug within `ChipCarousel` and/or `Chip`? That said, if using `Chip`, this issue may go away due to `children` being required.

src/DataTable/CollapsibleButtonGroup.jsx
```diff
@@ -51,7 +51,6 @@ function CollapsibleButtonGroup({
   const cloneAction = (action, index) => React.cloneElement(
     action.component,
     {
-      // eslint-disable-next-line react/no-array-index-key
       key: `${action}${index}`,
```
adamstankiewicz, line 54: [curious] What changed such that we could remove this rule without changing its implementation (e.g., `index` is still used)?

src/Form/tests/useCheckboxSetValues.test.jsx
```diff
@@ -14,10 +13,18 @@ function Example() {
   return (
     <>
       <span data-testid={VALUES}>{values.join(' ')}</span>
-      <button data-testid="add" onClick={() => add('provolone')}>Add</button>
-      <button data-testid="remove" onClick={() => remove('provolone')}>Remove</button>
-      <button data-testid="set" onClick={() => set(['cheddar', 'swiss', 'provolone'])}>Set</button>
-      <button data-testid="clear" onClick={() => clear()}>Clear</button>
+      <button data-testid="add" onClick={() => add('provolone')} type="button">
```
adamstankiewicz: [curious] Why not use `Button` (Paragon component) vs. `button` to avoid needing to pass `type="button"` here explicitly?

src/Input/index.jsx
```diff
@@ -32,7 +32,7 @@ class Input extends React.Component {
     if (process.env.NODE_ENV !== 'development') { return forwardedRef; }
 
     return (element) => {
-      if (forwardedRef) { forwardedRef.current = element; } // eslint-disable-line no-param-reassign
```
adamstankiewicz, line 35: [curious] What changed such that we could remove this ESLint rule?

src/Modal/ModalLayer.tsx
```diff
@@ -5,15 +5,14 @@ import { FocusOn } from 'react-focus-on';
...
-// istanbul ignore next
 function ModalBackdrop({ onClick }: { onClick?: () => void }) {
   return (
-    // eslint-disable-next-line jsx-a11y/no-static-element-interactions
     <div
       className="pgn__modal-backdrop"
       onClick={onClick}
       onKeyDown={onClick}
       data-testid="modal-backdrop"
+      role="presentation"
```
adamstankiewicz, line 15: [sanity check] From an a11y perspective, we should verify that `role="presentation"` makes sense here.

src/hooks/tests/useToggle.test.tsx
```diff
@@ -19,16 +18,15 @@ const resetHandlerMocks = () => {
...
 function FakeComponent({ defaultIsOn, handlers }: { defaultIsOn: boolean, handlers: ToggleHandlers }) {
   const [isOn, setOn, setOff, toggle] = useToggle(defaultIsOn, handlers);
 
   return (
     <div>
       <div data-testid="toggle-value">{isOn ? TOGGLE_IS_ON : TOGGLE_IS_OFF}</div>
-      <button id="set-on" onClick={setOn}>set on</button>
-      <button id="set-off" onClick={setOff}>set off</button>
-      <button id="toggle" onClick={toggle}>toggle</button>
+      <button id="set-on" onClick={setOn} type="button">set on</button>
```
adamstankiewicz: [curious] Would it make sense to use `Button` directly here vs. passing `type="button"?

www/src/types/index.d.ts
```diff
@@ -1,8 +1,7 @@
 // typing of properties and methods used in global
 
-// eslint-disable-next-line import/prefer-default-export
 export declare global {
-  // eslint-disable-next-line
+  // eslint-disable-next-line no-var, vars-on-top
   var analytics: Array;
```
adamstankiewicz, line 5: [curious] Would defining `analytics` as a TS `type` mitigate the ESLint rules?

adamstankiewicz (commented):
Thanks for tacking this, @PKulkoRaccoonGang! Left some curiosity questions, but overall looking great! Biggest comment is around a `role="presentation"` on the modal dialog backdrop component, where we're hoping to get additional a11y clarification.

example/src/MyComponent.jsx
peterkulko: Good catch! Removed unnecessary `object-curly-newline` rule
https://github.com/openedx/paragon/pull/3488/files#diff-5dcc0fb0dadec970ffd28d9d4cd5cdc6dd928f2edbf682be2df77bf1256320eeR2

src/DataTable/CollapsibleButtonGroup.jsx
peterkulko: I've been following [the history of this file](https://github.com/openedx/paragon/blob/54622e70b28a7db53516b0e617b7eaa7a0d230a8/src/DataTable/CollapsibleButtonGroup.jsx#L57) and it seems that the main difference is that the new version uses a separate `cloneAction` function, which is called for each element of the array. With such a call to `.map(cloneAction)`, ESLint cannot "see" the use of the `index` as a key directly, since the `cloneAction` function is defined separately.

But, this does not mean that the problem has disappeared - it still exists, I just think that ESLint has limitations in its code analysis. The linter does not track the use of the `index` due to an additional layer of abstraction (a separate function `cloneAction`).

src/Input/index.jsx
peterkulko: In this code `forwardedRef.current = element` we are changing the current property, not the `forwardedRef` parameter itself (as we did [before](https://github.com/openedx/paragon/blob/420e52ddafda3d25e1b897c6ced7c7ed90138246/src/Input/index.jsx#L132)). Thanks to `props: false` (which was [added](https://github.com/openedx/paragon/blob/release-23.x/.eslintrc.js#L59) to the ESLint linter configuration), such a change is now allowed, so ESLint does not throw an error and the comment is no longer needed.

src/hooks/tests/useToggle.test.tsx
peterkulko: Yes, replaced ✅

src/ChipCarousel/ChipCarousel.test.jsx
peterkulko: Replaced with Paragon Chip component
https://github.com/openedx/paragon/pull/3488/files#diff-2e5076eb3bcfd9fd360e2f2232b7cac8029b20e222ff833d2048d84685fb2c71R9

src/ChipCarousel/ChipCarousel.test.jsx
peterkulko: Since the presence of children is now required for [the Chip component](https://paragon-openedx.netlify.app/components/chip/#props-api), I replaced the empty div element with the Chip component with children

src/Form/tests/useCheckboxSetValues.test.jsx
peterkulko: Replaced with Paragon Button component
https://github.com/openedx/paragon/pull/3488/files#diff-2c605fb44f28a61fc741a2de395d5e4bd24ad269b6484f8ade8a5e5c91f051e6R18

www/src/types/index.d.ts
peterkulko: Please correct me if I misunderstood. In our case `analytics` should not be just a type, as it is a global variable used at runtime.
In this file (`index.d.ts`) we use `declare global` to extend the global namespace. This is the declarative file (`.d.ts`) that tells TS that the global variable `analytics` exists. We need the global variable declaration itself, so the current disabling of `no-var,` `vars-on-top` rules is expected.

src/CheckBox/index.jsx
peterkulko: Since we are using the latest version of [eslint-plugin-react](https://github.com/jsx-eslint/eslint-plugin-react) in [@edx/eslint-config](https://github.com/openedx/eslint-config/blob/master/package.json#L49) according to this [change log](https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/CHANGELOG.md?plain=1#L850) the rule `react/no-did-update-set-state` no longer works in the latest versions of React

src/TransitionReplace/DemoTransitionReplace.jsx
```diff
@@ -1,3 +1,4 @@
+// TODO: remove this file in the particular PR
```
peterkulko: https://github.com/openedx/paragon/pull/3534

peterkulko:
> Biggest comment is around a role="presentation" on the modal dialog backdrop component, where we're hoping to get additional a11y clarification.

The following decision was made at the [Paragon working group](https://openedx.atlassian.net/wiki/spaces/BPL/pages/5137235969/2025-07-30+Meeting+notes) meeting:

> presentation makes more sense. we have accessible ways to close the modal. adding the backdrop to the a11y tree would make things more confusing

This PR will be updated

src/Modal/ModalLayer.tsx
peterkulko: > presentation makes more sense. we have accessible ways to close the modal. adding the backdrop to the a11y tree would make things more confusing

[notes](https://openedx.atlassian.net/wiki/spaces/BPL/pages/5137235969/2025-07-30+Meeting+notes) from PWG

peterkulko:
@brian-smith-tcril @adamstankiewicz Hi! This PR has been updated, all comments have been answered. Please recheck when you have time. Thanks!

src/SelectableBox/utils.js
```diff
@@ -21,4 +20,5 @@ export const getInputType = (component, type) => {
         return Form.RadioSet;
     }
   }
+  return null;
```
brian-smith-tcril, line 23: [curious] this is being changed to return `null` and `useLayoutEffect` in `Sticky` is being changed to return `undefined`. Could you provide context as to why each was picked?

src/Sticky/index.jsx
```diff
@@ -45,6 +44,7 @@ const Sticky = React.forwardRef(({
         observer.unobserve(stickyElement);
       };
     }
+    return undefined;
```
brian-smith-tcril, line 47: [curious] this is being changed to return `undefined` and `getInputType` in `SelectableBox` is being changed to return `null`. Could you provide context as to why each was picked?

www/src/config.js
```diff
@@ -1,18 +1,10 @@
-/* eslint-disable */
-const hasFeatureFlagEnabled = require('./utils/hasFeatureFlagEnabled');
```
brian-smith-tcril, line 2: Did some digging to figure this out, looks like the feature flag check was removed in https://github.com/openedx/paragon/pull/1589

www/src/pages/foundations/typography.tsx
```diff
@@ -196,7 +196,7 @@ export default function TypographyPage({ pageContext }) {
             </tr>
             <tr>
               <td colSpan={2}>
-                <a href="#">Standalone Link</a>
+                <a href={location.pathname}>Standalone Link</a>
```
brian-smith-tcril: This changes the link to refresh the page where the anchor didn't before. If the issue was an invalid anchor could we instead update these to be

```suggestion
                <a href="#links">Standalone Link</a>
```

so they go to the links heading they're under?

brian-smith-tcril (commented):
This is great!

I left a couple comments but nothing major.

Thank you so much for cleaning this up!

src/Sticky/index.jsx
peterkulko: In the case of `useLayoutEffect` `return undefined;` is a cleanup function. According to [the React documentation](https://react.dev/reference/react/useLayoutEffect#returns) _useLayoutEffect returns undefined_, in this case if the linter requires something to be returned I think we should also return `undefined`.

src/SelectableBox/utils.js
peterkulko: Currently, the `getInputType` function returns a React component or `null`. When the function cannot find a matching component type, it returns `null` (i.e. "render nothing"). I believe that this way `null` explicitly indicates that the component was not found, which is more semantically correct.

I also found an excerpt from the article [Understanding “return null” vs “return false” in React](https://medium.com/@ibeanuhillary/understanding-return-null-vs-return-false-in-react-699e98b207d2) that might support my beliefs:

> When you return null from a React component, it tells React to render nothing for that component. This is useful for conditionally omitting elements without causing side effects.

Do you have any other thoughts on this?

www/src/pages/foundations/typography.tsx
peterkulko: Yes, these changes were added to resolve the error with the `jsx-a11y/anchor-is-valid` rule


> 227:17  error  The href attribute requires a valid value to be accessible. Provide a valid, navigable address as the href value. If you cannot provide a valid href, but still need the element to resemble a link, use a button and change it with appropriate styles. Learn more: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/HEAD/docs/rules/anchor-is-valid.md  jsx-a11y/anchor-is-valid


I added `href="#links"` to all the links in the "Links" block.

peterkulko:
@brian-smith-tcril Hi! I have made all the changes after your review. Thanks!

brian-smith-tcril (approved):
LGTM!

brian-smith-tcril:
I'm not 100% sure what codecov is finding on all of the files that are showing a negative change % (specifically not sure what is going on with `ModalDialogHeader`). I'm fine merging this as-is, but if you'd like to do a quick pass on it to see if you can get that codecov number up a bit I'll hold off for a bit.

peterkulko:
> I'm not 100% sure what codecov is finding on all of the files that are showing a negative change % (specifically not sure what is going on with ModalDialogHeader). I'm fine merging this as-is, but if you'd like to do a quick pass on it to see if you can get that codecov number up a bit I'll hold off for a bit.

Currently, in ModalDialogHeader, we’re using both TypeScript and PropTypes. When React renders the component, it applies `defaultProps` first, so the destructured default value (`as = 'div'`) is never actually executed - which is why that line isn’t covered by the tests.

I believe we’ll address the TypeScript and PropTypes issue in components when we remove PropTypes, similar to what was done here: [https://github.com/openedx/paragon/pull/3794](https://github.com/openedx/paragon/pull/3794)

Also, I see that codecov reports missing coverage in the following parts of the code:

image

image

image

The [codecov/patch](https://github.com/openedx/paragon/pull/3488/checks?check_run_id=52548817572) report shows the coverage for the modified files, highlighting the specific areas that aren’t currently covered by tests.

It seems to me that adding unit tests for these sections shouldn’t be part of this PR, since its purpose is just to remove eslint ignore comments. So, I’d suggest skipping this for now.

PR #3554, opened 2025-04-28
feat: adding extendedTour property to ProductTour 

kiram15:
## Description

Changing the header of the ProductTour component, along with the action row buttons. Previously, the action row was "Dismiss" and "Next", but now the dismiss functionality is available in a new close button, and the button is a "Back" button to go back a step in the tour. 
[Figma](https://www.figma.com/design/CDznc7PPoNiEQmfkkT57CV/ProductTour-proposal?node-id=0-1&t=jmJsQgaR9416hMAm-0) 

Before | After 
Screenshot 2025-05-06 at 11 15 27 AMScreenshot 2025-05-06 at 11 15 45 AM

## Testing instructions 
- Run `npm run start` in your local paragon repo after checking out this branch
- Navigate to `http://localhost:8000/components/producttour/`
- Go forward and back in the tour, validate the header matches Figma 
- Ensure that the overridden dismiss functionality (at step 3) works as expected

### Deploy Preview

## Merge Checklist

## Post-merge Checklist


src/ProductTour/CheckpointHeader.jsx
```diff
@@ -0,0 +1,62 @@
...
+    defaultMessage: 'Close tour',
+    description: 'Close alt text for ProductTour component',
+  });
+
+  return (
+    <div className="pgn__checkpoint-header-extended">
+      {index === 0 && (
+        <>
+          <CheckpointTitle>{title}</CheckpointTitle>
+          <p className="pgn__checkpoint-page-index">{oneBasedIndex} of {totalCheckpoints}</p>
```
adamstankiewicz: [suggestion] Looks like there could be an opportunity to create a sub-component like `CheckpointHeaderStatus` (or similar) to make this `p` more re-usable both here and below given they are identical. E.g.

```tsx
const CheckpointHeaderStatus = ({ currentStep, totalSteps }) => (
  <p className="pgn__checkpoint-page-index">{currentStep} of {totalSteps}</p>
);

<CheckpointHeaderStatus currentStep={oneBasedIndex} totalSteps={totalCheckpoints} />
```

adamstankiewicz: Related, we likely should use `react-intl` to tag this string for translation for improved i18n support, so the "of" can be displayed in the user's locale.

adamstankiewicz: On checkpoints with longer titles (e.g., see screenshot for `title` wrapping onto 2+ lines), the "X of Y" appears to reduce its width too far where the "X of Y" itself wraps.

image

This could likely be mitigated by using `flex-shrink: 0` on this element to tell CSS this element should not reduce its width; instead the `title`'s width would shrink accordingly.

src/ProductTour/Checkpoint.scss
```diff
@@ -92,11 +59,24 @@ $checkpoint-arrow-transparent: solid $checkpoint-arrow-width transparent;
     margin-bottom: map.get($spacers, 2\.5);
   }
 
+  .pgn__checkpoint-header-extended {
+    display: flex;
+    justify-content: space-between;
+    margin-bottom: map_get($spacers, 2\.5);
+    align-items: center;
+    height: 30px;
```
adamstankiewicz, line 61: Once the "X of Y" element doesn't shrink (see related comment about `flex-shrink: 0`), I noticed the spacing between the top of the tour contents and the title appears to be less than the current state and what's depicted in Figma when the `title` happens to wrap onto more than 1 line: 

image

Removing this `height: 30px` ensures the intended spacing remains in tact:

image

src/ProductTour/Checkpoint.scss
```diff
@@ -92,11 +59,24 @@ $checkpoint-arrow-transparent: solid $checkpoint-arrow-width transparent;
     margin-bottom: map.get($spacers, 2\.5);
   }
 
+  .pgn__checkpoint-header-extended {
+    display: flex;
+    justify-content: space-between;
+    margin-bottom: map_get($spacers, 2\.5);
+    align-items: center;
```
adamstankiewicz, line 60: Once the "X of Y" element doesn't shrink (see related comment about `flex-shrink: 0`), I noticed that the "X of Y" element is vertically centered within the height of the title. However, if the title happens to be quite long (e.g., wrapping onto 2-3+ lines), keeping the "X of Y" aligned to the top is preferable so it stays in a more "static location", regardless of title length.

src/ProductTour/Checkpoint.scss
```diff
@@ -92,11 +59,24 @@ $checkpoint-arrow-transparent: solid $checkpoint-arrow-width transparent;
     margin-bottom: map.get($spacers, 2\.5);
   }
 
+  .pgn__checkpoint-header-extended {
```
adamstankiewicz: [consideration] Given the "extended" variant is the default now per the design decision, is the `-extended` in this class name necessary? Could it be simplified to `.pgn__checkpoint-header`?

src/ProductTour/Checkpoint.scss
```diff
@@ -92,11 +59,24 @@ $checkpoint-arrow-transparent: solid $checkpoint-arrow-width transparent;
...
+
   #pgn__checkpoint-title {
     font-size: $h3-font-size;
     margin-inline-end: map.get($spacers, 2\.5);
     margin-bottom: 0;
   }
+
+  .pgn__checkpoint-page-index {
+    font-size: $small-font-size;
+    margin-bottom: 0;
```
adamstankiewicz, line 72: [consideration] Might opt to use a `span` vs. `p` for this element so you don't need to override `margin-bottom: 0;`.

src/ProductTour/CheckpointHeader.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const closeAltText = intl.formatMessage({
+    id: 'pgn.ProductTour.checkpointHeader.close',
+    defaultMessage: 'Close tour',
+    description: 'Close alt text for ProductTour component',
+  });
+
+  return (
+    <div className="pgn__checkpoint-header-extended">
+      {index === 0 && (
```
adamstankiewicz: nit: is it worth having a `const isFirstCheckpoint = index === 0;` helper variable (similar to other components) to use and in the conditional below?

```
{isFirstCheckpoint && ...
{!isFirstCheckpoint && ...
```

src/ProductTour/CheckpointHeader.jsx
```diff
@@ -0,0 +1,62 @@
...
+        <>
+          <CheckpointTitle>{title}</CheckpointTitle>
```
adamstankiewicz: [consideration] `CheckpointHeader` props API suggests `title` is optional, and defaults to empty string, whereas `CheckpointTitle` treats it as an optional prop, defaulting to `null`.

Couple observations/questions:
1. It looks like `CheckpointTitle` renders an `h2` even if it has no title to render, resulting in an empty `h2` in the DOM.
2. I wonder if:
    - We make `title` required in `CheckpointTitle`?
    - Ensure `CheckpointTitle` only gets (conditionally) rendered if `title` exists?

PR #3572, opened 2025-05-09
feat: changing the header of ProductTour 

kiram15:
## Description

Changing the header of the ProductTour component, along with the action row buttons. Previously, the action row was "Dismiss" and "Next", but now the dismiss functionality is available in a new close button, and the button is a "Back" button to go back a step in the tour. 
[Figma](https://www.figma.com/design/CDznc7PPoNiEQmfkkT57CV/ProductTour-proposal?node-id=0-1&t=jmJsQgaR9416hMAm-0) 

Before | After 
Screenshot 2025-05-06 at 11 15 27 AMScreenshot 2025-05-06 at 11 15 45 AM

## Testing instructions 
- Run `npm run start` in your local paragon repo after checking out this branch
- Navigate to `http://localhost:8000/components/producttour/`
- Go forward and back in the tour, validate the header matches Figma 

## Description

### Deploy Preview

## Merge Checklist

## Post-merge Checklist


src/ProductTour/CheckpointActionRow.jsx
```diff
@@ -4,65 +4,64 @@ import Button from '../Button';
...
+          {backButtonText}
+        </Button>
+      )}
       <Button
-        variant="tertiary"
-        className="pgn__checkpoint-button_dismiss"
-        onClick={onDismiss}
+        autoFocus
+        className="pgn__checkpoint-button_advance"
+        variant="primary"
```
peterkulko: Do we need to specify the "primary" variant if it is specified by default for buttons in Paragon?

src/ProductTour/CheckpointActionRow.jsx
```diff
@@ -4,65 +4,64 @@ import Button from '../Button';
...
-}, ref) => (
-  <div className="pgn__checkpoint-action-row" ref={ref}>
-    {(showDismissButton === undefined ? !isLastCheckpoint : showDismissButton) && (
+}, ref) => {
+  const isFirstCheckpoint = index === 0;
+  return (
+    <div className="pgn__checkpoint-action-row" ref={ref}>
+      {!isFirstCheckpoint && (
+        <Button
+          className="pgn__checkpoint-button_back"
```
peterkulko: Previously, the button had a class [pgn__checkpoint-button_dismiss](https://github.com/openedx/paragon/blob/release-22.x/src/ProductTour/Checkpoint.scss#L46) which provided appropriate `margin-inline-end` for the button. Now I see that there is no `margin-inline-end ` between the action buttons in the middle of the checkpoint. Does it make sense to change `pgn__checkpoint-button_dismiss` to `pgn__checkpoint-button_back`?

[Paragon production](https://paragon-openedx-v22.netlify.app/components/producttour/)
image

[Deploy preview](https://deploy-preview-3572--paragon-openedx-v23.netlify.app/components/producttour/) for this PR
image

src/ProductTour/CheckpointActionRow.jsx
```diff
@@ -4,65 +4,64 @@ import Button from '../Button';
...
   onEnd,
-  showDismissButton,
   index,
-}, ref) => (
-  <div className="pgn__checkpoint-action-row" ref={ref}>
-    {(showDismissButton === undefined ? !isLastCheckpoint : showDismissButton) && (
+}, ref) => {
+  const isFirstCheckpoint = index === 0;
+  return (
+    <div className="pgn__checkpoint-action-row" ref={ref}>
```
peterkulko: [question]: If it makes sense to use the [Paragon Stack component](https://paragon-openedx-v22.netlify.app/components/stack/) here?

src/ProductTour/Checkpoint.scss
```diff
@@ -89,12 +89,19 @@ $checkpoint-arrow-transparent: solid $checkpoint-arrow-width transparent;
...
+    margin-bottom: map_get($spacers, 2);
+    align-items: center;
+    height: 30px;
   }
 
   #pgn__checkpoint-title {
     font-size: $h3-font-size;
-    margin-inline-end: map.get($spacers, 2\.5);
+    margin-inline-end: map_get($spacers, 2\.5);
+    margin-bottom: .5rem;
```
peterkulko:
```suggestion
    margin-bottom: map_get($spacers, 2);
```

src/ProductTour/Checkpoint.scss
```diff
@@ -89,12 +89,19 @@ $checkpoint-arrow-transparent: solid $checkpoint-arrow-width transparent;
   .pgn__checkpoint-header {
     display: flex;
     justify-content: space-between;
-    margin-bottom: map.get($spacers, 2\.5);
+    margin-bottom: map_get($spacers, 2);
+    align-items: center;
+    height: 30px;
```
peterkulko: [question]: I don't have access to the design, but it seems to me that we shouldn't hardcode set the height for `pgn__checkpoint-header`. Maybe we should resize the Close button (which is currently stretching this block), or even remove the height altogether. These are just my thoughts. What do you think?

src/ProductTour/CheckpointHeader.jsx
```diff
@@ -0,0 +1,62 @@
...
+import CheckpointTitle from './CheckpointTitle';
+import messages from './messages';
+
+const CheckpointHeader = React.forwardRef(({
+  index, onDismiss, title, totalCheckpoints,
+}) => {
+  const intl = useIntl();
+  const oneBasedIndex = index + 1;
+
+  const closeAltText = intl.formatMessage({
```
peterkulko: Should these translations be in the [messages.js](https://github.com/openedx/paragon/blob/release-22.x/src/ProductTour/messages.js#L3) file?

src/ProductTour/CheckpointHeader.jsx
```diff
@@ -0,0 +1,62 @@
...
+            {...messages.pageIndexText}
+            values={{ step: oneBasedIndex, totalSteps: totalCheckpoints }}
+          />
+        </p>
+        <IconButton
+          iconAs={Icon}
+          src={Close}
+          alt={closeAltText}
+          onClick={onDismiss}
+          variant="primary"
```
peterkulko: Do we need to specify the "primary" variant if it is specified by default for buttons in Paragon?

src/ProductTour/CheckpointHeader.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const closeAltText = intl.formatMessage({
+    id: 'pgn.ProductTour.checkpointHeader.close',
+    defaultMessage: 'Close tour',
+    description: 'Close alt text for ProductTour component',
+  });
+
+  return (
+    <>
+      <div className="pgn__checkpoint-header">
```
peterkulko: [question]: If this is a checkpoint header, should there be a corresponding HTML semantic tag (`<header>`) instead of a `div`?

src/ProductTour/ProductTour.test.jsx
```diff
@@ -132,17 +127,58 @@ describe('<ProductTour />', () => {
...
         });
 
         // Verify no Checkpoints have rendered
         expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
       });
 
+      it('onClick of close icon button disables tour', async () => {
+        const { rerender } = render(<ProductTourWrapper tours={[tourData]} />);
+        // Advance the tour, close icon only appears after first step
+        await act(async () => {
```
peterkulko: Given that `userEvent` [already has a wrapper](https://flyingsquirrel.medium.com/when-should-i-use-act-in-react-testing-library-d7dd22a3340e) in `act` in its internal implementation, I suggest removing this additional wrapper.

src/ProductTour/ProductTour.test.jsx
```diff
@@ -132,17 +127,58 @@ describe('<ProductTour />', () => {
...
+          await userEvent.click(screen.getByRole('button', { name: 'Next' }));
+        });
+        rerender(<ProductTourWrapper tours={[tourData]} />);
+
+        const closeIcon = screen.getByRole('button', { name: 'Close tour' });
```
peterkulko: If we move the translation from the [CheckpointHeader.jsx](https://github.com/kiram15/paragon/blob/kiram15/ENT-10146-1/src/ProductTour/CheckpointHeader.jsx#L17) component to a translation file (messages.js), we will be able to use this translation in our tests.

dependent-usage.json
```diff
@@ -1,5 +1,5 @@
 {
-  "lastModified": 1733926086658,
+  "lastModified": 1734012486681,
```
peterkulko: [question]: It seems this file is automatically updated by the github action, should we delete this change?

PKulkoRaccoonGang (dismissed):
Thanks for this PR! I left a few suggestions

adamstankiewicz:
@kiram15 Reminder that Netlify unfortunately doesn't handle when the base branch is changed (went from `release-23.x` to `release-22.x`), so the deploy preview is still for v23 despite the PR being against `release-22.x`. The PR needs to be initially created from the start against `release-22.x` for the Netlify deploy preview against v22, without changing the base branch in the GitHub UI.

src/ProductTour/Checkpoint.scss
```diff
@@ -10,7 +10,7 @@ $checkpoint-arrow-transparent: solid $checkpoint-arrow-width transparent;
   background: $checkpoint-background-color;
   border-top: $checkpoint-border-width solid $checkpoint-border-color;
   border-radius: $border-radius;
-  padding: map.get($spacers, 4);
+  padding: map_get($spacers, 3\.5);
```
adamstankiewicz: Was the change from `map.get` to `map_get` intentional?

src/ProductTour/Checkpoint.scss
```diff
@@ -89,12 +89,19 @@ $checkpoint-arrow-transparent: solid $checkpoint-arrow-width transparent;
   .pgn__checkpoint-header {
     display: flex;
     justify-content: space-between;
-    margin-bottom: map.get($spacers, 2\.5);
+    margin-bottom: map_get($spacers, 2);
```
adamstankiewicz: Was the change from `map.get` to `map_get` intentional?

src/ProductTour/Checkpoint.scss
adamstankiewicz: I agree we should avoid hardcoding height, if possible; should be feasible for `.pgn__checkpoint-header` to inherit its height from its children (e.g., the close button, etc.).

src/ProductTour/Checkpoint.scss
```diff
@@ -89,12 +89,19 @@ $checkpoint-arrow-transparent: solid $checkpoint-arrow-width transparent;
...
+    align-items: center;
+    height: 30px;
   }
 
   #pgn__checkpoint-title {
     font-size: $h3-font-size;
-    margin-inline-end: map.get($spacers, 2\.5);
+    margin-inline-end: map_get($spacers, 2\.5);
```
adamstankiewicz: Was the change from `map.get` to `map_get` intentional?

src/ProductTour/Checkpoint.scss
```diff
@@ -89,12 +89,19 @@ $checkpoint-arrow-transparent: solid $checkpoint-arrow-width transparent;
...
+    margin-bottom: .5rem;
+  }
+
+  .pgn__checkpoint-page-index {
+    font-size: $small-font-size;
     margin-bottom: 0;
```
adamstankiewicz: As suggested [here](https://github.com/openedx/paragon/pull/3554#discussion_r2077705882), I would still recommend switching the associated element to a `span` vs `p` so we don't need to override `margin-bottom: 0;`.

src/ProductTour/CheckpointHeader.jsx
```diff
@@ -0,0 +1,62 @@
...
+        <p className="pgn__checkpoint-page-index">
+          <FormattedMessage
+            id="pgn.ProductTour.CheckpointHeader.pageIndex"
+            {...messages.pageIndexText}
```
adamstankiewicz: Spreading `messages.pageIndexText` already includes an `id`; is the `id` prop above on line 28 necessary?

src/ProductTour/ProductTour.test.jsx
```diff
@@ -132,17 +127,58 @@ describe('<ProductTour />', () => {
...
 
         // Verify no Checkpoints have rendered
         expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
       });
 
+      it('onClick of close icon button disables tour', async () => {
+        const { rerender } = render(<ProductTourWrapper tours={[tourData]} />);
+        // Advance the tour, close icon only appears after first step
+        await act(async () => {
+          await userEvent.click(screen.getByRole('button', { name: 'Next' }));
```
adamstankiewicz: Note Paragon uses the latest version v14 of `@testing-library/user-event`, where `userEvent.click(...)` is no longer recommended in favor of:

```jsx
const user = userEvent.setup();
await user.click(...);
```

The [RTL docs](https://testing-library.com/docs/user-event/setup#direct-apis) recommend using `userEvent.setup()` moving forward. Given this, I would recommend any _net-new tests_ you're adding to use the recommended `userEvent.setup()` approach; don't worry about refactoring existing tests, though.

src/ProductTour/ProductTour.test.jsx
```diff
@@ -132,17 +127,58 @@ describe('<ProductTour />', () => {
...
+        });
+        rerender(<ProductTourWrapper tours={[tourData]} />);
```
adamstankiewicz: Why is it necessary to use `rerender` here and in subsequent tests? Context: the test passes without the `rerender` 🤔

src/ProductTour/README.md
```diff
@@ -16,6 +16,10 @@ tours are enabled, `ProductTour` will only render the first enabled in the `tour
 `Checkpoints` are rendered in the order they're listed in the checkpoint array.
 The checkpoint objects themselves have additional props that can override the props defined in `ProductTour`.
 
+Best practices for ProductTour includes not overloading the user with a large amount of steps.
```
adamstankiewicz, line 20: 1. Would recommend adding a new "Usage Guidelines" heading here so the usage guidelines don't get lost in the general component description above. 
2. I would still also include no more than 5 steps as the general recommendation, given that's the current usage guidelines we have today versus just saying keeping the number of steps "to a minimum".

src/ProductTour/CheckpointActionRow.jsx
kiram15: Good point, but I actually think using the ActionRow component is more appropriate

src/ProductTour/CheckpointActionRow.jsx
kiram15: Adding in the ActionRow suggestion from earlier actually adds in padding so now it does have the buffer in between!

src/ProductTour/index.jsx
```diff
@@ -180,15 +190,13 @@ ProductTour.propTypes = {
       target: PropTypes.string.isRequired,
       /** The text displayed in the title of the Checkpoint */
       title: PropTypes.node,
-      /** Enforces visibility of the dismiss button under all circumstances */
-      showDismissButton: PropTypes.bool,
     })),
-    /** The text displayed on the button used to dismiss the tour. */
-    dismissButtonText: PropTypes.node,
```
adamstankiewicz, line 187: [consideration/suggestion] Removing support for the `dismissButtonText` prop is *technically* a breaking change, given some consumers use it today (both as a string and as a node), and might expect its output to exist (e.g., in an MFE's test assertion that could start failing, for instance our new Enterprise e2e Playwright tests that expects the "Dismiss" button to exist today, [example](https://github.com/edx/playwright-e2e/blob/b75adde0ea13610e95e6063d454c4d7c89e0ffa2/e2e/poms/admin-portal/LearnerCreditManagementPage.ts#L31-L42)). I think we need to keep support for the prop but re-purpose it. Just wrapped up a quick async chat with @brian-smith-tcril and aligned on a potential path forward to mitigate any breaking changes for consumers (e.g., trying to avoid causing any existing MFE tests to fail, etc.).

**Summary:**
1. Keep the prop `dismissButtonText`, but deprecate it with `withDeprecatedProp` wrapper that exists in Paragon to give dev console warning about the deprecation ([example](https://github.com/openedx/paragon/blob/7f522485d286c5192a87aa4427da97a3da984c04/src/Card/index.jsx#L79-L84)).
1. Treat `dismissButtonText` as a pass-thru to a new `dismissAltText` prop (i.e., `dismissAltText` = `dismissButtonText`).
1. `dismissAltText` should only accept a string; if `dismissButtonText` was originally passed in as a node (e.g., `frontend-app-learning` passes in a `<FormattedMessage>`), fallback to the default Paragon string "Close tour" so it's still a valid string message in the "X" button's `alt` for users.

Happy to clarify on a huddle if you have questions! beyond this (i.e., once we mitigate this potential breaking change for consumers), PR will be ready to go!

src/ProductTour/CheckpointHeader.jsx
```diff
@@ -0,0 +1,60 @@
...
+            values={{ step: oneBasedIndex, totalSteps: totalCheckpoints }}
+          />
+        </span>
+        <IconButton
+          size="sm"
+          iconAs={Icon}
+          src={Close}
+          alt={altText}
+          onClick={onDismiss}
+          data-testid="Dismiss tour"
```
adamstankiewicz: nit: testids are typically hyphenated like "dismiss-tour" vs. sentence case ("Dismiss tour").

The "Dismiss tour" type of assertions would ideally come from `getByText('Dismiss tour')` rather than `getByTestId('Dismiss tour')`, for example, to act more like a user relying on the rendered button text.

src/ProductTour/ProductTour.test.jsx
```diff
@@ -98,31 +96,25 @@ describe('<ProductTour />', () => {
...
         // Click the advance button
         const advanceButton = screen.getByRole('button', { name: 'Next' });
-        await act(async () => {
-          await userEvent.click(advanceButton);
-        });
-
-        rerender(<ProductTourWrapper tours={[tourData]} />);
-
-        const heading = screen.getByRole('heading', { name: 'Checkpoint 2' });
+        await userEvent.click(advanceButton);
```
adamstankiewicz, line 114: nit: one of a few other spots where `userEvent.click` could migrate to `user.click` via `userEvent.setup()`, similar to other tests in this file.

src/ProductTour/index.jsx
```diff
@@ -149,16 +177,20 @@ ProductTour.propTypes = {
   tours: PropTypes.arrayOf(PropTypes.shape({
     /** The text displayed on all buttons used to advance the tour. */
     advanceButtonText: PropTypes.node,
+    /** The text displayed on all buttons used to go back in the tour */
+    backButtonText: PropTypes.node,
```
adamstankiewicz: [nit/consider] We may want to require this `backButtonText` prop as a string for the net-new prop, to avoid situations like the now-deprecated `dismissButtonText` causing issues as a node. 

Thinking of a future hypothetical case where the back button switches to a back arrow icon, where we'd want to treat this prop as `alt` text similar to the migration to `dismissAltText`. Having it start out as a string would make supporting things like that easier in the future!

(Same feedback for the `backButtonText` inline within `checkpoints` array below).

src/ProductTour/index.jsx
adamstankiewicz: [update] This feedback has been addressed.

adamstankiewicz (approved):
LGTM! Great job with these changes, and I really appreciate you shepherding the contribution through the proper channels, and trying to encourage more design input from the Open edX community on these types of contributions 🙌 😄

PR #3598, opened 2025-05-20
feat: introduce tokens approach to overriding alert btn variants

adamstankiewicz:
## Description

Introduces a mechanism for brand packages (`@edx/brand`) to override the default button variants used in certain components (starting with `Alert`) without needing to modify upstream `openedx` MFE code.

For instance, below, the example MFE's usage renders a primary button by default:

```tsx
<Alert actions={[<Button>Hello world</Button>]}>
```

However, an `@edx/brand` package might opt to override the following `Alert` design token JSON file (i.e., `tokens/src/themes/light/components/Alert.json`):

```json
{
  "color": {
    "$type": "color",
    "alert": {
      "actions": {
        "overrides": {
          "button": {
            "variants": {
              "primary": "brand"
            }
          }
        }
      }
    }
  }
}
```

The `style-dictionary` build process was modified to parse these override tokens, generating an `overrides/alert.css` file with the following:

```css
/**
 * Do not edit directly, this file was auto-generated.
 */

.pgn__alert-message-wrapper .pgn__alert-actions .btn-primary,
.pgn__alert-message-wrapper-stacked .pgn__alert-actions .btn-primary {
  --pgn-btn-color: var(--pgn-color-btn-text-brand);
  --pgn-btn-bg: var(--pgn-color-btn-bg-brand);
  --pgn-btn-border-color: var(--pgn-color-btn-border-brand);
  --pgn-btn-hover-color: var(--pgn-color-btn-hover-text-brand);
  --pgn-btn-hover-bg: var(--pgn-color-btn-hover-bg-brand);
  --pgn-btn-hover-border-color: var(--pgn-color-btn-hover-border-brand);
  --pgn-btn-disabled-color: var(--pgn-color-btn-disabled-text-brand);
  --pgn-btn-disabled-bg: var(--pgn-color-btn-disabled-bg-brand);
  --pgn-btn-disabled-border-color: var(--pgn-color-btn-disabled-border-brand);
  --pgn-btn-active-color: var(--pgn-color-btn-active-text-brand);
  --pgn-btn-active-bg: var(--pgn-color-btn-active-bg-brand);
  --pgn-btn-active-border-color: var(--pgn-color-btn-active-border-brand);
  --pgn-btn-focus-outline-color: var(--pgn-color-btn-focus-outline-brand);
  --pgn-btn-focus-color: var(--pgn-color-btn-focus-text-brand);
  --pgn-btn-focus-border-color: var(--pgn-color-btn-focus-border-brand);
  --pgn-btn-focus-bg: var(--pgn-color-btn-focus-bg-brand);
}
```

Where, it overrides the base button CSS variables for the `.btn-<originalVariant>` used in the `openedx` MFE code to be the desired variant's CSS variables instead, e.g. `var(--pgn-color-btn-text-brand)` vs `var(--pgn-color-btn-text-primary)`.

**Open questions**
* How should this approach be best documented for theme authors?
    * `brand-openedx`?
      * **We will update the README for `brand-openedx`.
    * Paragon README?
    * `$description` attribute in the tokens JSON?

### Deploy Preview

* https://deploy-preview-3598--paragon-openedx-v23.netlify.app/
* https://deploy-preview-3598--paragon-openedx-v23.netlify.app/guides/theming/
* https://deploy-preview-3598--paragon-openedx-v23.netlify.app/components/alert/#with-action-buttons

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/lib/build-tokens.js
+++ b/lib/build-tokens.js
@@ -139,6 +139,14 @@ async function buildTokensCommand(commandArgs) {
               outputReferences,
             },
           },
+          {
+            format: 'css/button-variant-overrides-alert-actions',
+            destination: `themes/${themeVariant}/overrides/alert.css`,
+            filter: transformSourceTokensOnly ? 'isSource' : undefined,
+            options: {
+              outputReferences,
+            },
+          },
         ],
       },
     },
--- a/lib/build-tokens.js
+++ b/lib/build-tokens.js
@@ -140,8 +140,8 @@ async function buildTokensCommand(commandArgs) {
             },
           },
           {
-            format: 'css/button-variant-overrides-alert-actions',
-            destination: `themes/${themeVariant}/overrides/alert.css`,
+            format: 'css/component-button-variant-overrides',
+            destination: `themes/${themeVariant}/overrides/component-button-variants.css`,
             filter: transformSourceTokensOnly ? 'isSource' : undefined,
             options: {
               outputReferences,
--- a/src/Button/_mixins.scss
+++ b/src/Button/_mixins.scss
@@ -16,7 +16,29 @@
   @include border-radius($button-border-radius, 0);
 }
 
-// Mixin for button states that affect border radius
+/// Sets the CSS custom properties for a button variant
+/// @param {String} $variant - The variant name (e.g., 'primary', 'outline-primary')
+@mixin button-variant($variant) {
+  --pgn-btn-color: var(--pgn-color-btn-text-#{$variant});
+  --pgn-btn-bg: var(--pgn-color-btn-bg-#{$variant});
+  --pgn-btn-border-color: var(--pgn-color-btn-border-#{$variant});
+  --pgn-btn-hover-color: var(--pgn-color-btn-hover-text-#{$variant});
+  --pgn-btn-hover-bg: var(--pgn-color-btn-hover-bg-#{$variant});
+  --pgn-btn-hover-border-color: var(--pgn-color-btn-hover-border-#{$variant});
+  --pgn-btn-disabled-color: var(--pgn-color-btn-disabled-text-#{$variant});
+  --pgn-btn-disabled-bg: var(--pgn-color-btn-disabled-bg-#{$variant});
+  --pgn-btn-disabled-border-color: var(--pgn-color-btn-disabled-border-#{$variant});
+  --pgn-btn-active-color: var(--pgn-color-btn-active-text-#{$variant});
+  --pgn-btn-active-bg: var(--pgn-color-btn-active-bg-#{$variant});
+  --pgn-btn-active-border-color: var(--pgn-color-btn-active-border-#{$variant});
+  --pgn-btn-focus-outline-color: var(--pgn-color-btn-focus-outline-#{$variant});
+  --pgn-btn-focus-color: var(--pgn-color-btn-focus-text-#{$variant});
+  --pgn-btn-focus-border-color: var(--pgn-color-btn-focus-border-#{$variant});
+  --pgn-btn-focus-bg: var(--pgn-color-btn-focus-bg-#{$variant});
+}
+
+/// Mixin for button states that affect border radius
+/// @param {String} $side - The side of the button to affect ('left' or 'right')
 @mixin button-border-radius-0($side) {
   @if $side == 'right' {
     @include border-right-radius(0);
@@ -26,7 +48,7 @@
     @include border-left-radius(0);
   }
   
-  // Handle pseudo-classes
+  /// Handle pseudo-classes
   &.btn:not(:disabled, .disabled):focus,
   &.btn:not(:disabled, .disabled):focus::before {
     @if $side == 'left' {
@@ -49,7 +71,7 @@
     }
   }
   
-  // Handle .active class
+  /// Handle .active class
   &.btn:not(:disabled, .disabled).active,
   &.btn:not(:disabled, .disabled).active::before {
     @if $side == 'right' {
--- a/tokens/src/themes/light/components/Alert.json
+++ b/tokens/src/themes/light/components/Alert.json
@@ -21,6 +21,11 @@
         "info": { "source": "$alert-info-border-color", "$value": "{color.theme.border.info}" },
         "danger": { "source": "$alert-danger-border-color", "$value": "{color.theme.border.danger}" },
         "warning": { "source": "$alert-warning-border-color", "$value": "{color.theme.border.warning}" }
+      },
+      "actions": {
+        "variant-overrides": {
+          "primary": "brand"
+        }
       }
     }
   }
--- a/tokens/src/themes/light/components/Alert.json
+++ b/tokens/src/themes/light/components/Alert.json
@@ -21,6 +21,20 @@
         "info": { "source": "$alert-info-border-color", "$value": "{color.theme.border.info}" },
         "danger": { "source": "$alert-danger-border-color", "$value": "{color.theme.border.danger}" },
         "warning": { "source": "$alert-warning-border-color", "$value": "{color.theme.border.warning}" }
+      },
+      "actions": {
+        "overrides": {
+          "button": {
+            "variants": {}
+          },
+          "examples": {
+            "button": {
+              "variants": {
+                "primary": "brand"
+              }
+            }
+          }
+        }
       }
     }
   }
--- a/tokens/src/themes/light/components/Alert.json
+++ b/tokens/src/themes/light/components/Alert.json
@@ -21,16 +21,6 @@
         "info": { "source": "$alert-info-border-color", "$value": "{color.theme.border.info}" },
         "danger": { "source": "$alert-danger-border-color", "$value": "{color.theme.border.danger}" },
         "warning": { "source": "$alert-warning-border-color", "$value": "{color.theme.border.warning}" }
-      },
-      "actions": {
-        "overrides": {
-          "variants": {},
-          "examples": {
-            "variants": {
-              "primary": "brand"
-            }
-          }
-        }
       }
     }
   }
--- a/tokens/src/themes/light/components/Alert.json
+++ b/tokens/src/themes/light/components/Alert.json
@@ -23,8 +23,13 @@
         "warning": { "source": "$alert-warning-border-color", "$value": "{color.theme.border.warning}" }
       },
       "actions": {
-        "variant-overrides": {
-          "primary": "brand"
+        "overrides": {
+          "variants": {},
+          "examples": {
+            "variants": {
+              "primary": "brand"
+            }
+          }
         }
       }
     }
--- a/tokens/style-dictionary.js
+++ b/tokens/style-dictionary.js
@@ -364,10 +364,13 @@ const initializeStyleDictionary = async ({ themes }) => {
   StyleDictionary.registerFormat({
     name: 'css/button-variant-overrides-alert-actions',
     format: async ({ dictionary }) => {
-      let output = '';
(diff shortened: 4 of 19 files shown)


tokens/src/themes/light/components/Alert.json
```diff
@@ -21,6 +21,16 @@
...
         "warning": { "source": "$alert-warning-border-color", "$value": "{color.theme.border.warning}" }
+      },
+      "actions": {
+        "overrides": {
+          "variants": {},
+          "examples": {
+            "variants": {
+              "primary": "brand"
+            }
+          }
```
adamstankiewicz: [inform] This `examples` key serves as a way to provide some inline-documentation in the JSON tokens themselves around usage for the `actions.overrides.variants`.

lib/build-tokens.js
```diff
@@ -139,6 +139,14 @@ async function buildTokensCommand(commandArgs) {
               outputReferences,
             },
           },
+          {
+            format: 'css/button-variant-overrides-alert-actions',
```
adamstankiewicz: [inform] This custom format proposes a naming convention when this approach needs to apply to other components, too (e.g., modals might be a custom format `css/button-variant-overrides-modal-actions` or similar).

src/Button/_mixins.scss
```diff
@@ -16,7 +16,29 @@
   @include border-radius($button-border-radius, 0);
 }
 
-// Mixin for button states that affect border radius
+/// Sets the CSS custom properties for a button variant
+/// @param {String} $variant - The variant name (e.g., 'primary', 'outline-primary')
+@mixin button-variant($variant) {
```
adamstankiewicz, line 21: [inform] Makes the setting of these CSS variables for each `.btn-<variant>` class name definition more DRY/reusable. See usage in `src/Button/index.scss`.

tokens/style-dictionary.js
```diff
@@ -330,6 +330,58 @@ const initializeStyleDictionary = async ({ themes }) => {
     },
   });
 
+  /**
+   * Generates CSS custom properties for a button variant that matches the button-variant mixin
+   * @param {string} variant - The variant name (e.g., 'primary', 'brand')
+   * @returns {string} CSS custom properties for the button variant
+   */
+  const generateButtonVariantProperties = (variant) => {
+    const properties = [
```
adamstankiewicz: [inform] Same properties as the `button-variant` SCSS mixin.

lib/build-tokens.js
```diff
@@ -139,6 +139,14 @@ async function buildTokensCommand(commandArgs) {
...
+            destination: `themes/${themeVariant}/overrides/alert.css`,
```
adamstankiewicz: Similar to above, this proposed putting all generated, component-specific overrides under an `overrides` directory (e.g., this might include `/overrides/modal.css` in the future).

tokens/src/themes/light/components/Alert.json
```diff
@@ -21,6 +21,20 @@
         "info": { "source": "$alert-info-border-color", "$value": "{color.theme.border.info}" },
         "danger": { "source": "$alert-danger-border-color", "$value": "{color.theme.border.danger}" },
         "warning": { "source": "$alert-warning-border-color", "$value": "{color.theme.border.warning}" }
+      },
+      "actions": {
+        "overrides": {
+          "button": {
+            "variants": {}
```
adamstankiewicz, line 28: [inform] The token schema exists, but is a no-op for default (i.e., `variants` is empty object, so the style-dictionary format won't have any key/values to iterate over by default). The `examples` key below intends to act as in-line documentation around what to pass into `variants`, though the primary documentation for this schema will likely live in `brand-openedx` (PR still to come).

lib/build-tokens.js
brian-smith-tcril: Thinking about the naming convention:

It feels like we can generalize
```
css/button-variant-overrides-modal-actions
```

to something like

```
css/{component}-variant-overrides-{component-use}
```

I think it might make more sense to have it be
```
css/variant-overrides-{component}-{component-use}
```
which would change this specific case to
```
css/variant-overrides-button-modal-actions
```

But considering the `destination` would be `overrides/modal.css` and not `overrides/button.css` my `component` vs `component-use` structure doesn't seem to fully make sense.

And considering `Alert.json` has

```json
      "actions": {
        "overrides": {
          "button": {
            "variants": {}
```

maybe following that structure makes sense? So something like

```
css/alert-actions-overrides-button-variant
```

I'm just hoping to open up conversation around this one. Looking forward to hearing other thoughts!

src/Button/_mixins.scss
brian-smith-tcril: This is great! If you want to break this out into a tiny separate PR I'd be happy to land this change standalone while we work through the details of the rest of this PR.

brian-smith-tcril (commented):
Thanks for documenting the decisions!

I left a comment hoping to start a discussion around the custom format naming convention.

Overall this is looking great!

lib/build-tokens.js
adamstankiewicz: Ah, the original `css/button-variant-overrides-modal-actions` was originally generalized as `css/button-variant-overrides-{component}-{component-use}`. I like the suggestion to go off the token schema for the name; should be portable enough to other components, i.e.:

```
css/alert-actions-overrides-button-variant
```

Another consideration with fresh eyes this morning is whether we might want to consider a singular format that handles all the possible button variant related component overrides in a single output file, e.g. `overrides/component-button-variant-overrides.css`, including `Alert`, modals, etc.

For example, instead of having a custom format only handling `const alertButtonVariantOverrides = dictionary.tokens.color?.alert?.actions?.overrides?.button?.variants;`, perhaps a singular format could iterate over `alertButtonVariantOverrides`, `modalButtonVariantOverrides`, etc. something like:

```js
/**
   * Configuration for button variant overrides by component type.
   * Each entry defines:
   * - getTokens: Function that receives the dictionary tokens and returns the button variant overrides
   * - selectors: Array of CSS selectors where the button variants should be overridden
   */
  const BUTTON_VARIANT_OVERRIDES_CONFIG = [
    {
      name: 'Alert',
      getTokens: (tokens) => tokens?.color?.alert?.actions?.overrides?.button?.variants,
      selectors: [
        '.pgn__alert-message-wrapper .pgn__alert-actions',
        '.pgn__alert-message-wrapper-stacked .pgn__alert-actions',
      ],
    },
    // Add new component configurations here, for example:
    // {
    //   name: 'modal',
    //   getTokens: (tokens) => tokens?.color?.modal?.actions?.overrides?.button?.variants,
    //   selectors: ['.pgn__modal .pgn__modal-footer'],
    // },
  ];

  /**
   * Registers a custom format for generating CSS style overrides for button variants in different components.
   * All overrides are combined into a single CSS file.
   */
  StyleDictionary.registerFormat({
    name: 'css/component-button-variant-overrides',
    format: async ({ dictionary }) => {
      const { fileHeader } = await getStyleDictionaryUtils();
      const header = await fileHeader({
        file: 'overrides/component-button-variants.css',
        formatting: 'css',
      });
      let output = `${header}\n`;

      // Process each component configuration
      BUTTON_VARIANT_OVERRIDES_CONFIG.forEach((config) => {
        const buttonVariantOverrides = config.getTokens(dictionary.tokens);

        if (!buttonVariantOverrides || typeof buttonVariantOverrides !== 'object') {
          return; // No overrides found for this component, skip.
        }

        output += `// ${config.name}\n\n`;

        Object.entries(buttonVariantOverrides).forEach(([originalVariant, overrideVariant]) => {
          const selectorOutput = config.selectors
            .map(selector => `${selector} .btn-${originalVariant}`)
            .join(',\n');

          output += `${selectorOutput} {\n`;
          output += `  ${generateButtonVariantProperties(overrideVariant)}\n`;
          output += '}\n\n';
        });
      });

      return output;
    },
  });
```

Same output as before, but means we can extend an existing format for future components that might need this approach (e.g., modals) via `BUTTON_VARIANT_OVERRIDES_CONFIG` versus having to register/use a new custom format for each component.

brian-smith-tcril: > I like the suggestion to go off the token schema for the name; should be portable enough to other components

Cool! 

> Another consideration with fresh eyes this morning is whether we might want to consider a singular format that handles all the possible button variant related component overrides in a single output file, e.g. overrides/component-button-variant-overrides.css, including Alert, modals, etc.

I think we discussed something similar to that in PWG this week. I remember wanting to make sure that we didn't include `Alert` specific overrides in `Button`.

If I'm understanding correctly that wouldn't be a problem with this proposal, in which case `css/overrides/component-button-variant-overrides.css` sounds great.

If implementing that ends up being too complex I'm happy with `css/alert-actions-overrides-button-variant` too.

tokens/style-dictionary.js
```diff
@@ -330,6 +330,58 @@ const initializeStyleDictionary = async ({ themes }) => {
...
+      `--pgn-btn-active-bg: var(--pgn-color-btn-active-bg-${variant});`,
+      `--pgn-btn-active-border-color: var(--pgn-color-btn-active-border-${variant});`,
+      `--pgn-btn-focus-outline-color: var(--pgn-color-btn-focus-outline-${variant});`,
+      `--pgn-btn-focus-color: var(--pgn-color-btn-focus-text-${variant});`,
+      `--pgn-btn-focus-border-color: var(--pgn-color-btn-focus-border-${variant});`,
+      `--pgn-btn-focus-bg: var(--pgn-color-btn-focus-bg-${variant});`,
+    ];
+
+    return properties.join('\n  ');
+  };
```
MaxFrank13: At what point do you think it's called for to move something like this into a separate file? With this plus the potential `BUTTON_VARIANT_OVERRIDES_CONFIG` object, I feel it could get pretty massive. That being said, we probably won't edit that config object very often so maybe it's not really a problem that everything lives in this file

adamstankiewicz: I could see this `generateButtonVariantProperties` function moving to the `tokens/utils.js` file to keep `style-dictionary.js` more concise. I think we'd want to keep `BUTTON_VARIANT_OVERRIDES_CONFIG` co-located with the registered format, though.

More forward-looking, maybe once we start adding additional component types to `BUTTON_VARIANT_OVERRIDES_CONFIG` (e.g., beyond just say the initial alert and modal use cases), it could be refactored slightly. e.g., perhaps long-term this file can be broken up in terms in `tokens/style-dictionary/formats.js`,  `tokens/style-dictionary/transforms.js`,  `tokens/style-dictionary/index.js`, etc. to better modularize the entire file.

src/Button/_mixins.scss
adamstankiewicz: Opted to just keep this change as part of the same contribution vs. splitting out to standalone PR.

lib/build-tokens.js
adamstankiewicz: Pushed up commit (5bb5a2a34d0e76ffdd72ecac3f42544ae70c5716) to implement the above `BUTTON_VARIANT_OVERRIDES_CONFIG` approach, where button variant overrides across components will be co-located across all relevant components (`overrides/component-button-variants.css`) vs. having a standalone CSS override file output per component.

tokens/style-dictionary.js
adamstankiewicz: [inform] Pushed up commit (5bb5a2a34d0e76ffdd72ecac3f42544ae70c5716) to move the `generateButtonVariantProperties` helper into the utils.js file, per the above feedback.

brian-smith-tcril (approved):
I'm really happy with how elegant of a solution this became!

Let's land this!

PR #3603, opened 2025-05-22
feat(cli): add `exclude-core` flag

brian-smith-tcril:
## Description

Add a new `--exclude-core` flag for the `build-tokens` CLI command. When using this flag, core tokens are not transformed - only theme tokens are.

### DIFF
--- a/lib/__tests__/build-tokens.test.js
+++ b/lib/__tests__/build-tokens.test.js
@@ -63,6 +63,25 @@ describe('buildTokensCommand', () => {
     }));
   });
 
+  it('should handle the exclude-core flag', async () => {
+    await buildTokensCommand(['--exclude-core']);
+
+    // only light theme, not core
+    expect(StyleDictionary).toHaveBeenCalledTimes(1);
+    const callArgs = StyleDictionary.mock.calls[0][0];
+    for (const file of callArgs.platforms.css.files) {
+      expect(file.destination).toContain('themes/light');
+    }
+
+    // Verify only light theme index files are created
+    expect(createIndexCssFile).toHaveBeenCalledTimes(1);
+    expect(createIndexCssFile).toHaveBeenCalledWith(expect.objectContaining({
+      buildDir: defaultBuildDir,
+      isThemeVariant: true,
+      themeVariant: 'light',
+    }));
+  });
+
   it('should use custom build path', async () => {
     const customBuildDir = './custom-build/';
     await buildTokensCommand(['--build-dir', customBuildDir]);
--- a/lib/build-tokens.js
+++ b/lib/build-tokens.js
@@ -22,6 +22,7 @@ async function buildTokensCommand(commandArgs) {
     'build-dir': './build/',
     'source-tokens-only': false,
     'output-references': true,
+    'exclude-core': false,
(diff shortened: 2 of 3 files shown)


lib/build-tokens.js
```diff
@@ -146,7 +148,7 @@ async function buildTokensCommand(commandArgs) {
 
   // Create list of style-dictionary configurations to build (core + theme variants)
   const configs = [
-    { config: coreConfig },
+    ...((excludeCore ? [] : [{ config: coreConfig }])),
```
brian-smith-tcril: needed to use double parens here to get past an error:
```
An error occurred: TypeError: (intermediate value)(intermediate value)(intermediate value) is not iterable 
```

adamstankiewicz: [nit/consideration] Wondering if it might be worth refactoring slightly to mutate the `configs` array vs falling back to `[]`, etc. here. For example, something like:

```js
const configs = [];
if (!excludeCore) {
  configs.push({ config: coreConfig }).
}
parsedThemes.forEach((themeVariant) => {
  const config = getStyleDictionaryConfig(themeVariant);
  configs.push({ config, themeVariant });
});
```

No strong feelings, just wondering if this might be preferable over spreading arrays.

lib/__tests__/build-tokens.test.js
```diff
@@ -63,6 +63,21 @@ describe('buildTokensCommand', () => {
     }));
   });
 
+  it('should handle the exclude-core flag', async () => {
+    await buildTokensCommand(['--exclude-core']);
+
+    // only light theme, not core
+    expect(StyleDictionary).toHaveBeenCalledTimes(1);
```
adamstankiewicz, line 70: nit: how much effort might it be to have an assertion to verify it was explicitly the light theme vs relying solely on the count?

adamstankiewicz (approved):
LGTM, just 2 minor non-blocking nits/considerations.

lib/__tests__/build-tokens.test.js
brian-smith-tcril: addressed in https://github.com/openedx/paragon/compare/9bdb1ce8a3c783f77e301bd73f916785d95522d1..503833f68d2b0e9c52b9522e5b08e6e53fa9fa15

adamstankiewicz: nice, thanks for updating!

lib/build-tokens.js
brian-smith-tcril: I also don't have strong feelings about it. I do think the `push` flow is easier to follow, but I also like that the spread flow can be done by changing only 1 line.

adamstankiewicz: Discussed in Slack async, but for posterity: I would slightly lean towards the `push` flow for future maintainability.

brian-smith-tcril: Addressed in https://github.com/openedx/paragon/compare/503833f68d2b0e9c52b9522e5b08e6e53fa9fa15..369c03068ac4e226b2390460d63651a7905bb05f

adamstankiewicz (approved):

PR #3606, opened 2025-05-22
feat(cli): add `excludeCore` flag for `build-scss`

brian-smith-tcril:
## Description

Add a new `--excludeCore` flag for the `build-scss` CLI command. When using this flag, core css is not generated - only theme css is.


adamstankiewicz (approved):
~LGMT~ LGTM! :shipit:

PR #3629, opened 2025-06-04
fix: ensure Style Dictionary format css/component-button-variant-overrides work for brand packages

adamstankiewicz:
## Description

When trying to use https://github.com/openedx/paragon/pull/3598 within `@edx/elm-theme` to generate the overrides CSS file, no output was seen.

This issue was due to the `filter` used on the `build-tokens.js` script when configuring the `css/component-button-variant-overrides` format, which conditionally filters for source tokens only (i.e., the brand package tokens only, not core Paragon tokens).

When this format keys into `dictionary.tokens` with this filter applied, the relevant JSON for the `Alert` actions button variant overrides is not found.

Removing the `filter` from the format usage in `build-tokens.js` resolves the issue; alternatively, we could keep the `filter` and rely on `dictionary.unfilteredTokens`, too. That said, if we need the unfiltered tokens for this use case, we might as well just remove the `filter` from the format usage in `build-tokens.js`.

The `filter` for this custom format handling the button variant overrides isn't really necessary, given Paragon itself does not define any key/value mapping of button variant overrides (i.e., empty object). Given this, the output overrides CSS file when running the `build-tokens` CLI command is still only generated when the brand package explicitly provided a key/value mapping of button variant overrides.

### Deploy Preview

N/A

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/lib/build-tokens.js
+++ b/lib/build-tokens.js
@@ -144,10 +144,6 @@ async function buildTokensCommand(commandArgs) {
           {
             format: 'css/component-button-variant-overrides',
             destination: `themes/${themeVariant}/overrides/component-button-variants.css`,
-            filter: transformSourceTokensOnly ? 'isSource' : undefined,
-            options: {
-              outputReferences,
-            },
           },
         ],
       },


brian-smith-tcril (approved):
Based on the detailed PR description this makes sense to me!

:shipit:

PR #3650, opened 2025-06-14
fix: add missing forwardRefs

arbrandes:
A couple of components were wrapped with `React.forwardRef()` but never actually used the `ref`, making the function call not only pointless but causing noise in console logs:

```
console.error
  Warning: forwardRef render functions accept exactly two parameters: props and ref. Did you forget to use the ref parameter?
```


brian-smith-tcril (approved):
Great catch! Looks like The `CheckpointActionRow` one went away in https://github.com/openedx/paragon/pull/3572, and the `CheckpointHeader` component was new in that PR

PR #3663, opened 2025-06-19
feat: update theme switcher UX to use external CSS URLs

### LINKED ISSUE #1468: [discovery] Paragon docs site theme switcher improvements

adamstankiewicz:
## Description

The current theme switcher on the Paragon docs site only allows toggling between the default Open edX theme and a 2U/edX.org specific theme called Elm. As a result, this current approach to previewing a custom brand on the documentation site is not accessible outside of 2U/edX.org, limiting the utility of the documentation website for the broader Open edX community.

As previously discussed in several Paragon Working Groups, we'd like to remove the hardcoded usage of a 2U/edX.org brand package (`@edx/elm-theme`) in favor of allowing users to add a custom theme, including:
1. Theme name. Displayed in the menu sidebar as a status indicator, e.g. "Current theme: Elm".
2. Input one or more CSS URL(s) from your custom theme packages (e.g., `core.css`, `light.css`).

These settings will persist:
* Encoded `?themes` query parameter using `lz-string` compression and a shorthand object notation.
* `localStorage` for repeat visits without the `?themes` query parameter.

The `onRouteUpdate` callback function in `gatsby-browser.js` was updated to persist the `?themes` query parameter as the user navigates across the documentation site. This behavior ensures any currently viewed page includes the custom theme configuration if/when the URL is copied/shared.

#### v2

https://github.com/user-attachments/assets/6ccb1122-bc1b-4bbb-b9a0-a2751c27f71e

image

image

image

image

image

image

image

image

image

---

#### v1

image

image

image

image

image

https://github.com/user-attachments/assets/85135de9-968d-4aa4-b4aa-bfbe7cc521cf

### Deploy Preview

https://deploy-preview-3663--paragon-openedx-v23.netlify.app/components/button/

#### With `@edx/elm-theme`

https://deploy-preview-3663--paragon-openedx-v23.netlify.app/components/button/?themes=N4IgLiBcDaoHZRAeQA4FM4AI0BMAamAFACJoBmAhgK4A2YAlCADQhVTQC6Avk-IgKI0AtkVx4AdAHsATgHNGLNjBAALMGBQBnSAHodAYxxxxAK0040NAJYA3aeLhowOuCiE6AArgAeOy0IBaMBU0ITQPAEZxCIAGaJ0cK01nfRk0cX1NTWZVdS1dAyNTc0tbe0dnV3cvHF9-IJCwyOi4iISk52tZNQyskG4OFisoCK4gA

## Merge Checklist

## Post-merge Checklist


www/src/components/CustomThemesForm.tsx
```diff
@@ -0,0 +1,190 @@
...
+            required: 'Theme name is required.',
+            validate: (value) => value.trim().length > 0 || 'Theme name is required.',
+          })}
+        />
+        {errors.name && (
+          <Form.Control.Feedback type="invalid">
+            {errors.name.message}
+          </Form.Control.Feedback>
+        )}
+      </Form.Group>
```
brian-smith-tcril: Non-blocking question: Do we want name to be required here? Would it make sense to default to "Custom" instead?

www/src/components/CustomThemesForm.tsx
```diff
@@ -0,0 +1,190 @@
...
+          </Stack>
+          {errors.urls?.[idx] && (
+            <Form.Control.Feedback type="invalid">
+              {errors.urls[idx]?.url?.message}
+            </Form.Control.Feedback>
+          )}
+        </Form.Group>
+      ))}
+      <Button
+        variant="link"
```
brian-smith-tcril: Using the `link` variant here feels a little odd to me, specifically because of the underlined text when hovering (even when not hovering over the text)

image

www/src/components/CustomThemesForm.tsx
```diff
@@ -0,0 +1,190 @@
...
+        )}
+      </Form.Group>
+      {fields.map((field, idx) => (
+        <Form.Group
+          key={field.id}
+          controlId={`customThemeUrl${idx}`}
+          isInvalid={!!errors.urls?.[idx]}
+        >
+          {idx === 0 && (
+            <Form.Label>CSS URL</Form.Label>
```
brian-smith-tcril: Maybe

```suggestion
            <Form.Label>CSS URL(s)</Form.Label>
```

www/src/components/CustomThemesForm.tsx
```diff
@@ -0,0 +1,190 @@
...
+    }
+  };
+
+  useImperativeHandle(ref, () => ({
+    submitForm,
+  }));
+
+  const handleAddUrl = () => {
+    append({ url: '' });
+  };
```
brian-smith-tcril: It seems like something odd is happening when adding URLs without filling in fields


https://github.com/user-attachments/assets/6c3b03a3-3d82-45bb-b5d0-bbb3d6fc9925

www/src/components/ThemeSelector.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    <div>
+      <Stack gap={1}>
+        <div role="status">
+          <div className="small">Current theme:</div>
+          <div>
+            <Badge variant="primary">
+              {currentTheme?.name || 'Open edX'}
+            </Badge>
+          </div>
```
brian-smith-tcril: Couple things here
_____

It feels a little odd having the badge on a separate line from the "Current Theme:" text

image

vs switching both to `span` and adding `pr-1.5` to the current theme text element (vertical alignment doesn't feel perfect here, just a quick "I wonder what it would look like if..." screenshot)

image

_____

It'd be nice to indicate that the Open edX theme is the built-in default in some way

_____

Possible future idea: maybe we could switch this to a `Form.RadioSet` so going back to see the default theme wouldn't require clearing out the fields in the add custom theme modal (or opening a new tab which feels like a perfectly reasonable workaround - and is why I'm considering this suggestion a "possible future" one and non-blocking)

www/src/context/SettingsContext.tsx
```diff
@@ -1,21 +1,24 @@
...
+import { useDirection } from '../hooks/useDirection';
+import { useSettingsUI } from '../hooks/useSettingsUI';
 
 export interface IDefaultValue {
   settings: {
-    theme?: string,
     direction?: string,
     language?: string,
-    containerWidth?: ContainerSize,
+    containerWidth?: any,
```
brian-smith-tcril: Just curious as to why this change is in here, it doesn't feel directly related to the theme picker.

www/src/context/SettingsContext.tsx
```diff
@@ -75,29 +65,15 @@ function SettingsContextProvider({ children }) {
...
-          <link
-            key={id}
-            href={`/static/${stylesheet}.css`}
-            rel={`stylesheet${settings.theme === id ? '' : ' alternate'}`}
-            type="text/css"
-          />
-        ))}
-      </Helmet>
-      <IntlProvider messages={messages[settings.language]} locale={settings.language.split('-')[0]}>
+      <IntlProvider messages={messages[settings.language || 'en']} locale={settings.language?.split('-')[0] || 'en'}>
```
brian-smith-tcril, line 84: This change also feels unrelated to the theme picker, but does feel like a reasonable one. Any ideas as to why this hasn't caused issues in the past?

brian-smith-tcril (commented):
Overall this is looking really great! It also seems to be working as expected too which is awesome!

I'm very happy with the functionality, and I didn't find anywhere to leave specific notes there. I did leave a few comments about the UI elements being used for controlling the custom themes.

Thanks for doing this! Super excited to see the power of runtime theming on display here!

www/src/components/ThemeSelector.tsx
adamstankiewicz: ---

I agree, the display in the Menu here can definitely be improved. For context on the separate line from the "Current theme:" text, I figured that might be preferable to better handle if/when users use longer names for their custom themes. If we make it a single line between the text and the badge, we might want to have validation on max character count for the theme name so we don't risk having the custom theme name get into any weird style issues.

image

That said, I suppose we might want that max character count validation anyways because the same problem theoretically exists anyways even with the badge on its own line! What do you think?

---

> It'd be nice to indicate that the Open edX theme is the built-in default in some way

Maybe the default theme name is "Open edX (Default)"? E.g.:

image

---

> Possible future idea: maybe we could switch this to a Form.RadioSet so going back to see the default theme wouldn't require clearing out the fields in the add custom theme modal (or opening a new tab which feels like a perfectly reasonable workaround - and is why I'm considering this suggestion a "possible future" one and non-blocking)

Yeah, I agree it would still be nice to be able to switch between Open edX and your custom theme(s) without needing to "Reset to Default" in the modal every time you want to compare. Keeping such a switcher would be the foundation for the UX to support configuring more than 1 custom theme, as well. I'm going to explore the `Form.RadioSet` approach for switching between "Open edX (Default)" vs. "Custom Theme" as part of this PR. The lift to refactor shouldn't be too bad.

www/src/context/SettingsContext.tsx
adamstankiewicz: D'oh, good catch. I will remove this change! Not sure how it made it here.

www/src/components/CustomThemesForm.tsx
adamstankiewicz: Yeah, I'm OK with making name optional, and defaulting to "Custom". In the case of future support for adding multiple custom themes at once, I'm wondering if we might increment a number like "Custom 1", "Custom 2"?

www/src/components/ThemeSelector.tsx
brian-smith-tcril: Sounds like going the `Form.RadioSet` route on this PR will address everything in one fell swoop. If that ends up being too big of a lift then:
* Changing the default name from "Open edX" to "Open edX (Default)" sounds good to me!
* The long name problem definitely pushes the badge component out of its comfort zone - I know truncate is deprecated but if radio set ends up blowing scope it might make sense here as a temp solution

www/src/components/CustomThemesForm.tsx
adamstankiewicz: Ha, had a similar thought. That said, this label is specific to a single CSS URL input (not the entire group of URL inputs). When there is more than 1 input, this `Form.Label` is no longer displayed in favor of the `aria-label={idx === 0 ? undefined : 'Additional CSS URL'}` below.

To keep the semantics of the labels, but to have the visually displayed descriptor appear to relevant to all the inputs, I might propose I no longer treat the "CSS URL(s)" descriptor as an actual form label and have all the inputs use the `aria-label` approach. Then, the "CSS URL(s)" is more of a descriptor for the entire group vs. a specific input label.

www/src/components/CustomThemesForm.tsx
brian-smith-tcril: > increment a number like "Custom 1", "Custom 2"?

My thoughts exactly!

www/src/components/CustomThemesForm.tsx
```diff
@@ -0,0 +1,190 @@
...
+          )}
+          <Stack direction="horizontal" gap={1}>
+            <Controller
+              name={`urls.${idx}.url`}
+              control={control}
+              rules={{
+                required: 'CSS URL is required.',
+                validate: (value) => {
+                  if (!value || value.trim().length === 0) {
+                    return 'CSS URL is required?!?!.';
```
adamstankiewicz: [inform, note to self] The `?!?!` here needs to be removed; it was there only for testing.

www/src/context/SettingsContext.tsx
adamstankiewicz: My IDE is throwing a TypeScript error on the `settings.language.split` as `settings.language` might be `undefined`. As for why this hasn't caused issues in the past, I'm not entirely sure. I haven't been able to identify exactly why and opted to just resolve for this PR. It also fails `npm run type-check` for me locally, but the CI on latest Paragon PRs doesn't seem to be failing.

image

brian-smith-tcril: Cool! Thanks for the context!

www/src/components/CustomThemesForm.tsx
brian-smith-tcril: That sounds like a good plan. It would likely remove the need for some of the logic checking for `idx` being `0` throughout the map too.

www/src/components/CustomThemesForm.tsx
adamstankiewicz: This behavior is technically "as designed" in the current state. When you click "Add another URL", it moves focus into the newly rendered input. Then, when you click "Add another URL" again, the input field has been "touched" (blurred) so `react-hook-form` runs its validation to show the "CSS URL is required" message.

I think the UX improvement here would be to prevent clicking "Add another URL" (disabled button) if you there's have an empty input displayed and available to use.

brian-smith-tcril: The parts that felt odd to me:
1. "CSS URL is required." not showing up on the first input
2. Theme Name validation happening when the "Add another URL" button is clicked

> I think the UX improvement here would be to prevent clicking "Add another URL" (disabled button) if you there's have an empty input displayed and available to use.

That seems like a great solution. I think it would address 1 perfectly.

I do still think it's a bit odd that adding a URL triggers Theme Name validation, but if the plan is to move to have Custom as default instead of making it a required field then it'll be a non-issue.

www/src/components/CustomThemesForm.tsx
adamstankiewicz: I agree, if it's not acting as a hyperlink, it probably shouldn't be styled as a hyperlink (i.e., it doesn't navigate to a new page). I've updated it locally to use `variant="tertiary"`, and will be included in my next commit:

Default:

image

www/src/components/CustomThemesForm.tsx
adamstankiewicz: Will be updated in my next commit, but here's where I landed on this change visually:

image

Renders a display-only `<Form.Label>` for "CSS URL(s)" and then each of the url inputs define `aria-label="CSS URL"` as a label for the singular url input.

Clicking the `<Form.Label>` will still put focus into the first url input.

www/src/components/ThemeSelector.tsx
adamstankiewicz: I ended up adding some additional UX polish to clean this feature up further and not need to worry about truncation. Now, by default, the theme selector will be in view-only state to depict the current theme. Clicking "Edit" discloses the `Form.RadioSet` to switch between the default and custom themes, add a custom theme, or reset custom themes. Once done editing themes, the theme selector may be closed to transition back to the view-only state.

image<br>

image<br>

image

image<br>

image<br>

image<br>

https://github.com/user-attachments/assets/03ad098f-9425-4579-bb86-4730026fa824

www/src/components/CustomThemesForm.tsx
adamstankiewicz: [update] I changed this once more locally; ultimately settled on `outline-primary` :)

www/src/components/CustomThemesForm.tsx
adamstankiewicz: [update] Ultimately landed on using `outline-primary`:

image

www/src/components/CustomThemesForm.tsx
adamstankiewicz: Updated in latest commits!

www/src/components/ThemeSelector.tsx
```diff
@@ -0,0 +1,180 @@
...
+            alt="Finish editing"
+            size="sm"
+            onClick={toggleEditMode}
+            variant="secondary"
+          />
+        ) : (
+          <Button ref={editButtonRef} size="sm" variant="link" onClick={toggleEditMode}>
+            Edit
+          </Button>
+        )}
```
brian-smith-tcril: Using "Edit" and :x: for expanding/collapsing the radio set feels a little odd to me. 

https://github.com/user-attachments/assets/4da6da47-67e1-4ef7-b218-2d9fe27c16be

We use :arrow_down_small: and :arrow_up_small: in other places 

https://github.com/user-attachments/assets/d5b37748-f0e8-4c0a-b860-d9116cc3efc4

Would just using a [basic style `Collapsible`](https://paragon-openedx.netlify.app/components/collapsible/#basic-style) here work?

brian-smith-tcril (commented):
I haven't looked at all of the code changes since the new UX was implemented but I checked out the deploy preview and it's great!

I left one comment in there, and I noticed a couple more UX things
____

It's impossible to click edit at mobile widths

image

I'm not sure the best way to address this, and I can look into it more later, I just wanted to make sure the issue was noted for now.  
_____

The radioset doesn't stay expanded when navigating

https://github.com/user-attachments/assets/a7498a1f-6760-4ef5-8a56-41d0c93f2ee0

I wouldn't consider this blocking. We already lose some of the expanded/collapsed state of the menu when navigating 

https://github.com/user-attachments/assets/9b1444e3-d0a0-48f8-a30b-244b64a10eab
_____

### Overall though this is awesome! It's super cool to see the multi-custom-theme functionality in action!

www/src/components/ThemeSelector.tsx
adamstankiewicz: Yes, it will! Updated in most recent commit.

brian-smith-tcril (approved):
LGTM!

PR #3706, opened 2025-07-23
feat: add label colors

brian-smith-tcril:
> [!NOTE]
> ### This will stay in draft until a final decision on what colors to add is made.

## Description

This is a continuation of https://github.com/openedx/paragon/pull/3703

@PKulkoRaccoonGang I've requested your review on this because I'm wondering if you have any ideas on how to do this with a bit less copy/pasting. I was also originally hoping to be able to add the label colors using nesting, so instead of
```json
    "label-a": {
      "100": {
        "$value": "#F9E6E6",
        "$description": "Label color A of level 100.",
        [...]
```
we could have
```json
    "label": {
      "a": {
        "100": {
          "$value": "#F9E6E6",
          "$description": "Label color A of level 100.",
          [...]
```
but I ran into issues with some colors using item for level and others using sub-item, so any thoughts on how possible that'd be would be great!

@kblemel for some reason it's not letting me request a review from you so I'm just tagging you here in the description.

### Deploy Preview

https://deploy-preview-3706--paragon-openedx-v23.netlify.app/foundations/colors/

### DIFF
--- a/styles/css/themes/light/utility-classes.css
+++ b/styles/css/themes/light/utility-classes.css
@@ -2451,3 +2451,2163 @@ a.text-dark:focus {
   border-color: var(--pgn-color-dark-base) !important;
 }
 
+.bg-label-a-100 {
+  background-color: var(--pgn-color-label-a-100) !important;
+}
+
+a.bg-label-a-100:hover,
+a.bg-label-a-100:focus,
+button.bg-label-a-100:hover,
+button.bg-label-a-100:focus {
+  background-color: var(--pgn-color-action-default-label-a-100) !important;
+}
+
+.text-label-a-100 {
+  color: var(--pgn-color-label-a-100) !important;
+}
+
+a.text-label-a-100:hover,
+a.text-label-a-100:focus {
+  color: var(--pgn-color-action-default-label-a-100) !important;
+}
+
+.border-label-a-100 {
+  border-color: var(--pgn-color-label-a-100) !important;
+}
+
+.bg-label-a-200 {
+  background-color: var(--pgn-color-label-a-200) !important;
+}
+
+a.bg-label-a-200:hover,
+a.bg-label-a-200:focus,
+button.bg-label-a-200:hover,
+button.bg-label-a-200:focus {
+  background-color: var(--pgn-color-action-default-label-a-200) !important;
+}
+
+.text-label-a-200 {
+  color: var(--pgn-color-label-a-200) !important;
+}
+
+a.text-label-a-200:hover,
+a.text-label-a-200:focus {
+  color: var(--pgn-color-action-default-label-a-200) !important;
+}
+
+.border-label-a-200 {
+  border-color: var(--pgn-color-label-a-200) !important;
+}
+
+.bg-label-a-300 {
+  background-color: var(--pgn-color-label-a-300) !important;
+}
+
+a.bg-label-a-300:hover,
+a.bg-label-a-300:focus,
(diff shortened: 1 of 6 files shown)


peterkulko:
> but I ran into issues with some colors using item for level and others using sub-item, so any thoughts on how possible that'd be would be great!

It seems to me that we can try modifying the [filters](https://github.com/openedx/paragon/blob/release-23.x/tokens/style-dictionary.js#L284) in style-dictionary. Something like this (this is just my guess):

```
...

tokens = tokens.filter((token) => {
  if (attributeName === 'item' && token.attributes.type === 'label' && token.attributes.subitem) {
    retrn allowedValues.includes(token.attributes[attributeName]);
  }
  if (attributeName === 'subitem' && token.attributes.type === 'label') {
    return allowedValues.includes(token.attributes.subitem);
  }
  return allowedValues.includes(token.attributes[attributeName]);
})
...
```

But I think that such modifications would be breaking changes. Please correct me if I'm wrong.

> I was also originally hoping to be able to add the label colors using nesting

I'm not sure we should support nesting for something like `label-a`. We already have semantic types like `warning`, `danger`, `info` etc. By adding new semantic types we should continue the naming sequence, despite the name of the new types (`label-a`). I see that we have the type `accent`, but in the case of this type the item is its "a, b" for our new types `label-*` items and sub-items are different

image

I mean, for me, the duplication of `label-*` is expected since it is a type name and I don't think we should complicate the internal logic of the dictionary because of it.

Another idea. Maybe we should go back to thinking about naming the color palette we are adding. We already have the names `secondary`, `danger`, `info` associated with colors. The names `label-a` or `label-b` are not semantically associated with any color. Despite the meaning of these colors, it is difficult to understand what color `label-a` is, it is too abstract.

I suggest considering something like:

- label-a -> rose / blush
- label-b -> spring
- label-c -> azure / sky
- label-d -> amber / sand
- label-e -> cyan
- label-f -> magenta
- label-g -> mint
- label-h -> violet / purple
- label-i -> lime / fresh

I found some of these ideas for color names in [Material UI](https://mui.com/material-ui/customization/color/#color-palette)

Or if we have critically have neutral names `label-*` we can consider the creation of constant colors (`cyan`, `magenta`, `mint`...) and raising the level of abstraction to `label-*` which will contain constant colors. For example, how is it done in [alias/colors](https://github.com/openedx/paragon/blob/release-23.x/tokens/src/themes/light/alias/color.json).

```
    "active": {
      "source": "$component-active-color",
      "$value": "{color.white}", <- constant color white
      "$description": "Color for active element." <- any color is not tied to any semantic meaning
    }
```

@brian-smith-tcril, @edschema I would like to know what you think about my ideas 💯

brian-smith-tcril:
> I'm not sure we should support nesting for something like `label-a`.

Yeah, the more I think about it the less sense trying to support nesting for label colors makes sense.

 > The names label-a or label-b are not semantically associated with any color.
 
 This is intentional. When someone is using a label color in an MFE, we don't want them to think that it will always be a certain color. We also want theme authors to feel free to change the label colors to any other set of colors (with guidance around ensuring their set is accessible)
 
 > Or if we have critically have neutral names label-* we can consider the creation of constant colors (cyan, magenta, mint...) and raising the level of abstraction to label-* which will contain constant colors.
 
 I really like this plan. I'd like to make sure a final decision is made around how many label colors we need and which ones we're keeping first, then update this to have that extra layer of abstraction.

PR #3742, opened 2025-08-13
Remove `propTypes` + `defaultProps` from `<Button>`, improve typing

bradenmacdonald:
## Description

Like in #3738, this PR removes deprecated `propTypes` and `defaultProps` from `<Button>` and related components. It also makes the TypeScript definitions slightly more strict (e.g. requires a valid ARIA role for the `role` attribute, and a valid variant for the `variant` attribute). It also corrects the documentation to mention the missing (but valid) `tertiary` variant, and that you can include arbitrary HTML props.

I've actually _removed_ some props documentation in this case, e.g. `onClick` is no longer listed as an explicit prop, because it too difficult to give it the correct type (the detailed event type depends on whether or not users override the element type with `as`). So instead I omitted props like `onClick` and `role` from our explicit list - this means that they'll always have the right type (based on the `as` prop) in user's IDE and when type checking, but they're no longer documented explicitly on the docs page. To compensate, I put this informative message at the top of the props documentation:

Screenshot 2025-08-14 at 2 59 00 PM

### Deploy Preview

* `<Button>`: [Before](https://paragon-openedx.netlify.app/components/button/) vs [After](https://deploy-preview-3742--paragon-openedx-v23.netlify.app/components/button/)
* `<ButtonGroup>`: [Before](https://paragon-openedx.netlify.app/components/button/button-group/) vs [After](https://deploy-preview-3742--paragon-openedx-v23.netlify.app/components/button/button-group/)
* `<ButtonToolbar>`: [Before](https://paragon-openedx.netlify.app/components/button/button-group/#buttontoolbar) vs [After](https://deploy-preview-3742--paragon-openedx-v23.netlify.app/components/button/button-group/#buttontoolbar)

## Merge Checklist

n/a

### DIFF
--- a/src/Button/index.tsx
+++ b/src/Button/index.tsx
@@ -1,14 +1,30 @@
 import React from 'react';
-import PropTypes, { type Requireable } from 'prop-types';
 import classNames from 'classnames';
 import BaseButton, { type ButtonProps as BaseButtonProps } from 'react-bootstrap/Button';
 import BaseButtonGroup, { type ButtonGroupProps as BaseButtonGroupProps } from 'react-bootstrap/ButtonGroup';
-import BaseButtonToolbar, { type ButtonToolbarProps } from 'react-bootstrap/ButtonToolbar';
+import BaseButtonToolbar, { type ButtonToolbarProps as BaseButtonToolbarProps } from 'react-bootstrap/ButtonToolbar';
 import type { ComponentWithAsProp } from '../utils/types/bootstrap';
 
 import Icon from '../Icon';
 
-interface ButtonProps extends Omit<BaseButtonProps, 'size'> {
+type BaseVariant = (
+  | 'primary'
+  | 'secondary'
+  | 'tertiary'
+  | 'brand'
+  | 'success'
+  | 'danger'
+  | 'warning'
+  | 'info'
+  | 'dark'
+  | 'light'
+  | 'link'
+);
+
+export interface ButtonProps extends Omit<BaseButtonProps, 'size'> {
+  /** Set a custom element for this component (default: `button`, with `type="button"`). */
+  as?: React.ElementType;
+  size?: 'sm' | 'md' | 'lg' | 'inline';
   /**
    * An icon component to render. Example:
    * ```
@@ -25,18 +41,32 @@ interface ButtonProps extends Omit<BaseButtonProps, 'size'> {
    * ```
    */
   iconAfter?: React.ComponentType;
-  size?: 'sm' | 'md' | 'lg' | 'inline';
+
+  // The following are the same as in BaseButtonProps, but we re-define them to add documentation.
+  // The upstream type defintions do not have any comments/docs.
+
+  /** Disables the Button, preventing mouse events, even if the underlying component is an `<a>` element */
+  disabled?: boolean;
+  /** Optional: Specify additional class name(s) to apply to the button */
+  className?: string;
+  /** Specifies the text that is displayed within the button. */
+  children: React.ReactNode;
+  /** Specifies variant to use.
+   * Can be one of the base variants: `primary`, `secondary`, `tertiary`, `brand`, `success`, `danger`, `warning`,
+   * `info`, `dark`, `light`, `link`,
+   * as well as one of the customized variants (= base variant prefixed with `inverse-`, `outline-`
+   * or `inverse-outline-`)
+   * */
+  variant?: BaseVariant | `inverse-${BaseVariant}` | `outline-${BaseVariant}` | `inverse-outline-${BaseVariant}`;
 }
 
-type ButtonType = ComponentWithAsProp<'button', ButtonProps> & { Deprecated?: any };
-
-const Button: ButtonType = React.forwardRef<HTMLButtonElement, ButtonProps>(({
+const Button: ComponentWithAsProp<'button', ButtonProps> = React.forwardRef(({
   children,
   iconAfter,
   iconBefore,
   size,
   ...props
-}, ref) => (
+}: ButtonProps, ref: React.ForwardedRef<HTMLDivElement>) => (
   <BaseButton
     size={size as 'sm' | 'lg' | undefined} // Bootstrap's <Button> types do not allow 'md' or 'inline', but we do.
     {...props}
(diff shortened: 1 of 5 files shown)


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

www/src/components/PropsTable.tsx
```diff
@@ -100,10 +100,13 @@ function PropsTable({ props: componentProps, displayName, content }: IPropsTable
...
         <p className="px-4 small mb-3">
-          This is a pass through component from React-Bootstrap, see original props documentation{' '}
+          This is a pass through component from React-Bootstrap; see original props documentation{' '}
           <Hyperlink destination={bootstrapLink}>
             here.
-          </Hyperlink>
+          </Hyperlink>{' '}
+          You can also pass any HTML attributes, such as <code>onClick</code>, <code>role</code>, etc. Which attributes
+          are available depends on whether or not you have overriden the underlying HTML element using the{' '}
+          <code>as</code> property.
```
bradenmacdonald, line 109: As far as I can tell, this does apply to _all_ the passthrough components coming from Bootstrap, not just `Button`, so I think it makes sense to put here.

bradenmacdonald:
@brian-smith-tcril @PKulkoRaccoonGang would either of you like to review this?

Also, I notice the PR bot says 

> This repository is currently maintained by @openedx/paragon-working-group.

But that group doesn't exist. Should it be @openedx/committers-paragon ?

brian-smith-tcril:
> Also, I notice the PR bot says
> 
> > This repository is currently maintained by @openedx/paragon-working-group.
> 
> But that group doesn't exist. Should it be @openedx/committers-paragon ?

That sounds perfect! If you could make a PR updating catalog info I'll approve/merge it ASAP!

brian-smith-tcril:
Overall this looks awesome, the only part that feels slightly concerning to me is

> It also makes the TypeScript definitions slightly more strict (e.g. requires a valid ARIA role for the role attribute)

as that is technically a breaking change. I 100% agree that consumers *should* be using valid ARIA roles, but in the case a consumer isn't I'd prefer to not have renovate break a build.

Maybe for `release-23.x` we could use `React.AriaRole | string` and then remove the `| string` in `next`? I'm very open to other thoughts/ideas on how to handle this.

bradenmacdonald:
Ah, it turns out `React.AriaRole` has a clever solution for this - it already allows `| (string & {})` which means that it accepts any string but it only shows the valid roles in the dropdown. So it won't break anything. (But on the other hand it won't catch errors either.)

What about `variant` though? I did also add stricter typing for that, and I know of at least one MFE with an invalid Button variant set (`muted`). Should I add the same `| (string & {})` for variant so we accept any variant without ever flagging errors, but still show a dropdown of valid values in the IDE, or should we actually flag invalid variants?

brian-smith-tcril:
> What about `variant` though? I did also add stricter typing for that, and I know of at least one MFE with an invalid Button variant set (`muted`). Should I add the same `| (string & {})` for variant so we accept any variant without ever flagging errors, but still show a dropdown of valid values in the IDE, or should we actually flag invalid variants?

I think for `release-23.x` adding `| (string & {})` makes the most sense to me. That keeps us squarely in non-breaking territory.

We could even do something like
```tsx
type BaseVariant = (
  | 'primary'
  | 'secondary'
  | 'tertiary'
  | 'brand'
  | 'success'
  | 'danger'
  | 'warning'
  | 'info'
  | 'dark'
  | 'light'
  | 'link'
);

type Variant = BaseVariant | `inverse-${BaseVariant}` | `outline-${BaseVariant}` | `inverse-outline-${BaseVariant}`;

// This was added so these types could be added as a non-breaking change, it should be removed
// in the next major release, and all uses should be updated to the non-deprecated Variant.
type DeprecatedVariant = Variant | (string & {});
```

bradenmacdonald:
@brian-smith-tcril Nice suggestion, but I think it makes the docs a bit worse:

Screenshot 2025-08-18 at 3 32 37 PM

How about this?

```ts
/**
 * This was added so these types could be added as a non-breaking change, it should be removed
 * in the next major release, and all uses should be updated to the non-deprecated Variant.
 * @deprecated - remove in Paragon 24
 */
type OtherDeprecatedValue = string & {}; // Allow any other string value for now, even though it's invalid

export interface ButtonProps extends Omit<BaseButtonProps, 'size'> {
  ...
  variant?: BaseVariant | `inverse-${BaseVariant}` | `outline-${BaseVariant}` | `inverse-outline-${BaseVariant}` | OtherDeprecatedValue;
}
```

Screenshot 2025-08-18 at 3 43 10 PM

It's still not ideal, but I think it's a little more clear?  I pushed that change so you can see.

src/Button/index.tsx
```diff
@@ -1,14 +1,37 @@
...
+  | 'light'
+  | 'link'
+);
+
+/**
+ * This was added so these types could be added as a non-breaking change, it should be removed
+ * in the next major release, and all uses should be updated to the non-deprecated Variant.
+ * @deprecated - remove in Paragon 24
+ */
+type OtherDeprecatedValue = string & {}; // Allow any other string value for now, even though it's invalid
```
brian-smith-tcril:
```suggestion
/**
 * This was added so these types could be added as a non-breaking change.
 * @deprecated - remove in Paragon 24
 */
type OtherDeprecatedValue = string & {}; // Allow any other string value for now, even though it's invalid
```

brian-smith-tcril (commented):
> How about this?
> 
> ```ts
> /**
>  * This was added so these types could be added as a non-breaking change, it should be removed
>  * in the next major release, and all uses should be updated to the non-deprecated Variant.
>  * @deprecated - remove in Paragon 24
>  */
> type OtherDeprecatedValue = string & {}; // Allow any other string value for now, even though it's invalid
> 
> export interface ButtonProps extends Omit<BaseButtonProps, 'size'> {
>   ...
>   variant?: BaseVariant | `inverse-${BaseVariant}` | `outline-${BaseVariant}` | `inverse-outline-${BaseVariant}` | OtherDeprecatedValue;
> }
> ```

That looks great! I hadn't checked how my suggestion would look on the docs site, this looks much nicer!

The comment makes a little less sense with this change, I left a comment with a suggestion to simplify the comment.

bradenmacdonald:
Thanks, changed.

brian-smith-tcril (approved):
LGTM!

PR #3763, opened 2025-08-26
fix: DataTable filter "Clear all" bug fix

kiram15:
https://github.com/user-attachments/assets/550cfb6e-84e6-436b-9b7d-6ed18b9fb0dd

## Description

There was a [previous bug fix](https://github.com/openedx/paragon/pull/3639) for the filters on DataTable, but this fix introduced a [new bug](https://2u-internal.atlassian.net/browse/ENT-10858) where checkboxes were filtering correctly, but did not visually show as checked. This fix still provides correct functionality when clicking "Clear all" (all checkboxes get unchecked) but also fixes the new bug. The underlying issue was an empty array not being falsy in javascript. 

### Deploy Preview

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/DataTable/FilterStatus.jsx
+++ b/src/DataTable/FilterStatus.jsx
@@ -26,7 +26,7 @@ function FilterStatus({
         className={buttonClassName}
         variant={variant}
         size={size}
-        onClick={() => setAllFilters([])}
+        onClick={() => setAllFilters(undefined)}
       >
         {clearFiltersText === undefined
           ? (
--- a/src/DataTable/filters/CheckboxFilter.jsx
+++ b/src/DataTable/filters/CheckboxFilter.jsx
@@ -27,7 +27,7 @@ function CheckboxFilter({
   return (
     <Form.Group role="group" aria-labelledby={ariaLabel.current}>
       <FormLabel id={ariaLabel.current} className="pgn__checkbox-filter-label">{Header}</FormLabel>
-      <Form.CheckboxSet name={Header} value={checkedBoxes}>
+      <Form.CheckboxSet name={Header}>
         {filterChoices.map(({ name, number, value }) => (
           <Form.Checkbox
             key={`${headerBasedId}${name}`}
--- a/src/DataTable/tests/FilterStatus.test.jsx
+++ b/src/DataTable/tests/FilterStatus.test.jsx
@@ -50,7 +50,7 @@ describe('<FilterStatus />', () => {
     const button = screen.getByText(filterProps.clearFiltersText);
     await userEvent.click(button);
     expect(clearSpy).toHaveBeenCalledTimes(1);
-    expect(clearSpy).toHaveBeenCalledWith([]);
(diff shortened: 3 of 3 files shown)


brobro10000:
So something to consider as part of this PR after digging deeper into this implementation is how downstream consumers are utilizing the `filters` args. By setting it to `undefined` we run the risk of throwing an error if consumers assume the cleared filter state is an array when in actuality it is now `undefined`.

In this [location](https://github.com/openedx/paragon/blob/689dbf8be8cdfb8f3d8ffacfdf3b25084bedbc6f/src/DataTable/index.jsx#L173-L182), consider normalizing the return type to an array before fetching data.

``` jsx
  useEffect(() => {
    if (fetchData) {
      // Normalize filters to preserve downstream contract (always an array)
      const normalizedFilters = Array.isArray(tableStateFilters) ? tableStateFilters : [];
      fetchData({
        pageSize: tableStatePageSize,
        pageIndex: tableStatePageIndex,
        sortBy: tableStateSortBy,
        filters: normalizedFilters,
      });
    }
  }, [fetchData, tableStatePageSize, tableStatePageIndex, tableStateSortBy, tableStateFilters]);
```

Another option is to consider a secondary approach (on top of the array validation above) which utilizes the `resetFilters` prop in `FilterStatus`. This would allow the filter values to fall back to the default filters set. Here is an example output of the FilterStatus component.

``` jsx
function FilterStatus({
  className, variant, size, clearFiltersText, buttonClassName, showFilteredFields,
}) {
  const { state, setAllFilters, resetFilters, headers } = useContext(DataTableContext);
  if (!setAllFilters && !resetFilters) {
    return null;
  }

  const headersMap = headers.reduce((cur, acc) => {
    // eslint-disable-next-line no-param-reassign
    cur[acc.id] = acc.Header;
    return cur;
  }, {});
  const filterNames = state.filters ? state.filters.map((filter) => headersMap[filter.id]) : [];
  const filterTexts = <p>Filtered by {filterNames.join(', ')}</p>;

  return (
    <div className={className}>
      {showFilteredFields && filterTexts}
      <Button
        className={buttonClassName}
        variant={variant}
        size={size}
        onClick={() => {
          if (typeof resetFilters === 'function') {
            resetFilters(); // Resets all column filterValues to initial (usually clears)
          } else {
            setAllFilters([]); // Fallback to canonical clear using an empty array
          }
        }}
      >
        {clearFiltersText === undefined
          ? (
            <FormattedMessage
              id="pgn.DataTable.FilterStatus.clearFiltersText"
              defaultMessage="Clear filters"
              description="A text that appears on the `Clear filters` button"
            />
          )
          : clearFiltersText}
      </Button>
    </div>
  );
}
```

brian-smith-tcril (changes requested):
I tested this on the deploy preview and encountered an error

https://deploy-preview-3763--paragon-openedx-v22.netlify.app/components/datatable/#sidebar-filter

https://github.com/user-attachments/assets/b5f9acba-429b-4c4b-af04-a0193a92e17f

PR #3990, opened 2025-11-13
feat: add `dvh-100` and `min-dvh-100` utility classes

### LINKED ISSUE #3985: Add `min-dvh-*` utility class
`vh` doesn't account for mobile UI elements.

There are now units that can be used to address this exact problem:

* https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/length#relative_length_units_based_on_viewport 

| Unit | Description | Docs |
| - | - | - |
| `vh` | Existed before any of the others, same as `lvh` | https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/length#vh |
| `lvh` | Full viewport height, includes area covered by retractable mobile browser UI elements | https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/length#large_viewport_units |
| `svh` | Small viewport height, excludes area that can be used by retractable mobile browser UI elements | https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/length#small_viewport_units |
| `dvh` | Dynamic viewport height, if the mobile browser is showing retractable elements it's the height excluding those, if the mobile browser isn't showing them it's the full height | https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/length#dynamic_viewport_units |

Blog post with some good visuals https://web.dev/blog/viewport-units

_____

It looks like Bootstrap is [considering including these in v6](https://github.com/twbs/bootstrap/issues/41055#issuecomment-2787440420) (we're on 4), so adding this to Paragon itself feels reasonable to me

_____

It seems current best practice is to use a fallback

```css
// https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/length#dynamic_viewport_units
.min-dvh-100 {
   // fallback for browsers that don't support dynamic viewport units
   min-height: 100vh;

   // dynamic units for modern browsers
   min-height: 100dvh;
}
```

brian-smith-tcril:
## Description

Resolves https://github.com/openedx/paragon/issues/3985

### DIFF
--- a/styles/scss/core/_utilities.scss
+++ b/styles/scss/core/_utilities.scss
@@ -122,3 +122,27 @@ $indexes: map.merge(
     }
   }
 }
+
+// Dynamic viewport
+// 
+// Following the pattern of `vh-100` and `min-vh-100` from
+// https://github.com/twbs/bootstrap/blob/v4-dev/scss/utilities/_sizing.scss
+// 
+// More info on MDN
+// https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/length#dynamic_viewport_units
+
+.dvh-100 {
+   // fallback for browsers that don't support dynamic viewport units
+   height: 100vh !important;
+
+   // dynamic units for modern browsers
+   height: 100dvh !important;
+}
+
+.min-dvh-100 {
+   // fallback for browsers that don't support dynamic viewport units
+   min-height: 100vh !important;
+
+   // dynamic units for modern browsers
+   min-height: 100dvh !important;
+}


styles/scss/core/_utilities.scss
```diff
@@ -122,3 +122,27 @@ $indexes: map.merge(
...
+// More info on MDN
+// https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/length#dynamic_viewport_units
+
+.dvh-100 {
+   // fallback for browsers that don't support dynamic viewport units
+   height: 100vh !important;
+
+   // dynamic units for modern browsers
+   height: 100dvh !important;
+}
```
peterkulko, line 140: Do you think using `@supports` is a good way to avoid adding comments about browser support?

```suggestion
.dvh-100 {
   height: 100vh !important;

   @supports (height: 100dvh) {
     height: 100dvh !important;
   }
}
```

PKulkoRaccoonGang (changes requested):
LGTM! I left a comment, but it’s non-blocking for landing this PR.

styles/scss/core/_utilities.scss
brian-smith-tcril: I think it's a question of what we want the compiled CSS to look like, with your suggestion the compiled CSS would be
```css
.dvh-100 {
  height: 100vh !important;
}

@supports (height: 100dvh) {
  .dvh-100 {
    height: 100dvh !important;
  }
}
```

Which means it would minify into slightly more code

```css
.dvh-100{height:100vh!important;height:100dvh!important}
```

vs

```css
.dvh-100{height:100vh!important;}
@supports (height:100dvh){.dvh-100{height:100dvh!important;}}
```

If the issue is just having comments in there, then I'd be happy to just remove the comments and have

```suggestion
.dvh-100 {
   height: 100vh !important;
   height: 100dvh !important;
}
```

as anyone familiar with `dvh` will understand that the `vh` units are there as a fallback for browsers that don't have `dvh` support, and anyone unfamiliar would be able to intuit it after searching for `dvh`.

peterkulko: I would still like to keep at least some comments if you’re not fond of using `@supports`.

PKulkoRaccoonGang (approved):

Issue #3311, opened 2024-12-03
Truncate component breaks JSX conventions


bradenmacdonald:
## Feedback summary

The [`<Truncate>` component](https://paragon-openedx.netlify.app/components/truncate/) documentation says:

> Note: Truncate supports **only plain HTML children** and **not jsx**.

And a supposed example of this can be seen in the ["HTML markdown support" section](https://paragon-openedx.netlify.app/components/truncate/#html-markdown-support).

However, this is not true. A more accurate statement would be "Truncate supports only a buggy, idiosyncratic version of JSX."

## It is JSX, not HTML

First of all, if it was truly HTML instead of JSX, it would have to be passed into the Truncate component as a string; it is in fact being parsed by React's JSX parser. You can clearly see this if you use other JSX syntax like `{"string interpolation"}`:

![Screenshot 2024-12-03 at 10 35 15 AM](https://github.com/user-attachments/assets/f69343fb-412f-49e1-9a22-87d495e1a9b9)

## Bugs

Second, of all, it throws errors on things that are perfectly valid HTML or JSX:

![Screenshot 2024-12-03 at 10 37 43 AM](https://github.com/user-attachments/assets/9cb4d9d3-7810-44a9-bd26-89cd56923d48)

![Screenshot 2024-12-03 at 10 38 01 AM](https://github.com/user-attachments/assets/68cf3e37-f911-4a05-bcc2-22bbcbe00fc3)

## Incompatibility

Finally, because this _is_ actually JSX but it's not implemented correctly, TypeScript is always going to throw errors like this (which is how I discovered all this):

```
Type '{ children: Element; class: string; }' is not assignable to type 'DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>'.
  Property 'class' does not exist on type 'DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>'. Did you mean 'className'?
```

I don't think it is possible to tell TypeScript that some portion of the JSX is using our own special version of JSX and should be using `class` instead of `className`.

## Suggestion

At the very least, I suggest fixing the bugs and implementing support for `className` instead of `class`, `htmlFor` instead of `for`, etc. so that we can use this with TypeScript.

But now might be the time to replace the current complex JS implementation with [a simple CSS implementation](https://stackoverflow.com/a/41137262), [given that all major browsers have supported `-webkit-line-clamp` for a while](https://caniuse.com/css-line-clamp) (click "Usage Relative").

[...]

PR #4037, opened 2025-12-11
feat: truncate component rework

MaxFrank13:
## Description

Corresponding issue: #3311 

This refactors the `Truncate` component to use CSS properties instead of complex JS. The `-webkit-line-clamp` property has support for most common browsers ([source](https://caniuse.com/?search=-webkit-line-clamp)). Eventually, `line-clamp` should have support but that is not currently the case.

This was discussed in Paragon Working Group ([session notes](https://openedx.atlassian.net/wiki/x/AQBLPwE)).

Both `title` and `aria-label` are used for accessibility in the same way it was used before. The `assembleStringFromChildrenArray` utility function has been moved to a `utils.ts` file and thus converted to TypeScript.

The deprecated component as well as it's utils file still exist under the `Truncate` directory. The deprecated component is still accessible at `Truncate.Deprecated`. The `utils` file has been renamed to `utils.deprecated`. It didn't make sense to convert the entire utils file to TS when only the one utility function (`assembleStringFromChildrenArray`) was needed.

The tests for the deprecated component are under `TruncateDeprecated.test.jsx`.

```
src/
└── Truncate/
    ├── index.js                  
    ├── index.scss                
    ├── README.md                 
    ├── Truncate.test.jsx         
    ├── Truncate.tsx              
    ├── TruncateDeprecated.jsx    
    ├── TruncateDeprecated.test.jsx 
    ├── utils.deprecated.js       
    ├── utils.deprecated.test.js  
    ├── utils.test.ts             
    └── utils.ts 
```

### Deploy Preview

https://deploy-preview-4037--paragon-openedx-v23.netlify.app/components/truncate/

## Merge Checklist

## Post-merge Checklist


brian-smith-tcril:
On slack you mentioned needing to use inline styles. I wasn't able to *completely* get away from them but I was able to at least minimize it down to setting a css var

```diff
diff --git a/src/Truncate/Truncate.tsx b/src/Truncate/Truncate.tsx
index fa421b68f6..ab9dd95753 100644
--- a/src/Truncate/Truncate.tsx
+++ b/src/Truncate/Truncate.tsx
@@ -9,11 +9,6 @@ interface TruncateProps {
 }
 
 function Truncate({ children, lines = 1 }: TruncateProps) {
-  const style = {
-    lineClamp: lines,
-    WebkitLineClamp: lines,
-  };
-
   const initialText = Array.isArray(children) ? assembleStringFromChildrenArray(children) : String(children);
 
   return (
@@ -21,7 +16,7 @@ function Truncate({ children, lines = 1 }: TruncateProps) {
       title={initialText}
       aria-label={initialText}
       className="truncate-text"
-      style={style}
+      style={{ "--truncate-prop-lines": lines } as React.CSSProperties}
       data-testid="truncate-element"
     >
       {children}
diff --git a/src/Truncate/index.scss b/src/Truncate/index.scss
index a2f9fbde1f..418641008b 100644
--- a/src/Truncate/index.scss
+++ b/src/Truncate/index.scss
@@ -2,6 +2,6 @@
   overflow: hidden;
   display: -webkit-box;
   -webkit-box-orient: vertical;
-  -webkit-line-clamp: 1;
-  line-clamp: 1;
+  -webkit-line-clamp: var(--truncate-prop-lines);
+  line-clamp: var(--truncate-prop-lines);
 }
```

As for the naming of the variable, this is definitely setting a new pattern so I'd love to hear opinions. My reasoning behind `--truncate-prop-lines` is:
* I want to make sure it's clear that this CSS var isn't related to tokens (I *think* just not having the `--pgn-` prefix helps with that)
* I want someone reading the `scss` file to know where the var is coming from

MaxFrank13:
> I want to make sure it's clear that this CSS var isn't related to tokens (I think just not having the --pgn- prefix helps with that)

Yes I think this makes sense and would be a clear way of indicating that this is a local variable and not a design token.

W.r.t the [style conventions](https://github.com/openedx/paragon/blob/master/docs/decisions/0012-css-styling-conventions), it also mentions that all CSS classes should be prefixed with `pgn__`. I feel like this differs from the variable question. Should the CSS classes added in this PR have the `pgn__` prefix? My inclination is to think yes because they are classes from Paragon for a specific Paragon component, and are not [utility classes](https://paragon-openedx.netlify.app/foundations/css-utilities/?themes=N4IgLiBcDaoHZRAeQA4FM4AI0BMAamAFACJoBmAhgK4A2YAlCADQhVTQC6Avk-IgDIB7AMYUamAKI0AtkVx4AdIIBOAc0Ys2MEAAswYFJAD0RmiLE7BAZzCQAzAAYnR4SrQLhVq8137DJs1EaSxt7JwdTAEtVPQ8vEG5eEARIEH5IgDc0SRkfLWhfAytjFxw4BQArKxw0GkzlBTg0MCM4FGkjAAFcAA8jWukAWjAdNGk0IxxImxc3BWlI8s9vFj0ikuEyyura%2Bsbm1vau3v6ZYdHxyemWupiwecW4724OFkioACYuIA).

bradenmacdonald:
@MaxFrank13 What's the status of this PR?

MaxFrank13:
> @MaxFrank13 What's the status of this PR?

Hey @bradenmacdonald! Just getting back from leave today. I believe this left off with us still considering what the best naming convention would be for the variable and class. If there are no objections to what Brian and I have mentioned in the earlier comments, then I can go ahead and make those changes. Curious if you have any thoughts on this?

src/Truncate/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Retrieves plain string from children array and collects data
+ * to be able to restore original children in the future.
+ *
+ * @param {array} children
+ * @param {array} elementsData original data to restore children
+ * @returns string
+ */
+export function assembleStringFromChildrenArray(
+  children: Array<React.ReactNode>,
+  elementsData: Array<ElementDataEntry> = [],
```
bradenmacdonald: Nitpick: Modifying a passed array is a bit of an anti-pattern, unless the name of the function makes it obvious.

bradenmacdonald (commented):
@MaxFrank13 Great!

Yeah, I tend to agree that the CSS classes should have the `pgn__` prefix, and the CSS var is better as just `--truncate-prop-lines`, since it's locally scoped to this component anyways.

src/Truncate/utils.ts
MaxFrank13: Ah good call out! I've refactored so that we no longer use a passed array but instead initialize one within the function. In order for the recursion to work, this required returning the array as well as the string result. Because of this, the usage in `Truncate.tsx` had to be modified as well. I _believe_ it's working exactly the same but would invite any criticisms there as well!

bradenmacdonald: LGTM! I haven't tested this though. You also have a bunch of lint warnings now.

MaxFrank13: Ah! Something weird going on with VScode and the lint extension. Weird it only happens in Paragon repo for me. I can still run the `npm run lint` command though so I just did it that way 🤷

src/Truncate/Truncate.tsx
```diff
@@ -0,0 +1,33 @@
...
+    initialText = result;
+  } else {
+    initialText = String(children);
+  }
+
+  return (
+    <p
+      title={initialText}
+      aria-label={initialText}
+      className="truncate-text"
```
brian-smith-tcril, line 24: I checked the deploy preview (https://deploy-preview-4037--paragon-openedx-v23.netlify.app/components/truncate/) and it didn't appear to be working at all. I think this is likely the culprit.

```suggestion
      className="pgn__truncate-text"
```

src/Truncate/Truncate.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+function Truncate({ children, lines = 1 }: TruncateProps) {
+  let initialText: string = '';
+  if (Array.isArray(children)) {
+    const { result } = assembleStringFromChildrenArray(children);
+    initialText = result;
+  } else {
+    initialText = String(children);
+  }
```
brian-smith-tcril, line 18: This has some interesting side effects

Image

the deprecated component just doesn't work when passing in a react component

Image

which feels like a slightly better fail state than just having the title/aria-label get mangled.

src/Truncate/README.md
```diff
@@ -5,46 +5,44 @@ components:
...
     for using the platform and objectives to accomplish. To help members of each group learn about what edX 
     offers, reach goals, and solve problems, edX provides a variety of information resources.
     Learners, course teams, researchers, developers: the edX community includes groups with a range of reasons
     for using the platform and objectives to accomplish. To help members of each group learn about what edX
     offers, reach goals, and solve problems, edX provides a variety of information resources.
-  </Truncate.Deprecated>
+  </Truncate>
 ```
 
 ### With the custom ellipsis
```
brian-smith-tcril, line 28: This functionality doesn't exist in the new version. My initial thought was that we should just remove this example, but I think the best thing to do would be to have 2 docs site pages (one for the new version and one for the deprecated version) until we fully remove the deprecated component.

src/Truncate/Truncate.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  return (
+    <p
```
brian-smith-tcril, line 21: Having this hardcoded to `p` seems less than ideal. The deprecated version allows consumers to configure this with `elementType` (which feels like it's just an odd name, we generally use "`as`").

The other question is what a reasonable default would be, I'm thinking `span` probably makes sense so people can use as a child of other elements without making funky invalid html or needing to use the `as` prop.

src/Truncate/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Retrieves plain string from children array and collects data
+ * to be able to restore original children in the future.
+ *
+ * @param {array} children
+ * @param {array} elementsData original data to restore children
+ * @returns string
+ */
+export function assembleStringFromChildrenArray(
```
brian-smith-tcril, line 24: The `@returns` says this just returns a string, but it's actually returning quite a bit more than that. Could probably be simplified to just keep/return the string.

brian-smith-tcril (commented):
Sorry it took me so long to get to reviewing this one!

I'm super excited to see the CSS version of this coming together.

I noticed a few issues and left some comments, please let me know if you have any questions for me!

PR #4052, opened 2025-12-18
chore(ci): update semantic release workflow to use OIDC

brian-smith-tcril:
## Description

* Update `semantic-release` version used from `22` to `25` (we had it pinned at `22` to ensure it would work with node 18 - that is no longer needed)
* Remove `NPM_TOKEN` from workflow as those are deprecated
* Add `permissions` sections to match [the example from the `semantic-release` documentation](https://github.com/semantic-release/semantic-release/blob/master/docs/recipes/ci-configurations/github-actions.md#githubworkflowsreleaseyml-configuration-for-node-projects)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,10 +5,19 @@ on:
     branches:
     - next
     - release-*
+
+permissions:
+  contents: read # for checkout
+
 jobs:
   release:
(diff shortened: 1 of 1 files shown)


arbrandes (approved):
Looks good to me!

Issue #3949, opened 2025-10-28
[CLI] Can't interrupt `build-scss`


brian-smith-tcril:
`Ctrl` + `C` doesn't stop `build-scss`. Considering this step can take some time (anything more than a second or two is enough to justify ensuring it can be interrupted) we should make sure this works.

PR #4155, opened 2026-03-04
fix: allow Ctrl+C to interrupt build-scss command

PKulkoRaccoonGang:
## Description

**Issue:** https://github.com/openedx/paragon/issues/3949

`sass.compile()` is a synchronous function. While it runs, Node.js is frozen - it can't do anything else, including responding to Ctrl+C. Instead of launching all compilations at once, we run them one at a time with a `setTimeout(resolve, 0)` pause between each.

For example:

Step | Operation | Description
-- | -- | --
1 | sass.compile(core) | The first Sass compilation starts running
2 | Ctrl+C pressed | A SIGINT signal is triggered and queued by the runtime
3 | setTimeout(0) | The event loop processes the next task and detects the pending SIGINT
4 | Process exits | The program terminates due to the interrupt signal
5 | sass.compile(light) | This step is never executed because the process has already exited

image

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/lib/__tests__/build-scss.test.js
+++ b/lib/__tests__/build-scss.test.js
@@ -114,7 +114,7 @@ describe('buildScssCommand', () => {
     });
   });
 
-  it('should compile core and theme stylesheets', () => {
+  it('should compile core and theme stylesheets', async () => {
     const args = [
       '--corePath=styles/core.scss',
       '--themesPath=styles/themes',
@@ -122,7 +122,7 @@ describe('buildScssCommand', () => {
       '--defaultThemeVariants=light',
     ];
 
-    buildScssCommand(args);
+    await buildScssCommand(args);
 
     expect(sass.compile).toHaveBeenCalledWith(
       expect.stringContaining('core.scss'),
@@ -135,8 +135,8 @@ describe('buildScssCommand', () => {
     );
   });
 
-  it('should use default arguments when none provided', () => {
-    buildScssCommand([]);
+  it('should use default arguments when none provided', async () => {
+    await buildScssCommand([]);
 
     expect(sass.compile).toHaveBeenCalledWith(
       expect.stringContaining('core.scss'),
@@ -144,8 +144,8 @@ describe('buildScssCommand', () => {
     );
   });
 
-  it('should exclude core properly', () => {
-    buildScssCommand(['--excludeCore']);
+  it('should exclude core properly', async () => {
+    await buildScssCommand(['--excludeCore']);
 
     expect(sass.compile).not.toHaveBeenCalledWith(
       expect.stringContaining('core.scss'),
--- a/lib/__tests__/build-scss.test.js
+++ b/lib/__tests__/build-scss.test.js
@@ -117,7 +117,7 @@ describe('buildScssCommand', () => {
     });
   });
 
-  it('should compile core and theme stylesheets', async () => {
+  it('should compile core and theme stylesheets', () => {
     const args = [
       '--corePath=styles/core.scss',
       '--themesPath=styles/themes',
@@ -125,7 +125,7 @@ describe('buildScssCommand', () => {
       '--defaultThemeVariants=light',
     ];
 
-    await buildScssCommand(args);
+    buildScssCommand(args);
 
     expect(sass.compile).toHaveBeenCalledWith(
       expect.stringContaining('core.scss'),
--- a/lib/__tests__/build-scss.test.js
+++ b/lib/__tests__/build-scss.test.js
@@ -1,4 +1,7 @@
 const fs = require('fs');
+const path = require('path');
+const { spawn } = require('child_process');
+const os = require('os');
 const sass = require('sass');
 
 const buildScssCommand = require('../build-scss');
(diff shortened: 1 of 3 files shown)


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

lib/build-scss.js
```diff
@@ -151,7 +151,13 @@ const compileAndWriteStyleSheets = ({
  *
  * @param {Array<string>} commandArgs - Command line arguments for building SCSS stylesheets.
  */
-function buildScssCommand(commandArgs) {
+async function buildScssCommand(commandArgs) {
+  process.on('SIGINT', () => {
+    // eslint-disable-next-line no-console
+    console.log(chalk.yellow('\nBuild interrupted. Partial output may exist in the output directory.'));
+    process.exit(130);
+  });
```
brian-smith-tcril: Nothing else in this file uses `console.log`, and directly calling `process.exit` from the `lib` file seems less than ideal. We already have error handling in
https://github.com/openedx/paragon/blob/c4e95f0e0cbafb3f663c6fa9a69b2e90fdaa2150/bin/paragon-scripts.js#L275-L287
so it'd be nice to just surface the error from this function and hanve `bin/paragon-scripts` handle it.

lib/build-scss.js
```diff
@@ -168,27 +174,35 @@ function buildScssCommand(commandArgs) {
     defaultThemeVariants,
   } = minimist(commandArgs, { default: defaultArgs, boolean: ['excludeCore'] });
 
+  // eslint-disable-next-line no-console
+  console.log(chalk.grey('Press Ctrl+C to interrupt the build (takes effect between compilations).\n'));
```
brian-smith-tcril: Same thing about trying to keep `console.log` out of this file. I figure for this one we can just remove these lines.

I'd also like to look into what it would take to make it so we can actually interrupt the compilation. I was thinking we might be able to use `sass.compileAsync` but [the `sass` docs mention](https://sass-lang.com/documentation/js-api/functions/compileasync/)
> ⚠️ Heads up!
When using the sass npm package, [compile](https://sass-lang.com/documentation/js-api/functions/compile) is almost twice as fast as [compileAsync](https://sass-lang.com/documentation/js-api/functions/compileAsync), due to the overhead of making the entire evaluation process asynchronous.

lib/__tests__/build-scss.test.js
brian-smith-tcril, line 1: It'd be nice to add a test to this file checking the `SIGINT` functionality.

brian-smith-tcril (commented):
Thanks for doing this! This definitely is an improvement over the current state. I left a few comments with small cleanup suggestions, and one with a bigger question.

lib/build-scss.js
peterkulko: I also looked into using a worker thread to make the compilation truly interruptible. It would require a separate worker file with the sass importer logic duplicated, message passing for compilation results, and error handling across threads. It's more complexity than the current one-liner, and the benefit is just saving a few seconds on interruption...

brian-smith-tcril: Yeah, it's definitely worth landing an improvement instead of blocking this on a complicated rework.

Maybe just showing a "cancelling" or "exiting" message would be helpful? Just some form of immediate feedback so users know their Ctrl+C was picked up.

peterkulko: We're using the synchronous `sass.compile`, the event loop stays blocked during compilation, which means the SIGINT handler only gets a chance to run after the current theme finishes.

So even if we add an interrupt message, it wouldn't appear right away, it would show up once compilation of the current theme completes.

lib/build-scss.js
peterkulko: Moved the SIGINT handler to `bin/paragon-scripts.js`. Unfortunately we can't surface SIGINT as a thrown error through the catch block because `sass.compile()` is synchronous, `process.exit(130)` in the signal handler is the only way to interrupt it immediately. So the handler lives in the entry point instead.

lib/__tests__/build-scss.test.js
peterkulko: Added

lib/__tests__/build-scss.test.js
```diff
@@ -114,15 +117,15 @@ describe('buildScssCommand', () => {
     });
   });
 
-  it('should compile core and theme stylesheets', () => {
+  it('should compile core and theme stylesheets', async () => {
```
brian-smith-tcril: Now that `buildScssCommand` is back to not being `async` we can probably go back to a `sync` test to make the diff smaller

brian-smith-tcril (commented):
I'm happy with this PR. It's a solid improvement over the current state of things, and considering the tradeoffs required to make `sass.compile` async I think it's the best option we have.

I left one last comment about changing the tests back to sync from async now that `buildScssCommand` is staying sync, once that's cleaned up I'll be happy to merge this!

lib/__tests__/build-scss.test.js
peterkulko: Corrected, thanks

brian-smith-tcril (approved):

PR #4195, opened 2026-04-06
feat: add atlasTranslations to package.json

### LINKED ISSUE #176: i18n pipeline
### Description

For now, this is a discovery: is the frontend-base i18n pipeline fully developed?  We might be missing an `atlas pull` step in our publish pipeline not only for the apps, but for frontend-base itself.  Or maybe this should only happen in frontend-template-site.

Should we ship libraries with full translations?  It seems this would be a new thing: frontend-component-[header|footer] only extract i18n as part of the release workflow, leaving the atlas pull to the MFE.

See [this Slack thread](https://openedx.slack.com/archives/C04BM6YC7A6/p1771444376973389) for some more context and preliminary findings.

brian-smith-tcril:
* Part of https://github.com/openedx/frontend-base/issues/176

Adds an `atlasTranslations` field to `package.json` so that sites using `openedx translations:pull` (frontend-base PR #205) can resolve and pull Paragon's translations transitively.

```json
"atlasTranslations": {
  "path": "translations/paragon/src/i18n/messages"
}
```

No `dependencies` field is needed — Paragon has no further `atlasTranslations` deps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,9 @@
     "/lib"
   ],
   "sideEffects": false,
+  "atlasTranslations": {
+    "path": "translations/paragon/src/i18n/messages"
+  },
   "scripts": {
     "build": "make build",
     "build-docs": "make build-docs",


arbrandes (approved):
👍🏼

PR #4270, opened 2026-05-02
E0d/fix actionrow a11y

e0d:
## Description

### Deploy Preview

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/src/ActionRow/index.tsx
+++ b/src/ActionRow/index.tsx
@@ -32,7 +32,7 @@ function ActionRow({
 }
 
 function ActionRowSpacer() {
-  return <span className="pgn__action-row-spacer" />;
+  return <span className="pgn__action-row-spacer" aria-hidden="true" />;
 }
 
 ActionRow.Spacer = ActionRowSpacer;
--- a/src/ActionRow/index.tsx
+++ b/src/ActionRow/index.tsx
@@ -1,6 +1,23 @@
 import React from 'react';
 import classNames from 'classnames';
 
+/**
+ * Layout utility that arranges action controls in a horizontal row or vertical stack.
+ *
+ * When ActionRow groups interactive controls (buttons, links), add `role="toolbar"`
+ * and `aria-label` so assistive technology users can identify the group and
+ * distinguish multiple ActionRows on the same page:
+ *
+ * ```jsx
+ * <ActionRow role="toolbar" aria-label="Form actions">
+ *   <Button variant="tertiary">Cancel</Button>
+ *   <Button variant="primary">Submit</Button>
+ * </ActionRow>
+ * ```
+ *
+ * Both props are accepted via the standard HTML attributes spread; no extra
+ * prop definitions are required.
+ */
 interface ActionRowProps extends React.HTMLAttributes<HTMLElement> {
   /** Specifies the base element */
   as?: React.ElementType;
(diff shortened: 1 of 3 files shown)


src/ActionRow/_index.scss
```diff
@@ -25,7 +25,7 @@
   }
 
   & > * + * {
-    margin-bottom: var(--pgn-spacing-action-row-gap-y);
+    margin-bottom: var(--pgn-spacing-action-row-gap-y, .5rem);
   }
 }
```
brian-smith-tcril, line 30: The commit message on https://github.com/openedx/paragon/pull/4270/changes/5f410358215c70dfff5deb63ab5525eb2e10fe4f says
> **fix(ActionRow): add fallback values to spacing tokens**
>
> Without fallbacks, both gap tokens resolve to 0 if design tokens are
not compiled and loaded, collapsing actions flush with no visual
separation. .5rem matches the intent of the token and maintains
grouping cues for cognitive accessibility (WCAG 1.4.12 advisory).

Without values from tokens *nothing* in Paragon would be WCAG compliant. Paragon ships with these defined.

Using fallbacks here puts us in a position of needing to update this in 2 places (both the default in the token json, and the fallback here) if we want to make a change.

I also fed Claude this comment and got:
> Your reasoning checks out, and there's a sharper point you can make using prior art in the repo.
>
> Token IS defined with the same value:
> - tokens/src/core/components/ActionRow.json:6-7 — both gap tokens default to .5rem
> - styles/css/core/variables.css:150-151 — compiles to --pgn-spacing-action-row-gap-x: 0.5rem
> 
> So the PR's var(..., .5rem) is literally duplicating the value that already ships, which is your dual-source-of-truth concern.
>
> Existing fallback patterns in Paragon are different:                                                                                                                                                                                                                                                                                                                                                                                  
> 
> There are 28 fallbacks across Button/Alert/Badge/PageBanner/Bubble, but a spot-check shows they're all for tokens that don't have a base value in variables.css — variant/state-scoped tokens like --pgn-btn-bg, --pgn-alert-border-color, --pgn-badge-focus-bg. Those tokens are only set inside .btn-primary { ... }, etc., and the fallback represents the unset/default state.
>
> --pgn-spacing-action-row-gap-x is fundamentally different: it's a base token with a shipped value. Adding a fallback here would be the first instance in this directory of a fallback duplicating a base-token value — i.e., it diverges from the established pattern, not aligns with it.                                                                                                                                            
>
> On the WCAG framing: if "tokens not loaded" is the failure mode being defended against, every Paragon component fails — spacing, typography, color contrast, focus rings. Patching one component's gap doesn't materially improve the no-tokens experience; it just creates a maintenance trap for that one component.

e0d: I've rebased and dropped the commit.  I'm not sure why Claude created the branch from the release branch rather than master.

src/ActionRow/index.tsx
```diff
@@ -1,6 +1,23 @@
...
 
+/**
+ * Layout utility that arranges action controls in a horizontal row or vertical stack.
+ *
+ * When ActionRow groups interactive controls (buttons, links), add `role="toolbar"`
+ * and `aria-label` so assistive technology users can identify the group and
+ * distinguish multiple ActionRows on the same page:
+ *
+ * ```jsx
+ * <ActionRow role="toolbar" aria-label="Form actions">
```
brian-smith-tcril, line 12: This is using `role="toolbar"` but the README example has `role="group"`

@kblemel any thoughts on which we should be recommending (or the cases in which we should recommend one vs the other)?

PR #4271, opened 2026-05-02
fix(ActionRow): align DOM and visual order in stacked mode

e0d:
In isStacked mode, flex-direction: column-reverse placed the primary action (last DOM child) at the visual top but keyboard focus still reached it last, violating WCAG 1.3.2 Meaningful Sequence and 2.4.3 Focus Order.

Fix (Option A): reverse children in DOM order when isStacked is true and change the SCSS to flex-direction: column. The primary action is now first in both DOM and visual order — a sighted keyboard user tabs to it first, matching what they see.

Consumers continue passing the primary action as the last child (consistent with horizontal mode); the component handles reversal internally.

## Description

### Deploy Preview

## Merge Checklist

## Post-merge Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ www/public
 
 # Local Netlify folder
 .netlify
+.claude/
--- a/src/ActionRow/_index.scss
+++ b/src/ActionRow/_index.scss
@@ -17,7 +17,7 @@
   display: flex;
   flex-grow: 1;
   align-items: center;
-  flex-direction: column-reverse;
+  flex-direction: column;
   justify-content: center;
 
   & > * {
@@ -25,7 +25,7 @@
   }
 
   & > * + * {
-    margin-bottom: var(--pgn-spacing-action-row-gap-y);
+    margin-top: var(--pgn-spacing-action-row-gap-y);
   }
 }
 
--- a/src/ActionRow/ActionRow.test.tsx
+++ b/src/ActionRow/ActionRow.test.tsx
@@ -1,0 +1,33 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import ActionRow from '.';
+
+describe('<ActionRow />', () => {
+  describe('stacked mode', () => {
+    it('renders children in reverse DOM order so primary action leads focus', () => {
+      const { getAllByRole } = render(
+        <ActionRow isStacked>
+          <button type="button">Secondary</button>
+          <button type="button">Primary</button>
+        </ActionRow>,
+      );
+      const buttons = getAllByRole('button');
+      expect(buttons[0]).toHaveTextContent('Primary');
+      expect(buttons[1]).toHaveTextContent('Secondary');
+    });
+  });
+
+  describe('horizontal mode', () => {
+    it('renders children in original DOM order', () => {
+      const { getAllByRole } = render(
+        <ActionRow>
+          <button type="button">Secondary</button>
+          <button type="button">Primary</button>
+        </ActionRow>,
+      );
+      const buttons = getAllByRole('button');
+      expect(buttons[0]).toHaveTextContent('Secondary');
+      expect(buttons[1]).toHaveTextContent('Primary');
+    });
+  });
+});
--- a/src/ActionRow/README.md
+++ b/src/ActionRow/README.md
@@ -16,7 +16,7 @@ notes: |
 A layout utility for the common use case of aligning buttons, links or text
 in a row in a control bar or nav.
 
-ActionRow assumes that its last child is the primary action and lays out actions so that the last item is in a primary location. If horizontal, the primary action sits on the right. If stacked, the primary action sits at the top of the stack (this is done via `flex-direction: column-reverse;`).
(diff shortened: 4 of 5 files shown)


src/ActionRow/_index.scss
```diff
@@ -17,7 +17,7 @@
   display: flex;
   flex-grow: 1;
   align-items: center;
-  flex-direction: column-reverse;
+  flex-direction: column;
```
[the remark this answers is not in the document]
e0d: @codex Does the change at line 28 of this file resolve this issue satisfactorily?

e0d:
@codex review

brian-smith-tcril:
@kblemel in the WG meeting you mentioned finding documentation/discussions about `flex-direction: column-reverse` and `flex-direction: row-reverse` for interactive elements. Would you be able to link those here?

CLAUDE.md
```diff
@@ -0,0 +1,98 @@
+# CLAUDE.md
```
brian-smith-tcril: I know we've discussed adding `CLAUDE.md` and `AGENTS.md` files to repos in the org before, but that should definitely happen as a separate PR.

ACCESSIBILITY_AUDIT.md
```diff
@@ -0,0 +1,394 @@
+# WCAG 2.2 AA Accessibility Audit — Paragon Component Library
```
brian-smith-tcril: Very cool to see this file for context, but it shouldn't be part of this PR

.gitignore
```diff
@@ -20,3 +20,4 @@ www/public
 
 # Local Netlify folder
 .netlify
+.claude/
```
brian-smith-tcril, line 23: Not sure about "claude" as the section comment here but I figure making it clear this _isn't_ part of the "Local Netlify folder" would be good

```suggestion

# claude
.claude/
```

docs/a11y-audit-ActionRow.md
```diff
@@ -0,0 +1,154 @@
+# WCAG 2.2 AA Accessibility Audit — ActionRow
```
brian-smith-tcril: Very cool to see this file for context, but it shouldn't be part of this PR

brian-smith-tcril (commented):
Overall I think this makes sense.

I did notice one edge case when using the following example code to simulate an application that switches between stacked and non-stacked (I'm thinking this would likely be done for responsive design with `useMediaQuery`)

<details>
<summary>Code used on docs site to test</summary>

```jsx
() => {
  const [stackIt, setStackIt] = React.useState(false);
  
  React.useEffect(() => {
    const interval = setInterval(() => {
      setStackIt(prev => !prev);
    }, 2000); // toggles every 2 seconds
    
    return () => clearInterval(interval);
  }, []);
  
  return (
    <ActionRow isStacked={stackIt}>
      <p className="x-small">
        Bespoke leggings yuccie, portland umami readymade craft beer vaporware sriracha.
      </p>
      <Button variant="tertiary">
        Go back
      </Button>
      <Button variant="primary">
        Continue
      </Button>
    </ActionRow>
  );
}
```

</details>

#### Behavior before PR
The element selected by the keyboard stays selected when the component switches between stacked and non-stacked.

https://github.com/user-attachments/assets/0163548b-01e8-4f43-a068-4fa718b00d20

#### Behavior after PR
The element selected by the keyboard is no longer selected when the component switches between stacked and non-stacked

https://github.com/user-attachments/assets/229e28bc-0ec9-444c-aad6-654ef8fbeb6f

@kblemel - any thoughts on that change? Also it'd be great to know if any of the conversations/documentation you found mentioned something like that (I assume the plugin you mentioned does something similar to what this PR is doing and would result in similar behavior)

.gitignore
e0d: What's your recommendation?  I don't have a strong commitment to this change.

wittjeff:
I have vague memory of this being a design choice with external input -- someone wanted Next to be earlier than Back because Back is much less likely to be needed, so it's a usability improvement.  This is the kind of thing I don't have a strong opinion on, so defer to whatever Design wants. 
I have seen something similar somewhere with Desktop nav collapsing into a hamburger menu (somewhere, not necessarily with Paragon), where the focus order was not as expected. I have the same neutral opinion there. I think "may be confusing" is possible and technically correct but perhaps too strong in practical terms.

e0d:
> I have vague memory of this being a design choice with external input -- someone wanted Next to be earlier than Back because Back is much less likely to be needed, so it's a usability improvement. This is the kind of thing I don't have a strong opinion on, so defer to whatever Design wants. I have seen something similar somewhere with Desktop nav collapsing into a hamburger menu (somewhere, not necessarily with Paragon), where the focus order was not as expected. I have the same neutral opinion there. I think "may be confusing" is possible and technically correct but perhaps too strong in practical terms.

Maybe that's true, but I'd argue the visual order and tab order should still be the same in that case.  Do you disagree?

.gitignore
brian-smith-tcril: maybe `# agents`

Issue #4275, opened 2026-05-04
Decouple semantic token declarations from theme-specific values


brian-smith-tcril:
> [!NOTE]
> Generated by Claude

## Problem

Paragon's design tokens currently couple two separable concerns into the `themes/<variant>/` directory:

1. **The declaration that a path exists** — e.g. `color.gray.500` is a token of type `color`.
2. **The value of that path in a particular theme** — e.g. `#454545` in light, possibly something else in dark.

`tokens/src/themes/light/global/color.json` does both. The schema (which paths exist) is determined by whatever the theme files happen to contain.

This coupling was surfaced while implementing #4274 (app tokens). App tokens reference Paragon paths like `{color.gray.500}`, and the build needs those paths in scope for reference resolution. With the current architecture, the only way to expose them is to include a specific theme variant's files (`themes/light/**`) — which drags in concrete values that the app build doesn't actually care about, since the output is `var(--pgn-color-gray-500)` regardless. It's load-bearing in name only; the values are immediately filtered out.

## Proposal

Split the token tree into **semantic tokens** (the paths Paragon promises consumers can reference) and **primitive tokens** (the concrete values per theme variant):

```
tokens/src/
├── core/                  # primitives that are theme-invariant + semantic declarations
│   ├── colors.json        # e.g. color.gray.500 → {raw.color.gray.500}
│   └── ...
└── themes/
    └── light/
        └── colors.json    # raw.color.gray.500 → "#454545"
```

Concrete shape:

```json
// core/colors.json — schema, references primitives
{
  "color": {
    "gray": {
      "500": { "$value": "{raw.color.gray.500}", "$type": "color" }
    }
  }
}

// themes/light/colors.json — primitives only
{
  "raw": {
    "color": {
      "gray": {
        "500": { "$value": "#454545", "$type": "color" }
      }
    }
  }
}
```

After this split:
- `core/` is the stable interface Paragon exposes. Anything referencing `{color.…}` only needs `core/` in scope.
- Themes are pure value providers — adding a new theme variant means defining the `raw.*` set, no schema duplication.
- The app token build (#4274) stops depending on which theme happens to be included for vocabulary purposes.

## Tradeoffs / scope

This is a multi-PR effort, not a small change:

- Every existing `themes/<variant>/**/*.json` file needs migration: each `color.X.Y` becomes a `raw.color.X.Y` definition, and a corresponding `core/` entry references it.
[...]

PR #4276, opened 2026-05-05
feat: support application tokens

### LINKED ISSUE #4274: Handle non-`--pgn` prefixed tokens from MFEs
Theme authors should be able to use tokens to customize MFEs that use CSS variables. One example is [here](https://github.com/openedx/frontend-app-catalog/blob/0435acd978124f23085bd885f643819b4a7c712d/src/home/components/home-banner/index.scss#L14)
```scss
background-color: var(--catalog-home-page-banner-background-color, var(--pgn-color-gray-500));
```

Using the existing tooling, theme authors would end up with `--pgn-catalog-home-page-banner-background-color`.

MFEs shouldn't prefix their own CSS vars with `--pgn`, as those styles aren't part of Paragon, but the tooling should allow for theme authors to include tokens for MFEs.

brian-smith-tcril:
## Description

Adds a third top-level token directory `apps/` alongside `core/` and `themes/`. Tokens placed under `apps/<app-name>/` are emitted **without** the `--pgn-` prefix, so brand package authors can override CSS variables defined by individual MFEs (which use their own naming conventions, not `--pgn-`).

References from app tokens to Paragon tokens are preserved as `var(--pgn-…)` in the output, so theme variation flows through automatically without per-variant app token files. App `variables.css` files are inlined into each theme variant's bundle by `build-scss` via `@import`, so consuming MFEs need no configuration change.

Closes #4274. See `docs/decisions/0021-support-application-tokens.rst` (added in this PR) for the full design rationale, including rejected alternatives.

Follow-up architectural work for decoupling semantic token declarations from theme-specific values: #4275.

### Deploy Preview

This PR changes build tooling and adds an ADR; there are no Paragon component changes. End-to-end verification was performed manually against `frontend-app-catalog` using a sibling test branch — see [`app-tokens-test/test-theme`](https://github.com/brian-smith-tcril/paragon/tree/app-tokens-test/test-theme) for built fixture, screenshots, and the `env.config.jsx` used.

### DIFF
--- a/docs/decisions/0021-support-application-tokens.rst
+++ b/docs/decisions/0021-support-application-tokens.rst
@@ -1,0 +1,192 @@
+21. Support application tokens
+------------------------------
+
+Status
+======
+
+Accepted
+
+Context
+=======
+
+Paragon's design token system (see `ADR 0019: Scaling Paragon's styles
+architecture with design tokens
+<0019-scaling-styles-with-design-tokens.rst>`__) prefixes every emitted CSS
+variable with ``--pgn-``. This is intentional for tokens that belong to the
+Paragon design system: it namespaces them, keeps them from colliding with
+consumer CSS, and reflects their origin.
+
+However, Open edX micro-frontends (MFEs) define their own CSS variables for
+component-level customization, and those variables are *not* part of Paragon.
+They use names chosen by the MFE itself. For example, ``frontend-app-catalog``
+defines this in its home banner stylesheet:
+
+.. code-block:: scss
+
+  background-color: var(--catalog-home-page-banner-background-color, var(--pgn-color-gray-500));
+
+The fallback (``var(--pgn-color-gray-500)``) covers default rendering when no
+MFE-specific override is supplied. Brand package authors who want to customize
+the catalog banner separately are expected to set
+``--catalog-home-page-banner-background-color`` to a value of their choosing.
+
+With Paragon's existing token tooling, brand package authors cannot do this.
+Every token built through ``paragon build-tokens`` is unconditionally prefixed
+with ``pgn``. A brand package author defining a token at path
+``catalog.home-page.banner.background-color`` would emit a CSS variable named
+``--pgn-catalog-home-page-banner-background-color`` — a name no MFE actually
+reads.
+
+This is the gap reported in `issue #4274
+<https://github.com/openedx/paragon/issues/4274>`__: brand package authors
+should be able to use the existing token system to customize MFE CSS
+variables, but the system has no mechanism for tokens that don't belong to
+Paragon.
+
+Decision
+========
+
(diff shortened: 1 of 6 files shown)


docs/decisions/0021-support-application-tokens.rst
```diff
@@ -0,0 +1,192 @@
...
+
+**Rejected alternative: a per-app prefix (e.g. ``--<app-name>-…``) injected
+automatically by the build.** The reason this fails is that MFE CSS variable
+conventions vary across the Open edX codebase. Some MFEs use the unprefixed
+app slug (``--catalog-…``), some use the full package name
+(``--frontend-app-catalog-…``), some use neither. Any single auto-prefix
+scheme would either require brand package authors to write extra path
+segments to compensate, or would simply fail to match the MFE's actual
+variable name. Letting the JSON path be the source of truth is the only
+flexible answer.
```
peterkulko, line 75: Do we really want to allow this level of flexibility without requiring consumers to follow a naming convention for newly created design tokens?

Could there be conflicts if multiple tokens from different applications generate CSS variables with identical names, potentially overriding each other?

PKulkoRaccoonGang (commented):
Thanks for this PR! This is definitely a very useful change. I've tested the application token generation in [elm-theme](https://github.com/edx/elm-theme) (without MFE) and left a few questions.

docs/decisions/0021-support-application-tokens.rst
brian-smith-tcril: That's a good question, I was mostly going off of what we're already seeing.

I'm thinking it should be a naming convention we enforce on the app side of things. In `frontend-app-catalog` we have things like `--catalog-course-about-page-intro-card-min-height` and `--catalog-home-page-banner-background-color`.

We could probably add an ADR for app token naming similar to https://github.com/openedx/frontend-plugin-framework/blob/master/docs/decisions/0003-slot-naming-and-life-cycle.rst, but I know we started FPF with "wild west" slot names and used aliases when we landed on a decision for a naming convention.

With CSS vars "aliases" are super easy, we can just add the old name as a fallback.

xitij2000: I feel the flexibility would be useful here. For those creating plugins or XBlocks that use CSS variables, this would make it easy to add additional tokens for use in those custom cases as well.

xitij2000 (approved):
This is working well! 
I see a lot of 'non-app' usecases for this as well, such as for XBlocks and frontend plugins, however I don't know what would be a better name either. 

"scoped" / "scopes" perhaps? 🤷🏾

arbrandes (approved):
I'll keep this short: I read the ADR, agreed with everything, then asked Claude:

"Does the implementation match the ADR?"

"Yes."

"Are there any issues with the code?"

"No."

Approved.

PR #4282, opened 2026-05-09
feat: add support for :host pseudo-class in style-dictionary output

xitij2000:
## Description

This PR updates the generated CSS variables from design tokens to emit variables under both `:root` and `:host`.

Previously, design token CSS variables were generated only under `:root`, which makes them globally available at the document level. This works well for normal application usage, but it does not fully support stylesheets that are loaded inside a Shadow DOM.

The primary use case for this change is to support isolated XBlocks that render inside a Shadow DOM. By including `:host` alongside `:root`, the XBlock can load these stylesheets with the theme remaining isolated to the XBlock without spilling over to the rest of the page. 

### Deploy Preview

NA. This should have no visual impact. 

## Merge Checklist

* [x] Consider whether this change needs to reviewed/QA'ed for accessibility (a11y). If so, please request an a11y review for the PR in the [#wg-paragon](https://openedx.slack.com/archives/C02NR285KV4) Open edX Slack channel.

## Post-merge Checklist

### DIFF
--- a/tokens/style-dictionary.js
+++ b/tokens/style-dictionary.js
@@ -158,7 +158,7 @@ const createCustomCSSVariables = async ({ formatterArgs }) => {
     usesDtcg: true,
   });
   const header = await fileHeader({ file, formatting });
-  return `${header}:root {\n${variables}\n}\n`;
+  return `${header}:root, :host {\n${variables}\n}\n`;
 };
 
 /**


openedx-webhooks:
Thanks for the pull request, @xitij2000!


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

brian-smith-tcril:
I'm trying to make sure I understand this correctly:
* In an isolated XBlock
  * CSS var values set on `:root` can be read/consumed inside the shadow DOM without issue
  * New values cannot be written to `:root` from *within* the shadow DOM

This change makes it so if a theme that sets CSS vars is loaded *within* the shadow DOM those vars will be written to `:host`, instead of not being written anywhere because `:root` could not be targeted.

I think this makes sense, and I think this isn't a risky change.

It would be great if you could provide some example use cases for this. I'm confident this won't break anything, but it is adding functionality we'll need to support, and I'd like to verify this is the best way to support those use cases.

xitij2000:
Here is a summary of the issue we faced:
- We want to create an XBlock that uses Paragon and uses the MFE Config API to apply the same stylesheets as all the MFEs so that it visually matches the MFEs. (Relevant [PR](https://gitlab.com/opencraft/dev/exemplar-blocks/-/merge_requests/24))
- If we include the Paragon stylesheets in the XBlock they will be loaded for all XBlocks and they might not be styled correctly since the paragon stylesheets apply global styles. 
- Also the LMS stylesheet will still be in effect which will distort the styles of Paragon components
- If we wrap the component in a Shadow DOM then the outside styling won't leach in and vice versa. 
- If we load the stylesheets inside the Shadow DOM, then all the CSS classes apply, but the CSS "variables" do not! 
- CSS variables targeting `:root` will be picked up by content inside the shadow DOM **only if** the stylesheet is loaded in the main page and **not** when the stylesheet is loaded inside the shadow DOM.
- So we would want all the CSS classes etc. to be loaded inside the Shadow DOM and all the CSS variables to be loaded outside, which isn't possible since they are mixed together. 

Simply adding the [`:host` pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:host) somewhat solves this since now you can load all the stylesheet inside the Shadow DOM and it should work. It will cause negligible increase in the size of the stylesheets and have no effect unless someone explicitly loads it in a Shadow DOM, in which case this is likely want they want to do.

mphilbrick211:
@brian-smith-tcril when you have a moment, could you take another look at this one? Thanks!

PR #4412, opened 2026-08-10
feat: enable compatibility with react-intl v10.1.20+

bradenmacdonald:
## Description

This makes Paragon compatible with `react-intl` **10.1.20+** in addition to the existing support for react-intl v5 and v6.

The main motivating factor is that [`react-intl` v5 and v6 declare an annoying dependency on `typescript` v4/v5](https://github.com/formatjs/formatjs/blob/10db1843ba2801270cfba7ea74552813f476ef4b/packages/react-intl/package.json#L148), and since Paragon requires v5 or v6, Paragon is blocking upgrades of TypeScript above v5.

Regarding the parts of the `react-intl` API that Paragon uses, there have been no major/breaking changes between v5 and v10, so compatibility is not much of an issue. The main change is that `injectIntl` is gone in `react-intl` v10, but Paragon was no longer using it anyways. MFEs that haven't finished removing `injectIntl` yet can continue to use react-intl v6 until they are upgraded.

### Other versions

`react-intl` v10 between v10.0.0 and v10.1.20 require React 19, so they're excluded, but v10.1.20+ support React 18.

`react-intl` v9 was broken and v7/v8 work at runtime but have `typescript` dependencies that prevent their installation in this repo. I don't think it's worth trying to support them.

### Deploy Preview

https://deploy-preview-4412--paragon-openedx-v23.netlify.app/

## Merge Checklist

n/a for this sort of PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
   "peerDependencies": {
     "react": "^16.8.6 || ^17 || ^18",
     "react-dom": "^16.8.6 || ^17 || ^18",
-    "react-intl": "^5.25.1 || ^6.4.0"
+    "react-intl": "^5.25.1 || ^6.4.0 || ^10.1.20"
   },
   "devDependencies": {
     "@babel/cli": "^7.16.8",
@@ -156,6 +156,7 @@
     "jest-environment-jsdom": "^29.7.0",
     "markdown-loader-jest": "^0.1.1",
     "react": "^18",
+    "react-intl": "^10.1.20",
     "react-test-renderer": "^18",
     "regenerator-runtime": "^0.14.0",
     "semantic-release": "^25.0.0",
@@ -197,7 +198,7 @@
       "/component-generator/"
     ],
     "transformIgnorePatterns": [
-      "/node_modules/(?!(@openedx/paragon)/).*/"
+      "/node_modules/(?!(@openedx/paragon|react-intl|@formatjs|intl-messageformat)/).*/"
     ]
   },
   "browserslist": [
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Use `paragon help` to see more information.
 
 ## Internationalization
 
-Paragon supports internationalization for its components out of the box with the support of [react-intl](https://formatjs.io/docs/react-intl/). You may view translated strings for each component on the documentation website by switching languages in the settings.
+Paragon supports internationalization for its components out of the box with the support of [react-intl](https://formatjs.io/docs/react-intl/). ``react-intl`` 5.x, 6.x, or 10.1.20+ may be used. You may view translated strings for each component on the documentation website by switching languages in the settings.
 
 Due to Paragon's dependence on ``react-intl``, that means that your whole app needs to be wrapped in its provider, e.g.:
 
@@ -265,59 +265,30 @@ When developing a new component you should generally follow three rules:
       />
       ```
 
-   - For places where the display string has to be a plain JavaScript string use ``formatMessage``, this would require access to ``intl`` object from ``react-intl``, e.g.
+   - For places where the display string has to be a plain JavaScript string use ``formatMessage``, this would require access to ``intl`` object from ``react-intl``, which is provided by the ``useIntl`` hook (the ``injectIntl`` HOC must not be used, as it was removed in ``react-intl`` 10)
 
-      - For class components use ``injectIntl`` HOC
+      ```javascript
+      import { useIntl } from 'react-intl';
 
-          ```javascript
-          import { injectIntl } from 'react-intl';
+      const MyFunctionComponent = ({ altText }) => {
+        const intl = useIntl();
+        const intlAltText = altText || intl.formatMessage({
+          id: 'pgn.MyComponent.altText',
+          defaultMessage: 'Close',
+          description: 'Close label for Toast component',
+        });
 
-          class MyClassComponent extends React.Component {
-            render() {
-              const { altText, intl } = this.props;
-              const intlAltText = altText || intl.formatMessage({
-                id: 'pgn.MyComponent.altText',
-                defaultMessage: 'Close',
-                description: 'Close label for Toast component',
-              });
+        return (
+          <IconButton
+            alt={intlAltText}
+            onClick={() => {}}
+            variant="primary"
+          />
+        )
+      }
 
-              return (
-                <IconButton
-                  alt={intlCloseLabel}
-                  onClick={() => {}}
-                  variant="primary"
-                />
-              )
-            }
-          }
-
-          export default injectIntl(MyClassComponent);
-          ```
-
-      - For functional components use ``useIntl`` hook
-
-          ```javascript
-          import { useIntl } from 'react-intl';
-
-          const MyFunctionComponent = ({ altText }) => {
-            const intls = useIntl();
-            const intlAltText = altText || intl.formatMessage({
-              id: 'pgn.MyComponent.altText',
-              defaultMessage: 'Close',
-              description: 'Close label for Toast component',
-            });
-
-            return (
-              <IconButton
-                alt={intlCloseLabel}
-                onClick={() => {}}
-                variant="primary"
-              />
-            )
-          }
-
-          export default MyFunctionComponent;
-          ```
+      export default MyFunctionComponent;
+      ```
 
    **Notes on the format above**:
    - `id` is required and must be a dot-separated string of the format `pgn.<componentName>.<subcomponentName>.<propName>`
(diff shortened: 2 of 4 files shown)


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

package.json
```diff
@@ -197,7 +198,7 @@
       "/component-generator/"
     ],
     "transformIgnorePatterns": [
-      "/node_modules/(?!(@openedx/paragon)/).*/"
+      "/node_modules/(?!(@openedx/paragon|react-intl|@formatjs|intl-messageformat)/).*/"
```
brian-smith-tcril, line 201: This change stood out to me, so I did a bit of digging. It turns out the CJS vs ESM rabbit hole is real on this one. I'm happy to have this to unblock newer `react-intl`, and look into ESM test stuff when moving to newer `jest`.

<details>
<summary>Claude writeup of the digging</summary>

# Can Paragon use native ESM instead of `transformIgnorePatterns`? (PR #4412)

**Question:** PR [#4412](https://github.com/openedx/paragon/pull/4412) widens Jest's `transformIgnorePatterns` to transpile `react-intl` and friends. Could we drop that line and run Jest with `--experimental-vm-modules` instead?

**Answer:** Not on Jest 29. It's achievable in principle, but it requires a Jest 30+ upgrade plus source-level changes well beyond the scope of a react-intl compat PR. The `transformIgnorePatterns` carve-out is the right call here.

**Tested on:** PR branch `90d8c8c` (`feat: enable compatibility with react-intl v10.1.20+`), clean `npm ci`, Node v24.13.0 / npm 11.12.1, Jest 29.7.0.

---

## Why the carve-out is needed at all

`react-intl` 6.8.9 shipped dual builds — `main: index.js` (CJS) and `module: lib/index.js` (ESM). Jest resolves `main`, gets CommonJS, and never needs a transform. That's why the pre-PR pattern only listed `@openedx/paragon`.

`react-intl` 10.1.20 is **ESM-only**:

```
type:    "module"
exports: { ".": "./index.js", "./server": "./server.js" }
```

No `main`, no CJS build. Its entire new dependency tree is the same — `@formatjs/intl@4`, `@formatjs/icu-messageformat-parser@3`, `@formatjs/icu-skeleton-parser@2`, `@formatjs/fast-memoize@3`, `intl-messageformat@11` all declare `"type": "module"`.

## All three added prefixes are load-bearing

```
/node_modules/(?!(@openedx/paragon|react-intl|@formatjs|intl-messageformat)/).*/
```

`react-intl` alone is not enough, because the lockfile nests its deps — confirmed in the installed tree:

```
node_modules/react-intl/node_modules/@formatjs/fast-memoize
node_modules/react-intl/node_modules/@formatjs/icu-messageformat-parser
node_modules/react-intl/node_modules/@formatjs/icu-skeleton-parser
node_modules/react-intl/node_modules/@formatjs/intl
```

The pattern is an unanchored regex tested against the full path, so it is tried at **every** `/node_modules/` occurrence. For `node_modules/react-intl/node_modules/@formatjs/intl/index.js` the first occurrence fails the lookahead on `react-intl/`, but the second would match and ignore the file unless `@formatjs` is also listed. Same story for `intl-messageformat`. The bare `@formatjs` prefix covers all four nested packages.

## What happens as you remove it

Two representative suites, one per transformer — `src/Alert/Alert.test.tsx` (ts-jest) and `src/Dropzone/tests/DefaultContent.test.jsx` (babel-jest):

| Config | Result |
|---|---|
| PR as-is | **16/16 pass** |
| Pattern reverted to `(?!(@openedx/paragon)/)` | `SyntaxError: Cannot use import statement outside a module` at `node_modules/react-intl/index.js:2` |
| Reverted + `NODE_OPTIONS=--experimental-vm-modules` | `Must use import to load ES Module: .../node_modules/react-intl/index.js` (at `Runtime.requireModule`) |

The flag does something real — Jest stops mis-parsing the file and correctly identifies it as ESM. But Paragon's test files are still CommonJS, so they reach `react-intl` via `requireModule`, and Jest 29 has no `require(esm)` support. Node 22+ can `require()` a synchronous ESM graph natively; **jest-runtime only wired that up in Jest 30**. On Jest 29 the only way past this error is to make the importing files ESM too.

## Making the test files ESM: three walls

Attempted config (`extensionsToTreatAsEsm`, `ts-jest` `useESM: true`, Babel `modules: false`, `transformIgnorePatterns` back to the pre-PR value):

**Wall 1 — `.js` cannot be treated as ESM.** Jest rejects `.js` in `extensionsToTreatAsEsm` unless the nearest `package.json` sets `"type": "module"`. Flipping Paragon's root to `"type": "module"` would change the semantics of the *published* package, not just the tests. With Babel emitting ESM but Jest loading `.js` as CJS, you get:

```
src/Dropzone/utils.js:46
export { getTypesString, isMultipleTypes, formatBytes };
^^^^^^
SyntaxError: Unexpected token 'export'
```

Workable around by splitting the transform so `.js` → CJS and `.jsx` → ESM, which leads directly to:

**Wall 2 — the mixed CJS/ESM graph silently breaks imports.** With `.js` compiled to CJS and `.tsx` loaded as native ESM, `src/Alert/index.tsx` does:

```ts
import breakpoints from '../utils/breakpoints';   // breakpoints.js -> CJS
...
const isExtraSmall = useMediaQuery({ maxWidth: breakpoints.extraSmall.maxWidth });
```

Babel's CJS output marks the module with `__esModule` and puts the value on `exports.default`, but **native ESM ignores `__esModule`** — Node's CJS interop binds `default` to the whole `module.exports` object. So `breakpoints` becomes `{ default: {...}, __esModule: true }`, `breakpoints.extraSmall` is `undefined`, and you get:

```
TypeError: Cannot read properties of undefined (reading 'maxWidth')
  at src/Alert/index.tsx:100:73
```

Result of the attempt: the `.jsx` suite passed cleanly in ESM mode, but `Alert.test.tsx` failed **10 of 16 tests**. Note the failure mode — not a module error, a plausible-looking runtime `TypeError`. Any partial ESM migration will produce this class of breakage across the boundary, which is what makes it a whole-repo change rather than a config tweak.

**Wall 3 — not reached, but unavoidable: `jest.mock`.** 22 `jest.mock()` calls across 15 files under `src/`. In ESM mode these don't hoist and must become `jest.unstable_mockModule()` + `await import()`, with `jest` itself imported from `@jest/globals`.

## Recommendation

- **Keep the `transformIgnorePatterns` change in #4412.** It's the minimal correct fix for the ESM-only dependency, and it's verified working.
- **Defer native ESM to a Jest 29 → 30+ PR.** Jest 30's `require(esm)` support collapses Wall 1 and Wall 2 entirely (no `extensionsToTreatAsEsm`, no mixed-graph interop problem — CJS test files just `require` the ESM package). That upgrade would let the carve-out be deleted later as a cleanup, which is a much better-scoped change than doing it here.

## Separate downstream gap (not fixable in this repo)

Widening the peer range invites MFEs onto react-intl 10, where they hit the same problem via `@openedx/frontend-build`'s shared config:

```js
transformIgnorePatterns: ['/node_modules/(?!@(open)?edx)'],
```

That transforms the `@edx`/`@openedx` scopes only. `@openedx/paragon` gets transpiled, its `import { useIntl } from 'react-intl'` becomes a `require`, and that loads untransformed ESM → same error. Any test rendering an intl-using Paragon component (`Alert`, `ModalDialog`, `DataTable`, `Dropzone`, `ProductTour`, …) trips it, **even if the consumer never imports `react-intl` themselves**.

That pattern also has no `.*` tail, so nested copies aren't rescued either: for `node_modules/@openedx/paragon/node_modules/react-intl/index.js` the regex matches at the second `/node_modules/` and ignores the file.

Worth confirming whether there's a companion frontend-build PR, or at minimum a README note, before MFEs start taking v10.

</details>

brian-smith-tcril (approved):

package.json
bradenmacdonald: Oh absolutely. I have wasted a _lot_ of time trying to get Jest to play nice with ESM packages in the past. (In particular I never could get the `meilisearch` package bumped to newer ESM versions in the Authoring MFE for this reason). I was quite relieved that we "only" have to change `transformIgnorePatterns` here to get `react-intl` working.

If anyone else encounters the rabbit hole, my advice is to focus instead on [replacing Jest with Vitest](https://github.com/openedx/public-engineering/issues/486) which is [as close to a drop-in replacement as one can get](https://github.com/openedx/paragon/pull/4043), except with Vitest everything just works.

Thanks for the ✅!

